PR #13, opened 2023-04-13
[WIP] Serve Pagefind from a StarBook endpoint in dev

delucis:
Uses Pagefind’s 1.0.0 alpha release to programmatically create and serve the search index from an endpoint injected by StarBook in dev.

We might not end up shipping this (and choose to show a “search not available in dev” message or something) but for now this is a helpful experiment.

😎 **Why this is cool:** see your full site indexed and searchable locally while building — test out content strategies in real-time against the search index!

😬 **Why this is iffy:** Astro doesn’t give us an API for rendering the content of a content collection entry so we need to try and do this ourselves somehow. This implementation uses `@astrojs/markdown-remark`, but still won’t fully match page rendering in Astro.


delucis:
Closing for now.

PR #45, opened 2023-05-14
Set up size-limit CI

delucis:


delucis:
Failing check is expected because the action runs `size-limit` twice: once on the PR branch and once on `main`. Because `main` doesn’t contain `size-limit` yet, the second run fails. Should be fixed going forward.

Issue #56, opened 2023-05-16
Site title always links to `/` not to the current locale


delucis:
### What version of `starlight` are you using?

0.0.7

### What version of `astro` are you using?

2.4.3

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### What browser are you using?

Firefox/Chrome/Safari

### Describe the Bug

The site title/logo link (in the `SiteTitle.astro` component) currently always links to the website base.

This means for pages not in the root locale (like German in the current Starlight docs), clicking the title link takes you to a different language.

We should pass the current locale to the site title component and use it in the link so that users stay in their own language's part of the site when clicking it.

### Link to Minimal Reproducible Example

https://starlight.astro.build/de/getting-started/

### Participation

- [ ] I am willing to submit a pull request for this issue.

delucis:
Done in #57 🎉

PR #57, opened 2023-05-17
Add locale to SiteTitle link

BryceRussell:
**Closes**: https://github.com/withastro/starlight/issues/56

- Add `locale` prop to `<SiteTitle>`
- Use locale prop to to create link for `<SiteTitle>`

### DIFF
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -2,6 +2,10 @@
 import { logos } from 'virtual:starlight/user-images';
 import config from 'virtual:starlight/user-config';
 
+interface Props {
+  locale: string | undefined;
+}
+
 if (config.logo) {
   let err: string | undefined;
   if ('src' in config.logo) {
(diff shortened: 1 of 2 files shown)


packages/starlight/components/SiteTitle.astro
```diff
@@ -17,9 +21,15 @@ if (config.logo) {
   }
   if (err) throw new Error(err);
 }
+
+const { locale } = Astro.props
+
+const href = locale
+  ? `${import.meta.env.BASE_URL.replace(/\/$/, '')}/${locale}/`
```
delucis, line 28: Thanks @BryceRussell!

Did you find `BASE_URL` sometimes _doesn't_ have a trailing slash? I think I may have assumed it always does elsewhere in the code base, so could be worth double checking.

delucis: Did a quick investigation. Would probably be good to refactor everywhere we use `import.meta.env.BASE_URL` to use a helper like `withBase` so we’re consistent, i.e. this would become something like `withBase(locale ? '/locale/' : '/')`.

We can do that in a follow-up PR though. Thanks for this @BryceRussell!

delucis (approved):

PR #59, opened 2023-05-17
Update npm create command

morinokami:
Added `--` to correctly pass the `template` option (though `Template starlight is empty!` error still occurs because the starlight template is not yet provided😅).


delucis (approved):
Good catch! Hoping to get the template published soon.

Thanks @morinokami 🚀

Issue #58, opened 2023-05-17
TOC Incorrect Clicking Behavior


liruifengv:
### What version of `starlight` are you using?

0.0.7

### What version of `astro` are you using?

2.4.3

### What package manager are you using?

pnpm

### What operating system are you using?

Windows 10

### What browser are you using?

Chrome 113.0.5672.93

### Describe the Bug

The click event of markdown's TOC is incorrect.

When I click on an item, the item above it is selected.

And some items are not clickable.

### Link to Minimal Reproducible Example

https://starlight.astro.build/de/reference/markdown-syntax/

### Participation

- [ ] I am willing to submit a pull request for this issue.

delucis:
Thanks for the report!

> When I click on an item, the item above it is selected.

I think this is probably caused by us fixing the scroll margin in #54. Item selection is done using an [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API), and with the new margin, the heading doesn’t intersect when clicking it.

We might be able to adjust the `rootMargin` here to get what we want:

https://github.com/withastro/starlight/blob/0040a06a4dcea9348222617080410823d74dae09/packages/starlight/components/TableOfContents/starlight-toc.ts#L72

I found this tool super handy for debugging intersection observer behaviour: https://github.com/pomber/intersection-observer-debugger

> And some items are not clickable.

Could you say more? Are you referring to clicking on the last few items on the page not getting highlighted or are some links not working at all for you?

liruifengv:
> And some items are not clickable.

en... This seems to be caused by the same intersection observer.

delucis:
There may be something we can do to improve those last few items. In the current Astro Docs component, we add a click handler in addition to the intersection observer. We might be able ignore the intersections if a link was clicked or something if people think this is important.

PR #61, opened 2023-05-17
Fix toc headingsObserver rootMargin

liruifengv:
Fix https://github.com/withastro/starlight/issues/58

- Set rootMargin '-10% 0% -85%'


delucis (approved):
Thanks @liruifengv! I think I might change this to also use different values on mobile vs desktop, because I’m still seeing not quite the behaviour I want on mobile, but this is a great first step. Welcome to the Starlight contributors 🥳

PR #63, opened 2023-05-17
Add a collapse function to the sidebar

liruifengv:


packages/starlight/components/SidebarSublist.astro
delucis, line 1: Oh cool, thanks for tackling that!

Couldn’t we do it without JavaScript though? If we wrap each section in `<details open></details>` and headings in the `<summary>` tag we should get the same collapsing behaviour without needing JS.

liruifengv: OK, I can try it

packages/starlight/components/SidebarSublist.astro
```diff
@@ -43,6 +47,22 @@ interface Props {
     margin-top: 0.75rem;
   }
 
+  .sidebar-group-header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    cursor: pointer;
```
delucis, line 54: The designs we have for this have the padding around the full item:
image

We can add it here:

```suggestion
    padding-inline: var(--sl-sidebar-item-padding-inline);
    cursor: pointer;
```

We should also remove the `padding-inline` from `h2` in this case.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -15,10 +16,13 @@ interface Props {
             {entry.label}
           </a>
         ) : (
-          <>
-            <h2>{entry.label}</h2>
+          <details open>
+            <summary class="sidebar-group-header">
+              <h2>{entry.label}</h2>
+              <Icon name="down-caret" class="icon caret" />
```
delucis, line 22: The icons are also 20px in the designs, so we can make these a bit larger and I’m going to suggest using `right-caret` here so we can simplify the CSS selector:

```suggestion
              <Icon name="right-caret" class="caret" size="1.25rem" />
```

(Also removed the `icon` class as it’s not used here.)

packages/starlight/components/SidebarSublist.astro
```diff
@@ -43,6 +47,22 @@ interface Props {
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    cursor: pointer;
+    user-select: none;
+  }
+
+  details:not([open]) .caret {
+    transform: rotate(-90deg);
+  }
```
delucis, line 60: As suggested above, if we use `right-caret` we can simplify this a litte:

```suggestion
  [open] .caret {
    transform: rotateZ(90deg);
  }
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -43,6 +47,22 @@ interface Props {
...
   }
 
+  .sidebar-group-header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    cursor: pointer;
+    user-select: none;
+  }
+
```
delucis, line 57: One last thing, to display correctly for languages written right-to-left (Arabic or Hebrew for example), we should flip the caret 180deg so it points left in these scenarios:

```suggestion

  :global([dir='rtl']) .caret {
    transform: rotateZ(180deg);
  }
```

delucis (changes requested):
Thanks @liruifengv! Looks and works great.

Here are a few small tweaks to get this ready to merge.

liruifengv:
Thanks for your help. I have learned a lot.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -43,6 +47,30 @@ interface Props {
...
+    align-items: center;
+    justify-content: space-between;
+    padding-inline: var(--sl-sidebar-item-padding-inline);
+    cursor: pointer;
+    user-select: none;
+  }
+  
+  h2 {
+    padding-inline: unset
+  }
```
delucis, line 61: Oh, I think we don’t need this. We can just edit the `h2` style on line 43 above to remove `padding-inline`!

packages/starlight/components/SidebarSublist.astro
```diff
@@ -43,6 +47,30 @@ interface Props {
     margin-top: 0.75rem;
   }
```
delucis, line 49: Ah, and remembering we also need these styles to get rid of the default arrow in Safari:
```suggestion

  summary::marker,
  summary::-webkit-details-marker {
    display: none;
  }
```
Looks like this currently 😅 
image

liruifengv: The dreaded Safari

packages/starlight/components/SidebarSublist.astro
liruifengv: Sorry, I didn't see it earlier...

delucis (approved):

delucis:
Thanks @liruifengv — will release this now 🙌

PR #67, opened 2023-05-17
Fix background color on select component

### LINKED ISSUE #65: Hard to read text in dropdown menus when using dark theme

0.0.7

2.4.3

pnpm

Windows

Chrome

### Describe the Bug

**Description**

It's hard to read the text in dropdown menus when using dark theme. Also, the style of the dropdown menus should match with the rest of the site.

This behavior is same on *Microsoft Edge* and *Mozilla Firefox*.

**Screenshots**

image
---
image

**Additional context**

[G18: Ensuring that a contrast ratio of at least 4.5:1 exists between text (and images of text) and background behind the text](https://www.w3.org/TR/WCAG20-TECHS/G18.html#:~:text=If%20the%20background%20is%20a,contrast%20ratio%20with%20the%20background.)

_No response_

### Participation

TheOtterlord:
- Closes #65 
- Doesn't address [this comment](https://github.com/withastro/starlight/issues/65#issuecomment-1551771078) because I don't think we want to disable an outline.
> Yeah I also think that the focus state should be removed, I was trying to look if the astro.build website had this same behavior but they don't

After change (viewed in chromium):
image
image

### DIFF
--- a/.changeset/light-dingos-fry.md
+++ b/.changeset/light-dingos-fry.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix background color on select component
--- a/packages/starlight/components/Select.astro
+++ b/packages/starlight/components/Select.astro
@@ -72,6 +72,10 @@ interface Props {
     appearance: none;
   }
 
+  option {
+    background-color: var(--sl-color-bg-nav);
+  }
+
   @media (min-width: 50rem) {
     select {
       font-size: var(--sl-text-sm);


dreyfus92 (approved):
This looks way better, awesome job @TheOtterlord 😄

delucis (approved):
_Definitely_ not worse than what we had before!

Looks like some folks still see a small bit of off-colour background in the dropdown, but at least it’s legible now. Thanks for the quick fix @TheOtterlord!

PR #75, opened 2023-05-18
Fix create command with full template path

liruifengv:


delucis (approved):
D’oh, I’m stupid 😆 

Thanks @liruifengv!

PR #76, opened 2023-05-19
Scale down code block font size to match Figma design

lloydjatkinson:

- Something else!

#### Description

- This PR changes the code block font size to a smaller value.

### DIFF
--- a/packages/starlight/style/props.css
+++ b/packages/starlight/style/props.css
@@ -93,7 +93,7 @@
 
   --sl-text-body: var(--sl-text-base);
   --sl-text-body-sm: var(--sl-text-xs);
-  --sl-text-code: var(--sl-text-xs);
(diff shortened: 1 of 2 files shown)


packages/starlight/style/props.css
```diff
@@ -93,7 +93,7 @@
 
   --sl-text-body: var(--sl-text-base);
   --sl-text-body-sm: var(--sl-text-xs);
-  --sl-text-code: var(--sl-text-base);
+  --sl-text-code: var(--sl-text-xs);
```
delucis, line 96: Thanks @lloydjatkinson!

I think 13px here might be just too small. Let’s try `sm` which is 14px:

```suggestion
  --sl-text-code: var(--sl-text-sm);
```

delucis (approved):
Thanks! We’ll get this out in the next release 🙌

Issue #70, opened 2023-05-17
Sidebar titles are not being updated when switching locales.


dreyfus92:
### What version of `starlight` are you using?

0.0.7

### What version of `astro` are you using?

2.4.3

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

Titles on the sidebar are not being updated properly when you switch between locales. 

<img width="1475" alt="image" src="https://github.com/withastro/starlight/assets/85648028/1211da30-9e5a-4810-954f-48e046ab3e65">

<img width="1476" alt="image" src="https://github.com/withastro/starlight/assets/85648028/f883ad4c-6f54-45ea-9399-81400e4a1670">

### Participation

- [ ] I am willing to submit a pull request for this issue.

delucis:
Oops! This definitely worked before but I probably broke it in #55. Thanks for reporting!

dreyfus92:
@delucis I will take a look to that PR later.

delucis:
Oh wait, I remember now. Titles are translated for “autogenerated” sidebars currently (see the Reference section for example), but not yet for explicit entries in the sidebar.

So here’s the current sidebar config:

https://github.com/withastro/starlight/blob/c4731c5829ef23d43499856fc1fd5e8c36ef6f24/docs/astro.config.mjs#L42-L63

We’ll need to set up a way to translate the “Start Here” items. I was delaying this while figuring out other “string” translations (like `ui.ts` in the Astro docs). Might end up suggesting a format something like:

```js
{
  label: 'Welcome, world',
  link: '/',
  translations: {
    de: 'Wilkommen, Welt',
  },
}
```

PR #78, opened 2023-05-20
Implement UI string translation using data collections

delucis:

- Changes to Starlight code

#### Description

- Adds support for translating elements of Starlight’s UI.

Users can provide translations in JSON files in `src/content/i18n/` which is a data collection, for example:

```json
{
  "search.label": "Suchen",
  "search.shortcutLabel": "(Drücke / zum Suchen)"
}
```

In Starlight templates, we look up strings for the current locale from this data collection, falling back to the user-configured default locale, or to Starlight’s English defaults.


dreyfus92:
This looks great @delucis but I'm wondering if you need to change something else in the astro.config.mjs file to be able to see the updated UI in german?

delucis:
Ah, this doesn't yet fix #70! This is only for the static parts of the UI like the search bar, table of contents, etc. That looks like it's working ok in the preview, I think?

Will work to fix #70 as well soon!

dreyfus92:
It works in the preview 😃

PR #82, opened 2023-05-23
fix: change link for guides/i18n

ElianCodes:

#### Description

Weirdly enough, clicking on the link "i18n support" here, will throw a 404 page. This fixes that (but will lose translation?) changing it to `../../guides/i18n` will also fix it.

Screenshot 2023-05-23 at 18 03 25


delucis (approved):
Oh yeah, good catch! 🙌

PR #83, opened 2023-05-23
[docs] defaultLocale conflict when using root

sarah11918:
Just left a quick suggested docs PR that would have addressed the problem I ran into so I don't forget! Adjust as necessary!

### DIFF
--- a/docs/src/content/docs/guides/i18n.md
+++ b/docs/src/content/docs/guides/i18n.md
@@ -59,7 +59,7 @@ Starlight provides built-in support for multilingual sites, including routing, f
 
 You can use a “root” locale to serve a language without any i18n prefix in its path. For example, if English is your root locale, an English page path might look like `/about` instead of `/en/about`.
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/i18n.md
```diff
@@ -59,7 +59,7 @@ Starlight provides built-in support for multilingual sites, including routing, f
 
 You can use a “root” locale to serve a language without any i18n prefix in its path. For example, if English is your root locale, an English page path might look like `/about` instead of `/en/about`.
 
-To set a root locale, use the `root` key in your `locales` config:
+To set a root locale, use the `root` key in your `locales` config and remove `defaultLocale` (or set to `root`):
```
delucis, line 62: I might flesh this out just slightly. It is technically possible your default content is _not_ in the root (although pretty unlikely I’d expect).

```suggestion
To set a root locale, use the `root` key in your `locales` config. If the root locale is also the default locale for your content, remove `defaultLocale` or set it to `'root'`.
```

delucis (commented):
Thanks @sarah11918! One small suggestion.

PR #85, opened 2023-05-23
Improve outside click detection on search modal

BryceRussell:

- Changes to Starlight code

#### Description

- Improve outside click detection on search modal
- Check if click target exists inside the DOM before closing


delucis (approved):
Ohhhh right! Because when you click the “Load results” button, the button is removed from the DOM, and then that check isn’t true. Nice bug and nice job figuring it out. Thanks @BryceRussell!

PR #86, opened 2023-05-23
Add required title to complete code snippets

TheOtterlord:

#### Description

- What does this PR change? Adds `title` field to "complete" code snippets.

I've left code snippets that are not complete run-able files, because I'm not sure the intention for these is to be used as a complete configuration, but I'm open to suggestions. If we wanted to emphasise that a snippet isn't "complete" configuration, would we instead use an elipsis `...` notation, or something of the like? Open to ideas if you think this needs to cover more.


delucis (approved):
Thanks @TheOtterlord!

PR #98, opened 2023-05-30
Make the "Contribute" string in the right sidebar translatable

morinokami:

- Changes to Starlight code

#### Description

- Include the key for the "Contribute" string used in the right sidebar in `i18nSchema`.
  - The German and Spanish translations were copied and pasted from Astro's site. If any of these translations are incorrect, please let me know!

### DIFF
--- a/.changeset/many-mugs-lie.md
+++ b/.changeset/many-mugs-lie.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Make the "Contribute" string in the right sidebar translatable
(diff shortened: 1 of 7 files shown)


delucis:
Thanks for taking the time to do this @morinokami!

I’m actually going to move the edit link to the page footer and remove the “Contribute” heading (see #99), that’s why I hadn’t yet configured this heading for translation:

Screenshot of edit link in the footer next last updated date, in Japanese

I’m guessing it might cause conflicts if I merge this, so will close it instead. Sorry for not leaving a comment to explain my intention on that heading before you spent this time!

PR #128, opened 2023-06-04
Add file tree component & project structure guide to docs

delucis:

#### Description

- Adds a “Project Structure” page
- Adds a `<FileTree>` component similar to the one we use in Astro Docs. This is only available in the docs site currently, but the intention is to make this available as part of Starlight when we have time to polish it and solidify the public API.


docs/src/content/docs/guides/project-structure.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: Project Structure
+description: Learn how to organize files in your Starlight project.
+---
+
+This guide will show you how a Starlight project is organized and what different files in your project do. Starlight projects generally follow the same file and directory structure as other Astro projects. See [Astro’s project structure documentation](https://docs.astro.build/en/core-concepts/project-structure/) for more detail.
```
sarah11918, line 6:
```suggestion
This guide will show you how a Starlight project is organized and what the different files in your project do.

Starlight projects generally follow the same file and directory structure as other Astro projects. See [Astro’s project structure documentation](https://docs.astro.build/en/core-concepts/project-structure/) for more detail.
```
Nit: I like adding "the", and I like separating these concepts into their own paragraphs.

docs/src/content/docs/guides/project-structure.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Files and directories
+
+- `astro.config.mjs` — The Astro configuration file; includes the Starlight integration and configuration.
+- `src/content/config.ts` — Content collection configuration file; adds Starlight’s frontmatter schemas to your project.
```
sarah11918, line 11:
```suggestion
- `src/content/config.ts` — Content collections configuration file; adds Starlight’s frontmatter schemas to your project.
```
I'm honestly not sure here! Singlular or plural??

docs/src/content/docs/guides/project-structure.mdx
```diff
@@ -0,0 +1,46 @@
...
+- `src/content/docs/` — Content files. Starlight turns each file in this directory into a page on your site.
```
sarah11918, line 12:
```suggestion
- `src/content/docs/` — Content files. Starlight turns each `.md`, `.mdx` or `.mdoc` file in this directory into a page on your site.
```
Maybe we should specify which files become pages, for full honesty?

docs/src/content/docs/guides/project-structure.mdx
```diff
@@ -0,0 +1,46 @@
...
+- `src/content/i18n/` (optional) — Translation data to support [internationalization](/guides/i18n/).
+- `src/` — Other project source code (components, styles, etc.)
+- `public/` — Unprocessed, static assets (fonts, favicon, PDFs, etc.)
```
sarah11918, line 15:
```suggestion
- `src/` — Other project source code and files (components, styles, images, etc.) for your Astro project.
- `public/` — Static assets (fonts, favicon, PDFs, etc.)  that will not be processed by Astro.
```
I actually didn't dislike what you had, but you had periods at the end of all the other lines, and adding some text at the end avoids `etc.).`  :sweat_smile: 

Also, there was no explicit mention about images that *will* be processed, so I snuck "images" into the list of other project source code. I also considered an optional `src/assets` line itself. I think it's worth saying where most of your images actually are going to go.

sarah11918 (commented):
Looks great! Left some comments!

docs/src/content/docs/guides/project-structure.mdx
delucis: Editing your suggestion to remove the double “project” in the first bullet point and also removed the “Astro” as I don’t think that’s crucial in this context:

```suggestion
- `src/` — Other source code and files (components, styles, images, etc.) for your project.
- `public/` — Static assets (fonts, favicon, PDFs, etc.)  that will not be processed by Astro.
```

I did consider mentioning `src/assets/`. Now that we support images anywhere in the project, I didn’t think it necessary to go into detail. I do think an “Authoring Content” guide could be good that could specifically go into linking to images from Markdown and therefore where they’re stored (I think now that it’s supported, storing an image alongside its content is probably preferable often, e.g. `src/content/docs/guide.md` and `src/content/docs/guide/image.png`).

docs/src/content/docs/guides/project-structure.mdx
delucis: Eh, could go either way couldn’t it. I’ll go with the instinct that given you bothered to comment on this, that’s a signal it should be plural 😅

PR #134, opened 2023-06-04
Add Portuguese language support

Yan-Thomas:

- Changes to Starlight code

#### Description

Added Portuguese UI labels to Starlight and updated the i18n page in docs to account for the change.

### DIFF
--- a/docs/src/content/docs/guides/i18n.mdx
+++ b/docs/src/content/docs/guides/i18n.mdx
@@ -120,7 +120,7 @@ If a translation is not yet available for a language, Starlight will show reader
 
 Some of Starlight’s default UI requires text labels to work.
 For example, the table of contents on this page has an “On this page” heading in English.
-We aim to ship these labels in as many languages as possible but currently only have support for English, German, Japanese, Spanish, and Portuguese.
+We aim to ship these labels in as many languages as possible. Currently, English, German, Japanese, Portuguese, and Spanish are supported out of the box.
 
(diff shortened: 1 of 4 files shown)


docs/src/content/docs/guides/i18n.mdx
```diff
@@ -120,7 +120,7 @@ If a translation is not yet available for a language, Starlight will show reader
 
 Some of Starlight’s default UI requires text labels to work.
 For example, the table of contents on this page has an “On this page” heading in English.
-We aim to ship these labels in as many languages as possible but currently only have support for English, German, Japanese, and Spanish.
+We aim to ship these labels in as many languages as possible but currently only have support for English, German, Japanese, Spanish, and Portuguese.
```
delucis, line 123: I guess maybe we can drop “only” now that we have 5 languages. What do you think?

```suggestion
We aim to ship these labels in as many languages as possible. Currently, English, German, Japanese, Portuguese, and Spanish are supported out of the box.
```

delucis (commented):
Thanks @Yan-Thomas! Would have felt funny to launch without this given how much you’ve contributed in this area already. Glad to include it.

Made one small suggestion for a tweak to the docs, but otherwise this looks great!

docs/src/content/docs/guides/i18n.mdx
Yan-Thomas: Agree!

delucis (approved):

PR #143, opened 2023-06-05
config reference intro and example

sarah11918:
Adds a short intro explaining that Starlight is an Astro integration, plus generic example for configuring it at the top of the page.
deli

### DIFF
--- a/docs/src/content/docs/reference/configuration.md
+++ b/docs/src/content/docs/reference/configuration.md
@@ -5,6 +5,22 @@ description: An overview of all the configuration options Starlight supports.
 
 ## Configure the `starlight` integration
 
+Starlight is an [Astro integration](https://docs.astro.build/en/guides/integrations-guide/), configured in `astro.config.mjs`.
+
+```js
+// astro.config.mjs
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/reference/configuration.md
```diff
@@ -5,6 +5,22 @@ description: An overview of all the configuration options Starlight supports.
 
 ## Configure the `starlight` integration
 
+Starlight is an [Astro integration](https://docs.astro.build/en/guides/integrations-guide/), configured in `astro.config.mjs`.
```
FredKSchott, line 8: I wouldn't overwhelm the user with that page of docs, or feeling the need to understand much about Astro, integrations, etc. Instead, I might reframe to something like this:

```suggestion
Starlight is built on top the [Astro](https://astro.build) web framework. You can configure your project inside the Astro `astro.config.mjs` configuration file, like this:
```

FredKSchott (approved):
1 comment but I trust your judgement to resolve, +1!

sarah11918:
Used (mostly) Fred's suggestion... will merge when checks pass!

PR #157, opened 2023-06-05
feat: Adds print styles for Starlight docs pages

tony-sull:

- Changes to Starlight code
- Something else!

#### Description

- Adds print-specific styles for pages built with Starlight
- Hides navigation elements like the sidebars, search, and mobile menu
- Forces the light theme color palette when printing


delucis:
I think my preference is likely for pulling this out to a separate global file to avoid extra downloads. (We’d need a `?url` import I’d guess to pass that to the appropriately configured style tag.)

I am a bit torn because as you highlighted, that does come with extra maintenance complexity. Especially as this is exactly the kind of thing we’re likely to forget to check as we add stuff 😅 That said, most stuff is done pretty safely via `print:hidden`, so we should be OK with that approach?

(There’s a chance even doing that will fail the CI check here — it’s a bit of a rough tool at the moment, just globbing `dist/*.css` and measuring that gzipped, so even CSS that doesn’t actually get loaded is liable to get taken into account. Tried another approach and found it unreliable, but this is definitely on a list of tooling to improve.)

tony-sull:
@delucis Yep that's no problem, I'll refactor that this morning and pull the styles out into a global sheet 👍

tony-sull:
@delucis done!  Everything is moved out to the separate `print.css` file

Looks like the HTML size-limit is failing by ~1kb, that should just be from the extra `print:` class names + the stylesheet `<link>`

packages/starlight/index.astro
```diff
@@ -32,6 +32,8 @@ import './style/asides.css';
 // Important that this is the last import so it can override built-in styles.
 import 'virtual:starlight/user-css';
 
+import printHref from './style/print.css?url';
```
delucis, line 35: Hah, looks like this is under Vite’s inline assets limit, so is getting inlined as a data URI in the build output 😅

```html
<link rel="stylesheet" href="[data:text/css;base64,OnJvb3QsCjo6YmFja2Ryb3AgewogIC8qIENvbG91cnMgKGxpZ2h0IG1vZGUpICovCiAgLS1zbC1jb2xvci13aGl0ZTogaHNsKDIyMywgMTMlLCAxMCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmF5LTE6IGhzbCgyMjIsIDEzJSwgMTYlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItZ3JheS0yOiBoc2woMjIyLCAxMSUsIDIzJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktMzogaHNsKDIyNSwgOSUsIDM2JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNDogaHNsKDIyMSwgOCUsIDU2JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNTogaHNsKDIyOCwgOCUsIDc3JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNjogaHNsKDIzMSwgMjMlLCA5NCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmF5LTc6IGhzbCgyNDAsIDIwJSwgOTclKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmxhY2s6IGhzbCgwLCAwJSwgMTAwJSkgIWltcG9ydGFudDsKCiAgLS1zbC1jb2xvci1vcmFuZ2UtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1vcmFuZ2UpLCA4MCUsIDI1JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLW9yYW5nZTogaHNsKHZhcigtLXNsLWh1ZS1vcmFuZ2UpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLW9yYW5nZS1sb3c6IGhzbCh2YXIoLS1zbC1odWUtb3JhbmdlKSwgOTAlLCA4OCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmVlbi1oaWdoOiBoc2wodmFyKC0tc2wtaHVlLWdyZWVuKSwgODAlLCAyMiUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmVlbjogaHNsKHZhcigtLXNsLWh1ZS1ncmVlbiksIDkwJSwgNDYlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItZ3JlZW4tbG93OiBoc2wodmFyKC0tc2wtaHVlLWdyZWVuKSwgODUlLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ibHVlLWhpZ2g6IGhzbCh2YXIoLS1zbC1odWUtYmx1ZSksIDgwJSwgMzAlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmx1ZTogaHNsKHZhcigtLXNsLWh1ZS1ibHVlKSwgOTAlLCA2MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ibHVlLWxvdzogaHNsKHZhcigtLXNsLWh1ZS1ibHVlKSwgODglLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1wdXJwbGUtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1wdXJwbGUpLCAzOSUsIDMwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXB1cnBsZTogaHNsKHZhcigtLXNsLWh1ZS1wdXJwbGUpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXB1cnBsZS1sb3c6IGhzbCh2YXIoLS1zbC1odWUtcHVycGxlKSwgODAlLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1yZWQtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1yZWQpLCA4MCUsIDMwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXJlZDogaHNsKHZhcigtLXNsLWh1ZS1yZWQpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXJlZC1sb3c6IGhzbCh2YXIoLS1zbC1odWUtcmVkKSwgODAlLCA5MCUpICFpbXBvcnRhbnQ7CgogIC0tc2wtY29sb3ItYWNjZW50LWhpZ2g6IGhzbCh2YXIoLS1zbC1odWUtYWNjZW50KSwgODAlLCAzMCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1hY2NlbnQ6IGhzbCh2YXIoLS1zbC1odWUtYWNjZW50KSwgOTAlLCA2MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1hY2NlbnQtbG93OiBoc2wodmFyKC0tc2wtaHVlLWFjY2VudCksIDg4JSwgOTAlKSAhaW1wb3J0YW50OwoKICAtLXNsLWNvbG9yLXRleHQtYWNjZW50OiB2YXIoLS1zbC1jb2xvci1hY2NlbnQpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci10ZXh0LWludmVydDogdmFyKC0tc2wtY29sb3ItYmxhY2spICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1iZy1uYXY6IHZhcigtLXNsLWNvbG9yLWdyYXktNykgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWJnLXNpZGViYXI6IHZhcigtLXNsLWNvbG9yLWJnKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmctaW5saW5lLWNvZGU6IHZhcigtLXNsLWNvbG9yLWdyYXktNikgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWhhaXJsaW5lLXNoYWRlOiB2YXIoLS1zbC1jb2xvci1ncmF5LTYpICFpbXBvcnRhbnQ7CgogIC0tc2wtY29sb3ItYmFja2Ryb3Atb3ZlcmxheTogaHNsYSgyMjUsIDklLCAzNiUsIDAuNjYpICFpbXBvcnRhbnQ7CgogIC8qIFNoYWRvd3MgKGxpZ2h0IG1vZGUpICovCiAgLS1zbC1zaGFkb3ctc206IDBweCAxcHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSwKICAgIDBweCAycHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSAhaW1wb3J0YW50OwogIC0tc2wtc2hhZG93LW1kOiAwcHggOHB4IDRweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksCiAgICAwcHggNXB4IDJweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksIDBweCAzcHggMnB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSwKICAgIDBweCAxcHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSAhaW1wb3J0YW50OwogIC0tc2wtc2hhZG93LWxnOiAwcHggMjVweCA3cHggcmdiYSgwLCAwLCAwLCAwLjAxKSwKICAgIDBweCAxNnB4IDZweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksIDBweCA5cHggNXB4IGhzbGEoMjIzLCAxMyUsIDEwJSwgMC4wOCksCiAgICAwcHggNHB4IDRweCBoc2xhKDAsIDAlLCAwJSwgMC4xNiksIDBweCA0cHggMnB4IGhzbGEoMCwgMCUsIDAlLCAwLjA0KSAhaW1wb3J0YW50Owp9CgoucHJpbnRcOmhpZGRlbiB7CiAgZGlzcGxheTogbm9uZSAhaW1wb3J0YW50Owp9Ci5wcmludFw6ZmxleCB7CiAgZGlzcGxheTogZmxleCAhaW1wb3J0YW50Owp9Ci5wcmludFw6YmxvY2sgewogIGRpc3BsYXk6IGJsb2NrICFpbXBvcnRhbnQ7Cn0KCi8qIGluZGV4LmFzdHJvICovCm1haW4gewogIHBhZGRpbmctYm90dG9tOiAwICFpbXBvcnRhbnQ7Cn0KbWFpbiA+IC5jb250ZW50LXBhbmVsIHsKICBwYWRkaW5nLWJsb2NrLXN0YXJ0OiAwICFpbXBvcnRhbnQ7Cn0KCi8qIGNvbXBvbmVudHMvTWFya2Rvd25Db250ZW50LmFzdHJvICovCi5jb250ZW50IHByZSB7CiAgb3ZlcmZsb3cteDogaGlkZGVuICFpbXBvcnRhbnQ7CiAgd2hpdGUtc3BhY2U6IHByZS13cmFwICFpbXBvcnRhbnQ7CiAgcGFnZS1icmVhay1pbnNpZGU6IGF2b2lkOwp9CgovKiBsYXlvdXQvUGFnZUZyYW1lLmFzdHJvICovCi5wYWdlID4gaGVhZGVyIHsKICBwb3NpdGlvbjogcmVsYXRpdmUgIWltcG9ydGFudDsKfQoucGFnZSA+IC5tYWluLWZyYW1lIHsKICBwYWRkaW5nLWlubGluZS1zdGFydDogMCAhaW1wb3J0YW50Owp9CgovKiBsYXlvdXQvVHdvQ29sdW1uQ29udGVudC5hc3RybyAqLwoubWFpbi1wYW5lIHsKICAtLXNsLXNpZGViYXItd2lkdGg6IDBweCAhaW1wb3J0YW50OwogIC0tc2wtY29udGVudC13aWR0aDogMTAwJSAhaW1wb3J0YW50Owp9Cg==](view-source:data:text/css;base64,OnJvb3QsCjo6YmFja2Ryb3AgewogIC8qIENvbG91cnMgKGxpZ2h0IG1vZGUpICovCiAgLS1zbC1jb2xvci13aGl0ZTogaHNsKDIyMywgMTMlLCAxMCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmF5LTE6IGhzbCgyMjIsIDEzJSwgMTYlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItZ3JheS0yOiBoc2woMjIyLCAxMSUsIDIzJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktMzogaHNsKDIyNSwgOSUsIDM2JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNDogaHNsKDIyMSwgOCUsIDU2JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNTogaHNsKDIyOCwgOCUsIDc3JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWdyYXktNjogaHNsKDIzMSwgMjMlLCA5NCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmF5LTc6IGhzbCgyNDAsIDIwJSwgOTclKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmxhY2s6IGhzbCgwLCAwJSwgMTAwJSkgIWltcG9ydGFudDsKCiAgLS1zbC1jb2xvci1vcmFuZ2UtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1vcmFuZ2UpLCA4MCUsIDI1JSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLW9yYW5nZTogaHNsKHZhcigtLXNsLWh1ZS1vcmFuZ2UpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLW9yYW5nZS1sb3c6IGhzbCh2YXIoLS1zbC1odWUtb3JhbmdlKSwgOTAlLCA4OCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmVlbi1oaWdoOiBoc2wodmFyKC0tc2wtaHVlLWdyZWVuKSwgODAlLCAyMiUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ncmVlbjogaHNsKHZhcigtLXNsLWh1ZS1ncmVlbiksIDkwJSwgNDYlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItZ3JlZW4tbG93OiBoc2wodmFyKC0tc2wtaHVlLWdyZWVuKSwgODUlLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ibHVlLWhpZ2g6IGhzbCh2YXIoLS1zbC1odWUtYmx1ZSksIDgwJSwgMzAlKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmx1ZTogaHNsKHZhcigtLXNsLWh1ZS1ibHVlKSwgOTAlLCA2MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1ibHVlLWxvdzogaHNsKHZhcigtLXNsLWh1ZS1ibHVlKSwgODglLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1wdXJwbGUtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1wdXJwbGUpLCAzOSUsIDMwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXB1cnBsZTogaHNsKHZhcigtLXNsLWh1ZS1wdXJwbGUpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXB1cnBsZS1sb3c6IGhzbCh2YXIoLS1zbC1odWUtcHVycGxlKSwgODAlLCA5MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1yZWQtaGlnaDogaHNsKHZhcigtLXNsLWh1ZS1yZWQpLCA4MCUsIDMwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXJlZDogaHNsKHZhcigtLXNsLWh1ZS1yZWQpLCA5MCUsIDYwJSkgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLXJlZC1sb3c6IGhzbCh2YXIoLS1zbC1odWUtcmVkKSwgODAlLCA5MCUpICFpbXBvcnRhbnQ7CgogIC0tc2wtY29sb3ItYWNjZW50LWhpZ2g6IGhzbCh2YXIoLS1zbC1odWUtYWNjZW50KSwgODAlLCAzMCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1hY2NlbnQ6IGhzbCh2YXIoLS1zbC1odWUtYWNjZW50KSwgOTAlLCA2MCUpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1hY2NlbnQtbG93OiBoc2wodmFyKC0tc2wtaHVlLWFjY2VudCksIDg4JSwgOTAlKSAhaW1wb3J0YW50OwoKICAtLXNsLWNvbG9yLXRleHQtYWNjZW50OiB2YXIoLS1zbC1jb2xvci1hY2NlbnQpICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci10ZXh0LWludmVydDogdmFyKC0tc2wtY29sb3ItYmxhY2spICFpbXBvcnRhbnQ7CiAgLS1zbC1jb2xvci1iZy1uYXY6IHZhcigtLXNsLWNvbG9yLWdyYXktNykgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWJnLXNpZGViYXI6IHZhcigtLXNsLWNvbG9yLWJnKSAhaW1wb3J0YW50OwogIC0tc2wtY29sb3ItYmctaW5saW5lLWNvZGU6IHZhcigtLXNsLWNvbG9yLWdyYXktNikgIWltcG9ydGFudDsKICAtLXNsLWNvbG9yLWhhaXJsaW5lLXNoYWRlOiB2YXIoLS1zbC1jb2xvci1ncmF5LTYpICFpbXBvcnRhbnQ7CgogIC0tc2wtY29sb3ItYmFja2Ryb3Atb3ZlcmxheTogaHNsYSgyMjUsIDklLCAzNiUsIDAuNjYpICFpbXBvcnRhbnQ7CgogIC8qIFNoYWRvd3MgKGxpZ2h0IG1vZGUpICovCiAgLS1zbC1zaGFkb3ctc206IDBweCAxcHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSwKICAgIDBweCAycHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSAhaW1wb3J0YW50OwogIC0tc2wtc2hhZG93LW1kOiAwcHggOHB4IDRweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksCiAgICAwcHggNXB4IDJweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksIDBweCAzcHggMnB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSwKICAgIDBweCAxcHggMXB4IGhzbGEoMCwgMCUsIDAlLCAwLjA2KSAhaW1wb3J0YW50OwogIC0tc2wtc2hhZG93LWxnOiAwcHggMjVweCA3cHggcmdiYSgwLCAwLCAwLCAwLjAxKSwKICAgIDBweCAxNnB4IDZweCBoc2xhKDAsIDAlLCAwJSwgMC4wMyksIDBweCA5cHggNXB4IGhzbGEoMjIzLCAxMyUsIDEwJSwgMC4wOCksCiAgICAwcHggNHB4IDRweCBoc2xhKDAsIDAlLCAwJSwgMC4xNiksIDBweCA0cHggMnB4IGhzbGEoMCwgMCUsIDAlLCAwLjA0KSAhaW1wb3J0YW50Owp9CgoucHJpbnRcOmhpZGRlbiB7CiAgZGlzcGxheTogbm9uZSAhaW1wb3J0YW50Owp9Ci5wcmludFw6ZmxleCB7CiAgZGlzcGxheTogZmxleCAhaW1wb3J0YW50Owp9Ci5wcmludFw6YmxvY2sgewogIGRpc3BsYXk6IGJsb2NrICFpbXBvcnRhbnQ7Cn0KCi8qIGluZGV4LmFzdHJvICovCm1haW4gewogIHBhZGRpbmctYm90dG9tOiAwICFpbXBvcnRhbnQ7Cn0KbWFpbiA+IC5jb250ZW50LXBhbmVsIHsKICBwYWRkaW5nLWJsb2NrLXN0YXJ0OiAwICFpbXBvcnRhbnQ7Cn0KCi8qIGNvbXBvbmVudHMvTWFya2Rvd25Db250ZW50LmFzdHJvICovCi5jb250ZW50IHByZSB7CiAgb3ZlcmZsb3cteDogaGlkZGVuICFpbXBvcnRhbnQ7CiAgd2hpdGUtc3BhY2U6IHByZS13cmFwICFpbXBvcnRhbnQ7CiAgcGFnZS1icmVhay1pbnNpZGU6IGF2b2lkOwp9CgovKiBsYXlvdXQvUGFnZUZyYW1lLmFzdHJvICovCi5wYWdlID4gaGVhZGVyIHsKICBwb3NpdGlvbjogcmVsYXRpdmUgIWltcG9ydGFudDsKfQoucGFnZSA+IC5tYWluLWZyYW1lIHsKICBwYWRkaW5nLWlubGluZS1zdGFydDogMCAhaW1wb3J0YW50Owp9CgovKiBsYXlvdXQvVHdvQ29sdW1uQ29udGVudC5hc3RybyAqLwoubWFpbi1wYW5lIHsKICAtLXNsLXNpZGViYXItd2lkdGg6IDBweCAhaW1wb3J0YW50OwogIC0tc2wtY29udGVudC13aWR0aDogMTAwJSAhaW1wb3J0YW50Owp9Cg==)" media="print">
```

Let me check if there’s a workaround for that.

tony-sull:
Best laid plans!  It turns out there isn't currently a great way to get around the CSS being inlined there, I'm going to close this out for now as blocked.

@delucis, @bluwy, and I were catching up on Discord and it sounds like this may be worthy of a feature in Vite to be able to `?url` import and make sure it's never inlined.  My vote is still on `?no-really-i-want-a-url`, but I'll leave that up to the Vite team to decide 😄

packages/starlight/index.astro
delucis: A feature to fix this just merged in Vite!

https://github.com/vitejs/vite/pull/15454

Looks like this might work once that is released:

```suggestion
import printHref from './style/print.css?no-inline';
```

bluwy: It should be `?url&no-inline` to mark the CSS as an asset URL first, and then force not inline it as a data URL.

delucis:
Will need to wait on https://github.com/withastro/astro/pull/12758 to use the `?url&no-inline` import query without type errors.

delucis:
A type check is failing while we wait on #2736, but I marking this ready for review nonetheless, as I think everything else is ready.

Points to consider:

- How do the styles look when printing the deploy preview? Are there any components or contexts that are not working as desired?
- This PR uses `print:*` utility classes on some elements to control visibility on print. Are we happy with that? It’s the easiest way to do this I think (and allows users to even use these utilities in their custom components), but does mean print styles impact HTML a little.
- Do we think the styles are simple and flexible enough to be compatible with customizations to user sites?

packages/starlight/style/print.css
```diff
@@ -0,0 +1,140 @@
+@media print {
+	/**
+	This is increased in specificity so it works even when injected into pages before regular styles.
+	TODO: Just use `:root` once we move to @layer.
+	Same goes for the use of `!important` in this file.
+	*/
+	:root[data-theme]:is(:root) {
+		/* Colours (light mode) */
```
HiDeoo, line 8:
```suggestion
```

I'm assuming a comment left from copying the block of CSS?

HiDeoo (commented):
Wow, such an amazing work and so hyped to finally see this PR being ready for review after so long! 🎉

Tested across multiple browsers, devices, etc. and most of the work is amazing! I also tested a bit on iOS, which seems fine, but it's super difficult to test without actually printing as the print preview is super small so I mostly focused on iPadOS which seems a more realistic use case to print from a tablet vs printing from a phone.

Here are a few questions and feedbacks:

- Should a banner text color be `--sl-color-white`?
- Code block terminal frame looks weird imo with that empty window title bar, should we hide it or maybe make sure the 3 buttons are a bit visible (note that they are visible on Safari iOS and iPadOS only (not Safari desktop))?

  SCR-20250109-qsgd

- Code block with a file name are missing a bottom bottom border below the tab bar (which is visible in terminal code blocks). I think it could be helpful if visible and also for consistency.
  
  SCR-20250109-qtaa

- I would probably remove the tab top border in the code block with a file name?
  
  SCR-20250109-qtej

- Steps vertical line indicator is invisible.
  
  SCR-20250109-qtkp

- Badges text is a bit difficult to read imo.

  SCR-20250109-qtts

- Should `blockquote` get an `break-inside: avoid;`?
- Same question for lists?
- Same question for tables?
- Same question for preformatted text and figures?
- Card grids may need a tweak regarding the breakpoint used as it seems a 1 column layout is always used?
- The Copy button in code blocks should be hidden (if printing from a tablet, it would be visible and not useful).

Regarding some of the questions:

> This PR uses print:* utility classes on some elements to control visibility on print. Are we happy with that? It’s the easiest way to do this I think (and allows users to even use these utilities in their custom components), but does mean print styles impact HTML a little.

I personally like the approach even more as a plugin author :thumbsup:

> Do we think the styles are simple and flexible enough to be compatible with customizations to user sites?

I think they are, tested on a few different projects and I think the choice are flexible enough and thoughtfully done to not be too opinionated but just obvious enough when it comes to printing.

packages/starlight/style/print.css
delucis: Kind of but I think it was left here semi-intentionally. Probably a clearer use of this comment would be this:
```suggestion
		/* Colours — force light mode when printing */
```

delucis:
Thank you for the thorough testing and thoughtful comments! Lots of good stuff spotted and I think I’ve addressed almost all of this in the latest commits. Here are my notes one by one. Most of these I implemented and the couple I didn’t I left some comments.
 
> Should a banner text color be `--sl-color-white`?

Yes, probably and also the banner background reset — even though generally browsers do this automatically in my testing, setting it explicitly seems safest. Done in 1342f866.

> Code block terminal frame looks weird imo with that empty window title bar, should we hide it or maybe make sure the 3 buttons are a bit visible (note that they are visible on Safari iOS and iPadOS only (not Safari desktop))?

Yeah, I also considered hiding it, but then realised that there can be a title on that bar potentially, so it’s not really safe to always hide it. Your comment made me think and I realised I could force the dots to appear in print styles by applying an inset box-shadow instead of a background. That way the SVG mask still works. Done in 1a36d9a8.

> - Code block with a file name are missing a bottom bottom border below the tab bar (which is visible in terminal code blocks).
> - I would probably remove the tab top border in the code block with a file name?

Done in 20e47c91. I’m a little hesitant about these styles as they get into EC’s internal class names a bit. May be worth eventually chatting with @hippotastic about upstreaming some print styles to EC.

Here’s an example of a print preview with these styles:

code block with a file name of src/content.config.ts

> Steps vertical line indicator is invisible.

Fixed in bb2c313d using the same box-shadow technique as the terminal frame dots.

> Badges text is a bit difficult to read imo.

Agreed! Fixed in cd76d017 to use a more appropriate text color:

image

> - Should `blockquote` get an `break-inside: avoid;`?
> - Same question for lists?
> - Same question for tables?
> - Same question for preformatted text and figures?

Expressive Code blocks do already have this set, as do the file tree, asides, and tab components. On the whole when I played with it, I felt like you don’t want to _over_ do `break-inside` as you can end up with lots of unnatural breaks with large amounts of whitespace, so felt it need to be done with some balance, or where cases were essential (code blocks for example because without it, often the title would end up on one page and the code on the next).

To answer the elements one by one:

- `<blockquote>` — I think this should mostly be covered by the widows/orphans rule applied to paragraphs and list items. Markdown renders a `<p>` inside blockquotes, so shorter quotes should be kept together, and only when it’s safe (i.e. ≥4 lines) would they break, which feels a reasonable compromise between cohesion and avoiding excess whitespace.

- `<ul>`/`<ol>` — These can get pretty long (especially if you imagine lists like we use with `<Steps>` so I don’t think this is essential. May occasionally look odd for really short groups of bullet points, but at least in our docs I didn’t spot really bad things without this.

- `<table>` — Browsers seem to handle these pretty elegantly, even repeating table headers, so I think it’s probably OK to let these wrap. Example:
   
   two document pages with a table running across the page break. on the second page, the table header is repeated for clarity

- `<pre>`/`<figure>` — I think you’re probably right here, certainly figures where the caption and image could easily split across pages. Added styles for these in 104e853a. 

> Card grids may need a tweak regarding the breakpoint used as it seems a 1 column layout is always used?

I _think_ this is expected because of the page width? Do you think it’s problematic to have these stack like on smaller viewports in print?

> The Copy button in code blocks should be hidden (if printing from a tablet, it would be visible and not useful).

Ah, good point couldn’t notice this testing only on a laptop. Added a style that I _assume_ should fix this in 43ed243b, but couldn’t reproduce on my computer — will need to double check on a phone later.

hippotastic:
I'd gladly move any code snippet-related print styles into EC! Just let me know when you've reached a point where you are happy with the code block print styles, and don't worry about getting into EC's territory. I did not think about print styles yet and am thankful for the input.

HiDeoo:
Thanks for all the great explanations. As most of the print stylesheets I've done in the past was me mostly applying suggestions from various sources and I never really print documents myself, my comment included a lot of questions and I appreciate the reasoning behind the decisions.

Just tested all the points I reported on various devices and can confirm that your latest changes fixed all of them.

I guess this also a topic where now that we have an initial version, we can iterate on it as we get more feedback from users that actually print documents.

> even though generally browsers do this automatically in my testing, setting it explicitly seems safest.

Safari on iOS (not iPadOS) seems to be inconsistent in this regard so definitely worth setting it explicitly :thumbsup:

> I _think_ this is expected because of the page width? Do you think it’s problematic to have these stack like on smaller viewports in print?

I'm not really sure what is expected. I noticed this when checking the print differences between https://deploy-preview-157--astro-starlight.netlify.app/resources/showcase/ which prints as I would personally expect vs https://deploy-preview-157--astro-starlight.netlify.app/resources/plugins/ so my initial thought was "Oh, we may need to combine some print media query with a size query to make it better" but without strong opinions on what is expected, I'm not really sure.

delucis:
Thanks for re-testing the changes! 💖 

> now that we have an initial version, we can iterate on it as we get more feedback

Definitely. I'm in a similar situation and although we've had some requests for print styles, I don't think they've included much detail so far. Will be happy to hear what people think.

> my initial thought was "Oh, we may need to combine some print media query with a size query to make it better" but without strong opinions on what is expected, I'm not really sure.

Yeah, I'm not sure. I think it's more a coincidence that those two grid breakpoints fall either side of the page width. Given it's a bit of a pain to adjust media queries (usually needs a bunch of style duplication), I'd probably lean towards accepting it as it is, I think.

HiDeoo (approved):
Forgot to approve with my last comment. Amazing work 👏

virzak:
This PR is absolutely awesome!

PR #158, opened 2023-06-05
add word break to search ui

kevinzunigacuellar:

- Changes to Starlight code

#### Description

- Fixes the galaxy fold test 🚀

### DIFF
--- a/.changeset/swift-files-retire.md
+++ b/.changeset/swift-files-retire.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Improve word wrapping for optimal display on smaller mobile screens
--- a/.changeset/swift-files-retire.md
+++ b/.changeset/swift-files-retire.md
@@ -2,4 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
-Improve word wrapping for optimal display on smaller mobile screens
+Fix word wrapping in search modal on narrow screens
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -294,6 +294,7 @@ const t = useTranslations(Astro.props.locale);
 
   #starlight__search .pagefind-ui__result-excerpt {
     font-size: var(--sl-text-body-sm);
+    word-break: break-all;
   }
 
   #starlight__search mark {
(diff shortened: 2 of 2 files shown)


packages/starlight/components/Search.astro
```diff
@@ -294,6 +294,7 @@ const t = useTranslations(Astro.props.locale);
 
   #starlight__search .pagefind-ui__result-excerpt {
     font-size: var(--sl-text-body-sm);
+    word-break: break-all;
```
delucis, line 297: Thanks for catching this Kevin!

I think we might prefer `break-word` here (always enjoyed the weird palindrome of this property–value pair 😅). Basically that lets word breaks _when necessary_ but not always:

```suggestion
    word-break: break-word;
```

| `break-all` | `break-word` |
|--------|--------|
| image | image |

kevinzunigacuellar: I see. I found that `word-break: break-word;` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values) and an alternative would be:

```css
word-break: normal;
overflow-wrap: anywhere;
```

It seems to work as well
image

I am okay with both. I leave the final call to you ⭐

kevinzunigacuellar:
```suggestion
    word-break: normal;
    overflow-wrap: anywhere;
```

delucis: TIL!

I guess in this case we don't need to explicitly set `word-break: normal` as that's the default.

Also reading those CSS docs I think behaviour is the same in this context but it might be best to choose `break-word` instead of `anywhere` because of [how it impacts implicit content width calculation](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#values)
```suggestion
    overflow-wrap: break-word;
```

.changeset/swift-files-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Improve word wrapping for optimal display on smaller mobile screens
```
delucis, line 5: Small tweak to explain the search modal context for people reading the changelog.
```suggestion
Fix word wrapping in search modal on narrow screens
```

packages/starlight/components/Search.astro
```diff
@@ -294,6 +294,7 @@ const t = useTranslations(Astro.props.locale);
 
   #starlight__search .pagefind-ui__result-excerpt {
     font-size: var(--sl-text-body-sm);
+    overflow-wrap: break-word;
```
delucis, line 297: Deprecated but it works?

```suggestion
    word-break: break-word;
```

delucis (approved):
Thanks for figuring this one out @kevinzunigacuellar 🙌

PR #167, opened 2023-06-07
Add base to Pagefind bundlePath

BryceRussell:

- Changes to Starlight code

#### Description

Currently if a `base` option has been set the search modal throws errors that it can't find `/_pagefind/pagefind.js`

image
(Base option: `/testing`)

This PR adds a [`bundlePath`](https://pagefind.app/docs/search-config/#bundle-path) to the Pagefind config that includes the `base` option from Astro to fix this issue.

### DIFF
--- a/.changeset/violet-bags-carry.md
+++ b/.changeset/violet-bags-carry.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Add `bundlePath` option to Pagefind configuration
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks @BryceRussell! Weird that `baseUrl` isn’t enough here — we’re essentially setting the same thing twice. Confirmed this is a bug though, so appreciate you fixing this 🙌

PR #168, opened 2023-06-07
Fix bottom padding on left sidebar

BryceRussell:

- Changes to Starlight code

#### Description

**Before**:

image

**After**:

image

### DIFF
--- a/packages/starlight/components/Sidebar.astro
+++ b/packages/starlight/components/Sidebar.astro
@@ -28,6 +28,10 @@ const { sidebar, locale } = Astro.props;
     gap: 1rem;
   }
 
+  .sidebar > :global(ul) {
+    padding-bottom: 1rem;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/starlight/components/Sidebar.astro
```diff
@@ -28,6 +28,10 @@ const { sidebar, locale } = Astro.props;
     gap: 1rem;
   }
 
+  .sidebar > :global(ul) {
```
delucis, line 31: Thanks for this! I think we want to target the sidebar’s last child for this: a) it means we don’t add a space between the links and the theme/language pickers on mobile and b) it means if we ever support adding some other content to the bottom of the menu, this will also handle that.

```suggestion
  .sidebar > :global(:last-child) {
```

delucis (approved):
Thanks for spotting and solving this, Bryce! 🙌

PR #177, opened 2023-06-08
Handle table overflow

### LINKED ISSUE #176: Tables may cause horizontal overflow

0.1.2

2.5.0

npm

Mac

Chrome

### Describe the Bug

Wide tables create horizontal overflow where the entire body is widened, causing content like the search/nav buttons to be rendered out of the viewport.

For example, https://har.fyi/reference/tables/pages/ with mobile emulation:

image

image

The offending table is under the [Schema](https://har.fyi/reference/tables/pages/#schema) heading.

FWIW I've also seen this affect the desktop layout for a really wide table (not used in prod anymore)

_No response_

### Participation

rviscomi:

- Style fixes

#### Description

- Closes #176
- What does this PR change? Adds styles to allow table elements to hide their overflow without pushing out the body

Before:

image

After:

image

### DIFF
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -95,6 +95,8 @@
   }
 
   .content :global(table) {
+    display: block;
+    overflow: auto;
     border: 1px solid var(--sl-color-gray-5);
     border-collapse: collapse;
   }
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -96,6 +96,9 @@
 
   .content :global(table) {
     display: block;
+    width: 100%;
+    width: max-content;
+    max-width: 100%;
     overflow: auto;
     border: 1px solid var(--sl-color-gray-5);
     border-collapse: collapse;
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -96,11 +96,7 @@
 
   .content :global(table) {
     display: block;
-    width: 100%;
-    width: max-content;
-    max-width: 100%;
     overflow: auto;
-    border: 1px solid var(--sl-color-gray-5);
     border-collapse: collapse;
   }
   .content :global(tr:nth-child(2n)) {


rviscomi:
Borrowed this approach from GitHub's CSS:

image

Tested only in Chrome. I'm wary of the effects of setting `display: block` on tables in other browsers. Would appreciate a closer look.

TheOtterlord (approved):
Works in Firefox :+1:

rviscomi:
A safer option may be to wrap the table in a `div` with `overflow: auto`.

delucis:
Yeah, definitely. Requires reaching in with a remark plug-in to do that though. So if we can confirm this is a safe change, I'd be happy to avoid that extra complexity. Definitely doable though!

Yan-Thomas:
Tested with NVDA (Windows 11 + Google Chrome), table works as expected! Heads up that with `display: block` the table fits the container size according to the border.
image

delucis:
Awesome, thanks for testing @Yan-Thomas! OK, I’m feeling reasonably confident this is safe — if we hear it isn’t from users, we can explore the alternative solutions.

That means we just need to tidy up the table border styling and should be good to go. @rviscomi want to take a look at that or would you like me to?

tony-sull:
Everything looks good on my end as well 🎉 

I only have access to VoiceOver for testing, but it was giving all the same context related to column & row count, column headers spanning multiple columns, etc.  Firefox's accessibility tree looks untouched as well, all the context I'd expect related to columns/headers/layout all look to be there.

🤞 I think this *may* be tentative confirmation that changing `display` on a `<table>` is no longer an accessibility concern in modern browsers!

rviscomi:
> That means we just need to tidy up the table border styling and should be good to go. @rviscomi want to take a look at that or would you like me to?

Should be fixed by applying the [other](https://github.com/withastro/starlight/pull/177#issuecomment-1582570137) GitHub styles related to `width`. Commit incoming.

packages/starlight/components/MarkdownContent.astro
```diff
@@ -95,6 +95,11 @@
   }
 
   .content :global(table) {
+    display: block;
+    width: 100%;
+    width: max-content;
+    max-width: 100%;
+    overflow: auto;
     border: 1px solid var(--sl-color-gray-5);
```
delucis, line 103: Thanks @rviscomi!

I think this can be even simpler! If we remove the border from the table, we get the effect we want:

```suggestion
    display: block;
    overflow: auto;
```

This is also maybe the preferable fix because setting `display: block` disabled collapsing the table’s border with cell borders so we’re seeing a double border currently. This branch on the left, live site on the right:

Screenshot of two tables, one with a 2 pixel border, one with 1 pixel as desired

rviscomi:
Ah ok I see what you mean. PTAL

delucis (approved):
Thanks for handling the different feedback @rviscomi and thanks everyone for jumping in and testing! 💖

PR #178, opened 2023-06-08
Add support for translating the Pagefind search modal

delucis:

- Changes to Starlight code

#### Description

- Adds Pagefind’s translation keys to the i18n schema and passes them to Pagefind UI if provided — adding support for languages that Pagefind doesn’t provide out of the box.
- Aliases our existing `search.label` string to Pagefind’s `placeholder` string, which fixes the input placeholder not being translated on non-English sites.
- Adds docs on the above to the i18n guide.

### DIFF
--- a/.changeset/funny-weeks-cry.md
+++ b/.changeset/funny-weeks-cry.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": minor
+---
+
+Add support for translating the Pagefind search modal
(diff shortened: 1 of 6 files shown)


Yan-Thomas (approved):
Tested locally and everything worked just fine! Smart work getting the values inserted by Pagefind working well with the translation system 👌 

I'll point out that I found a string that isn't translated (open search in `docs/` while in dev mode). It's intended to show only for developers so I understand not translating it, just to be sure you know about it in case you want to have even these translated.

image

delucis:
Thanks for calling that out @Yan-Thomas! My thinking was that given it’s only a dev notice it wasn’t worth translating, but I guess that may be making assumptions about our users. Don’t think we need to block this PR for this, but maybe you’re right that we should add this as an option. If you want to wire that up in a separate PR, I’d be happy to merge!

PR #187, opened 2023-06-09
pagefind: ignore inner file tree text

kevinzunigacuellar:

#### Description

- Ignore file tree inner text

| Before                                                                                                                          | After                                                                                                                           |
|---------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
| image | image |

### DIFF
--- a/docs/src/components/file-tree.astro
+++ b/docs/src/components/file-tree.astro
@@ -17,7 +17,7 @@ const processedContent = await fileTreeProcessor.process({
 });
 ---
 
-<file-tree set:html={processedContent} />
+<file-tree set:html={processedContent} data-pagefind-ignore="index" />
 
 <style is:global>
   file-tree {


dreyfus92 (approved):
LGTM ✅

docs/src/components/file-tree.astro
```diff
@@ -17,7 +17,7 @@ const processedContent = await fileTreeProcessor.process({
 });
 ---
 
-<file-tree set:html={processedContent} />
+<file-tree set:html={processedContent} data-pagefind-ignore="index" />
```
delucis, line 20: Looks like we can leave out the attribute value here according to the docs: https://pagefind.app/docs/indexing/#removing-individual-elements-from-the-index

```suggestion
<file-tree set:html={processedContent} data-pagefind-ignore />
```

delucis (commented):
@Yan-Thomas convinced me this is the right thing to do!

One small suggestion.

delucis (approved):
Thanks @kevinzunigacuellar!

PR #189, opened 2023-06-10
Use `sharp` for image processing

### LINKED ISSUE #186: webp animation not works in assets

0.1.2

2.6.2

pnpm

Linux

Chrome

### Describe the Bug

When whe try to compile an image in a .md file, the WEBP animation image does not work propely and crash the dev server with this error:
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Cannot read properties of null (reading 'data')".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

https://codesandbox.io/p/sandbox/github/PabloBagliere/example-error-astro/tree/main

### Participation

delucis:

- Changes to Starlight code

#### Description

- ~~Adds `sharp` to Starlight’s dependencies and configures Astro to use it to process images~~
- Adds `sharp` to Starlight’s docs & the starter template for image processing
  - Because of how `sharp` is loaded internally in Astro, it _must_ be a direct dependency of the Astro site and can’t be bundled with Starlight. This change means new Starlight sites will have `sharp` enabled by default.
- Closes #186 (`sharp` handles animated WebP correctly)

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,5 +9,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["starlight-docs"]
+  "ignore": ["starlight-docs", "@example/*"]
 }
--- a/.changeset/slow-elephants-tan.md
+++ b/.changeset/slow-elephants-tan.md
@@ -1,5 +1,0 @@
----
-"@astrojs/starlight": minor
----
-
-Faster image optimisation: switches to use `sharp` for image processing instead of `squoosh`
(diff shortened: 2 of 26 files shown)


delucis:
Ack, looks like Sharp _must_ be a direct dependency of a project and maybe can’t be built into Starlight currently? 🤔 Curious why Astro can only load from the root, but will need some investigation.

opswiz:
Sharp is the default image processor in Astro v3. Is it safe to remove the image entrypoint config?

delucis:
> Sharp is the default image processor in Astro v3. Is it safe to remove the image entrypoint config?

Yes! Good point. Didn’t mention that in our changelog for or remember to remove it in our `examples/`. Would you be up for a PR removing Sharp in our example templates?

opswiz:
Sure.

PR #193, opened 2023-06-12
Fix bottom padding for sidebar on larger screens

BryceRussell:

- Changes to Starlight code

#### Description

This PR https://github.com/withastro/starlight/pull/168 added `padding-bottom` to the `:last-child` in the sidebar, but the last child is the `mobile-preferences` element which is hidden on larger screen sizes making the padding only work in the mobile view.

This PR adds a `padding-bottom` to the second to last element on larger screen sizes so that the padding works for both mobile and non-mobile views.

**Before**:

image

**After**: 

image

### DIFF
--- a/.changeset/thirty-emus-check.md
+++ b/.changeset/thirty-emus-check.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix bottom padding for sidebar on larger screen sizes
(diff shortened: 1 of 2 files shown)


BryceRussell:
Should I remove this code from the previous PR? 

```css
  .sidebar > :global(:last-child) {
    padding-bottom: 1rem;
  }
```

This only works in the mobile view and the last element (`mobile-preferences`) already has its own padding, it would only makes sense to keep this if you could slot content under `mobile-preferences` in the future.

delucis (commented):
Note to self: trust @BryceRussell's initial PR and don't try to be clever next time 😅

Yes, you're right in mobile this was never essential so should be safe to remove that code as you suggest. Thank you!

delucis (approved):
Thanks for the fix @BryceRussell!

PR #195, opened 2023-06-13
Automatically bump example deps versions on release

bluwy:

- Changes to Starlight code

#### Description

- Make `examples/*` part of pnpm workspace
- Automatically bump example deps versions

This is super strange, but it turns out changesets would auto bump the example deps by itself? I tested it with `pnpm changeset version` locally. I think something must be up in the Astro core repo that's preventing this.

But anyways this PR should be good to go.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,5 +9,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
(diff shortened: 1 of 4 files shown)


delucis (approved):
Thank you for this! Also means I can run examples against the local dev version of Starlight for testing which is super helpful.

PR #196, opened 2023-06-13
Install dependencies after `changeset version` runs

delucis:

- Something else!

#### Description

#195 made Changesets automatically bump example versions as part of release PRs, but looks like builds fail in the release PR now (e.g. #194) because the lockfile is out of date (Changesets bumped `package.json` in examples but _not_ the lock file).

This PR runs `pnpm i` after `changeset version` in the GitHub action, which I think should fix this?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,7 @@ jobs:
       - name: Create Release Pull Request
         uses: changesets/action@v1
         with:
+          version: pnpm changeset version && pnpm i
           publish: pnpm changeset publish
           commit: '[ci] release'
           title: '[ci] release'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - name: Create Release Pull Request
         uses: changesets/action@v1
         with:
+          version: pnpm changeset version && pnpm i
```
bluwy, line 35: Ah my bad 🙈 I think we also need to add `--no-frozen-lockfile` so it can be updated in CI (Astro core has the same config)
```suggestion 
          version: pnpm changeset version && pnpm i --no-frozen-lockfile
```

bluwy (approved):

Issue #180, opened 2023-06-08
Setting Astro `build.format` config breaks sidebar links


jamescmartinez:
### What version of `starlight` are you using?

0.1.2

### What version of `astro` are you using?

2.6.1

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

**Expected**: sidebar links should not have a trailing slash when setting [Astro's build.format config](https://docs.astro.build/en/reference/configuration-reference/#buildformat) to `file`

**Actual**: they have a trailing slash, which causes the links to 404 in production builds

**How to reproduce**:

1. Clone the `build-format-file` branch of [this repo](https://github.com/mergentlabs/docs)
2. Comment out [line 8 of the Astro config](https://github.com/mergentlabs/docs/blob/main/astro.config.mjs#L8) that sets the base path to `/docs` (sorry, it's a temporary workaround for GitHub pages)
3. `npm run dev` to see the trailing `/` OR `npm run build && npm run preview` to get the 404s
4. Click any link in the sidebar
5. Smile at everyone's favorite sad piece of paper :)
<img width="135" alt="Screenshot 2023-06-08 at 5 40 36 PM" src="https://github.com/withastro/starlight/assets/1286597/97bfc892-e583-45a0-84be-f788a162f198">

### Link to Minimal Reproducible Example

https://github.com/mergentlabs/docs/tree/build-format-file

### Participation

- [X] I am willing to submit a pull request for this issue.

delucis:
Thanks for the report @jamescmartinez! I can confirm we’re not currently respecting `build.format` or [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) when building links. Should be do-able though and would welcome a PR!

Our general method for making user configuration from `astro.config` available to components etc. is via virtual modules and you can see that in place here:

https://github.com/withastro/starlight/blob/7f4e948ee20a48d79ce6297e2aa7be923164ff6d/packages/starlight/index.ts#L85-L94

We currently expose the project `root` via the `virtual:starlight/project-context` virtual module, but could add `trailingSlash` and `build.format` to that module so that Starlight code can do something like:
```js
import project from 'virtual:starlight/project-context';

const link = project.build.format === 'file' ? '/foo.html' : '/foo/';
```

jamescmartinez:
I didn't have time to wrap this up tonight, but here's a start: https://github.com/withastro/starlight/pull/207

[...]

PR #207, opened 2023-06-15
[WIP] Respect Astro `build.format` when building URLs

jamescmartinez:
🚧 WIP for #180 

This commit makes Starlight respect Astro's `build.format` when building URLs.

This is considered WIP because there is still more work to do with the routes; we should be seeing `https://localhost:3000/` without the trailing slash, but it's still there.


delucis:
Thanks for tackling this @jamescmartinez! Let me know when you want me to take a look.

delucis:
Thanks again for the PR @jamescmartinez! Feel free to re-open if you have time to tackle this, but going to close as stale for now. No worries if you don’t have time, hopefully someone else can fix #180 soon!

PR #210, opened 2023-06-15
Use `_top` instead of `starlight__overview` as page title ID

delucis:

- Changes to Starlight code

#### Description

Updates the ID used on the page title to be `_top` instead of `starlight__overview`.

Got some feedback from @joshpollara that it’s a bit ugly for the URL to have `#starlight__overview` in when clicking back to the overview, which makes sense. Another option would be to make this customisable, but given it touches a few different spots, a more minimal ID seems a good quick compromise for now.

(For context, the original idea behind `starlight__overview` was to avoid conflicts with actual page headings. Using `_top` with an underscore here should hopefully achieve the same.)

### DIFF
--- a/.changeset/mighty-wolves-jog.md
+++ b/.changeset/mighty-wolves-jog.md
@@ -2,6 +2,6 @@
 '@astrojs/starlight': minor
 ---
 
-Change page title ID to `top` for cleaner hash URLs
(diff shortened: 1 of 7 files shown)


TheOtterlord (approved):
looks good!

PR #212, opened 2023-06-15
i18n(ja): Add showcase.md

morinokami:

#### Description

- Add showcase.md.


docs/astro.config.mjs
```diff
@@ -84,12 +84,12 @@ export default defineConfig({
...
-              //   es: '',
-              //   ja: '',
-              //   fr: '',
-              //   it: '',
-              // },
+              translations: {
+                es: '',
+                ja: 'ショーケース',
+                fr: '',
+                it: '',
```
delucis, line 91: This works because the empty string is being ignored, but I might suggest commenting these out just to be explicit:

```suggestion
                // de: '',
                // es: '',
                ja: 'ショーケース',
                // fr: '',
                // it: '',
```

delucis (commented):
Thanks for adding this @morinokami! Just made one small suggestion.

morinokami:
@delucis Thanks for your suggestion! I have added another placeholder for German with [this commit](https://github.com/withastro/starlight/pull/212/commits/0d1f8a3e36a4d6c5f01a1662b814858ddae213e1) as a reminder.

PR #215, opened 2023-06-15
Add translation contribution docs

delucis:

- Something else!

#### Description

Updates our contributing guide with more details on how to contribute translation work to Starlight

### DIFF
--- a/packages/starlight/translations/README.md
+++ b/packages/starlight/translations/README.md
@@ -1,0 +1,14 @@
+# Starlight UI translation files
+
+This directory contains translation data for Starlight’s UI.
+Each language has its own JSON file and follows the [translation structure described in Starlight’s docs](https://starlight.astro.build/guides/i18n/#translate-starlights-ui).
+
+## Add a new language
+
+1. Create a JSON file named using the BCP-47 tag for the language, e.g. `en.json` or `ja.json`.
+
+2. Fill the file with translations for each UI string. You can base your translations on [`en.json`](./en.json).
+
+3. Add your file in [`index.ts`](./index.ts).
+
+4. Open a pull request on GitHub to add your file to Starlight!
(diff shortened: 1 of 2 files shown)


packages/starlight/translations/README.md
```diff
@@ -0,0 +1,14 @@
...
+This directory contains translation data for Starlight’s UI.
+Each language has its own JSON file and follows the [translation structure described in Starlight’s docs](https://starlight.astro.build/guides/i18n/#translate-starlights-ui).
+
+## Add a new language
+
+1. Create a JSON file named using the BCP-47 tag for the language, e.g. `en.json` or `ja.json`.
+
+2. Fill the file with translations for each UI string. You can base your translations on [`en.json`](./en.json).
+
+3. Add your file in [`index.ts`](./index.ts).
```
sarah11918, line 12: This felt a *little* light maybe? Given that an addition is required in two different places, it might be safe to spell out that two changes are needed.

```suggestion
3. Import your file in [`index.ts`](./index.ts) and add your language to the `Object.entries`.
```
(Not sure exactly how to be "correct" without being way too long here re: "add to this")

sarah11918 (approved):
Looks great, Chris! Just added a couple of thoughts for your consideration!

packages/starlight/translations/README.md
delucis: You spotted why I opted for the “add” cop-out ;-)

I actually think you text is pretty good! I wasn’t sure how important detail was here because I mainly care about getting a PR with the translations. If we have to fix up the plumbing, that’s OK and easy to do. But I do like this wording, so will update.

Thanks Sarah!

Yan-Thomas (approved):
LGTM! Great work 🚀

PR #216, opened 2023-06-15
Encode heading id when finding current link

morinokami:

- Changes to Starlight code

#### Description

- Currently, highlighting in the right sidebar is partially not working for non-English languages (e.g. https://starlight.astro.build/es/getting-started/#agregar-p%C3%A1ginas, https://starlight.astro.build/ja/environmental-impact/#%E7%94%BB%E5%83%8F). This change will make the highlighting work properly.
  - The reason why the highlighting is not working correctly is that the `link.hash` in `starlight-toc.ts` is URL encoded, whereas the `heading.id` is not. As a result, the comparison between the two strings is not accurate sometimes. To address this issue, we can use the `encodeURIComponent` function on `heading.id` before performing the comparison. This ensures that both strings are encoded in the same way, allowing for proper comparison and accurate highlighting.

### DIFF
--- a/.changeset/orange-moose-impress.md
+++ b/.changeset/orange-moose-impress.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Encode heading id when finding current link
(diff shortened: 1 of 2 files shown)


delucis (approved):
Ah, great fix @morinokami! I’d noticed this on a German heading containing `ü` a while back but forgotten about it. Looks like the fix is working perfectly — thank you 💜

PR #223, opened 2023-06-16
Fix labeler action config

TheOtterlord:

- Something else!

#### Description

- What does this PR change? Fixes the action configured paths to allow it to label things with `i18n`


Waxer59 (approved):
LGTM

PR #226, opened 2023-06-17
Support custom 404 pages

delucis:
#### What kind of changes does this PR include?

- Changes to Starlight code

#### Description

- Adds support for custom 404 pages. Users can create a `404.md` or `404.mdx` file in `src/content/docs/` to display content of their choice on the 404 page.

PR #227, opened 2023-06-17
Add i18n to Search component dev warning

Yan-Thomas:

- Changes to Starlight code

#### Description

This PR adds a new i18n label targetting the dev warning for the Search component (plus `pt` Pagefind translations and docs update).

To keep the line break between paragraphs without making it two separate labels, I added `white-space: pre-line;` to not remove the newline character from the source string. We can change this in case it might raise any issues.

Note that it was necessary to add the key/value to the other languages for the schema to pass correctly, perhaps we might want to change this to avoid adding new labels to be a breaking change (I'm not adding a changeset yet because of this).


afonsojramos (approved):

packages/starlight/translations/pt.json
```diff
@@ -16,5 +17,15 @@
...
+  "pagefind.clear_search": "Limpar",
+  "pagefind.load_more": "Carregar mais resultados",
+  "pagefind.search_label": "Pesquisar neste site",
+  "pagefind.filters_label": "Filtros",
+  "pagefind.zero_results": "Sem resultados para [SEARCH_TERM]",
+  "pagefind.many_results": "[COUNT] resultados para [SEARCH_TERM]",
+  "pagefind.one_result": "[COUNT] resultado para [SEARCH_TERM]",
+  "pagefind.alt_search": "Sem resultados para [SEARCH_TERM]. Mostrando resultados para [DIFFERENT_TERM] no lugar",
+  "pagefind.search_suggestion": "Sem resultados para [SEARCH_TERM]. Tente uma das pesquisas a seguir:",
+  "pagefind.searching": "Procurando por [SEARCH_TERM]..."
```
delucis, line 30: Do we actually need to set these here? Pagefind should provide them itself: https://github.com/CloudCannon/pagefind/blob/main/pagefind_ui/translations/pt.json

The aim of the `pagefind.` keys here is mainly for languages Pagefind doesn’t currently support, but it looks like Portuguese translations should be provided.

delucis, line 30:
```suggestion
  "page.nextLink": "Próximo"
```

packages/starlight/components/Search.astro
```diff
@@ -43,9 +43,8 @@ const pagefindTranslations = {
       </button>
       {
         import.meta.env.DEV ? (
-          <div style="margin: auto; text-align: center;" dir="ltr">
-            <p>Search is only available in production builds.</p>
-            <p>Try building and previewing the site to test it out locally.</p>
+          <div style="margin: auto; text-align: center; white-space: pre-line;" dir="ltr">
```
delucis, line 46: Given this is now using a translated string, maybe we should remove the `dir="ltr"`? It’s a actually a tricky case I haven’t really accounted for: no way to know the direction of the string that `t()` returns, e.g. on an RTL page, if it’s LTR fallback content or an RTL translation.

Here’s what was happening on RTL pages prior to adding `dir`:

Search modal showing English text with punctuation at the start of the sentences instead of at the end as expected

Not sure. I guess it might be safest to keep it for now given we don’t have any RTL translations of the warning yet.

delucis (commented):
Thanks for this @Yan-Thomas! Looks great.

> To keep the line break between paragraphs without making it two separate labels, I added `white-space: pre-line;`

I think that’s valid here. I might reevaluate if this were for some user-facing UI, but this is fine given this is dev-only. (Same reason for the inline styles — no need to ship those to end users.)

> It was necessary to add the key/value to the other languages for the schema to pass correctly, perhaps we might want to change this to avoid adding new labels to be a breaking change

Yeah, this is kind of intentional to ensure we ship full support when adding a key. It does make adding keys a bit of a lift — I spent time translating a bunch of stuff myself for #226 — but I think that’s OK and hopefully won’t be too common. Again, given this particular one is dev-only, I think it’s OK for it to be English-only for now (and NWTWWHB!)

packages/starlight/components/Search.astro
Yan-Thomas: Yeah, I guess it's safest to keep it for now. Wondering if we should build a function like `isFallbackTranslation("search.devWarning")` to handle these cases?

delucis (commented):
LGTM — just needs a changeset.

I think this can be considered a patch given it’s a) not impacting production sites and b) “fixing” a missing i18n support. Do you think that’s fair?

Yan-Thomas:
> I think this can be considered a patch given it’s a) not impacting production sites and b) “fixing” a missing i18n support. Do you think that’s fair?

Looks fair to me!

PR #237, opened 2023-06-20
Use file-system path in auto-generated sidebar item configuration

HiDeoo:

- Changes to Starlight code

#### Description

This pull request fixes an issue that I reported [on Discord](https://discord.com/channels/830184174198718474/1120359166574338058/1120359226372526100) regarding auto-generated sidebar items. The `directory` configuration property is currently expecting the slugified version of the path and not the actual path on disk.

For example, for a `test test` directory on disk, one would need to use `test-test` in the Astro configuration to auto-generate the sidebar items for this directory.

This PR changes this behavior to use the actual directory name on disk by using the [Astro Content Collection `id`](https://docs.astro.build/en/reference/api-reference/#id) which is the relative nested path which includes the file extension as we are using a collection defined as `type: 'content'`.

I did not find any tests so I tested the following scenarios manually. Let me know if you think I should add tests for this.

With the default locale:

1. Rename `docs/src/content/docs/reference` to `docs/src/content/docs/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Navigate to `http://localhost:3000/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

---

1. Rename `docs/src/content/docs/reference` to `docs/src/content/docs/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Add a new file `docs/src/content/docs/reference reference/index.md`.
1. Navigate to `http://localhost:3000/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

---

1. Rename `docs/src/content/docs/reference` to `docs/src/content/docs/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Add a new file `docs/src/content/docs/reference reference.md`.
1. Navigate to `http://localhost:3000/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

With the `fr` locale:

1. Rename `docs/src/content/docs/fr/reference` to `docs/src/content/docs/fr/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Navigate to `http://localhost:3000/fr/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

---

1. Rename `docs/src/content/docs/fr/reference` to `docs/src/content/docs/fr/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Add a new file `docs/src/content/docs/fr/reference reference/index.md`.
1. Navigate to `http://localhost:3000/fr/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

---

1. Rename `docs/src/content/docs/fr/reference` to `docs/src/content/docs/fr/reference reference`.
1. Update the reference sidebar item group in `docs/astro.config.mjs` to auto-generate from the directory `reference reference`.
1. Add a new file `docs/src/content/docs/fr/reference reference.md`.
1. Navigate to `http://localhost:3000/fr/getting-started/` and ensure the expected pages are listed in the sidebar with proper links.

I also tested including a `nested nested` directory in both the `docs/src/content/docs/reference reference` and `docs/src/content/docs/fr/reference reference` directories and the sidebar items were generated as expected.

Regarding versionning, I guess this is a breaking change if some people were relying on the current behavior and explicitely used the slugified version of the path. Altho, as semver [states](https://semver.org/#spec-item-4) that major version zero is for initial development and that anything may change at any time, I decided to mark this change as a minor version bump. Please let me know if you think this should not be the case.

### DIFF
--- a/.changeset/five-cooks-develop.md
+++ b/.changeset/five-cooks-develop.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": minor
+---
+
+Use path instead of slugified path for auto-generated sidebar item configuration
--- a/.changeset/five-cooks-develop.md
+++ b/.changeset/five-cooks-develop.md
@@ -3,3 +3,12 @@
 ---
 
 Use path instead of slugified path for auto-generated sidebar item configuration
+
+⚠️ Potentially breaking change. If your docs directory names don’t match their URLs, for example they contain whitespace like `docs/my docs/`, and you were referencing these in an `autogenerate` sidebar group as `my-docs`, update your config to reference these with the directory name instead of the slugified version:
+
+```diff
+autogenerate: {
+- directory: 'my-docs',
++ directory: 'my docs',
+}
+```
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -69,9 +69,9 @@ function groupFromAutogenerateConfig(
   const dirDocs = routes.filter(
     (doc) =>
       // Match against `foo.md` or `foo/index.md`.
-      doc.slug === localeDir ||
+      stripSupportedExtension(doc.entry.id) === localeDir ||
       // Match against `foo/anything/else.md`.
-      doc.slug.startsWith(localeDir + '/')
+      doc.entry.id.startsWith(localeDir + '/')
   );
   const tree = treeify(dirDocs, localeDir);
   return {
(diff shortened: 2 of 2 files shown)


.changeset/five-cooks-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": minor
+---
+
+Use path instead of slugified path for auto-generated sidebar item configuration
```
delucis, line 5: We can probably add a little migration note here to help anyone who does hit it (although I think we’re pretty safe — I doubt anyone’s run into this yet).

````suggestion
Use path instead of slugified path for auto-generated sidebar item configuration

⚠️ Potentially breaking change. If your docs directory names don’t match their URLs, for example they contain whitespace like `docs/my docs/`, and you were referencing these in an `autogenerate` sidebar group as `my-docs`, update your config to reference these with the directory name instead of the slugified version:

```diff
autogenerate: {
- directory: 'my-docs',
+ directory: 'my docs',
}
```
````

delucis (commented):
Thanks @HiDeoo! I think this mostly looks good — left one suggestion.

Re: testing — it’s on my to-do list to set up testing, with routing and navigation the top priority areas to test first, but haven’t got to it yet. If you’d be interested in tackling some of that, please feel free! It can be in a follow-up PR rather than delaying this PR for it though.

Agreed this is a minor. I think it’s conceptually a bug fix, but could theoretically break for someone, so minor while we’re pre-v1 makes sense. I’ve suggested adding a small migration note just in case.

HiDeoo:
Thanks for the great feedback, I just pushed changes to address the points you raised.

Regarding the extensions, wow, I didn't even know there was so many variations ^^ Altho, after testing it locally, your assumption seems to be correct and we can safely remove all of them.

For the tests, I think this would be indeed better suited in a separate PR. This would be something that I could be interested in when I get the time altho I wonder if I should not open an issue or discussion first to discuss the details and make sure we are on the same page, e.g. what kind of tests, the tools to use, etc. Feels like I should also take a closer look on Astro tests as even tho I wrote a few of them for various PRs to Astro, I didn't really pay attention at the whole setup.

delucis (approved):
Thanks!

Yes, an issue to discuss testing strategy sounds good. I kicked one off here for discussion, so I could bypass the usual “bug report” form: https://github.com/withastro/starlight/issues/240

PR #238, opened 2023-06-21
Document how to add Starlight to an existing project

ematipico:

#### Description

I noticed we don't document `astro add`, which is an easier way to add an integration to an existing project, so I added it 😃


docs/src/content/docs/getting-started.mdx
```diff
@@ -121,4 +121,10 @@ yarn upgrade @astrojs/starlight --latest
 </TabItem>
 </Tabs>
 
+Or, alternatively, you can use `astro add` directly:
+
+```shell
+astro add starlight
```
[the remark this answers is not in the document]
ematipico: > Does `astro add` _update_ packages?! 👀
> 

Yes, it does. It also updates the `astro.config.mjs` file

sarah11918:
Love that we can just `astro add` Starlight!

I guess I can play around with this to test, but what happens if you add Starlight to an existing project using content collections already? Is there anything to worry about? What if you already have a `src/content/docs/` folder? Are there any warnings or notes we should be including with these instructions?

delucis:
Ohhh, yes, now makes more sense 😅 

Right, as Sarah says, we’ll probably need to flesh this out a bit more as their are a couple extra steps needed:

- Add at least a `title` to the integration config (that’s technically the “minimum viable config”)
- Configure the `docs` collection in `src/content/config.ts`
- Add at least one content file to `src/content/docs/` (probably an `index.md` file)

And for _existing_ sites there may be extra context to help. Like, if you want Starlight to show up at `/docs` you currently need to put all your content in `src/content/docs/docs/` 😅 (hopefully we’ll improve that in the future).

We may actually want a separate page for manual set up like that?

sarah11918:
Yeah, I think a great pattern would be to have only a link on the Getting Started page, something like:

> Want to add Starlight as an integration to an existing Astro project? Follow the [manual instructions for adding Starlight](link-to-new-page).

This way, we can include all the necessary steps and caveats. (`/docs/docs/` is the new `/integrations/integrations` lol!)

docs/src/content/docs/getting-started.mdx
```diff
@@ -50,6 +50,33 @@ Try Starlight in your browser:
 [open the template on StackBlitz](https://stackblitz.com/github/withastro/starlight/tree/main/examples/basics).
 :::
 
+
+## Add starlight to an existing project
```
ElianCodes, line 54:
```suggestion
## Add Starlight to an existing project
```

ElianCodes (commented):
I just left a couple of notes, but I guess Sarah and Chris will still refine this somewhat more!

delucis:
Finally got a chance to circle back to this and add a new “Manual Setup” page!

docs/src/content/docs/manual-setup.mdx
```diff
@@ -0,0 +1,128 @@
...
+    </TabItem>
+    <TabItem label="Yarn">
+        ```sh
+        yarn astro add starlight
+        ```
+    </TabItem>
+
+</Tabs>
+
+This will install required dependencies and add Starlight to the `integrations` array in your Astro config file.
```
Yan-Thomas, line 39: Missing article
```suggestion
This will install the required dependencies and add Starlight to the `integrations` array in your Astro config file.
```

Yan-Thomas (commented):
Just two small nits, overall looks great to me!

sarah11918 (approved):
I saw Yan caught the nits! This is absolutely good enough to merge!  :raised_hands:

delucis:
Let’s go! Thanks for kicking this off @ematipico and thanks for everyone who jumped in reviewing along the way 🚀

PR #246, opened 2023-06-23
Fix various docs broken links

HiDeoo:

#### Description

This PR fixes various broken links in the documentation.

I just released an [Astro integration for Starlight](https://github.com/HiDeoo/starlight-links-validator) to validate internal links and for testing purposes I ran it against the documentation. It reported 8 broken links and this PR fixes 6 of them.

The last 2 remaining broken links are not fixable yet as they are in the deutsch version of the getting started (`docs/src/content/docs/de/getting-started.mdx`) and link to pages that do not have translations for the moment (`docs/src/content/docs/de/guides/project-structure.mdx` and `docs/src/content/docs/de/reference/frontmatter.md`). I guess a possible solution could be to link to the english version until the pages are translated but I'm not sure if it's the best solution.


delucis:
Super cool! Do you think it could make sense to integrate link checking directly into Starlight’s core?

delucis:
Yeah, that all makes sense. Extending `astro check` is an interesting idea — don’t think it’s possible currently.

Spun up an issue to discuss what an API could look like.

PR #248, opened 2023-06-23
i18n(es): Translated `404.md`

Waxer59:

#### Description

Translated `404.md`


delucis (approved):
Thanks @Waxer59! I hadn’t thought about this showing up in the i18n tracker, but makes sense!

We’ll need to also configure Netlify’s redirects to know this is the `/es/` 404 (by default they just use the root `404.html`), but we can do that later.

PR #254, opened 2023-06-26
Expose `<Icon>` component

HiDeoo:

- Changes to Starlight code

#### Description

This pull request is a proposal to open a discussion regarding the addition of an icons list to the documentation as [discussed](https://discord.com/channels/830184174198718474/1120359166574338058/1120650956271534090) on Discord. I am totally fine with going back to the drawing board if this is not the way to go.

In order to do so, various changes were made:

1. Expose the `<Icon>` component as a user-facing component usable from `@astrojs/starlight/components`. If this component is not exposed, we would need to rely on another component using icons to render the list, e.g. `<Card>`, which would not be ideal. I also personally do not see any reason to not expose this component, I even had a use case for a project where my only option was to literally copy the SVG code from the source code.
1. Add `<Icon>` documentation to the Components guide with an example.
1. Add a subsection to this new `<Icon>` documentation with a list of all available icons, including their name and a preview of the icon. This uses a new docs component `<IconList>`.
1. Update the existing `one of Starlight’s built-in icons` link in the Components guide to point to the new icons list.

Here is a preview of the new changes to the docs:

image

Remarks:

- The new docs component `<IconList>` uses a hardcoded list of icons rather than exposing `packages/starlight/components/Icons.ts` and using it to dynamically generate the list. I don't think they are a lot of use cases where a user would need to have access to this list. I also added a comment in
  `packages/starlight/components/Icons.ts` to specify that when this list is updated, the list in the docs should be updated as well. If this is not acceptable, I can change this.
- The number of columns in the icons list is dynamic to have a better responsive behavior.

Possible improvements:

I thought of some possible improvements but I wanted first to see if we were on the same page regarding the icons list and how to implement it. I also think they could come in a separate PR as I think the main priority is to have the list in the docs.

- Each icon block could be replaced by a button copying the icon name to the clipboard on click.
- Adding a search bar to filter the icons list (altho I am not sure the current number of icons justifies this).


docs/src/components/icons-list.astro
```diff
@@ -0,0 +1,71 @@
...
+      );
+    })
+  }
+</div>
+
+<style>
+  .icons-grid {
+    display: grid;
+    gap: 1rem;
+    grid-template-columns: repeat(auto-fit, minmax(175px, 1fr));
```
delucis, line 58: I think we can afford to have a narrower minimum width here (128px vs 175px). There’s one icon name that is too long and has to wrap, but I think that’s OK for having being able to show two columns on most mobile viewports.

```suggestion
    grid-template-columns: repeat(auto-fit, minmax(8rem, 1fr));
```

docs/src/components/icons-list.astro
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  .icons-grid .icon-preview {
+    align-items: center;
```
delucis, line 62: A couple of extra properties to keep things aligned when text does wrap.
```suggestion
    align-items: center;
    justify-content: center;
    text-align: center;
```

docs/src/content/docs/guides/components.mdx
delucis, line 1: Docs here look great! I half wondered if we should have some kind of icon “reference” page for the other places that can use the icon names (I think only the `actions` links in the `hero` frontmatter currently if I’m not forgetting something), but I think this looks great for now and we can link here.

packages/starlight/components/Icons.ts
```diff
@@ -1,3 +1,4 @@
+// When updating this file, please also update `docs/src/components/icons-list.astro` to keep the docs list of icons in sync.
```
delucis, line 1: We can remove this comment if importing and using the icon names directly.

```suggestion
```

delucis (commented):
This looks fantastic @HiDeoo — thank you for tackling it! Doubly helpful in exposing the component and properly documenting the icons that are available 💖 

Left some small suggestions below, but mostly looks great. In general I agree with all the decisions here.

Re: the improvements you mention, copy behaviour could be cool as a future enhancement and I agree that search is probably not needed just yet. We can always revisit as the number of icons grows.

delucis (approved):
Just realised I didn’t explicitly approve this! I think we can release tomorrow, so will merge right before so we don’t update the docs site too early. Thanks again for this 💖

PR #256, opened 2023-06-27
Add `lastUpdated` global and frontmatter options

HiDeoo:

- Changes to Starlight code

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1070481941863878697/1123159250747797596) on Discord, this pull request adds more configuration regarding the last updated date visible in the footer of each page.

- A new global `lastUpdated` option to enable/disable the feature. This option defaults to `false` as some deployment platforms performs shallow clones with no option to configure this or easy possibility to unshallow, e.g. Vercel, which could lead to inaccurate dates.
- A new frontmatter `lastUpdated` field to override the generated date. This field is optional and can be used to provide a more accurate date, e.g. when the Git history is not available.

This PR also adds docs for these new features to the configuration and frontmatter reference pages.


docs/src/content/docs/reference/configuration.md
```diff
@@ -317,3 +317,12 @@ interface HeadConfig {
   content?: string;
 }
 ```
+
+### `showUpdateDate`
```
delucis, line 321: Small API detail: I think it would be nice if the config file and frontmatter options had the same name — I’d suggest `lastUpdated` for both.

It’s been bugging me that I ended up with `editLink` in config and `editUrl` in frontmatter — I might still make a breaking change at some point to align those 😅 

```suggestion
### `lastUpdated`
```

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -133,3 +133,18 @@ interface HeroConfig {
   }>;
 }
 ```
+
+### `lastUpdated`
+
+**type:** `Date`
```
delucis, line 139: I’d suggest allowing us to set an override here either with a `Date` (for an explicit override to a specific value) or a `boolean` (to enable/disable the default git-based date for specific pages). This is also how the frontmatter [`editUrl`](https://starlight.astro.build/reference/frontmatter/#editurl) option works.

```suggestion
**type:** `Date | boolean`
```

The docs below would also need a small tweak to reflect that.

packages/starlight/components/Footer.astro
```diff
@@ -24,7 +24,11 @@ const prevNextLinks = getPrevNextLinks(sidebar);
         <EditLink data={entry.data} id={entry.id} {locale} />
       )
     }
-    <LastUpdated id={entry.id} {lang} {locale} />
+    {
+      config.showUpdateDate && (
```
delucis, line 28: Based on that, I’d guess we do something like the following so that the frontmatter value takes precedence if set when deciding whether or not to show this:

```suggestion
      (entry.data.lastUpdated ?? config.lastUpdated) && (
```

packages/starlight/components/LastUpdated.astro
```diff
@@ -8,19 +8,22 @@ import { useTranslations } from '../utils/translations';
...
 );
 
-let date: Date | undefined;
+let date = lastUpdated;
 try {
-  ({ date } = getFileCommitDate(currentFilePath, 'newest'));
+  if (!date) {
+    ({ date } = getFileCommitDate(currentFilePath, 'newest'));
+  }
 } catch {}
```
delucis, line 27: If `lastUpdated` can also be a boolean we’ll need to tweak this to still `getFileCommitDate` when it’s `true`.

packages/starlight/schema.ts
```diff
@@ -113,5 +113,13 @@ export function docsSchema() {
...
         })
         .optional(),
+      
+      /**
+       * The last update date of the current page.
+       * Overrides the date generated from the Git history.
+       */
+      lastUpdated: z
+        .date()
+        .optional(),
```
delucis, line 123:
```suggestion
      lastUpdated: z.union([z.date(), z.boolean()]).optional(),
```

packages/starlight/utils/user-config.ts
```diff
@@ -231,6 +231,15 @@ const UserConfigSchema = z.object({
...
   customCss: z.string().array().optional().default([]),
+
+  /** Define if the last update date should be visible in the page footer. */
+  showUpdateDate: z
+    .boolean()
+    .optional()
+    .default(false)
+    .describe(
+      'Define if the last update date should be visible in the page footer.'
+    ),
```
delucis, line 242: We can skip the `optional()` call here as it is implied by `default()`:

```suggestion
  lastUpdated: z
    .boolean()
    .default(false)
    .describe(
      'Define if the last update date should be visible in the page footer.'
    ),
```

delucis (commented):
Thank you for this @HiDeoo! I have some opinions on a couple details of the API, but this basically looks good.

HiDeoo:
Thanks for the review. Pushed an update that should address all your comments, great refinements to the API.

packages/starlight/components/LastUpdated.astro
HiDeoo: I choose to add the logic in the caller (`<Footer>`) and not include a `boolean` in this component which I think makes more sense. If not, I'm fine moving it here.

docs/src/content/docs/reference/configuration.md
```diff
@@ -317,3 +317,12 @@ interface HeadConfig {
   content?: string;
 }
 ```
+
+### `lastUpdated`
+
+**type:** `boolean`  
+**default:** `false`
+
+Define if the last update date should be visible in the page footer.
```
sarah11918, line 326:
```suggestion
Control whether the footer shows when the page was last updated.
```
I kind of like something like this! It makes it explicit that the display will ONLY happen in the footer or not, and doesn't duplicate "update date".

HiDeoo:
Thanks for the great feedback @sarah11918 , really appreciated.

sarah11918 (approved):
The docs working looks good to e!

docs/src/content/docs/reference/configuration.md
```diff
@@ -317,3 +317,12 @@ interface HeadConfig {
...
 ```
+
+### `lastUpdated`
+
+**type:** `boolean`  
+**default:** `false`
+
+Control whether the footer shows when the page was last updated.
+
+By default, this feature relies on the Git history and may not be accurate on some deployment platforms performing [shallow clones](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt). A page’s frontmatter can override this setting or the generated date using the [`lastUpdated`](/reference/frontmatter/#lastupdated) field.
```
delucis, line 328:
```suggestion
By default, this feature relies on your repository’s Git history and may not be accurate on some deployment platforms performing [shallow clones](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt). A page can override this setting or the Git-based date using the [`lastUpdated` frontmatter field](/reference/frontmatter/#lastupdated).
```

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -133,3 +133,16 @@ interface HeroConfig {
   }>;
 }
 ```
+
+### `lastUpdated`
+
+**type:** `Date | boolean`
+
+Overrides either the [global `lastUpdated` option](/reference/configuration/#lastupdated) or the last update date of the current page generated from the Git history. If a date is specified, it should be a valid [YAML timestamp](https://yaml.org/type/timestamp.html).
```
delucis, line 141:
```suggestion
Overrides the [global `lastUpdated` option](/reference/configuration/#lastupdated). If a date is specified, it must be a valid [YAML timestamp](https://yaml.org/type/timestamp.html) and will override the date stored in Git history.
```

delucis (approved):
I think this one should be ready to go out too! Going to make two small tweaks to the docs and then can also be in v0.4 aka The @HiDeoo Edition 😅

delucis (approved):

sarah11918:
Nice docs improvements, @delucis!

PR #259, opened 2023-06-28
Add support for collapsed sidebar groups

HiDeoo:

- Changes to Starlight code

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1120359166574338058/1120650956271534090) on Discord, this pull request adds support for collapsed sidebar groups:

- Sidebar groups can be collapsed by default using a new `collapsed` option defaulting to `false`.
- Autogenerated subgroups respect the `collapsed` property of their parent group by default.
- Autogenerated subgroups can override this behaviour using the `autogenerate.collapsed` property.

The reasoning behind the third point is to support the use case where an integration could generate a `ManualSidebarGroup` containing multiple `AutoSidebarGroup`, e.g. my [`starlight-typedoc`](https://github.com/HiDeoo/starlight-typedoc) integration where I could want a default output similar to:

```plaintext
- API (not collapsed)
  - module 1 (collapsed)
    - classes (collapsed)
    - functions (collapsed)
  - module 2 (collapsed)
    - classes (collapsed)
    - functions (collapsed)
```

I think having the ability to override the default behaviour of autogenerated subgroups is great for flexibility but it may also adds too much complexity to the API. If you think it's too much, I can remove this feature and only support the default behaviour.

I am not hyper confident on the wording of the documentation so feel free to suggest changes.

Here are some screenshots of the new default behaviour with the following file structure:

```plaintext
├─ getting-started.md
├─ showcase.md
├─ reference
│  ├─ test-a
│  │  ├─ index.md
│  │  ├─ test-a-a
│  │  │  ├─ index.md
│  ├─ test-b
│  │  ├─ index.md
│  │  ├─ test-b-b
│  │  │  ├─ index.md
```

<table>
  <tr>
    <th>Configuration</th>
    <th>Default Preview</th>
    <th>Preview after expanding</th>
  </tr>
  <tr>
    <td>
      <pre>
{
  label: 'Start Here',
  collapsed: true,
  items: [
    {
      label: 'Getting Started',
      link: 'getting-started',
    },
    {
      label: 'Showcase',
      link: 'showcase',
    },
  ],
}
      </pre>
    </td>
    <td>
      SCR-20230628-podo
    </td>
    <td>
      SCR-20230628-pobc
    </td>
  </tr>
  <tr>
    <td>
      <pre>
{
  label: 'Reference',
  autogenerate: {
    directory: 'reference'
  },
  collapsed: true,
}
      </pre>
    </td>
    <td>
      SCR-20230628-phtn
    </td>
    <td>
      SCR-20230628-phrs
    </td>
  </tr>
  <tr>
    <td>
      <pre>
{
  label: 'Reference',
  autogenerate: {
    directory: 'reference',
    collapsed: true
  },
  collapsed: false
}
      </pre>
    </td>
    <td>
      SCR-20230628-piqm
    </td>
    <td>
      N/A
    </td>
  </tr>
</table>


packages/starlight/utils/user-config.ts
```diff
@@ -35,6 +35,11 @@ const SidebarBaseSchema = z.object({
   translations: z.record(z.string()).default({}),
 });
 
+const SidebarCollapsibleItemSchema = z.object({
+  /** Whether this item should be collapsed by default. */
+  collapsed: z.boolean().default(false),
+});
```
delucis, line 41: I think I might suggest extending here already, something like this:

```suggestion
const SidebarGroupSchema = SidebarBaseSchema.extend({
  /** Whether this item should be collapsed by default. */
  collapsed: z.boolean().default(false),
});
```

And then using `SidebarGroupSchema` _instead_ of the base schema for the individual group types. Would avoid some of the awkwardness of the `&` intersections and the `merge` call for `ManualSidebarGroupSchema`.

docs/src/content/docs/reference/configuration.md
delucis, line 1: I think the best thing to do here is probably to separate out the documentation of `collapsed` into a dedicated sections instead of mixing it in with the basics — kind of like we do for the “Translating labels” section below.

I can take a pass at doing that if it’s OK with you?

docs/src/content/docs/reference/configuration.md
```diff
@@ -157,8 +160,11 @@ type SidebarItem = {
   translations?: Record<string, string>;
 } & (
   | { link: string }
-  | { items: SidebarItem[] }
-  | { autogenerate: { directory: string } }
+  | { items: SidebarItem[]; collapsed?: boolean }
+  | {
+      autogenerate: { directory: string; collapsed?: boolean };
```
delucis, line 165: Definitely feeling pleased with myself for making `autogenerate` an object 😁 

I had a suspicion some day we might want stuff in addition to the directory!

packages/starlight/utils/navigation.ts
```diff
@@ -64,7 +66,7 @@ function groupFromAutogenerateConfig(
   routes: Route[],
   currentPathname: string
 ): Group {
-  const { directory } = item.autogenerate;
+  const { collapsed: collapsedGeneratedEntries, directory } = item.autogenerate;
```
delucis, line 69: This is totally a tiny detail, so please excuse me being fussy! I can understand renaming here for clarity, but I’d suggest this naming as it’s a bit more compact (collapsed!) and I think reads nicely alongside `item.collapsed`.
```suggestion
  const { collapsed: subgroupCollapsed, directory } = item.autogenerate;
```

packages/starlight/utils/navigation.ts
```diff
@@ -77,7 +79,8 @@ function groupFromAutogenerateConfig(
   return {
     type: 'group',
     label: pickLang(item.translations, localeToLang(locale)) || item.label,
-    entries: sidebarFromDir(tree, currentPathname, locale),
+    entries: sidebarFromDir(tree, currentPathname, locale, collapsedGeneratedEntries ?? item.collapsed),
```
delucis, line 82:
```suggestion
    entries: sidebarFromDir(tree, currentPathname, locale, subgroupCollapsed ?? item.collapsed),
```

delucis (commented):
Lovely! Agree with adding `collapsed` also on the `autogenerate` object. It does increase the API surface slightly, but I can see the use case and it’s not something everyone will have to use if they don’t need it.

I can take a pass at the docs if that’s helpful!

packages/starlight/components/SidebarSublist.astro
```diff
@@ -21,7 +21,7 @@ interface Props {
             {entry.label}
           </a>
         ) : (
-          <details open>
+          <details open={!entry.collapsed}>
```
delucis, line 24: I realised something while walking the dog this morning: what happens when you visit a page inside a collapsed section? And what would people expect?

In Vitepress or Docusaurus, the `collapsed` setting applies on initial load but as you expand sections and navigate pages, the sidebar stays in place, keeping its state. In Starlight, when you navigate to a new page, you’ll always return the server-rendered sidebar state.

I think we’d probably want to have the current page link visible on load, so maybe we need to account for that when setting the `open` attribute?

```suggestion
          <details
            open={entry.entries.flat(Infinity).some(i => i.isCurrent) || !entry.collapsed}
          >
```

packages/starlight/utils/user-config.ts
HiDeoo: Way better indeed 👍🏼

docs/src/content/docs/reference/configuration.md
HiDeoo: Great call indeed.

packages/starlight/utils/navigation.ts
HiDeoo: No worries, on the contrary, I even knew this was not a good name after hitting the refactor shortcut at least 4 times for this one ^^

HiDeoo:
Pushed changes to include your great suggestions. Thanks for the review.

> I can take a pass at doing that if it’s OK with you?

If you find the time, that would be really appreciated. I'm not really happy with all the attempts I made and I think I am ending with something more complicated than it should be. If not, ping me and I will take another pass at it.

packages/starlight/components/SidebarSublist.astro
delucis: Oh yeah! In fact my naive `flat` version probably would not have worked at all now I think about it — nice one.

delucis (approved):
Took a pass at the docs here and @sarah11918 also says they’re OK so that’s good enough for me 😄

Should be ready to merge tomorrow 🚀

PR #261, opened 2023-06-28
Fix autogenerated nav for pages of fallback content

delucis:

- Changes to Starlight code

#### Description

#237 switched our navigation auto-generation logic to use file `id`s instead of `slug`s so that user config can match the authored on-disk file names. However, I missed that this meant that we were reaching into fallback content to retrieve the `id` which would still contain the original content locale in the path instead of the locale of the current language (e.g. in Starlight’s docs, `guides/customization.md` would be used in the French docs instead of `fr/guides/customization.md`). This led to fallback content simply not being picked up by our tree-building function.

This PR fixes this by adding a localized `id` to the top level of each route as we already do for `slug` and uses that `id` instead of the original entry `id` during navigation generation.


natemoo-re (approved):
LGTM

PR #268, opened 2023-06-29
feat: Added `isOpen` option to ManualSideBarGroup

nikcio:

- Changes to Starlight code

#### Description

Added an `isOpen` option to the groups in the sidebar. Sometimes you don't want a group to be expanded/open by default only if a user navigates to it or needs to see the topics.

Example:
image

The first `Start Here` group is without the option to defaults to `true` and acts like before. The second `Start Here` group has `isOpen` set to `false` in the `astro.config.mjs`.


delucis:
Thanks for this PR @nikcio! We're actually already working on more or less this feature in #63! OK if we close this in favour of that?

nikcio:
@delucis Oh didn't see that. Go with that 😄

PR #274, opened 2023-06-30
i18n(ja): Add 404.md

morinokami:

#### Description

- Add 404.md.


delucis (approved):
Thank you @morinokami — we would be lost without you 😄

PR #281, opened 2023-06-30
Add support for translating the site title

delucis:

- Changes to Starlight code

#### Description

Adds a `titleTranslations` config option that allows multilingual sites to set a different title for each language they support:

```js
starlight({
  title: 'My Docs',
  titleTranslations: {
    pt: 'Meus Docs',
  },
  defaultLocale: 'en',
  locales: {
    en: { label: 'English' },
    pt: { label: 'Português' },
  },
});
```


sarah11918 (approved):
Approving the config docs!

delucis (commented):
Got some feedback [on Discord](https://discord.com/channels/830184174198718474/1124360008835018884/1124363502614749265) that this API might not scale as well as we add more translatable strings. Going to put this on hold for a future release as we discuss.

lorenzolewis:
I'm not sure about the lifecycle here, but would it be possible to expose a `t()` function similar to the Astro recipe and use that within the config: https://docs.astro.build/en/recipes/i18n/#translate-ui-strings

So a similar config could look like this:
```ts
import { t } from '@astrojs/starlight'
starlight({
  title: t('My Docs'),
  defaultLocale: 'en',
  locales: {
    en: { label: 'English' },
    pt: { label: 'Português' },
  },
});
```

`t()` giving access to any values in the existing i18n system.

delucis:
This has got rather stale and we likely want a different approach, so will close for now.

PR #284, opened 2023-06-30
Downgrade `vfile` to version `5.3.7`

HiDeoo:

- Changes to Starlight code

#### Description

This PR downgrades the version of the `vfile` package to `^5.3.7`.

While working on the `<Tabs>` component, I noticed various TypeScript errors:

SCR-20230630-otjz

Turns out the following code from `packages/starlight/user-components/rehype-tabs.ts`:

```ts
declare module "vfile" {
  interface DataMap {
    panels: Panel[];
  }
}
```

fails with the following error:

```
error TS2664: Invalid module name in augmentation, module 'vfile' cannot be found.

10 declare module 'vfile' {
                  ~~~~~~~
```

This error appeared after [the upgrade](https://github.com/withastro/starlight/pull/208) from `vfile@5.3.7` to `vfile@6.0.0`. Downgrading to `vfile@5.3.7` fixes the issue.

SCR-20230630-olnu

I did not take the time to look into the issue in details yet but it looks like [this commit](https://github.com/vfile/vfile/commit/f72469b5412c2e2e8480993b2e4a7c2e1530c80f) totally dropped the type definitions from their package.json file.

In the meantime, I think it may be better to downgrade to `vfile@5.3.7` to avoid dealing with this issue (Astro is also still using this version).

### DIFF
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -48,6 +48,6 @@
     "unified": "^10.1.2",
     "unist-util-remove": "^3.1.1",
     "unist-util-visit": "^4.1.2",
-    "vfile": "^6.0.0"
+    "vfile": "^5.3.7"
   }
 }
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks for the fix!

PR #290, opened 2023-07-01
Open external social icon header links in a new window or tab

paramdeo:

- Changes to Starlight code

#### Description

Opens external social icon header links in a new window or tab.


delucis:
Thank you for the PR @paramdeo! It is intentional that those links don’t open in a new tab by default as per [user research that shows that this rarely a good idea](https://www.nngroup.com/articles/new-browser-windows-and-tabs/). Soon we’ll be offering ways for people to customize these components for their own website, so it would then be possible for folks to opt in to this behaviour if they really need it, but for now we’ll stick with the current behaviour as the default.

PR #291, opened 2023-07-01
Include additional instructions for using custom fonts

paramdeo:

#### Description

Hey everyone. It's not explicitly mentioned that the user needs to also include their custom CSS file (containing font families and other variable overrides) within `astro.config.mjs` under the [Use fonts](https://starlight.astro.build/guides/customization/#use-fonts) section in the docs.

I think it's a good idea to have this step outlined, as starting the Astro dev server without including the custom CSS in the config will result in the fonts not being loaded. This may lead to some head-scratching for new users, and so removing any potential points of friction or ambiguity is a good thing IMHO.

### DIFF
--- a/docs/src/content/docs/guides/customization.mdx
+++ b/docs/src/content/docs/guides/customization.mdx
@@ -449,3 +449,28 @@ main {
   font-family: 'IBM Plex Serif', serif;
 }
 ```
+
+Finally, add your custom CSS file to Starlight’s `customCss` array in `astro.config.mjs` after the CSS file containing your `@font-face` declarations:
+
+```js
+// astro.config.mjs
+import { defineConfig } from 'astro/config';
+import starlight from '@astrojs/starlight'
+
+export default defineConfig({
+  integrations: [
+    starlight({
+      title: 'Docs With a Custom Typeface',
+      customCss: [
+        // Path to your @font-face CSS file (if using local fonts)
+        '/src/fonts/font-face.css',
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/customization.mdx
```diff
@@ -449,3 +449,28 @@ main {
...
+        // Path to your @font-face CSS file (if using Fontsource)
+        '@fontsource/ibm-plex-serif/400.css',
+        '@fontsource/ibm-plex-serif/600.css',
+        // Path to your custom CSS file
+        '/src/styles/custom.css'
+      ],
+    }),
+  ],
+});
+```
```
delucis, line 476: Thanks for this feedback! Do you think we can link to the existing guide to adding custom CSS here instead? The order of font-face and custom CSS files is unimportant, so the process shown there should be sufficient. Maybe something like this:

````suggestion
Follow the [custom CSS instructions](#custom-css-styles) to add your styles to your site.
````

paramdeo: Hey! Sure thing, linking to the CSS section can definitely work if you prefer.

I realize the added markup is a bit verbose/redundant but I wasn't sure if it was left out on purpose (folks new to Astro might forget to add the stylesheet entirely).

But if linking is sufficient then that's okay.

delucis (approved):
Thanks again for highlighting this @paramdeo! 🙌

PR #297, opened 2023-07-03
Improve `<Tabs>` component keyboard interactions

### LINKED ISSUE #233: Tabs component keyboard interaction improvements

0.2.0

2.6.5

pnpm

Mac

Chrome

### Describe the Bug

One of our keyboard-only users reported a behavior he was not expecting with the tabs component. As expected, when focus moves into the tab list and the active tab element is focused, the left and right arrow can be used to move focus to the previous and next tab elements. However, when the tabbed content is fairly large, he expected to use the Down and Up arrow keys to use the normal scrolling behavior to read the entire content of the tab panel while keeping focus on the tab element so that he could easily move to the next tab element when he is done reading. At the moment, this is not possible as the down key automatically moves the focus to the tabpanel.

Looking at the [ARIA Authoring Practices Guide for the Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), a note mentions the following:

> If the tab list is horizontal, it does not listen for `Down Arrow` or `Up Arrow` so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

Following this guidance would allow such behavior.

In the same guide, in the keyboard interaction section, 2 optional keyboard shortcuts are also mentioned:

> `Home` (Optional): Moves focus to the first tab. Optionally, activates the newly focused tab (See note below).

> `End` (Optional): Moves focus to the last tab. Optionally, activates the newly focused tab (See note below).

I think these 2 shortcuts could be a nice addition to the tabs component.

With these changes in mind, I propose the following changes to the tabs component `keydown` event handler:

```diff
  // Work out which key the user is pressing and
  // Calculate the new tab's index where appropriate
  const dir =
    e.key === 'ArrowLeft'
      ? index - 1
      : e.key === 'ArrowRight'
      ? index + 1
-     : e.key === 'ArrowDown'
-     ? 'down'
+     : e.key === 'Home'
+     ? 0
+     : e.key === 'End'
+     ? this.tabs.length - 1
      : null;
  if (dir === null) return;
- // If the down key is pressed, move focus to the open panel,
- // otherwise switch to the adjacent tab
- if (dir === 'down') {
-   e.preventDefault();
-   this.panels[i]?.focus();
- } else if (this.tabs[dir]) {
+ if (this.tabs[dir]) {
    e.preventDefault();
    this.switchTab(this.tabs[dir], dir);
  }
```

If this is acceptable and I am not missing something for the current implementation, I would be happy to submit a PR.

https://stackblitz.com/edit/github-ybbceq?file=src%2Fcontent%2Fdocs%2Findex.mdx

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #233

As discussed in #233, this pull request stops intercepting the down arrow key in the `<Tabs>` component to follow the [ARIA Authoring Practices Guide for the Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/):

> If the tab list is horizontal, it does not listen for `Down Arrow` or `Up Arrow` so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

Altho, this is not the entire story as the guide also states that _when the tabpanel does not contain any focusable elements or the first element with content is not focusable, the tabpanel should set tabindex="0" to include it in the tab sequence of the page._ The current implementation is always setting a `tabindex` of `-1` through the rehype tabs processor.

This PR also introduces a change to follow this behavior. For example, with the following `<Tabs>` component:

```mdx
<Tabs>
<TabItem label="tab 1">

basic text

</TabItem>
</Tabs>
```

Tabbing through the page will first focus the tab list and then the tab panel (which does not contain any focusable element).

<video src="https://github.com/HiDeoo/vercel-env-push/assets/494699/7505e1e8-ba1b-4bc5-a00f-6f61c6607cc0" loop controls></video>

Altho, with the following code:

```mdx
<Tabs>
<TabItem label="tab 1">

<input type="text" placeholder="email" />

</TabItem>
</Tabs>
```

Tabbing from the tab list will focus the input element immediately.

<video src="https://github.com/HiDeoo/vercel-env-push/assets/494699/f53a031e-1061-4ec7-a37c-895fa8fd0777" loop controls></video>

Finally, this PR adds support for the `Home` and `End` keys to navigate to the first and last tab respectively.

Accessibility-wise, I tested the changes with VoiceOver on macOS and the behavior seems to be correct and match the one from the ARIA Authoring Practices Guide and the React Aria implementation altho I do not have a machine with Windows to test with NVDA or JAWS.

Notes:

- The rules to detect if a tabpanel contains a focusable element are borrowed from [React Aria](https://github.com/adobe/react-spectrum/blob/99ca82e87ba2d7fdd54f5b49326fd242320b4b51/packages/%40react-aria/focus/src/FocusScope.tsx#L256-L275) which follows this pattern as [described](https://react-spectrum.adobe.com/react-aria/Tabs.html#focusable-content) in their documentation.
- The rules are CSS selectors and used through the [`hast-util-select` package](https://github.com/syntax-tree/hast-util-select). They could all be converted to JS conditions (e.g. `node.tagName === 'input' && node.attributes.type !== 'hidden' && // etc.`) but I think the CSS selectors are way more readable in this case. Any thoughts?
- The changeset I added is fairly generic, should it contain a list of all the exact changes instead?

### DIFF
--- a/.changeset/curvy-points-juggle.md
+++ b/.changeset/curvy-points-juggle.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": minor
+---
+
+Improve `<Tabs>` component keyboard interactions
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -41,6 +41,7 @@
     "@types/mdast": "^3.0.11",
     "bcp-47": "^2.1.0",
     "execa": "^7.1.1",
+    "hast-util-select": "^5.0.5",
     "hastscript": "^7.2.0",
     "pagefind": "^1.0.0-alpha.5",
     "rehype": "^12.0.1",
--- a/packages/starlight/user-components/Tabs.astro
+++ b/packages/starlight/user-components/Tabs.astro
@@ -101,23 +101,16 @@ const { html, panels } = processPanels(panelHtml);
           const index = this.tabs.indexOf(e.currentTarget as any);
           // Work out which key the user is pressing and
           // Calculate the new tab's index where appropriate
-          const dir =
+          const nextIndex =
             e.key === 'ArrowLeft'
               ? index - 1
               : e.key === 'ArrowRight'
               ? index + 1
-              : e.key === 'ArrowDown'
-              ? 'down'
               : null;
-          if (dir === null) return;
-          // If the down key is pressed, move focus to the open panel,
-          // otherwise switch to the adjacent tab
-          if (dir === 'down') {
+          if (nextIndex === null) return;
+          if (this.tabs[nextIndex]) {
             e.preventDefault();
-            this.panels[i]?.focus();
-          } else if (this.tabs[dir]) {
-            e.preventDefault();
-            this.switchTab(this.tabs[dir], dir);
+            this.switchTab(this.tabs[nextIndex], nextIndex);
           }
         });
       });
--- a/packages/starlight/user-components/Tabs.astro
+++ b/packages/starlight/user-components/Tabs.astro
@@ -106,6 +106,10 @@ const { html, panels } = processPanels(panelHtml);
               ? index - 1
               : e.key === 'ArrowRight'
               ? index + 1
+              : e.key === 'Home'
+              ? 0
+              : e.key === 'End'
+              ? this.tabs.length - 1
               : null;
           if (nextIndex === null) return;
           if (this.tabs[nextIndex]) {
--- a/packages/starlight/user-components/rehype-tabs.ts
+++ b/packages/starlight/user-components/rehype-tabs.ts
@@ -1,3 +1,4 @@
+import { select } from 'hast-util-select';
 import { rehype } from 'rehype';
 import { CONTINUE, SKIP, visit } from 'unist-util-visit';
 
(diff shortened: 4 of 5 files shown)


Yan-Thomas:
Thanks @HiDeoo, this looks great! 

All the patterns seem to work in NVDA, the only issue is that with this change, using the shortcut to read the next line (<kbd>↓</kbd>) doesn't work, since we aren't capturing the down arrow and moving it to the tab panel. 

I did some research, and most tabbed interface implementations don't do this as well, I know this is a recommendation from the Inclusive Components book but it might as well be an oversight for not considering the need to scroll the page.

Letting Chris know that this tradeoff is worth taking, considering the down arrow works again the moment the user focus (<kbd>Tab</kbd>) inside the tab panel.

HiDeoo:
> All the patterns seem to work in NVDA

Thanks for taking the time to check this out, really appreciate it.

> I did some research, and most tabbed interface implementations don't do this as well

Yeah, this is pretty much the conclusion I came to as well, although I guess it does not 100% mean that it's the right thing to do considering it is such a complex topic with lots of considerations.

delucis (approved):
Thanks for this @HiDeoo and for the NVDA testing @Yan-Thomas! I put a [call out on Mastodon](https://m.webtoo.ls/@swithinbank/110650792255594347) to see if we’d get some more feedback, but so far not much.

I’m happy we’ve made a reasonable assessment of the options given the information we have and that this is a good approach for now. We can always reevaluate later if we get more user feedback.

Also happy with your code decisions hear @HiDeoo! The CSS selector approach looks reasonable, changeset is clear enough, and home/end key handling looks good 👍

PR #300, opened 2023-07-03
Fix broken link on 404 page when `defaultLocale: 'root'` is set

cbontems:

- Changes to Starlight code

#### Description

- When `defaultLocale` is set to `root` in `astro.config.mjs` instead of just removing it, this results in broken site logo link on  the root language 404 page.
- To fix this, I reproduced the simple approach used in the `localizedUrl` function, which sets locale to an empty string if locale is `root`.
- I applied the fix to the `SiteTitle.astro` component in case the same situation arrises from other pages than 404.

### DIFF
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -23,8 +23,7 @@ if (config.logo) {
   if (err) throw new Error(err);
 }
 
-const locale =  Astro.props.locale === 'root' ? '' : Astro.props.locale;
-const href = pathWithBase(locale || '/');
+const href = pathWithBase(Astro.props.locale || '/');
 ---
 
 <a {href} class="site-title flex">
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -23,7 +23,8 @@ if (config.logo) {
   if (err) throw new Error(err);
 }
 
-const href = pathWithBase(Astro.props.locale || '/');
+const locale =  Astro.props.locale === 'root' ? '' : Astro.props.locale;
+const href = pathWithBase(locale || '/');
 ---
 
 <a {href} class="site-title flex">
--- a/.changeset/many-tips-battle.md
+++ b/.changeset/many-tips-battle.md
@@ -2,4 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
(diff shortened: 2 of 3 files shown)


cbontems:
> I think we actually need to fix this in `404.astro` because it includes some other uses of `locale` that might also be buggy with `locale` being `'root'`:
> 
> https://github.com/withastro/starlight/blob/2598b8ffe2754f905c0e10b2479b0065cc70494c/packages/starlight/404.astro#L9-L11
> 
> I guess we can do something like this?
> 
> ```diff
> - const { lang = 'en', dir = 'ltr', locale } = config.defaultLocale || {};
> + const { lang = 'en', dir = 'ltr' } = config.defaultLocale || {};
> + let locale = config.defaultLocale?.locale;
> + if (locale === 'root') locale = undefined;
> ```

Sure,
Maybe we should also keep it in `SiteTitle.astro` as a safe guard in case a similar situation happens from another source, no ?

cbontems:
New commit also fixes an unrelated TS error for missing id attribute on the Page component.

packages/starlight/components/SiteTitle.astro
```diff
@@ -23,7 +23,8 @@ if (config.logo) {
   if (err) throw new Error(err);
 }
 
-const href = pathWithBase(Astro.props.locale || '/');
+const locale =  Astro.props.locale === 'root' ? '' : Astro.props.locale;
+const href = pathWithBase(locale || '/');
```
delucis, line 27: I think we can safely revert this here:
```suggestion
const href = pathWithBase(Astro.props.locale || '/');
```
`locale` is really a property of the route, so if this fails here, it should be fixed elsewhere. And `SiteTitle` is only used in one place, so again, if we get a bad value here, chances are there’s a bad value somewhere else too.

delucis (commented):
Thanks for fixing that `id` — missed that in a previous PR. I think we can revert the `SiteTitle` changes and the only thing this would still need is a changeset. I can add that if you like, otherwise you can run `pnpm changeset add` locally or click the link to add a new changeset in the bot comment on this PR.

cbontems:
OK, this makes sense.
SiteTitle is back to original and a changeset was created.

delucis (approved):
Thanks again for catching and fixing this @cbontems! Will get a patch release with this out now 🙌

PR #301, opened 2023-07-03
Update Astro and fix inline stylesheet support in Astro ≥2.6

delucis:

- Something else!

#### Description

- Updates the version of Astro used in development and in the starter template from 2.5.0 to latest (2.7.3).
- Adds the now stable version of inline stylesheet support config so users of Astro 2.6 and above still benefit from that performance improvement.


sarah11918 (approved):
I don't see anything out of place!

dreyfus92 (approved):
LGTM ✅

PR #303, opened 2023-07-05
Add pagination configuration

HiDeoo:

- Changes to Starlight code

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1120359166574338058/1120359226372526100) on Discord, this pull request adds configuration regarding the previous and next page links visible in the footer.

##### Global config

This PR adds a new global `pagination` option accepting a boolean and defaulting to `true` to define whether or not the previous and next page links are shown in the footer.

##### Frontmatter fields

Two new frontmatter fields are added to allow overriding the global configuration on a per-page basis: `prev` and `next`.

- If a boolean is provided, it will override the global configuration.
- If a string is provided, it will be used as the link text for the generated link if any. I think most users will want to modify the link text most of the time, so I think having this shorter syntax is nice to have instead of always having to provide an object.
- If an object is provided, both the link text and URL can be overridden for the generated link. If a page does not have a previous or next page and both a link text and URL are provided, a link will be generated.

##### Notes

We noted in #256 the benefits of having the config option and frontmatter fields having the same name but I think in this case, this could lead to a bit more confusing API. For example, if using two fields `prev` & `next`, if a user wants to quickly disable pagination, having to set both `prev` and `next` to `false` in the Astro config would not be that great imo. On the other hand, nesting everything under a `pagination` key in the frontmatter might be a bit too verbose / nested?

Considering that the frontmatter fields should be a sporadic use case / override, I think having a single global config option and two frontmatter fields makes sense. Any thoughts or suggestions?

### DIFF
--- a/docs/src/content/docs/reference/frontmatter.md
+++ b/docs/src/content/docs/reference/frontmatter.md
@@ -149,7 +149,7 @@ lastUpdated: 2022-08-09
 
 ### `prev`
 
-**type:** `boolean | string | { href?: string; label?: string }`
+**type:** `boolean | string | { link?: string; label?: string }`
 
 Overrides the [global `pagination` option](/reference/configuration/#pagination). If a string is specified, the generated link text will be replaced and if an object is specified, both the link and the text will can be overridden.
 
@@ -171,14 +171,14 @@ prev: Continue the tutorial
 ---
 # Override both the previous page link and text
 prev: 
-  href: /unrelated-page/
+  link: /unrelated-page/
   label: Check out this other page
 ---
 ```
 
 ### `next`
 
-**type:** `boolean | string | { href?: string; label?: string }`
+**type:** `boolean | string | { link?: string; label?: string }`
 
 Same as [`prev`](#prev) but for the next page link.
 
--- a/docs/src/content/docs/reference/frontmatter.md
+++ b/docs/src/content/docs/reference/frontmatter.md
@@ -146,3 +146,45 @@ title: Page with a custom last update date
 lastUpdated: 2022-08-09
 ---
 ```
+
+### `prev`
+
+**type:** `boolean | string | { href?: string; label?: string }`
+
+Overrides the [global `pagination` option](/reference/configuration/#pagination). If a string is specified, the generated link text will be replaced and if an object is specified, both the link and the text will can be overridden.
+
+```md
+---
+# Hide the previous page link
+prev: false
+---
+```
+
+```md
+---
+# Override the previous page link text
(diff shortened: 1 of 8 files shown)


docs/src/content/docs/reference/frontmatter.md
```diff
@@ -146,3 +146,45 @@ title: Page with a custom last update date
...
+# Override the previous page link text
+prev: Continue the tutorial
+---
+```
+
+```md
+---
+# Override both the previous page link and text
+prev: 
+  href: /unrelated-page/
```
delucis, line 174: I don’t know if this was the correct choice or not but in the `hero.actions` frontmatter and in the sidebar config we use `link` rather than `href`, so we should probably be consistent with this feature and use `link` here too.

packages/starlight/schemas/prevNextLink.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const PrevNextLinkConfigSchema = () =>
+  z
+    .union([
+      z.boolean(),
+      z.string(),
+      z
+        .object({
+          /** The navigation link URL. */
+          href: z.string().optional(),
```
delucis, line 11: So that would mean updating the schema here:
```suggestion
          link: z.string().optional(),
```

packages/starlight/schemas/prevNextLink.ts
```diff
@@ -0,0 +1,20 @@
...
+          /** The navigation link text. */
+          label: z.string().optional(),
+        })
+        .strict(),
```
delucis, line 15: Reading the code, internally `href` is probably nicer to have, so if you want, we can transform user config to still have `href` in Starlight code and only use `link` for the public API:

```suggestion
        .strict()
        .transform(({ link, label }) => ({ href: link, label })),
```

delucis (commented):
Docs, API decisions, and code all looks great @HiDeoo! Just making one suggestion to rename `href` to `link` in the frontmatter config. I don’t feel strongly either way really, but we ended up with `link` for some other settings, so feels like we should be consistent.

HiDeoo:
Thanks for the review.

I definitely agree with the suggested changes to favor consistency. I chose to not transform the user config, I think it may add a small but unnecessary layer of complexity for future maintainers hitting this code so might as well keep the user config and internal types aligned imo.

delucis (approved):
Looks great! And thanks for choosing the refactor path — I agree it’s the better one 🙌 

Hope to get this and the other minor PRs out of the door tomorrow.

PR #308, opened 2023-07-05
Improve code block and inline code default font stack

### LINKED ISSUE #304: Inline code with links UI confusion

latest

latest

pnpm

Mac

Chrome

### Describe the Bug

I am not sure if this really categorizes as a bug, but considering I got a report for it, I thought I'd report it here in case this should be changed.

For a website containing generated code documentation with TypeScript types linked to their definitions, e.g. [`SidebarItem[]`](https://starlight.astro.build/reference/configuration/#sidebaritem), the report mentioned wrong types and even invalid TypeScript syntax.

It took me a little bit to figure out the issue, but to take an example from from the official docs, heres a potential screenshot of the issue:

repro

The reader thought the code was `SidebarItem[.].` (notice the dots in the middle and after the array brackets).

There are multiple easy solutions to fix this from a user perspective (which is what I did), e.g. changing the font for inline code, removing underline for links in inline code, adding an underline offset, etc. altho considering this is also visible in the official docs with the default theme, I thought I should still mention it here.

https://starlight.astro.build/reference/configuration/#sidebar

### Participation

delucis:

- Changes to Starlight code

#### Description

- I forgot to wire up our default monospace fontstack to be used in code blocks and inline code. This resulted in some browsers using suboptimal fonts like Courier. This PR fixes that.
- I also renamed a couple of internal CSS custom properties which were using an outdated prefix.
- Closes #304 by mostly using a better font that risks less confusion than Courier when combined with link underlines. This is not a silver bullet, but other options would require targeting all link styles in a way that doesn’t seem desirable. Will be curious to hear if this change seems sufficient or if more work is needed to properly address #304.

Here are some before/after comparisons on macOS using the live site and the branch preview:
- https://starlight.astro.build/reference/configuration/#sidebar
- https://deploy-preview-308--astro-starlight.netlify.app/reference/configuration/#sidebar

Would be curious if anyone wants to share on a different platform/browser combo!

| Browser | Before | After |
|---|---|---|
| Safari | image | image |
| Edge​/​Chrome | image | image |
| Firefox | image | image |

One additional thing to call out is that this now means that Japanese pages also use this fontstack now, whereas the default monospace before was quite different:

| Before | After |
|---|---|
| image | image |

I’d love to hear from @morinokami or @Kyosuke whether this is an issue and whether the old behaviour is more expected in Japanese text.

### DIFF
--- a/.changeset/itchy-pugs-live.md
+++ b/.changeset/itchy-pugs-live.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix use of default monospace font stack
--- a/docs/src/components/file-tree.astro
+++ b/docs/src/components/file-tree.astro
@@ -29,7 +29,7 @@ const processedContent = await fileTreeProcessor.process({
     padding: 1rem;
     background-color: var(--sl-color-gray-6);
     font-size: var(--sl-text-xs);
-    font-family: var(--__sb-font-mono);
+    font-family: var(--__sl-font-mono);
     overflow-x: auto;
   }
 
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -87,6 +87,7 @@
 
   .content :global(pre code) {
     all: unset;
+    font-family: var(--__sb-font-mono);
   }
 
   .content :global(blockquote) {
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -87,7 +87,7 @@
 
   .content :global(pre code) {
     all: unset;
-    font-family: var(--__sb-font-mono);
+    font-family: var(--__sl-font-mono);
   }
 
   .content :global(blockquote) {
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -77,6 +77,10 @@
     background-color: var(--sl-color-bg-inline-code);
     margin-block: -0.125rem;
     padding: 0.125rem 0.375rem;
+    font-size: var(--sl-text-code-sm);
+  }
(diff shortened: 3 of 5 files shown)


morinokami:
I think the new behavior is preferable for the Japanese doc. When it comes to Japanese, the choice of monospace font depends on the site's design, so it's generally not a problem to use any monospace font. However, it can feel a bit odd when different fonts are used for the same elements like inline code across different languages within the same site, losing some uniformity. With this PR, the Starlight's code-related theme will be consistent across all languages, including Japanese. So, I believe we can consider this an improvement for the Japanese doc.

morinokami (approved):
So LGTM!!

kyosuke:
I think this pull request is a good change. When code is included in a Japanese document, it is easier to read if fonts are specified for the code.

The same set of fonts is specified in the GitHub code view.
Many Japanese web developers seem to prefer the GitHub code view, even if it contains Japanese comments.

kyosuke (approved):

HiDeoo:
The changes look great and I think this is a good approach to close #304.

The only detail "bothering" me, specially when checking out the deploy preview is the size change, everything involving code feels so big now. It also feels like this emphasize the code so much compared to the text for inline code blocks.

image

Could just be me, and I guess the user can tweak this in the end, just wondering if this is a good default. If I am correct, GitHub slightly reduces its font size for inline code, maybe this could be an approach too?

delucis:
Thanks for the feedback @morinokami and @kyosuke — really appreciate it!

@HiDeoo Yes, I also noticed this font-size change last night and will address it. It’s a bit odd — my guess is that overriding the default font-family for `<code>` also implicitly disables the default font-size.

delucis:
OK, made the inline code match the previous sizing. Should hopefully look OK now!

HiDeoo:
> OK, made the inline code match the previous sizing. Should hopefully look OK now!

Looks perfect to me now.

PR #311, opened 2023-07-06
[i18nIgnore] Tweak internationalization documentation

HiDeoo:

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1126169409648603266) on Discord, this PR adds a few changes to the internationalization documentation:

- Add a definition for the `LocaleConfig` configuration and link to it from the `locales` configuration reference.
- Add link to the internationalization guide from the `locales` configuration reference.
- Add links to the `locales` and `defaultLocale` configuration references from the internationalization guide.

A few notes:

- It has been mentioned that a link inside of `<code>` may not be valid. If I get this right, the inline code element [permitted content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code#technical_summary) is [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) which includes `<a>` if it contains only phrasing content which should be the case here.
- The escaping with a `\` of the brackets in `<code>{ \[dir: string\]: [LocaleConfig](#localeconfig) }</code>` is not mandatory afaiu but avoid some linters, e.g. the one built into VS Code, to complain with a warning about a potentially missing link definition: `No link definition found: 'dir: string'(link.no-such-reference)`.
- These changes should not introduce translation changes for existing translations as I tried to update them all accordingly as it's mostly code and links.

### DIFF
--- a/docs/src/content/docs/de/reference/configuration.md
+++ b/docs/src/content/docs/de/reference/configuration.md
@@ -142,7 +142,7 @@ interface LinkItem {
 
 **Typ:** <code>{ \[dir: string\]: [LocaleConfig](#localeconfig) }</code>
 
-Konfiguriere die Internationalisierung (i18n) für Ihre Website, indem du festlegst, welche `Locales` unterstützt werden.
+[Konfiguriere die Internationalisierung (i18n)](/de/guides/i18n/) für Ihre Website, indem du festlegst, welche `Locales` unterstützt werden.
 
 Jeder Eintrag sollte das Verzeichnis, in dem die Dateien der jeweiligen Sprache gespeichert sind, als Schlüssel verwenden.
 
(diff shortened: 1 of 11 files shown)


delucis (approved):
Spot on! And exciting to learn we can link inside `<code>` — hopefully not needed too often but still helpful for cases like this.

In return you get to learn a new secret about our i18n tracking system. Because this change edits translations that are not up-to-date yet we don’t want to mark those as updated just because we made the edits in this PR. Instead, I’ll add the `i18nIgnore` keyword to the commit message so that this PR is disregarded in deciding whether pages are in or out of date. I hope this fascinating knowledge is a worthy exchange for this nice PR 😀

HiDeoo:
> In return you get to learn a new secret about our i18n tracking system.

Oh, that makes total sense, very clever. Thanks for taking the time to explain this 👍

PR #313, opened 2023-07-07
Add a `not-content` CSS class for opting out of default styles

### LINKED ISSUE #241: Additional styles added to MDX Components
0.2.0

2.6.6

npm

Mac

Safari, Chrome

### Describe the Bug

**Expected:** When using components in MDX CSS is sandboxed to the component. Alternatively, providing an opt-out class for the `.content` class, for example, `.not-content` could allow components to be exempt from the MDX margin processing.

**Actual:** Components and their contents have a "margin-top: 1.5em" rule added.

**How to reproduce:**
Add a component of any type to an MDX file, include at least one div. Upon rendering and processing, an unexpected margin-top rule is added. This can cause components to break visual styles.

———

Issue also referenced by @marcjulian in https://github.com/withastro/starlight/issues/88#issuecomment-1562461482

———

Edit: Font sizes are also affected (e.x. an `<h2>` element). There are likely other styles impacted by the markdown processing.

delucis:

- Changes to Starlight code

#### Description

- Closes #241
- Adds support for a `not-content` class name that can be used to opt out of Starlight’s content styling within some HTML markup. Primarily useful in custom components that don’t want Starlight’s spacing and sizing applied to their children. (Most common issue being the `margin-top` applied to siblings.)
- Added docs to the components guide as this seemed most useful in that context, but open to suggestions if people think there’s a better spot for this: https://deploy-preview-313--astro-starlight.netlify.app/guides/components/#compatibility-with-starlights-styles

### DIFF
--- a/docs/src/content/docs/guides/components.mdx
+++ b/docs/src/content/docs/guides/components.mdx
@@ -36,7 +36,7 @@ Learn more about [using components in MDX](https://docs.astro.build/en/guides/ma
 ### Compatibility with Starlight’s styles
 
 Starlight applies default styling to your Markdown content, for example adding margin between elements.
-If these styles conflict with your component’s appearance, set the `.not-content` class on your component to disable them.
+If these styles conflict with your component’s appearance, set the `not-content` class on your component to disable them.
 
 ```astro
 ---
--- a/docs/src/content/docs/guides/components.mdx
+++ b/docs/src/content/docs/guides/components.mdx
@@ -33,6 +33,20 @@ import AnotherComponent from '../../components/AnotherComponent.astro';
 Because Starlight is powered by Astro, you can add support for components built with any [supported UI framework (React, Preact, Svelte, Vue, Solid, Lit, and Alpine)](https://docs.astro.build/en/core-concepts/framework-components/) in your MDX files.
 Learn more about [using components in MDX](https://docs.astro.build/en/guides/markdown-content/#using-components-in-mdx) in the Astro docs.
 
+### Compatibility with Starlight’s styles
+
+Starlight applies default styling to your Markdown content, for example adding margin between elements.
+If these styles conflict with your component’s appearance, set the `.not-content` class on your component to disable them.
+
+```astro
+---
+// src/components/Example.astro
+---
+<div class="not-content">
+  <p>Not impacted by Starlight’s default content styling.</p>
+</div>
+```
+
 ## Built-in components
 
 Starlight provides some built-in components for common documentation use cases.
--- a/packages/starlight/components/MarkdownContent.astro
+++ b/packages/starlight/components/MarkdownContent.astro
@@ -4,15 +4,7 @@
   .content
     > :global(
       :not(a, strong, em, del, span, input)
-        + :not(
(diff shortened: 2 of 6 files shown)


docs/src/content/docs/guides/components.mdx
```diff
@@ -33,6 +33,20 @@ import AnotherComponent from '../../components/AnotherComponent.astro';
 Because Starlight is powered by Astro, you can add support for components built with any [supported UI framework (React, Preact, Svelte, Vue, Solid, Lit, and Alpine)](https://docs.astro.build/en/core-concepts/framework-components/) in your MDX files.
 Learn more about [using components in MDX](https://docs.astro.build/en/guides/markdown-content/#using-components-in-mdx) in the Astro docs.
 
+### Compatibility with Starlight’s styles
+
+Starlight applies default styling to your Markdown content, for example adding margin between elements.
+If these styles conflict with your component’s appearance, set the `.not-content` class on your component to disable them.
```
dreyfus92, line 39:
```suggestion
If these styles conflict with your component’s appearance, set the `not-content` class on your component to disable them.
```

dreyfus92 (changes requested):
I leave you just a small suggestion, aside from that it LGTM 😄

docs/src/content/docs/guides/components.mdx
delucis: Ah, yup, I agree with you!

packages/starlight/components/MarkdownContent.astro
```diff
@@ -2,112 +2,139 @@
...
+          :where([class~='not-content'] *)
+        )
+    ) {
     display: block;
     max-width: 100%;
     height: auto;
   }
 
-  .content :global(h1) {
+  .content :global(h1:not(:where([class~='not-content'] *))) {
```
tony-sull, line 77: The `*` looks like it might throw this selector off but I can't quite place my finger on it 😅 

Is this making sure the `h1` style won't be applied if any parent has the `not-content` class?

I think that may need to be done by *undoing* the style when a `.not-content` parent is found since the `h1` selector itself couldn't look back up the tree

Something like,

```css
.content :global(h1) {
  font-size: var(--sl-text-h1);
}

.content :global(.not-content h1) {
  font-size: unset; /* or initial? */
}
```

delucis: Yeah this combo is not super easy to read but to pull it apart:

```css
h1:not(:where([class~='not-content'] *)) {}
```

Select all `h1` elements which are not _also_ descendents of an element with the `not-content` class. `:where()` effectively let’s you do an intersection of conditions, which is pretty powerful when combined with `:not()`!

Aside: realising now that the attribute selector (`[class~=...]`) has the same specificity as a plain `.not-content`, so might as well just use that I think.

```css
h1:not(:where(.not-content *)) {}
```

tony-sull: Well played, that's some magic right there!  I never realized you could effectively do a check up the tree with a combination of `:not()` and `:where()`!  🪄 👏

PR #318, opened 2023-07-10
Support relative paths in Starlight config

### LINKED ISSUE #285: Build fails to resolve customCss with relative paths

0.4.0

2.5.0

pnpm

Mac

Arc

### Describe the Bug

Relative paths in `astro.config.mjs` work during development, with `astro dev`, but `astro build` fails with the following error:

```
08:19:25 PM [content] "i18n" is not a collection. Check your content config for typos.
08:19:25 PM [content] Types generated 217ms
08:19:25 PM [build] output target: static
08:19:25 PM [build] Collecting build info...
08:19:25 PM [build] Completed in 389ms.
08:19:25 PM [build] Building static entrypoints...
Could not resolve "./src/styles/custom.css" from "virtual:starlight/user-css"
file: virtual:starlight/user-css
 error   Could not resolve "./src/styles/custom.css" from "virtual:starlight/user-css"
  File:
    virtual:starlight/user-css
  Stacktrace:
RollupError: Could not resolve "./src/styles/custom.css" from "virtual:starlight/user-css"
    at error (file:///Users/hasparus/workspace/sen-o-pograniczu/website/node_modules/.pnpm/rollup@3.26.0/node_modules/rollup/dist/es/shared/node-entry.js:2245:30)
    at ModuleLoader.handleInvalidResolvedId (file:///Users/hasparus/workspace/sen-o-pograniczu/website/node_modules/.pnpm/rollup@3.26.0/node_modules/rollup/dist/es/shared/node-entry.js:24632:24)
    at file:///Users/hasparus/workspace/sen-o-pograniczu/website/node_modules/.pnpm/rollup@3.26.0/node_modules/rollup/dist/es/shared/node-entry.js:24594:26
```

Changing paths to start with `/` fixed the problem, but it would be nice if the relative paths in Astro config were resolved consistently in both dev and build.

_No response_

### Participation

delucis:

- Changes to Starlight code

#### Description

- Closes #285 
- Adds support for specifying custom CSS files with a relative path, for example if you had some shared styles in a monorepo:
   ```js
   starlight({
     customCss: ['../shared/styles/base.css'],
   })
   ```
- Extends the same logic to logo images, so relative paths are supported:
   ```js
   starlight({
     logo: {
       src: './src/assets/my-logo.svg',
     },
   });
   ```
- How does it work? ~~It prepends any relative path with a leading `/`, i.e. the path in the example above becomes `/../shared/styles/base.css`. As far as I can tell, Vite treats any path with a leading slash as relative to the project root. We were already using this to resolve files like `/src/file.css`. Prepending the slash seems to let Vite resolve the relative path relative to that same base directory when imported.~~ It resolves any path with a leading `.` against the configured project root.

### DIFF
--- a/packages/starlight/index.ts
+++ b/packages/starlight/index.ts
@@ -88,13 +88,15 @@ function vitePluginStarlightUserConfig(
   opts: StarlightConfig,
   { root }: AstroConfig
 ): NonNullable<ViteUserConfig['plugins']>[number] {
+  const resolveRelativeId = (id: string) =>
+    id.startsWith('.') ? '/' + id : id;
   const modules = {
     'virtual:starlight/user-config': `export default ${JSON.stringify(opts)}`,
     'virtual:starlight/project-context': `export default ${JSON.stringify({
       root,
     })}`,
     'virtual:starlight/user-css': opts.customCss
-      .map((id) => `import "${id}";`)
+      .map((id) => `import "${resolveRelativeId(id)}";`)
       .join(''),
     'virtual:starlight/user-images': opts.logo
       ? 'src' in opts.logo
--- a/packages/starlight/index.ts
+++ b/packages/starlight/index.ts
@@ -100,8 +100,8 @@ function vitePluginStarlightUserConfig(
       .join(''),
     'virtual:starlight/user-images': opts.logo
       ? 'src' in opts.logo
-        ? `import src from "${opts.logo.src}"; export const logos = { dark: src, light: src };`
-        : `import dark from "${opts.logo.dark}"; import light from "${opts.logo.light}"; export const logos = { dark, light };`
+        ? `import src from "${resolveRelativeId(opts.logo.src)}"; export const logos = { dark: src, light: src };`
+        : `import dark from "${resolveRelativeId(opts.logo.dark)}"; import light from "${resolveRelativeId(opts.logo.light)}"; export const logos = { dark, light };`
       : 'export const logos = {};',
   };
   const resolutionMap = Object.fromEntries(
--- a/packages/starlight/index.ts
+++ b/packages/starlight/index.ts
@@ -6,7 +6,7 @@ import type {
   ViteUserConfig,
 } from 'astro';
 import { spawn } from 'node:child_process';
-import { dirname, relative } from 'node:path';
+import { dirname, relative, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { starlightAsides } from './integrations/asides';
 import { starlightSitemap } from './integrations/sitemap';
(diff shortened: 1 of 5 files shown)


packages/starlight/index.ts
```diff
@@ -88,13 +88,15 @@ function vitePluginStarlightUserConfig(
   opts: StarlightConfig,
   { root }: AstroConfig
 ): NonNullable<ViteUserConfig['plugins']>[number] {
+  const resolveRelativeId = (id: string) =>
+    id.startsWith('.') ? '/' + id : id;
```
bluwy, line 92: This is really hacking into Vite 😄 I think it's safer to use `path.resolve` here instead so it's an absolute path.

delucis: My first approach was indeed `path.resolve(config.root, id)` which worked for me on Mac, but I wondered if it breaks in Windows. Is `import 'C:\user\dir\file.css'` valid Javascript?

bluwy: > Is `import 'C:\user\dir\file.css'` valid Javascript?

It won't work in node because `C:` looks like a protocol, but it will work in Vite as it's plugin can handle that. So it should also work in Windows.

delucis: Ahh, Vite, the beautiful land where dreams come true. Thanks @bluwy!

Will update to this (adding the `resolve` import up top):

```suggestion
  const resolveRelativeId = (id: string) =>
    id.startsWith('.') ? resolve(fileURLToPath(root), id) : id;
```

packages/starlight/index.ts
```diff
@@ -88,18 +88,20 @@ function vitePluginStarlightUserConfig(
...
+  const resolveRelativeId = (id: string) =>
+    id.startsWith('.') ? resolve(fileURLToPath(root), id) : id;
   const modules = {
     'virtual:starlight/user-config': `export default ${JSON.stringify(opts)}`,
     'virtual:starlight/project-context': `export default ${JSON.stringify({
       root,
     })}`,
     'virtual:starlight/user-css': opts.customCss
-      .map((id) => `import "${id}";`)
+      .map((id) => `import "${resolveRelativeId(id)}";`)
```
bluwy, line 99: Optional note: When constructing imports, you can also do:

```js
`import ${JSON.stringify(someId)}`
```

too. It's usually more robust when handling quotes, but it's also a rather edge case.

bluwy (approved):

packages/starlight/index.ts
delucis: Innnnteresting. You mean in case someone has a `/User/"my docs"/path/`? That’s funny. I guess we can just update the resolve method to do that.

delucis: Honestly, that seems like such an edge case that I’m happy to ignore it for now and we can fix if someone actually runs into it.

bluwy: That's usually how I treat those kind of issues too 😄 So fair enough

PR #337, opened 2023-07-14
Add a Tailwind plugin and docs

### LINKED ISSUE #88: Support Tailwind CSS integration and styles

0.0.11

2.5.5

npm

Mac

Chrome, Firefox

### Describe the Bug

Adding `@astrojs/tailwind` and using Tailwind CSS utilities in the content should apply the generated CSS.
The integration was added using `npx astro add tailwind` and the following tag and utility classes are added to `src/content/docs/index.md`.

```html
<p class="text-red-500 font-bold">Tailwind CSS should work</p>
```

The `base.css` is not loaded and thus the styles are not applied, see the Stackblitz demo.

https://stackblitz.com/edit/github-ypma4c?file=src%2Fcontent%2Fdocs%2Findex.md

### Participation

delucis:

- Changes to Starlight code

#### Description

- Closes #88  
- Adds a Tailwind plugin, available by installing `@astrojs/starlight-tailwind`, which will configure Tailwind for compatibility with Starlight
  - Disables Preflight (but re-enables its important `border` styling)
  - Configures Tailwind’s dark mode for compatibility with Starlight’s `data-theme` attribute
  - Links Starlight’s color custom properties to `gray` and `accent` in Tailwind’s `theme.colors` config.
  - Links Starlight’s font custom properties to `sans` and `mono` in Tailwind’s `theme.fontFamily` config.
- Adds a Tailwind starter project at `examples/tailwind/`

Big shout out to @marcjulian for many of the tips in #88 that helped shape the work here. Very curious to hear feedback from folks about whether this solution works or not.

#### To do

- [x] Write docs for how to set up and use Tailwind alongside Starlight
- [x] Remove workarounds in the starter template due to https://github.com/withastro/astro/issues/7563 once that is fixed

#### Try it

Keeping this a draft while we wait for withastro/astro#7563 to be fixed, but you can try the starter template by cloning this branch, installing dependencies, then running `cd examples/tailwind && pnpm dev`.


sarah11918 (approved):
Docs LGTM! So excited for this feature!

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -0,0 +1,196 @@
+---
+title: CSS & Styling
```
Yan-Thomas, line 2: Wondering if it was intentional to name the file "CSS and Tailwind" but keep the title as "CSS & Styling"

Yan-Thomas (approved):
Docs and example Markdown files all look great to me! Only added a small, non-blocking comment 🙌

docs/src/content/docs/guides/css-and-tailwind.mdx
delucis: It kind of was, but maybe not a good decision? Perhaps I should stick to matching the title and rename it `css-and-styling.mdx`.

HiDeoo:
Super neat and clean ✨ 

I assume the `withastro/starlight/examples/tailwind` template would get aliased at some point in `create-astro` like Starlight?

Small question: any specific reason to use a `.cjs` file for the Tailwind configuration file considering they're [supporting ESM](https://tailwindcss.com/blog/tailwindcss-v3-3#esm-and-type-script-support) now? Could be 1 less file type maybe for the end user?

delucis:
> any specific reason to use a `.cjs` file for the Tailwind configuration file 

Mainly that it's what `astro add tailwind` currently generates. Tailwind docs also show CJS syntax like importing with `require` and exporting with `module.exports` (although usually with a `.js` extension).

I actually tried and all extensions work here (also `.ts`) and, perhaps most weirdly, you can use import or require in all of them thanks to how Tailwind processes the file during loading. But I thought it might be easiest to stick to what Astro does to have a better chance of matching what a user has.

HiDeoo:
> But I thought it might be easiest to stick to what Astro does to have a better chance of matching what a user has.

Makes total sense, thanks.

> Would be nice to support anything after a slash to shorten things

Ah yeah, that would be great indeed.

marcjulian:
Hey @delucis this looks awesome. Providing a `starlightPlugin` for preparing Tailwind to work well with Starlight styles is is perfect. It's event easy to swap out `accent` and `gray` color by just changing it in the tailwind config.

> Would be nice to support anything after a slash to shorten things, e.g. --template starlight/tailwind.

Even shorter would be `starwind` ⭐ 🌬️

marcjulian:
I noticed two things while playing around with a custom theme.

1. Changing `accent` to `colors.red` in `tailwind.config.cjs` updates the accent while in dev mode. However, the `accent` is overridden in the build output by the default Starlight colors.

Dev mode with `accent: colors.red`
CleanShot 2023-08-08 at 12 00 39

Build output with `accent: colors.red`
CleanShot 2023-08-08 at 12 01 04

I checked the generated CSS content. It does contain `--sl-color-accent: #dc2626;` (`theme(colors.red.600`) at the beginning of the CSS file in `:root`. But after all tailwind styles there is a block `:root,::backdrop` which redefines `--sl-color-accent: hsl(var(--sl-hue-accent), 100%, 60%);`, the default Starlight theme (coming from `packages/starlight/style/props.css`).

A workaround would be to add the `customCss` file back with the tailwind directives (https://github.com/withastro/starlight/pull/337/commits/8e3602b9d73014de444d308b58a5d025dc44587a). This works because the tailwind styles are added last to the generated CSS file.

2. When the `starlightPlugin` is enabled, I cannot override the theme styles with providing a `customCss` file.
For example, if the theme should have a darker toolbar/sidebar and a lighter content area in dark mode or when I need to apply my own theme:

```css
/* needs base directive otherwise vite throws: `@layer base` is used but no matching `@tailwind base` directive is present. */
@tailwind base;

@layer base {
  /* Dark Theme */
  :root {
    --sl-color-white: theme('colors.white');
    --sl-color-gray-1: theme('colors.gray.200');
    --sl-color-gray-2: theme('colors.gray.300');
    --sl-color-gray-3: theme('colors.gray.500');
    --sl-color-gray-4: theme('colors.gray.600');
    --sl-color-gray-5: theme('colors.gray.700');
    --sl-color-gray-6: theme('colors.gray.950'); /* darker toolbar/sidebar */
    --sl-color-black: theme('colors.gray.800'); /* brighter content */
  }
}
```

The custom theme styles are overridden by the styles provided by the `starlightPlugin`, but only in dev mode again.
CleanShot 2023-08-08 at 12 22 36

In build output, the custom styles take effect over the `starlightPlugin` styles.

CleanShot 2023-08-08 at 12 31 07

Perhaps allowing `starlightPlugin` to receive an option to disable the default style mapping from starlight would help here. This also eliminates having duplicate definitions of CSS variables and thus leading to a smaller CSS file.

delucis:
Thanks for testing @marcjulian! Which Astro version was this with?

marcjulian:
`astro@2.10.3` I forked this project and test it with the tailwind example.

delucis:
Thanks for the testing and feedback @marcjulian! I’ve moved the docs and starter template back to the `customCSS` approach as this seems to be the best way to resolve all these issues.

marcjulian:
Yes using `customCss` option solves both issues for dev mode and build.

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+    ```sh
+    yarn astro add tailwind
+    ```
+
+    </TabItem>
+
+    </Tabs>
+
+2.  Install the Starlight Tailwind plugin:
```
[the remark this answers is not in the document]
delucis: Thanks for the suggestion. Just added a note like this to the docs.

PR #351, opened 2023-07-17
Allow to not include site-search component in desired docs (feature)

kpetravicius:

- Changes to Starlight code

#### Description

Introduced **showSearch** prop in schemas. It allows a user to "hide" site-search component inside header element. 
Use case: index.mdx (or other docs) now can be more like landing pages than documentation front page.
image


delucis:
Thanks for the PR @kpetravicius! I think this needs to be tackled as part of a wider consideration of search functionality in #357, so will close this one for now as we discuss the bigger picture.

PR #354, opened 2023-07-17
feat: Add code block title

yjose:

- Changes to Starlight code

#### Description

This PR add support to show code block title using `remark-code-title` 

Light theme 

image

Dark theme 

image


delucis:
Thank you for the PR @yjose! We actually already have plans to integrate a tool called [`expressive-code`](https://github.com/expressive-code/expressive-code) that will add code titles like this as well as other features. Sorry that this wasn’t clear from the repo and you spent time on this. We’ll have a draft PR up soon in case you’re curious how that one works and can share it here.

Really appreciate you coming in with the PR though. If you have any other ideas, don’t hesitate to open an issue to discuss them — or join us in the `#starlight` channel in the [Astro Discord server](https://astro.build/chat).

yjose:
@Delucis No problem, and thank you for the updates. The package seems good and should be the right solution for such a use case. Looking forward to trying it!

PR #359, opened 2023-07-18
Define the order of links in sidebar

IDurward:

- Changes to Starlight code

#### Description

- What does this PR change? 
  This allows the user to define the order that document links are displayed in, in the sidebar, when autogenerate is used. If no order is defined then the default alphabetical order is used.


packages/starlight/schema.ts
```diff
@@ -38,6 +38,14 @@ export function docsSchema() {
        */
       description: z.string().optional(),
 
+      /**
+       * The order of this page in the navigation.
+       * Pages are sorted by this value in ascending order. Then by slug.
+       * If not provided, pages will be sorted alphabetically by slug.
+       * If two pages have the same order value, they will be sorted alphabetically by slug.
+       */
+      order: z.number().optional(),
```
delucis, line 47: Can we move this under a `sidebar` key? I think we’re very likely to want more things here, so this would allow for a nice grouping and I think also clarifies a bit what this is the order of. For example frontmatter might end up looking like this eventually:
```yml
title: Page title may be a bit longer
sidebar:
  order: 1
  label: Short nav title
```

delucis (commented):
Thank you for this PR @IDurward and for your patience waiting for a review!

I think this looks good in principle! I’ve left a suggestion regarding the exact frontmatter syntax.

We also just added unit testing to the repo ([guide/notes here](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#unit-tests)) so now have the opportunity to add some tests to ensure this is working as expected. Would you be able to add some tests for this? Happy to provide guidance if there’s anything unclear there.

packages/starlight/schema.ts
IDurward: Yeah, I agree that is better. I will make the required changes

delucis: Great — thank you! Took the liberty of resolving conflicts and updating your branch, so you should now have test files on here too.

delucis (commented):
Thanks again for this @IDurward! I took a quick pass at writing a test and gave `sidebar` a default value of `{}` so that pages aren’t forced to set it if they don’t need it.

This still needs documenting in the frontmatter reference if you’d like to pick that up. Otherwise, I’ll get it to it tomorrow 🙌

delucis (approved):
Thanks again for kicking this off @IDurward! And apologies for commandeering your PR — I realised there were some edge cases with the approach sorting in the routing code, so I’ve moved this to our sidebar code. Luckily, these changes will also make it easier in the future to add more features to this sidebar frontmatter config.

IDurward:
That's ok @delucis, I don't mind you commandeering as my time is a bit limited at the moent.
You may want to add a recomendation to the docs around ordering the pages in counts of 5 or 10 so that if a page needs to be added between existing pages then it reduces the need to change the numbering of every other page.

PR #371, opened 2023-07-21
Add prettier and format files

lorenzolewis:

- Changes to Starlight code

#### Description

- Add `prettier` and `prettier-plugin-astro`
- Add prettier config (based on the main Astro repo)
- Format all files in repo

### DIFF
--- a/.github/workflows/congratsbot.yml
+++ b/.github/workflows/congratsbot.yml
@@ -9,7 +9,7 @@ jobs:
     if: ${{ github.repository_owner == 'withastro' }}
     uses: withastro/automation/.github/workflows/congratsbot.yml@main
     with:
-      EMOJIS: '🎉,🎊,🧑‍🚀,🥳,🙌,🚀,🤩,☄️,💫,<:starlight:1107431075543797802>,<:houston_pride:1048276366518390814>'
+      EMOJIS: '🎉,🎊,🧑‍🚀,🥳,🙌,🚀,🤩,☄️,💫,<:starlight:1107431075543797802>,<:houston_pride:1130504824673284107>'
       COAUTHOR_TEMPLATES: >
         [
           "Thanks <names> for helping! ✨",


HiDeoo:
The plugin autoloading feature from Prettier was [removed](https://github.com/prettier/prettier/pull/14759) in v3 and the Prettier configuration file does not explicitly specify the Prettier plugin for Astro so I don't think it's running on Astro files (which may be why the diff does not contain any Astro file changes).

Altho, I do not think specifying the plugin in the config would be enough as there is currently [a bug](https://github.com/withastro/prettier-plugin-astro/issues/334#issuecomment-1638672927) in Prettier where you have to manually specify the files handled by a plugin for it to format these files, e.g. `prettier --write **/*.astro`.

I wonder if it would not be wise to wait until this issue is fixed before introducing Prettier to the codebase?

lorenzolewis:
Thanks @HiDeoo , didn't even think to double-check the Astro files.

I updated the prettier config to include the plugin and added a `format` script that will at least catch it all (from what I can tell). Also got all those `*.astro` files formatted.

Completely understand if you want to hold off on this until https://github.com/prettier/prettier/issues/15079 / https://github.com/withastro/prettier-plugin-astro/issues/334#issuecomment-1638672927 are resolved.

pnpm-lock.yaml
delucis, line 1: I’m fairly sure we shouldn’t be formatting the lock file, so should probably ignore that from formatting?

docs/src/content/docs/es/guides/components.mdx
```diff
@@ -22,9 +22,7 @@ import OtroComponente from '../../componentes/OtroComponente.astro';
 
 <UnComponente prop="algo" />
 
-<OtroComponente>
-  Los componentes también pueden contener **contenido anidado**.
-</OtroComponente>
+<OtroComponente>Los componentes también pueden contener **contenido anidado**.</OtroComponente>
```
delucis, line 25: Ahhh, the joys of auto-formatting content. We often use a shorter line length in code sample code than in regular code so it stays legible on narrower screens. Is it possible to use a shorter line length in `.md` and `.mdx`?

Example of this change overflowing on the left and current version on the right:
Screenshot comparison

docs/src/content/docs/es/guides/i18n.mdx
delucis, line 1: Looking at the code samples now they’re using tabs, we should probably set [`tab-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size) for Starlight code blocks.

We can do it in a follow-up PR though so we’re not mixing formatting with meaningful code changes.

docs/src/content/docs/fr/reference/configuration.md
```diff
@@ -155,11 +155,7 @@ sidebar: [
 type SidebarItem = {
   label: string;
   translations?: Record<string, string>;
-} & (
-  | { link: string }
-  | { items: SidebarItem[] }
-  | { autogenerate: { directory: string } }
-);
+} & ({ link: string } | { items: SidebarItem[] } | { autogenerate: { directory: string } });
```
delucis, line 158: Another example where a shorter line length in content would be better.

delucis (commented):
Thanks for this @lorenzolewis! Left a few observations. And… of course this is now conflicting 😅 Hopefully easy enough to resolve by reverting the formatting, updating branch, and then re-formatting?

lorenzolewis:
I found another magical button on GitHub that apparently closed this one 😅 I'll re-open another one with the suggestions included.

PR #383, opened 2023-07-22
Add unit testing

### LINKED ISSUE #240: Add tests to help guard against regressions
We haven’t added tests to any of Starlight’s functionality just yet, but this would be extremely helpful to help us assure we don’t accidentally break features as we work.

Let’s use this issue to discuss approaches!

Some quick thoughts:

- Astro components themselves are a little tricky to test. There are older tools like https://github.com/Princesseuh/astro-component-tester, but it may be better to start by building for example the `examples/basics` project and asserting it looks like we expect.

- We also had the suggestion at one point of using snapshot testing for things like this as a quick way to track general build output stability.

- A lot of Starlight’s code depends on loading virtual modules like `astro:content`, `astro:assets`, and Starlight’s own `virtual:*` modules for user config etc. Can we make these available in a testing environment for unit testing code? Or do we need to refactor things like the navigation/routing logic to be able to pass the data that depends on virtual modules into those functions?

- Testing navigation and routing logic feels high priority to me as there are quite a few variable inputs and it can be hard in manual testing to remember to test them in all their combinations.

delucis:

- Changes to Starlight code
- Something else!

#### Description

- Closes #240 
- Adds unit testing with Vitest
- Uses [Vitest workspaces](https://vitest.dev/guide/workspace.html) to support multiple test directories each using their own instance of Starlight’s virtual modules plugin.
- Adds a utility to mock out `getCollection` calls with fake content or data entries.
- Test coverage is not super high yet, but focused on getting the infrastructure in place and testing some key utilities such as routing and sidebar logic.
- Adds notes for contributors on testing: https://github.com/withastro/starlight/blob/dx-196/testing/CONTRIBUTING.md#testing


packages/starlight/__tests__/basics/base.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  // TODO: Stubbing BASE_URL is not currently possible.
+  // Astro controls BASE_URL via its `vite-plugin-env`, which prevents Vitest’s stubbing from
+  // working and there’s also no way to pass in Astro config in Astro’s `getViteConfig` helper.
+  describe.todo('with base', () => {
+    test('prepends base', () => {
+      vi.stubEnv('BASE_URL', '/base/');
+      expect(fileWithBase('/img.svg')).toBe('/base/img.svg');
+      vi.unstubAllEnvs();
+    });
+  });
```
delucis, line 23: Vitest provides a [stubbing helper](https://vitest.dev/api/vi.html#vi-stubenv) to set `import.meta.env` values in tests, but this does not work for values controlled by Astro’s [`vite-plugin-env`](https://github.com/withastro/astro/blob/e80896a67c08c1c7916417d2f385a185c773d97e/packages/astro/src/vite-plugin-env/index.ts) such as `BASE_URL`.

It is also not possible to set Astro’s `base` option via a `vitest.config.ts` file using Astro’s `getViteConfig` helper — instead Astro side loads an `astro.config` from disk. I couldn’t find a way to get it to detect a local test-suite-specific `astro.config`, so for now testing for custom `base` settings seems to be blocked.

packages/starlight/__tests__/test-config.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function defineVitestConfig(config: z.input<typeof StarlightConfigSchema>) {
+  return getViteConfig({
+    plugins: [
+      vitePluginStarlightUserConfig(StarlightConfigSchema.parse(config), {
+        root: new URL(import.meta.url),
+      }),
+    ],
+  });
+}
```
delucis, line 16: This config helper wraps Astro’s `getViteConfig` to set virtual modules in the test environment using Starlight’s Vite plugin. If each test suite could contain an `astro.config` instead, we could perhaps use the Starlight integration directly instead.

packages/starlight/package.json
```diff
@@ -2,7 +2,10 @@
   "name": "@astrojs/starlight",
   "version": "0.5.5",
   "description": "Build beautiful, high-performance documentation websites with Astro",
-  "scripts": {},
+  "scripts": {
+    "test": "vitest",
+    "test:coverage": "vitest run --coverage.enabled --coverage.all --coverage.reportsDirectory=./__coverage__"
```
delucis, line 7: In theory you can configure Vitest’s `coverage` settings in `vitest.config.ts`, but I couldn’t get this to work, so resorted to CLI arguments here.

packages/starlight/__tests__/basics/base.test.ts
delucis: Yeah, that would be the hardcore way (although given https://github.com/withastro/astro/issues/7561 might also be tricky to set-up without cluttering the repo with fixture directories at the base level), but seems wrong for testing a single function output 😅

packages/starlight/package.json
bluwy: Perhaps you need to use the `--coverage` CLI option, and configure the rest of the options in the config?

packages/starlight/__tests__/test-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+        z.literal('jpeg'),
+        z.literal('tiff'),
+        z.literal('webp'),
+        z.literal('gif'),
+        z.literal('svg'),
+      ]),
+    }),
+});
+
+function mockDoc(
```
bluwy, line 24: This gave me a good chuckle 😄

bluwy (commented):
Setup looks great! Not leaving an approval since I don't know the entire scope of the changes, but that's a lot of tests for the initial implementation.

packages/starlight/package.json
delucis: Yeah, tried that, but it still wouldn’t pick them up. `coverage` config is not supported in a root `vitest.workspace.ts` only in `vitest.config.ts`, so I wondered if maybe using workspaces is the issue. Unless… can you have a root config file and then the workspace config too? 🤔 That’s the one permutation I didn’t try.

delucis: YES! A root-level `vitest.config.ts` works! Nice.

PR #385, opened 2023-07-23
Fix Nested content css styles

lorenzolewis:

- Changes to Starlight code

#### Description

- Fixes an issue where the CSS property `margin-top` isn't applied to content within imported components

Before:
Screenshot 2023-07-23 at 16 00 39

After:
Screenshot 2023-07-23 at 16 00 15


delucis (approved):
No idea why the `>` snuck in — thanks for fixing!

PR #386, opened 2023-07-24
Fix: Prevent keypress from firing on inputs

huijing:

- Changes to Starlight code

#### Description

Currently, when `/` is pressed, regardless of whether the active element is an input or not, the search modal will pop up. If the input field requires things like a URL, this is a disruptive behaviour. This PR adds an input check on the active element, and if it is an input, the search modal will not pop up.

### DIFF
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -95,11 +95,16 @@ const pagefindTranslations = {
       closeBtn.addEventListener('click', closeModal);
 
       // Listen for `/` and `cmd + k` keyboard shortcuts.
+      const inputs = ["input", "select", "textarea"];
+      const isInput =
+        document.activeElement &&
+        inputs.indexOf(document.activeElement.tagName.toLowerCase()) !== -1;
+      
       window.addEventListener('keydown', (e) => {
         if (e.metaKey === true && e.key === 'k') {
           dialog.open ? closeModal() : openModal();
           e.preventDefault();
-        } else if (e.key === '/' && !dialog.open) {
+        } else if (e.key === '/' && !dialog.open && !isInput) {
           openModal();
           e.preventDefault();
         }
--- a/.changeset/quick-peas-deny.md
+++ b/.changeset/quick-peas-deny.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/starlight/components/Search.astro
```diff
@@ -95,11 +95,16 @@ const pagefindTranslations = {
       closeBtn.addEventListener('click', closeModal);
 
       // Listen for `/` and `cmd + k` keyboard shortcuts.
+      const inputs = ["input", "select", "textarea"];
+      const isInput =
+        document.activeElement &&
+        inputs.indexOf(document.activeElement.tagName.toLowerCase()) !== -1;
```
delucis, line 101: Thanks for flagging this and opening the PR — appreciate it!

Would you mind sharing how you tested this? In the deploy preview, if I focus one of the `<select>` menus in the site header and press `/`, the search modal is still triggered.

delucis: Oh, I see the issue now! The `isInput` check runs once when the script initialises and never again, whereas we want to run it on each `keydown` event. Will suggest some fixes.

packages/starlight/components/Search.astro
```diff
@@ -95,11 +95,16 @@ const pagefindTranslations = {
...
+      
       window.addEventListener('keydown', (e) => {
         if (e.metaKey === true && e.key === 'k') {
```
delucis, line 104:
```suggestion
        const isInput =
          document.activeElement &&
          ['input', 'select', 'textarea'].includes(document.activeElement.tagName.toLowerCase());
        if (e.metaKey === true && e.key === 'k') {
```

packages/starlight/components/Search.astro
delucis, line 101:
```suggestion
```

delucis (commented):
I think this should fix the behaviour:

packages/starlight/components/Search.astro
huijing: Those make sense to me. I've commited them. Let me know if you want me to squash and rebase this PR?

delucis (approved):
Perfect. Looks like it’s working now — thank you!

And no, no need to squash/rebase — I’ve updated the branch and will squash the commit when merging.

PR #393, opened 2023-07-24
[i18nIgnore] Add Prettier and format files

lorenzolewis:

- Changes to Starlight code

#### Description

- Add `prettier` and `prettier-plugin-astro`
- Add prettier config (based on the main Astro repo)
- Format all files in repo
- Add `format` script

Originally did this in https://github.com/withastro/starlight/pull/371 but had to reset my fork


HiDeoo:
Some small quick thoughts I just got:

- The `.astro` & `__coverage__` directories should be added to the `.prettierignore`.
- The `format` npm script should maybe use the [`--cache` option](https://prettier.io/docs/en/cli.html#--cache).

delucis:
(Don’t worry too much about the failing coverage check — evidently the number of lines dropped due to formatting, we can bump those budgets down so that this PR passes.)

lorenzolewis:
Thanks @HiDeoo, just added those both in

@delucis you snuck that coverage check in quickly ;)

delucis:
Just tried running `pnpm format` locally to format newly updated files, but seem to be seeing changes to spacing in Japanese, which you weren’t. For example:
```diff
- Starlightはまだ開発の初期段階にあります。何か不具合を見つけた場合は、[GitHubでissueを開く](https://github.com/withastro/starlight/issues/new/choose)か、または[Discord](https://astro.build/chat)でお知らせください。
+ Starlight はまだ開発の初期段階にあります。何か不具合を見つけた場合は、[GitHub で issue を開く](https://github.com/withastro/starlight/issues/new/choose)か、または[Discord](https://astro.build/chat)でお知らせください。
```

Any idea why that might be?

lorenzolewis:
Weird, I just merged from main to get all the latest updates and re-ran the command. The only thing it writes are the two updated files since the last time I merged.

I'm running `pnpm format` these in the very root directory of the workspace

delucis:
Very odd… Ah! If I change the package script like this it works:
```diff
-    "format": "pnpm prettier -w --cache --plugin prettier-plugin-astro ."
+    "format": "prettier -w --cache --plugin prettier-plugin-astro ."
```

I guess by adding the extra `pnpm` it’s using a different version of Prettier somehow?

lorenzo-crabnebula:
Open to changing that and the above script. I only think you need to include the package manager in the script if you're running some commands deeper down in sub-projects in a workspace

package.json
```diff
@@ -4,13 +4,16 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "version": "pnpm changeset version && pnpm i --no-frozen-lockfile"
+    "version": "pnpm changeset version && pnpm i --no-frozen-lockfile",
+    "format": "prettier -w --cache --plugin prettier-plugin-astro ."
```
Princesseuh, line 8: Make sure this actually formats `.astro` files because there's a bug in Prettier 3 where it doesn't take into account file types added by plugins when just doing `prettier -w .` and you need to manually specify something like `prettier -w **/*.astro`.

lorenzolewis: I _think_ by adding `--plugin prettier-plugin-astro` to the script this takes care of it. I got it to format all of the astro files here by only running `pnpm format` at least.

Princesseuh (approved):
Didn't go through *all* the `.astro` files, but all of which I did looked fine to me. At this point it's unlikely that the Prettier plugin generates invalid code, considering how used it is, anyway. Awesome work!

HiDeoo:
Should this somehow be checked or enforced in a GitHub Actions too? Wouldn't we get unrelated diffs in PRs if someone submit a PR without formatting a file and someone else edit the same file later and format it?

delucis:
Thanks for reviewing, @Princesseuh!

> Should this somehow be checked or enforced in a GitHub Actions too? Wouldn't we get unrelated diffs in PRs if someone submit a PR without formatting a file and someone else edit the same file later and format it?

Yeah, I haven’t mentioned it here because I think it’s OK to get the config right first and add this in a follow-up PR. The way we generally approach this in other Astro repos is to have a CI task which runs after every commit to `main` which runs Prettier (plus usually also https://www.npmjs.com/package/organize-imports-cli) and commits any changes. That approach means we don’t have to pester contributors about formatting but the codebase still stays formatted.

delucis (approved):

PR #397, opened 2023-07-25
Add LinkCard Component

lorenzolewis:

- Changes to Starlight code

#### Description

Opening as a draft for now to get some feedback. Relevant discussion on Discord: https://discord.com/channels/830184174198718474/1133135977628577813

- Create a `LinkCard` component for a nicely styled clickable card
- Update `.content` CSS to allow styling for components explicitly marked as `.content` (this was done because the `LinkCard` component is an `a` element and `top-margin` was not being applied. Open to a different way on this if some better suggestions)


HiDeoo:
Neat, this would be a great addition, I needed something like this recently. Some quick thoughts/questions:

- I wonder if I wouldn't expect the `LinkCard` props to extends `HTMLAttributes<'a'>` so a user can pass in any valid `<a>` props. This would mean that `link` could be removed as the user could just pass in `href` instead.
- `title` may be a bit confusing as it's the card title and not the link title which is a valid link attribute. If we decide to accept any `<a>` props, then `title` would be a valid prop for the link and we may need another prop for the card title.

delucis:
Thanks for this! Going to wait for some feedback from @doodlemarks before looking too much into the implementation — looped Mark in in the Discord thread: https://discord.com/channels/830184174198718474/1133135977628577813

lorenzolewis:
Thanks @HiDeoo , I put that in. Always a pleasure learning from you 💜

Updated design based on discussion in the Discord thread

HiDeoo:
Well, this was mostly my own thoughts / questions, maybe that's just bad ideas ^^ We will see what the team thinks about it.

lorenzolewis:
> Well, this was mostly my own thoughts / questions, maybe that's just bad ideas ^^ We will see what the team thinks about it.

No worries, can always revert it if someone says otherwise

lorenzolewis:
Alright, this one is ready for review. These are the things that I know of that didn't get included in this one:

- 3 column layout for `CardGrid`: I'm not 100% sure how this could be accomplished without exposing a new prop in CardGrid or making a different component. Maybe a separate PR for this one if it's wanted?
- Image for leading edge: Would still like to add this to fill the use case for the deploy page in the mockup, but would prefer getting this merged in first and then tackling that one separately.

The code block on the documentation page felt a bit long to me, so if anyone else agrees I can refactor that to reduce the number of cards.

delucis:
@lorenzolewis Thank for your patience on this once! Was hoping to get a release out with this today so commandeered your work, only to run out of time in the day 😅 

So apologies for committing and not reviewing! Let me at least leave some notes on what I changed:

- Pretty sure we’ll need that RTL support elsewhere in the future, so moved it to a utility class. Also made it use a `matrix()` transform to mirror the element across the vertical axis instead of rotating 180°. That’s only subtly different and doesn’t impact our symmetrical arrow icon in this case, but could be relevant for some icons, think 👉 vs 👈 for example.

- Made some small tweaks to the colors used to more closely match our existing hover styles. Super subtle changes as what you had was already great.

- Refactored the component HTML and CSS a bit to simplify it. A summary:
	- Removed the nested `<span>`s to use `set:html` directly on the parent `<span>`
	- Wrapped each link in a `<div>` — this solves it being treated correctly with spacing in the document flow.
	- Removed some CSS that wasn’t having an effect anymore.

- Streamlined the docs as you suggested above: smaller code example showing with/without `description` and inside/outside `<CardGrid>`. Also simplified the text a bit.

Given I basically invented the rush in my head and we actually do have time to discuss, I do have a question: how important do we think it is for someone to have the ability to set `title` on the resulting links? I don’t love the API difference of `<Card title="Starlight" />` and `<LinkCard heading="Starlight" />` and I also can’t imagine a valid use case for the regular `title` attribute with this component. Output like the following would be odd right?

```html
<a title="Titled link" href="/">Link text</a>
```

packages/starlight/user-components/LinkCard.astro
```diff
@@ -0,0 +1,65 @@
+---
+import Icon from './Icon.astro';
+import type { HTMLAttributes } from 'astro/types';
+
+interface Props extends HTMLAttributes<'a'> {
+	heading: string;
+	description?: string;
+}
```
delucis, line 8:
```suggestion
interface Props extends Omit<HTMLAttributes<'a'>, 'title'> {
	title: string;
	description?: string;
}
```

packages/starlight/user-components/LinkCard.astro
```diff
@@ -0,0 +1,65 @@
...
+
+const { heading, description, ...attributes } = Astro.props;
```
delucis, line 10:
```suggestion
const { title, description, ...attributes } = Astro.props;
```

packages/starlight/user-components/LinkCard.astro
```diff
@@ -0,0 +1,65 @@
...
+---
+
+<div>
+	<a {...attributes}>
+		<span class="flex stack">
+			<span class="title" set:html={heading} />
```
delucis, line 16:
```suggestion
			<span class="title" set:html={title} />
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -125,6 +125,44 @@ Add the `stagger` attribute to shift the second column of cards vertically and a
 
 :::
 
+### Link Cards
+
+Use the `<LinkCard>` component to link prominently to different pages.
+
+A `<LinkCard>` requires a `heading` and an [`href`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href) attribute. You can optionally include a short `description` or other link attributes such as `target`.
```
delucis, line 132:
```suggestion
A `<LinkCard>` requires a `title` and an [`href`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href) attribute. You can optionally include a short `description` or other link attributes such as `target`.
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -125,6 +125,44 @@ Add the `stagger` attribute to shift the second column of cards vertically and a
...
+<LinkCard
+	heading="Customizing Starlight"
+	description="Learn how to make your Starlight site your own with custom styles, fonts, and more."
+	href="/guides/customization/"
+/>
+
+<CardGrid>
+	<LinkCard heading="Authoring Markdown" href="/guides/authoring-content/" />
+	<LinkCard heading="Components" href="/guides/components/" />
+</CardGrid>
```
delucis, line 164:
````suggestion
```mdx
import { LinkCard, CardGrid } from '@astrojs/starlight/components';

<LinkCard
	title="Customizing Starlight"
	description="Learn how to make your Starlight site your own with custom styles, fonts, and more."
	href="/guides/customization/"
/>

<CardGrid>
	<LinkCard title="Authoring Markdown" href="/guides/authoring-content/" />
	<LinkCard title="Components" href="/guides/components/" />
</CardGrid>
```

The above code generates the following on the page:

import { LinkCard } from '@astrojs/starlight/components';

<LinkCard
	title="Customizing Starlight"
	description="Learn how to make your Starlight site your own with custom styles, fonts, and more."
	href="/guides/customization/"
/>

<CardGrid>
	<LinkCard title="Authoring Markdown" href="/guides/authoring-content/" />
	<LinkCard title="Components" href="/guides/components/" />
</CardGrid>
````

delucis (commented):
If people agree we can stick with `title`, we can make these changes:

HiDeoo:
> If people agree we can stick with `title`, we can make these changes:

I've seen [this technique](https://www.w3.org/TR/WCAG20-TECHS/H33.html) used a lot but it's not super recommended (even in this link) so personally I think it should be fine.

delucis:
Great link — thank you! Yeah, I think that case should be covered by being able to add a `description`. The other generic use case I can imagine is things like `<a title="Page about stuff">Read more</a>` but that too seems like a bad idea to me and not really relevant to this component.

I’m generally in favour of “the most accessible text is visible and accessible to all technologies” an a component restricting usage a little bit to encourage correct output also seems part of the motivation for making components, so leaning towards API consistency here.

delucis (approved):
OK, let’s do it! Made that change and will get this out in the next release.

Thanks again for the great work here @lorenzolewis!

lorenzolewis:
Woot! Then I can work on my next evil plan for a Table of Contents component 😈 😆 (will open an issue to discuss first <3 )

PR #399, opened 2023-07-25
Add favicon configuration

### LINKED ISSUE #363: Enable support for non-svg format favicons

0.5.4

### What is your idea?

Add favicon file path as a configuration setting

svg is not supported by all browsers. Visiting a starlight docs site in Safari for example does not load any favicon.

https://caniuse.com/link-icon-svg

_No response_

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #363.

This pull request adds a new `favicon` option to the `starlight` integration. This option allows you to specify the path of the default favicon for your site. If not specified, it defaults to `'/favicon.svg'` which means that the current behavior is unchanged.

I decided to use a `transform` on the schema to do the validation and inference of the `type` at the same place which means `config.favicon.href` contains the path to the favicon while `config.favicon.type` contains the inferred type of the favicon.

Note that properly testing this in Safari is not as easy as in other browsers due to Safari hyper agressive favicon cache. To test this, you need to clear this cache:

1. Enable the Develop menu in Safari (Safari > Settings > Advanced > Show Develop menu in menu bar)
1. Clear some caches (Develop > Empty Caches)
1. Quit Safari
1. Remove the favicon cache folder (`rm -rf ~/Library/Safari/Favicon\ Cache`)
1. Start Safari


docs/src/content/docs/reference/configuration.md
```diff
@@ -379,3 +379,29 @@ By default, this feature relies on your repository’s Git history and may not b
 Define if the footer should include previous and next page links.
 
 A page can override this setting or the link text and/or URL using the [`prev`](/reference/frontmatter/#prev) and [`next`](/reference/frontmatter/#next) frontmatter fields.
+
+### `favicon`
+
+**type:** `string`  
+**default:** `'/favicon.svg'`
+
+Set the path of the default favicon for your website.
```
lorenzolewis, line 388: Do you want to specify the accepted file extensions here?

lorenzolewis:
Were you wanting to do anything about that pesky Safari mask icon with this: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html

Or is that a rabbit hole Starlight doesn't want to venture down? 😅

docs/src/content/docs/reference/configuration.md
HiDeoo: Hmm, that is a good point. I wonder tho if we should explicitly list them or link to a reference of some sort, e.g. this one https://en.wikipedia.org/wiki/Favicon#Image_file_format_support ?

HiDeoo:
> Were you wanting to do anything about that pesky Safari mask icon

I think the issue with `mask-icon` is that the `.svg` file must be a single color, and Safari will overlay that with the color attribute that you provide.

lorenzolewis:
> I think the issue with `mask-icon` is that the `.svg` file must be a single color, and Safari will overlay that with the color attribute that you provide.

Ah yeah, forgot about that one. And contrast against dark/light mode chrome also makes this impossible to derive. I wonder if this might be a reference to include as well in case someone thinks it's a bug?

packages/starlight/__tests__/basics/head.test.ts
```diff
@@ -63,4 +63,29 @@ describe('createHead', () => {
...
+    const userFavicon = {
+      tag: 'link',
+      attrs: {
+        rel: 'icon',
+        href: '/favicon.ico',
+        sizes: '32x32'
+      }
+    } as const
+
+    expect(createHead([defaultFavicon, userFavicon])).toMatchObject([
```
delucis, line 86: The way the default and custom head configs are handled in `HeadSEO.astro` is actually more like this:

```suggestion
    expect(createHead([defaultFavicon], [userFavicon])).toMatchObject([
```

packages/starlight/__tests__/basics/head.test.ts
```diff
@@ -63,4 +63,29 @@ describe('createHead', () => {
...
+    } as const
+
+    const userFavicon = {
+      tag: 'link',
+      attrs: {
+        rel: 'icon',
+        href: '/favicon.ico',
+        sizes: '32x32'
+      }
+    } as const
```
delucis, line 84: This one will also need an explicit `content` set too, because `createHead` expects user arrays to already have been parsed.
```suggestion
    const userFavicon = {
      tag: 'link',
      attrs: {
        rel: 'icon',
        href: '/favicon.ico',
        sizes: '32x32'
      },
      content: '',
    } as const
```

packages/starlight/schemas/favicon.ts
```diff
@@ -0,0 +1,40 @@
...
+        return z.NEVER;
+      }
+
+      return {
+        href: favicon,
+        type: faviconTypeMap[ext],
+      };
+    })
+    .describe(
+      'The default favicon for your site which should be a path to an image in the `public` directory.'
```
delucis, line 35: Super tiny nit given most people won’t see this but we usually use a trailing slash when mentioning directories:

```suggestion
      'The default favicon for your site which should be a path to an image in the `public/` directory.'
```

packages/starlight/utils/head.ts
```diff
@@ -89,7 +89,19 @@ function getImportance(entry: HeadConfig[number]) {
...
+    // want to use the SVG icon when supported.
+    if (
+      entry.tag === 'link' &&
+      'rel' in entry.attrs &&
+      entry.attrs.rel === 'shortcut icon'
+    ) {
+      return 70;
+    }
+    return 80;
+  };
```
delucis, line 104: Not sure if this is sufficient given the test fix above — we’ll find out!

packages/starlight/utils/user-config.ts
```diff
@@ -267,6 +268,9 @@ const UserConfigSchema = z.object({
     .describe(
       'Define if the previous and next page links should be visible in the page footer.'
     ),
+
+  /** The default favicon for your site which should be a path to an image in the `public` directory. */
```
delucis, line 272: As above:
```suggestion
  /** The default favicon for your site which should be a path to an image in the `public/` directory. */
```

docs/src/content/docs/reference/configuration.md
```diff
@@ -379,3 +379,29 @@ By default, this feature relies on your repository’s Git history and may not b
...
+      tag: 'link',
+      attrs: { 
+        rel: 'icon',
+        href:'/images/favicon.ico',
+        sizes: '32x32',
+      },
+    },
+  ],
+});
+```
```
delucis, line 407: Suggestion to first show the simple case and then separately the version with fallbacks to underline that using `head` is not required:


````suggestion
Set the path of the default favicon for your website which should be located in the `public/` directory and be a valid (`.ico`, `.gif`, `.jpg`, `.png`, or `.svg`) icon file.

```js
starlight({
  favicon: '/images/favicon.svg',
}),
```

If you need to set additional variants or fallback favicons, you can add tags using the [`head` option](#head):

```js
starlight({
  favicon: '/images/favicon.svg',
  head: [
    // Add ICO favicon fallback for Safari.
    {
      tag: 'link',
      attrs: { 
        rel: 'icon',
        href:'/images/favicon.ico',
        sizes: '32x32',
      },
    },
  ],
}),
```
````

delucis (commented):
Thank you for picking this up @HiDeoo! I think this mostly looks good except I think the test is using `createHead` differently from how it is used in practice, so need to double check if updating that throws up any work still needed re: order of favicons.

HiDeoo:
Thanks for the thorough review as always. Great catch on the `createHead` arguments. I don't "think" this could change the output as sorting is done at the end but this is now properly tested thanks to your review.

delucis (approved):
Thanks @HiDeoo — looks good! I guess it’s a tiny bit odd that we’re depending on Starlight using `rel="shortcut icon"` and users using `rel="icon"` (also because our own use of `shortcut` is apparently no longer recommended/needed), but I’m OK with that for now. We can always revisit later if it becomes an issue.

PR #412, opened 2023-07-26
Add copy to clipboard to docs icons

vasucp1207:
Copy icon to clipboard

clipboard


docs/src/components/icons-list.astro
```diff
@@ -9,15 +9,40 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
...
+			iconText?.classList.toggle('icon-name');
+
+			intervalId = setTimeout(() => {
+				confirm?.classList.toggle('copied');
+				iconText?.classList.toggle('icon-name');
+			}, 1000)
+			
+    });
+		clearInterval(intervalId);
+  });
```
Yan-Thomas, line 43: To work with our new approach, we have to change the script to instead of toggling between two spans, use only one and change its `textContent` property through the click event.

I've also made a few additional changes:
1. To be more expressive and clear, instead of getting the icon's text through the children array, I used a querySelector based on the icon's id and span element.
2. The `clearInterval` function doesn't work with `setTimeout`, for that we'd use `clearTimeout`, and with our current implementation, we don't need to use it.
```suggestion
  const icons = document.querySelectorAll('.icon-preview');
  icons.forEach((icon) => {
    icon.addEventListener('click', () => {
      const iconName = icon.id;
      const copiedValue = `<Icon name="${iconName}" size="2rem" />`;
      navigator.clipboard.writeText(copiedValue);

      const iconLabel = document.querySelector(`#${iconName} span`);
      iconLabel!.textContent = 'Copied!';

      setTimeout(() => {
        iconLabel!.textContent= iconName;
      }, 1000);
    });
  });
```

docs/src/components/icons-list.astro
```diff
@@ -37,4 +62,18 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
 		margin: 0;
 		padding: 0.75rem;
```
Yan-Thomas, line 63: Now the `<div>` has been changed for a `<button>`, we need to reset its default background properties.
```suggestion
		padding: 0.75rem;
		background: none;
```

docs/src/components/icons-list.astro
```diff
@@ -37,4 +62,18 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
...
+		background: var(--sl-color-gray-5);
+	}
+
+	.copied {
+		display: none;
+	}
+
+	.icon-name {
+		display: none;
+	}
```
Yan-Thomas, line 78: Now that the element is focusable through keyboard, we can add the same effect from when it is hovered over. And with the code changes made, we can also remove the `.copied` and `.icon-name` classes.
```suggestion
	.icon-preview:hover, .icon-preview:focus {
		cursor: pointer;
		border: 1px solid var(--sl-color-gray-4);
		background: var(--sl-color-gray-5);
	}
```

Yan-Thomas (changes requested):
Thanks @vasucp1207, this is super useful and lovely!

I see the current implementation has some issues in regard to accessibility:
1. The icon card can't be used only using the keyboard nor is properly announced to screen readers.
2. The dynamically shown "copied" text isn't announced to screen readers.

Making accessibility improvements might seem daunting or complicated, so do not worry, I'll help you out with code suggestions and explainers below!

vasucp1207:
@Yan-Thomas, thanks for your suggestions.

docs/src/components/icons-list.astro
```diff
@@ -9,15 +9,33 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
...
 		})
 	}
 </div>
 
+<script>
+  const icons = document.querySelectorAll('.icon-preview');
+  icons.forEach((icon) => {
+    icon.addEventListener('click', () => {
+      const iconName = icon.id;
+      const copiedValue = `<Icon name="${iconName}" size="2rem" />`;
```
delucis, line 26: Do you think `size` is important to copy? It’s an optional prop (defaults to `1em`), so maybe we can keep this simple and just copy the component and name?

```suggestion
      const copiedValue = `<Icon name="${iconName}" />`;
```

docs/src/components/icons-list.astro
```diff
@@ -9,15 +9,33 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
...
+      navigator.clipboard.writeText(copiedValue);
+
+      const iconLabel = document.querySelector(`#${iconName} span`);
+      iconLabel!.textContent = 'Copied!';
+
+      setTimeout(() => {
+        iconLabel!.textContent= iconName;
+      }, 1000);
```
delucis, line 34: Might be best to avoid the `!` assertion here and just move the related code behind an `if` check. That way if for some reason the span isn’t found, nothing crashes.

We can also update the `querySelector` usage to query directly inside the `icon` which both simplifies our code and should perform a bit better. I’ve updated it to explicitly look for our `aria-live` attribute — that means our visual behaviour should break if we break our screen reader behaviour, which is a good way to help prevent accessibility regressions.

```suggestion
      const iconLabel = icon.querySelector('[aria-live]');
      if (iconLabel) {
        iconLabel.textContent = 'Copied!';
        setTimeout(() => {
          iconLabel.textContent = iconName;
        }, 1000);
      }
```

docs/src/components/icons-list.astro
```diff
@@ -9,15 +9,33 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
...
 			);
 		})
 	}
 </div>
 
+<script>
+  const icons = document.querySelectorAll('.icon-preview');
+  icons.forEach((icon) => {
+    icon.addEventListener('click', () => {
+      const iconName = icon.id;
```
delucis, line 25:
```suggestion
      const iconName = icon.dataset.name;
```

delucis (commented):
Thank you for this lovely addition @vasucp1207! And for the thorough accessibility review @Yan-Thomas 💜 

Left a few code suggestions. Really appreciate you taking the time to make this improvement.

docs/src/components/icons-list.astro
```diff
@@ -9,15 +9,33 @@ const icons = Object.keys(Icons) as (keyof typeof Icons)[];
 	{
 		icons.map((icon) => {
 			return (
-				<div class="icon-preview">
+				<button class="icon-preview" aria-label={`Copy ${icon} icon to clipboard`} id={icon}>
```
delucis, line 12: Oops, comment here went missing.

We can use a `data-*` attribute instead of an `id` to store the icon name. Not a huge difference in practice, but in general I prefer using them for data like this we want to use.

```suggestion
				<button class="icon-preview" aria-label={`Copy ${icon} icon to clipboard`} data-name={icon}>
```

delucis (approved):
Thanks again for this super useful PR @vasucp1207!

PR #419, opened 2023-07-27
Improve styling of sidebar entries that wrap

lorenzolewis:

- Changes to Starlight code

#### Description

If a sidebar entry wraps to a new line then it can be difficult to distinguish that entry from the ones above/below it. This puts in a fix to visually indent any subsequent lines.

Before:
Screenshot 2023-07-27 at 14 17 49
Screenshot 2023-07-27 at 14 23 08

After:
Screenshot 2023-07-27 at 14 17 37
Screenshot 2023-07-27 at 14 23 17


delucis (changes requested):
Thanks for this @lorenzolewis! I’ve noticed this issue before. I actually think the issue could be better addressed by a narrower line height? That would mean the wrapped lines nestle closer together and form a single visual unit more clearly. Been meaning to tackle that.

I think the requirement would be: 
- narrower line-height
- a little extra vertical padding so that the current appearance for single-line links is preserved.

lorenzolewis:
Sure, I can get behind that. I'll play around with it a bit to dial it in and update

lorenzolewis:
Screenshot 2023-07-28 at 21 52 49
Screenshot 2023-07-28 at 21 52 45

@delucis here's those changes, feel free to dial them as you see fit. I tried to keep the overall vertical height of an item as close as possible.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -87,7 +87,8 @@ interface Props {
 		border-radius: 0.25rem;
 		text-decoration: none;
 		color: var(--sl-color-gray-2);
-		padding: 0.1em var(--sl-sidebar-item-padding-inline);
+		padding: 0.3em var(--sl-sidebar-item-padding-inline);
+		line-height: 1.4em;
```
delucis, line 91: Nit but we generally use unit-less line-height values (has the same effect anyway):

```suggestion
		line-height: 1.4;
```

delucis (approved):
That was quick! Looks fantastic, thank you @lorenzolewis. One stylistic nit but that’s all I got.

PR #425, opened 2023-07-29
Add my site to the showcase

dhruvkb:

- Something else!

#### Description

This PR adds a documentation site for a Rust tool that's built with Starlight/Astro to the showcase. The tool itself has about ~600 stars (hopefully more once people see how cool the docs are!).

Hopefully a fuller showcase can convince more people to make the switch. This site is a bit different from the more cookie-cutter sites in the showcase so I hope it qualifies.


delucis (approved):
Looks fantastic — thanks for sharing @dhruvkb!

PR #434, opened 2023-07-29
Add dev container configuration

fflaten:

- Something else!

#### Description

Adds a dev container to the repo to make it easier to contribute using VSCode, Github Codespaces etc.
Also updates contributing docs to include Github Codespaces.

Both based on similar setup in astro-repo.


.devcontainer/welcome-message.txt
```diff
@@ -0,0 +1,3 @@
+🌟 Welcome to Starlight!
+
+🛠️  Your environment is fully setup with all required software installed.
```
fflaten, line 3: Double space is intentional. First space doesn't show properly with this emoji.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "Contribute to Starlight",
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "postCreateCommand": "pnpm install",
+  "waitFor": "postCreateCommand",
+  "customizations": {
+    "codespaces": {
+      "openFiles": ["README.md", "CONTRIBUTING.md"]
```
delucis, line 10: I think it may be helpful to just show the contributor guide here. There’s not much of use in the README:

```suggestion
      "openFiles": ["CONTRIBUTING.md"]
```

CONTRIBUTING.md
```diff
@@ -38,9 +38,16 @@ This repo is a “monorepo,” meaning it contains several projects in one. It c
...
+
 #### Developing locally
 
-**Prerequisites:** Developing Starlight requires [Node.js](https://nodejs.org/en) and [pnpm](https://pnpm.io/). Make sure you have these installed before following these steps.
+**Prerequisites:** Make sure you have these installed before continuing.
+
+```shell
+node: "^>=16.12.0"
+pnpm: "^8.2.0"
+```
```
delucis, line 50: It’s a shame to lose the links as those both include install instructions for those who need them. How about this if we want to mention the supported versions?

```suggestion
**Prerequisites:** Developing Starlight requires [Node.js](https://nodejs.org/en) (v16 or higher) and [pnpm](https://pnpm.io/) (v8.2 or higher). Make sure you have these installed before following these steps.
```

CONTRIBUTING.md
```diff
@@ -62,9 +69,18 @@ This repo is a “monorepo,” meaning it contains several projects in one. It c
    pnpm i
    ```
 
-#### Developing using Gitpod
+#### Developing using GitHub Codespaces
+
+To get started, create a codespace for this repository by clicking this 👇
 
-Instead of working locally on your machine, you can also contribute using an online coding development environment like Gitpod.
```
delucis, line 67: I think this line was important to distinguish the Gitpod instructions from the local ones.

How about a “Developing online” heading followed by a variation of this sentence before we get to the GitHub and Gitpod headings? (Those would then become h5 instead of h4.)

CONTRIBUTING.md
```diff
@@ -62,9 +69,18 @@ This repo is a “monorepo,” meaning it contains several projects in one. It c
...
+[![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://codespaces.new/withastro/starlight)
+
+Your new codespace will open in a web-based version of Visual Studio Code. All development dependencies will be preinstalled.
+
+> **Note**
+> Dev containers is now an open spec which is supported by [GitHub Codespaces](https://github.com/codespaces) and [other supporting tools](https://containers.dev/supporting).
+
```
delucis, line 82: I think we can remove this as it’s not really relevant to someone wanting to contribute.

```suggestion
```

delucis (commented):
Thanks @fflaten! Were you able to work with this set-up? I tried `cd docs && pnpm dev` on your branch in a codespace, but wasn’t able to load the dev server only getting a 502 error.

CONTRIBUTING.md
fflaten: I personally agree. Changed to be consistent with https://github.com/withastro/astro/blob/main/CONTRIBUTING.md#prerequisites .

.devcontainer/devcontainer.json
fflaten: Sounds good. It also looks a bit weird because of the symlink. Shows diff line marking even though it works and saves modifications to the correct file.

image

CONTRIBUTING.md
fflaten: Except providing a hint that the same devcontainer can be used in other ways, like VSCode with Docker Desktop (how I work).

Also based on https://github.com/withastro/astro/blob/main/CONTRIBUTING.md#using-github-codespaces-for-development

fflaten:
> Thanks @fflaten! Were you able to work with this set-up? I tried `cd docs && pnpm dev` on your branch in a codespace, but wasn’t able to load the dev server only getting a 502 error:
> 
> Screenshot of GitHub codespaces 502 error

Oh right, you need `--host`

.devcontainer/devcontainer.json
fflaten: Awesome. Thanks for the tip!

delucis (approved):
Thanks again for the contribution @fflaten! Took a quick pass at making the codespaces instructions align more closely with the style of our Gitpod instructions and now LGTM! 🙌

PR #436, opened 2023-07-31
Add support for mobile safe areas

### LINKED ISSUE #433: Support safe area on mobile devices

0.6.0

### What is your idea?

Update default styling to stretch beyond "safe area" on devices so styling is consistent on desktop and ex. mobile devices in landscape view.

Required changes:
- Include `viewport-fit=cover` in the `viewport` meta-tag
- Increase padding on `body` or `.page` + any fixed-positioned items like `<header>` to avoid conflicts with UI-elements/camera on devices
- Might need to adjust mobile/desktop breakpoints

CSS environment-variables named `safe-area-inset-*` exist to dynamically provide the correct padding-values per device.

_Copied from https://github.com/withastro/docs/issues/2270. Same result in a new Starlight-project_

On touch devices like iPhones, iPads etc. the website is padded by the browser to avoid content under UI-elements like the "slide for home screen"-bar, camera notch etc. In this area the gradient background is not visible and the website might also get double padding in some elements close to the edges.

image

- https://developer.mozilla.org/en-US/docs/Web/CSS/env#usage
- https://github.com/pester/docs/pull/234/files (example of similar adjustments in a Docusaurus-project)

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #433.

This pull request is an attempt to add support for mobile safe areas to Starlight. It is a draft for two main reasons:

- I think it would be nice to have more people test it as I don't have that many devices with notches to test it on, and it's fairly easy to forget some edge cases.
- As mentioned in #433, we may want to tweak some breakpoints.

To talk a bit about the changes, the syntax to support mobile safe areas is pretty verbose and it's even more verbose when we use some logical CSS properties as there is no equivalent for the safe area inset properties, they are only physical.

When dealing with safe areas I always refer to this [blog post](https://webkit.org/blog/7929/designing-websites-for-iphone-x/) that was released by the Webkit team when the iPhone X with its notch was released which I think does a good job at explaining how it works so sharing it here in case it can be useful and I also included it in the changeset.

Here are some screenshots from a device with a notch, starting with the home page:

|                        Before                        |                        After                        |
| :--------------------------------------------------: | :-------------------------------------------------: |
| home page before | home page after |

A doc page:

|                       Before                        |                       After                        |
| :-------------------------------------------------: | :------------------------------------------------: |
| doc page before | doc page after |

A doc page with the toc opened:

|                            Before                            |                            After                            |
| :----------------------------------------------------------: | :---------------------------------------------------------: |
| doc page with toc before | doc page with toc after |

Even if we somehow forget some edge cases, I think it'll be nice to have support for this in Starlight and we can always fix the edge cases when we find them.


HiDeoo:
> If only the device emulator in Chromium included option for safe area. So painful to use physical devices for testing this. Please let me know if there's an option I haven't found yet :)

Yeah, this is a pain. The only other "option" I have found for iOS is having a mac, installing the **huge** bundle that is Xcode just to get the iOS emulators, cry at every update that it takes forever to update, run manually an emulator for a device with a notch, start Safari on the simulator and visit the webpage.

delucis:
Thanks for tackling this @HiDeoo! Yeah, I also have no devices with this and will have to open up the iOS simulator. Pretty gross that this requires SO MUCH CSS for their dumb design flaw 😅 

Haven’t had a chance to dive in to the code in detail yet, so this may not work, but one trick I sometimes reach for when media queries are getting out of hand like this is to see if there’s a way to move things into a reusable custom property, e.g. this kind of concept:

```css
:root {
  /* default value */
  --some-prop: 1rem;
}

@media (prefers-notchy-screen: true) {
  :root {
    /* value when condition is met */
    --some-prop: 1.5rem;
  }
}
```

Then in most styles you just use your value without worrying about media queries. Maybe that could help conciseness here too.

HiDeoo:
Excellent tip, I'll make a second pass tomorrow now that I have identified most of areas requiring changes and see if we can apply this in some of these cases.

HiDeoo:
Third pass done :sweat_smile:

I managed to remove most `@supports (padding: max(0px))` directives and only keep a single one in `packages/starlight/style/props.css`. It only hit me today that I could move some media queries to this file nested in this directive even tho I was already doing that in some components ^^ I guess the step back helped.

All references to `env(safe-area-inset-*)` and `max()` are properly located in this file and in this `@supports` directive so we should not hit any issues with some older browsers not supporting those.

We still have to rely unfortunately a lot more than before on `:global([dir='rtl'])` at various places and this is one thing I don't know how to avoid.

I also refactored some previous changes I introduced to reduce the number of changes overall and fix a padding that I previously forgot to adjust.

I think at this point I'll open this for review as I think I need some other eyes to look at this ^^ I'll be happy tho to fix any issue or places that I may have missed or try other approaches if some of you have other ideas.

delucis:
I think it’s time to put this PR out of its misery. We haven’t had time to really invest in doing this elegantly and it has become very out of date (and with CSS layers coming soon, it will be even _more_ out of date).

PR #441, opened 2023-08-01
Add frontmatter to hide page from sidebar autogeneration

### LINKED ISSUE #420: Set sidebar title and hide pages from autogenerated sidebar

0.5.6

### What is your idea?

Add the following frontmatter properties after https://github.com/withastro/starlight/pull/359 is merged that will help users shape sidebars autogenerated:
- `sidebar > hide` that will hide the entry from being rendered in the sidebar
- `sidebar > label` that will allow a user to specify the text they want to be used in a sidebar (can also make i18n for sidebar items much easier)

Deriving the sidebar from the filesystem can be super helpful for large collections, but right now Starlight doesn't give you much flexibility to shape the derived sidebar.

Adding `sidebar > label` could also make it easier to localise sidebar labels outside of the config. This could be coupled with https://github.com/withastro/starlight/issues/384 to help reduce the amount of page-specific data the config is required to contain and instead offload that to page-specific frontmatter.

https://docusaurus.io/docs/sidebar/autogenerated#doc-item-metadata

### Participation

lorenzolewis:

- Changes to Starlight code

#### Description

- Closes #420
- Add `sidebar > hidden` to the frontmatter schema (defaults to false) that will skip adding that entry to any autogenerated sidebars


.changeset/quiet-rules-drum.md
```diff
@@ -0,0 +1,13 @@
+---
+'@astrojs/starlight': minor
+---
+
+Add support for hiding entries fromt the autogenerated sidebar:
```
HiDeoo, line 5:
```suggestion
Add support for hiding entries from the autogenerated sidebar:
```

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -224,3 +224,18 @@ sidebar:
...
+**type:** `boolean`
+**default:** `false`
+
+Hides this page from being included in an autogenerated sidebar group.
+
+```md
+---
+title: Page to hide from autogenerated sidebar
+sidebar:
+  hidden: 1
```
HiDeoo, line 239: Shouldn't this be `true` ?

packages/starlight/utils/navigation.ts
```diff
@@ -79,13 +79,16 @@ function groupFromAutogenerateConfig(
...
+	const dirDocs = routes
+		.filter(
+			(doc) =>
+				// Match against `foo.md` or `foo/index.md`.
+				stripExtension(doc.id) === localeDir ||
+				// Match against `foo/anything/else.md`.
+				doc.id.startsWith(localeDir + '/')
+		)
+		// Remove any entries that should be hidden
+		.filter((doc) => !doc.entry.data.sidebar.hidden);
```
HiDeoo, line 91: Any specific reason to not do that in the `filter` right above?

HiDeoo (commented):
Super neat, left a few comments.

> Hides this page from being included in an autogenerated sidebar group.

I wonder if we should somehow emphasize this is not a `draft` mode and that the page will still be generated. Or maybe it's already obvious and I'm a little bit paranoid.

docs/src/content/docs/reference/frontmatter.md
lorenzolewis: good catch, copy paste got me on that one

packages/starlight/utils/navigation.ts
lorenzolewis: At first I did that to make it more readable instead of nesting logic, but I think it'd be better to include it all so that the array isn't iterated over twice.

lorenzolewis:
Got those changes in.

RE: the phrasing, I tried to reference that it would hide it from any navigation/autogenerated sidebar in the docs so I feel like it avoids that confusion. But also open to any suggestions to adapt that.

packages/starlight/__tests__/sidebar/navigation-hidden.test.ts
```diff
@@ -0,0 +1,61 @@
...
+			['environmental-impact.md', { title: 'Eco-friendly docs' }],
+			['reference/configuration.md', { title: 'Config Reference' }],
+			['reference/frontmatter.md', { title: 'Frontmatter Reference', sidebar: { hidden: true } }],
+			['guides/components.mdx', { title: 'Components' }],
+		],
+	})
+);
+
+describe('getSidebar', () => {
+	test('returns sidebar entries sorted by frontmatter order', () => {
```
delucis, line 17: Test name needs updating here:

```suggestion
	test('excludes sidebar entries with hidden: true in frontmatter', () => {
```

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -224,3 +224,18 @@ sidebar:
   order: 1
 ---
 ```
+
+#### `hidden`
+
+**type:** `boolean`
+**default:** `false`
+
+Hides this page from being included in an autogenerated sidebar group.
```
delucis, line 233: Small wording tweak here:
```suggestion
Prevents this page from being included in an autogenerated sidebar group.
```

packages/starlight/schema.ts
```diff
@@ -142,6 +142,12 @@ export function docsSchema() {
 					 * Defaults to the page `title` if not set.
 					 */
 					label: z.string().optional(),
+
+					/**
+					 * Hides the page from any navigation that is autogenerated
+					 * Ignored if explicitly referenced in the sidebar config
```
delucis, line 148: I think we’re OK to leave out the second line here as we’re already saying when this applies in the first.

```suggestion
					 * Prevents this page from being included in autogenerated sidebar groups.
```

delucis (commented):
Looks good @lorenzolewis — thank you! Left some small comments, but implementation and testing all look spot on.

delucis (approved):
Thanks again @lorenzolewis! I think this looks good — we’ll get it out in the next minor release 🙌

PR #447, opened 2023-08-02
Add `titleDelimiter` option to the api and a new design for `<title>`

### LINKED ISSUE #400: Support site title to appear after page title

0.5.2

### What is your idea?

Starlight currently prints the title of a page as written in the frontmatter. I want to set it such that the site title appears after the page title. So a page called "Introduction" on my site actually renders the title "Introduction • Dhruv".

Site titles are a useful part of the title as that's visible in the tab bar. It helps to disambiguate two pages from different sites that have the same name.

The Astro documentation prints the site title after page title using this line in the layout:

https://github.com/withastro/docs/blob/17ff293f3f753404cce925b9c6f2d62edad6ca60/src/layouts/SplashLayout.astro#L24

As a Starlight user, I don't have access to change the layout, so I must at least have a config option to specify a format for `<title>` and Starlight populatesthe site title and page title into the format automatically.

### Participation

andremralves:
Closes #400

This PR adds a new option for Starlight's API, the `titleDelimiter`, and updates the `<title>` tag to use `${pageTitle} ${titleDelimiter} ${title}` by default. This behavior is similar to [how Docusaurus handles titles](https://docusaurus.io/docs/api/docusaurus-config#titleDelimiter).

Here is an example of the usage of this option:
```ts
export default defineConfig({
	integrations: [
		starlight({
			title: 'Starlight',
			titleDelimiter: '🌟' // Defaults to `-`
		}),
	],
});
```

The output will be `<title>pageTitle 🌟 Starlight</title>`.

I had to update the `index.mdx` in docs so the title keeps the same.

Other docs frameworks that I saw doesn't include an option in the API to remove the site title from the `title` tag. So I also thought about adding another option `titlePosition?: right|left|disabled`. What do you think?


packages/starlight/components/HeadSEO.astro
delucis, line 1: `title` is now always the site title but previously was the page title most of the time. We should update usages in the open graph and twitter meta tags to use `pageTitle`.

packages/starlight/utils/user-config.ts
```diff
@@ -271,6 +271,13 @@ const UserConfigSchema = z.object({
 
 	/** The default favicon for your site which should be a path to an image in the `public/` directory. */
 	favicon: FaviconSchema(),
+
+	/** Will be used as title delimiter in the generated `<title>` tag. */
+	titleDelimiter: z
+		.string()
+		.optional()
```
delucis, line 278: We can use Zod’s `default()` method here to set the default delimiter. Also suggesting we use a `|` pipe character instead of `-` to match Docusaurus:

```suggestion
		.default('|')
```

packages/starlight/components/HeadSEO.astro
```diff
@@ -15,7 +15,9 @@ interface Props {
 const { data, lang } = Astro.props;
 
 const canonical = Astro.site ? new URL(Astro.url.pathname, Astro.site) : undefined;
-const title = data.title || config.title;
+const title = config.title;
+const pageTitle = data.title;
+const titleDelimiter = config.titleDelimiter || '-';
```
delucis, line 20: And thanks to the default value, we can just use the delimiter directly here:

```suggestion
const titleDelimiter = config.titleDelimiter;
```

delucis (commented):
Thanks for this PR @andremralves and for your patience waiting for a review! I think on the whole this API seems really good.

One thing we were just discussing on our weekly Talking and Doc’ing call, was whether this should somehow handle homepages differently. Basically, should the override behaviour needed in `/index.mdx` in this PR be something we can handle better directly for the user? Not sure what that would look like though.

Instead, can we add the same `head` workaround you added to `index.mdx` to the index pages of our two `examples/`? Seems like a common customisation, so would be good to have it in place in the starter project.

I don’t think we need to worry about a `titlePosition` option. In general, displaying site title first (e.g. what `left` might do) is a bad practice.

andremralves:
> One thing we were just discussing on our weekly Talking and Doc’ing call, was whether this should somehow handle homepages differently. Basically, should the override behaviour needed in `/index.mdx` in this PR be something we can handle better directly for the user? Not sure what that would look like though.

I remember that in Docusaurus they use this behavior of `${siteTitle} ${delimiter} ${pageTitle}` for all pages, even the home page. Maybe we should consider doing the same. Then the home page would be: `Build documentation sites with Astro 🌟 Starlight`, that, for me, also looks good.

packages/starlight/components/HeadSEO.astro
```diff
@@ -24,7 +26,7 @@ const headDefaults: z.input<ReturnType<typeof HeadConfigSchema>> = [
 		tag: 'meta',
 		attrs: { name: 'viewport', content: 'width=device-width, initial-scale=1' },
 	},
-	{ tag: 'title', content: title },
+	{ tag: 'title', content: `${pageTitle} ${titleDelimiter} ${title}` },
```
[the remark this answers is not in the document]
andremralves: I don't have much experience with RTL languages. What could be different? Maybe the order?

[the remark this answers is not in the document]
delucis: I can do a test then. Yeah, basically that in RTL the reading order would be reversed. So should check that this template displays correctly as `title | pageTitle` rather than the opposite.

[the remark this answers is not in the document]
delucis: OK, did a test and this does seem to work as expected 👍

delucis:
> Maybe we should consider doing the same. Then the home page would be: `Build documentation sites with Astro 🌟 Starlight`

Personally, I prefer having the site title first in this case on the landing page. It makes it super clear from the tab which site you’re on and also looks good in search results. Obviously, is personal preference though. Perhaps there’s no need to update starter templates to include this workaround.

packages/starlight/components/HeadSEO.astro
delucis: Looks like this still needs updating. Could you make sure other parts of the code that referenced `title` now use `pageTitle`?

delucis (approved):
Thanks for kicking this off @andremralves! Just spent some time cleaning this up and adding documentation. Will get it out in our next release 🙌

PR #460, opened 2023-08-04
Sidebar current page highlight with trailing slashes

HiDeoo:

- Changes to Starlight code

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1070481941863878697/1136652592026497024) on Discord, this PR fixes the current page highlight in the sidebar for URLs with no trailing slash.

This [Discord thread](https://discord.com/channels/830184174198718474/1136696964281532417) also discusses why this PR does not handle Astro `trailingSlash` and `build.format` options for the sidebar `isCurrent` check which is purposefully lenient. These options should be handled in #207.

My initial suggestion on Discord was to use `const isCurrent = href === currentPathname.replace(/\/?$/, '/');` but instead I used the same approach as the existing `ensureLeadingAndTrailingSlashes()` function for multiple reasons:

- Consistency
- Might be slightly easier to understand for someone not used to regex
- According to this [bench](https://jsperf.app/kivusa/1/preview), this is faster. Not sure it's properly benchmarked or if it even matters in this case but might as well use the faster option ^^

### DIFF
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -6,6 +6,7 @@ import { pickLang } from './i18n';
 import { getLocaleRoutes, type Route } from './routing';
 import { localeToLang, slugToPathname } from './slugs';
 import type { AutoSidebarGroup, SidebarItem, SidebarLinkItem } from './user-config';
+import { ensureLeadingAndTrailingSlashes, ensureTrailingSlash } from './path';
 
 const DirKey = Symbol('DirKey');
 
@@ -97,19 +98,6 @@ function groupFromAutogenerateConfig(
 
 /** Check if a string starts with one of `http://` or `https://`. */
 const isAbsolute = (link: string) => /^https?:\/\//.test(link);
-
-/** Ensure the passed path ends with a trailing slash. */
-function ensureTrailingSlash(href: string): string {
-	if (href[href.length - 1] !== '/') href += '/';
-	return href;
-}
-
-/** Ensure the passed path starts and ends with trailing slashes. */
-function ensureLeadingAndTrailingSlashes(href: string): string {
-	if (href[0] !== '/') href = '/' + href;
-	href = ensureTrailingSlash(href);
-	return href;
-}
 
 /** Create a link entry from a user config object. */
 function linkFromConfig(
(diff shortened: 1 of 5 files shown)


packages/starlight/utils/navigation.ts
```diff
@@ -98,10 +98,16 @@ function groupFromAutogenerateConfig(
 /** Check if a string starts with one of `http://` or `https://`. */
 const isAbsolute = (link: string) => /^https?:\/\//.test(link);
 
+/** Ensure the passed path ends with a trailing slash. */
+function ensureTrailingSlash(href: string): string {
+	if (href[href.length - 1] !== '/') href += '/';
+	return href;
+}
```
delucis, line 105: Might it make sense to centralise these helpers? We also have these `strip` methods in our base utils:
https://github.com/withastro/starlight/blob/c344ee28b11dfeb951af3da21f3332ac7b202568/packages/starlight/utils/base.ts#L15-L20
If the non RegExp version is that much faster, we can refactor these.

delucis (commented):
Thanks @HiDeoo — LGTM! Left one refactor suggestion, but doesn’t have to be part of this PR if you don’t think it’s important.

packages/starlight/utils/navigation.ts
HiDeoo: Good catch, only saw the ones in the same file. I think we might as well do that now and be done with it while we're at it and it does not even need a changeset.

This would means we would have:

- `ensureLeadingSlash()`
- `ensureTrailingSlash()`
- `ensureLeadingAndTrailingSlashes()` _(using `ensureLeadingSlash()` and `ensureTrailingSlash()`)_
- `stripLeadingSlash()`
- `stripTrailingSlash()`

Now the big question ^^ Do we want to centralize them in `utils/navigation.ts`, `utils/base.ts` or a new file e.g. `utils/path.ts`, `utils/string.ts`?

For the regexp, I'll check with a new benchmark for stripping but like I said, I am not sure it's even properly benchmarked.

delucis: It probably doesn’t matter _hugely_ where these end up. I’d say they deserve a dedicated file, but don’t mind too much about naming!

I wouldn’t be surprised if the RegExp version is slower than the simple index lookups — after all they should require compiling and running the pattern vs a quick reference in the character array.

HiDeoo:
Moved all these helpers to a new dedicated file and indeed, as expected, according to the same benchmark tool, the stripping regexp version was also [slower](https://jsperf.app/tuvaxi/1/preview).

delucis (approved):
Thank you @HiDeoo!

PR #469, opened 2023-08-07
i18n(pt-BR): Initial setup + `getting-started.mdx`

Yan-Thomas:

#### Description

This PR starts the PT-BR translation for Starlight. By tomorrow, most pages will have open PRs already.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -9,6 +9,7 @@ export const locales = {
 	fr: { label: 'Français', lang: 'fr' },
 	it: { label: 'Italiano', lang: 'it' },
 	zh: { label: '简体中文', lang: 'zh' },
+	'pt-br': { label: 'Português do Brasil', lang: 'pt'}
 };
 
 const site = 'https://starlight.astro.build/';
@@ -60,6 +61,7 @@ export default defineConfig({
 						fr: 'Commencez ici',
 						it: 'Inizia qui',
 						zh: '从这里开始',
+						'pt-br': 'Comece Aqui',
 					},
 					items: [
 						{
@@ -72,6 +74,7 @@ export default defineConfig({
 								fr: 'Mise en route',
 								it: 'Iniziamo',
 								zh: '开始使用',
+								'pt-br': 'Introdução',
 							},
 						},
 						{
@@ -84,6 +87,7 @@ export default defineConfig({
 								// fr: 'Manual Setup',
 								// it: 'Manual Setup',
 								zh: '手动配置',
+								'pt-br': 'Instalação Manual',
 							},
 						},
 						{
@@ -96,6 +100,7 @@ export default defineConfig({
 								fr: 'Impact environnemental',
 								it: 'Impatto ambientale',
 								zh: '环境影响',
+								'pt-br': 'Impacto Ambiental',
 							},
 						},
 						{
@@ -120,6 +125,7 @@ export default defineConfig({
 						fr: 'Guides',
 						it: 'Guide',
 						zh: '指南',
+						'pt-br': 'Guias',
 					},
 					autogenerate: { directory: 'guides' },
 				},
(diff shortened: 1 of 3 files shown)


docs/astro.config.mjs
```diff
@@ -9,6 +9,7 @@ export const locales = {
 	fr: { label: 'Français', lang: 'fr' },
 	it: { label: 'Italiano', lang: 'it' },
 	zh: { label: '简体中文', lang: 'zh' },
+	'pt-br': { label: 'Português do Brasil', lang: 'pt'}
```
delucis, line 12: We use `pt-BR` in the lang tag in docs. Should that be the same here? Or do we just use `pt` and also drop the region specifier in the `label`?

delucis (commented):
Thanks @Yan-Thomas — looks good!

One additional set-up thing required in this repo: adding an entry for this subdirectory for the `i18n` label in `.github/labeler.yml`:

https://github.com/withastro/starlight/blob/54a76f70ae1949404271440a8450ad96c5372cc9/.github/labeler.yml#L6-L12

docs/astro.config.mjs
Yan-Thomas: Oh, that's right, technically the correct lang is `pt-BR` but our links in docs are `pt-br`, so I wasn't sure which one to use. I have no strong feelings about changing to just `pt`, though I'd still keep the label as "Brazilian Portuguese" for disambiguation, does that seem like a bad idea?

Alynva: Personally I prefer the specific version, mainly because in the past it was more common to find `pt-PT` translations when it was just `pt` so seing that gives me bad flashbacks, but I think nowadays it has changed. Google Translate, for example, does not distinguish between the two and there is only the `pt` language (with "Portuguese" label), and as far as I know, it is the `pt-BR` version for the most part. Idk what portuguese people think about that, there is already that debate around portugueses starting to speak only `pt-BR` and "losing" their native language, but its clear that the brazilian community has much more presence now than in the past.

docs/src/content/docs/pt-br/getting-started.mdx
```diff
@@ -0,0 +1,134 @@
+---
+title: Introdução
+description: Aprenda como começar a construir seu próximo site de documentação com Starlight pelo Astro.
+---
+
+import { Tabs, TabItem } from '@astrojs/starlight/components';
+
+:::caution[Trabalho em progresso]
+Starlight está em desenvolvimento antecipado.
```
Alynva, line 9: Não sei se "antecipado" fique adequado aqui, pois parece que é "antes da hora", e não faz muito sentido pra mim um desenvolvimento ser feito antes da hora. Talvez eu trocaria por "inicial", pra representar que não está finalizado

```suggestion
Starlight está em desenvolvimento inicial.
```

Alynva (changes requested):
LGTM, just a little suggestion!

docs/astro.config.mjs
delucis: I'll let you make the final call Yan, but just to clarify what the config does:

```
// 👇 Used as this language's directory and URL path — should be lowercase but doesn't *have* to be a correct BCP-47 tag
'pt-br': {
  // 👇 Used as the HTML Lang attribute for this language — should be a valid BCP-47 tag
  lang: 'pt-BR'
}
```
My only research was to see that Wikipedia only as a `pt` subdomain, uses `pt` as the `lang` tag and labels as "Português".

Alynva: Reading [this](https://meta.m.wikimedia.org/wiki/Requests_for_new_languages/Wikipedia_Brazilian_Portuguese) made me think about it, but idk the reasoning about `zh` and `zh-cn`

docs/src/content/docs/pt-br/getting-started.mdx
Yan-Thomas: Justo, eu usei "antecipado" mas por lembrar da interface da Steam que traduz assim para "Early Access", mas inicialmente eu também pensei em inicial.

Alynva (approved):
LGTM!

delucis:
Was about to approve and then BOOM merged! Thanks for taking the initiative and adding this @Yan-Thomas and thank you for reviewing @Alynva 🌟

PR #473, opened 2023-08-08
Fix issue with nested `<Tabs>` components

HiDeoo:

- Changes to Starlight code

#### Description

This pull request fixes an issue [reported](https://discord.com/channels/830184174198718474/1070481941863878697/1138524019361075322) on Discord regarding nested tabs.

The panels `hidden` property was not being set correctly due to the query selector matching all nested panels instead of just the direct children.

I updated it to use the [:scope CSS pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope#direct_children) to only match the direct children.

I don't think we have a proper way to test this yet so here is a video of a manual test:

https://github.com/withastro/starlight/assets/494699/0c89c448-0daf-4b74-a8b2-c57871504f43

### DIFF
--- a/packages/starlight/user-components/Tabs.astro
+++ b/packages/starlight/user-components/Tabs.astro
@@ -77,7 +77,7 @@ const { html, panels } = processPanels(panelHtml);
 			super();
 			const tablist = this.querySelector<HTMLUListElement>('[role="tablist"]')!;
 			this.tabs = [...tablist.querySelectorAll<HTMLAnchorElement>('[role="tab"]')];
-			this.panels = [...this.querySelectorAll<HTMLElement>('[role="tabpanel"]')];
+			this.panels = [...this.querySelectorAll<HTMLElement>(':scope > [role="tabpanel"]')];
 
 			this.tabs.forEach((tab, i) => {
 				// Handle clicks for mouse users
--- a/.changeset/wise-boats-flow.md
+++ b/.changeset/wise-boats-flow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fix issue with nested `<Tabs>` components


packages/starlight/user-components/Tabs.astro
```diff
@@ -77,7 +77,7 @@ const { html, panels } = processPanels(panelHtml);
 			super();
 			const tablist = this.querySelector<HTMLUListElement>('[role="tablist"]')!;
 			this.tabs = [...tablist.querySelectorAll<HTMLAnchorElement>('[role="tab"]')];
```
delucis, line 79: Don’t we need a similar approach for this selector too?

delucis (commented):
Thanks for the quick fix @HiDeoo! First time I’ve seen `:scope` — super handy.

I’ll admit seeing your demo gives me UI anxiety — not sure anyone _should_ be using nested tabs 😅 — but if people want to I guess we enable them 😆 

One quick question: did you test the keyboard navigation by any chance to make sure it also works as expected with this set-up?

packages/starlight/user-components/Tabs.astro
HiDeoo: We actually do not (which is quite neat as those are not direct children and would need longer selectors):

- The first one is using `querySelector()` and not `querySelectorAll()` so it only return the first result which is what we want.
- The second one is re-using the result of the first one so it's also what we want.

Basically only the last one was an issue as it was based on the component (`this`) and was using `querySelectorAll()`.

HiDeoo:
> First time I’ve seen `:scope` — super handy

Ah yeah definitely not the most common pseudo-class ^^

> I’ll admit seeing your demo gives me UI anxiety

I lost myself in the tabs making that video 😆 

> did you test the keyboard navigation by any chance to make sure it also works as expected with this set-up?

I actually did and forgot to mention it. Here is a new video of only keyboard navigation which work as expected:

https://github.com/withastro/starlight/assets/494699/ac5debb1-2ae2-4a6f-adf8-1dc1516dd600

delucis (approved):
This fix is much more lovely than nested tabs deserve. Thanks for that and the clear tests and explanations. Great work @HiDeoo 🙌

PR #475, opened 2023-08-08
Include built-in UI translation for regional docs

Yan-Thomas:

- Changes to Starlight code

#### Description

This PR makes it possible for regional languages, e.g. `pt-BR` (Brazilian Portuguese) to use the built-in UI translation for the main language (e.g `pt`).

I've intentionally made that this would only work for the built-in translations, so you still can create an `i18n/pt-BR.json` file and override the built-in translation in case you want.

I've made a fairly simple case-insensitive function looking for the expected BCP-47 tag, we could throw an error if it isn't a valid string or make some more complex type checking for it, but I went simple to not overengineer for now.

TO-DO: docs! (let's see if we're going to change the API or something before)

#### Testing

The feature was tested with a copy branch of https://github.com/withastro/starlight/pull/469/, with the following steps:
- Add `getting-started.mdx` translation for `pt-BR`, the UI elements should be translated with the built-in UI i18n ✅ 
- Add `i18n/pt-BR.json`, the UI elements should ignore the built-in UI and use the user-provided file instead ✅


packages/starlight/utils/translations.ts
```diff
@@ -33,7 +43,11 @@ const defaults = buildDictionary(
 export function useTranslations(locale: string | undefined) {
 	// TODO: Use better mapping, e.g. so that `en-GB` matches `en`.
```
delucis, line 44: We can delete this now 🥳 
```suggestion
```

delucis (commented):
This looks great @Yan-Thomas! Would be up for writing a couple of tests for the cases you mentioned?

You could expand on [`translations.test.ts`](https://github.com/withastro/starlight/blob/6a7692ae3178f9f9f727cc17b8ae860604afd78f/packages/starlight/__tests__/basics/translations.test.ts) and [`translations-with-user-config.test.ts`](https://github.com/withastro/starlight/blob/6a7692ae3178f9f9f727cc17b8ae860604afd78f/packages/starlight/__tests__/basics/translations-with-user-config.test.ts).

Let me know if you have questions about this. In the test with user config, the fake i18n dictionaries are created here:

https://github.com/withastro/starlight/blob/6a7692ae3178f9f9f727cc17b8ae860604afd78f/packages/starlight/__tests__/basics/translations-with-user-config.test.ts#L5-L9

packages/starlight/utils/translations.ts
```diff
@@ -19,10 +19,20 @@ try {
 const defaults = buildDictionary(
 	builtinTranslations.en!,
 	userTranslations.en,
-	builtinTranslations[defaultLocale],
+	builtinTranslations[stripLangRegion(defaultLocale)],
```
delucis, line 22: Oh actually, one thought: what if we had a built-in language with a regional specifier. We would no longer match it here right?

Thinking for example of Traditional Chinese. If that was built in at `zh-TW` and a user uses `lang: zh-TW`, it would get truncated and they’d get our `zh` strings instead I think?

delucis: Maybe this kind of approach?
```suggestion
	builtinTranslations[defaultLocale] || builtinTranslations[stripLangRegion(defaultLocale)],
```

Yan-Thomas: Good catch, yeah, I think that should do the trick!

delucis: Great. And then I guess the same applies the second place we use `stripLangRegion` too.

Yan-Thomas:
> Would you be up for writing a couple of tests for the cases you mentioned?

Sure, will take care of making some tests!

> Re: docs, I don’t know how much are needed. In general, I don’t think there’s new API here, just smoothing over use of regional specifiers.

Yeah, I feel like probably just a note on the internationalization page would suffice! I am leaving for today but will add that as part of the PR tomorrow!

delucis (changes requested):
Adding a note so I don’t forget — this needs a changeset. I’d suggest this is a patch as it fixes usage that is arguably expected.

Yan-Thomas:
Note that I haven't added tests to ensure the built-in regional variant gets precedence over the base language one, it seems we don't even have support for them yet, since the `translations/index.ts` implementation won't allow for dashes to match with the language.

```js
// import statements can't have dashes
// and renaming to something like `zhTw`
// won't match with the user-provided lang.
import zh-TW from './zh-TW.json';

const { parse } = builtinI18nSchema();

export default Object.fromEntries(
	Object.entries({ cs, en, es, de, ja, pt, fr, it, nl, da, tr, ar, nb, zh }).map(([key, dict]) => [
		key,
		parse(dict),
	])
);
```

.changeset/two-experts-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Now regional language variants (e.g. `pt-BR`) receives its base language's (e.g. `pt`) built-in translations by default.
```
delucis, line 5: I think we can expand on this slightly to be extra clear:
```suggestion
Locales whose language tag includes a regional subtag now use built-in UI translations for their base language. For example, a locale with a language of `pt-BR` will use our `pt` UI translations.
```

docs/src/content/docs/guides/i18n.mdx
```diff
@@ -145,6 +145,10 @@ In addition to hosting translated content files, Starlight allows you to transla
 
 English, Czech, French, German, Italian, Japanese, Portuguese, Dutch, Danish, Spanish, Turkish, Arabic, Norwegian, and Simplified Chinese translated UI strings are provided out of the box, and we welcome [contributions to add more default languages](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md).
 
+:::tip
+By default, regional variants (e.g. `pt-BR`) will receive the built-in translations for its base language (e.g. `pt`).
+:::
+
```
delucis, line 151: I’m not sure this is the right place for this where a user is focused on _adding_ a translation.

```suggestion
```

It might make most sense in the [configuration reference section for `lang`](https://deploy-preview-475--astro-starlight.netlify.app/reference/configuration/#lang), something like this:

> The BCP-47 tag for this language, e.g. `"en"`, `"ar"`, or `"zh-CN"`. If not set, the language’s directory name will be used by default. Language tags with regional subtags (e.g. `"pt-BR"` or `"en-US"`) will use built-in UI translations for their base language if no region-specific translations are found.

delucis (approved):
Thanks @Yan-Thomas — this looks good! Suggesting two docs tweaks and then should be good to go 🙌

delucis:
> Note that I haven't added tests to ensure the built-in regional variant gets precedence over the base language one, it seems we don't even have support for them yet, since the `translations/index.ts` implementation won't allow for dashes to match with the language.

Oh we could still support it! We just couldn’t rely on the shorthand object notation:

```js
import en from './en.json';
import zhTW from './zh-TW.json';

const { parse } = builtinI18nSchema();

export default Object.fromEntries(
	Object.entries({ en, 'zh-TW': zhTW }).map(([key, dict]) => [
		key,
		parse(dict),
	])
);
```

Yan-Thomas:
> Oh we could still support it! We just couldn’t rely on the shorthand object notation:

Oh nice! Tested locally and worked perfectly, we're good to ship then! 🚀

docs/src/content/docs/reference/configuration.md
```diff
@@ -255,7 +255,7 @@ The label for this language to show to users, for example in the language switch
 
 **type:** `string`
 
-The BCP-47 tag for this language, e.g. `"en"`, `"ar"`, or `"zh-CN"`. If not set, the language’s directory name will be used by default.
+The BCP-47 tag for this language, e.g. "en", "ar", or "zh-CN". If not set, the language’s directory name will be used by default. Language tags with regional subtags (e.g. "pt-BR" or "en-US") will use built-in UI translations for their base language if no region-specific translations are found.
```
delucis, line 258: Quick fix for some missing backticks here before merging!

```suggestion
The BCP-47 tag for this language, e.g. `"en"`, `"ar"`, or `"zh-CN"`. If not set, the language’s directory name will be used by default. Language tags with regional subtags (e.g. `"pt-BR"` or `"en-US"`) will use built-in UI translations for their base language if no region-specific translations are found.
```

PR #483, opened 2023-08-10
Fix target languages in tracker

TheOtterlord:

- Something else!

#### Description

- What does this PR change? Changes the tracker to use the language keys, fixing the filter determining which directories should be marked as language directories.


delucis (approved):
Nice catch — thanks for the quick fix @TheOtterlord!

PR #488, opened 2023-08-10
[LinkCard] use only title as the link text

### LINKED ISSUE #487: accessibility issue with LinkCard

0.7.0

2.10.4

npm

Mac

Firefox, Chrome, Safari

### Describe the Bug

Love the addition of the `LinkCard` component, but it includes the entire contents of the card as part of the link, instead of just the title. This results in verbose announcements by screen readers, when really only the title should be announced.

To fix the markup but keep the whole card clickable, a pseudo-element can be used.
```html
<div class=card>
  <a href="/">the title</a>
  <div>the other stuff</div>
</div>
```
```css
a::before {
  content: '';
  position: absolute;
  inset: 0;
}
```

Further reading: https://inclusive-components.design/cards/

Happy to contribute if this is something desirable.

_No response_

### Participation

mayank99:

- Changes to Starlight code

#### Description

- Closes #487
- Uses pseudo content trick from https://inclusive-components.design/cards#thepseudocontenttrick
- No visual changes

### DIFF
--- a/packages/starlight/user-components/LinkCard.astro
+++ b/packages/starlight/user-components/LinkCard.astro
@@ -11,25 +11,36 @@ const { title, description, ...attributes } = Astro.props;
 ---
 
 <div>
-	<a {...attributes}>
-		<span class="flex stack">
+	<span class="flex stack">
+		<a {...attributes}>
 			<span class="title" set:html={title} />
-			{description && <span class="description" set:html={description} />}
-		</span>
-		<Icon name="right-arrow" size="1.333em" class="icon rtl:flip" />
-	</a>
+		</a>
+		{description && <span class="description" set:html={description} />}
+	</span>
+	<Icon name="right-arrow" size="1.333em" class="icon rtl:flip" />
 </div>
 
 <style>
-	a {
+	div {
 		display: grid;
 		grid-template-columns: 1fr auto;
 		gap: 0.5rem;
 		border: 1px solid var(--sl-color-gray-5);
 		border-radius: 0.5rem;
 		padding: 1rem;
-		text-decoration: none;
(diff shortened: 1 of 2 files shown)


mayank99:
@delucis This is the approach I was thinking. The content may be short for the few use cases you have in mind, but this is intended to be a generic reusable component, so you never know how much content the users are passing in.

packages/starlight/user-components/LinkCard.astro
```diff
@@ -11,25 +11,36 @@ const { title, description, ...attributes } = Astro.props;
 ---
 
 <div>
```
kevinzunigacuellar, line 13: Not an a11y expert but could we use an `<article>` instead of a `<div>` for a more semantic a11y tree.

mayank99: i like that. articles need should generally have labels though.
```html
<article aria-labelledby='title'>
  <a id='title'>...</a>
```

is there a utility i can use for generating a unique id?

delucis (approved):
This is great! Thank you @mayank99 — ticks all my boxes ✅ 

I’m not 100% sure about the `<article>` approach. Could that be too opinionated for some use cases? I think of articles as containing more information than _just_ a link (which when used without `description`, is what these are), but maybe I’m thinking too literally.

Ideally, I’d love to have these automatically be list items when used inside `<CardGrid>`, but that would require some hacker-y inside `<CardGrid>` so I’ve held off doing that for now.

mayank99:
Sounds good to me, `<article>` and `<li>` could potentially be explored in the future. Until then, users always have the ability to manually wrap their `<LinkCard>`.

PR #492, opened 2023-08-10
fix pagefind build

rishi-raj-jain:
remove -y as that's causing build errors with new starlight projects


delucis:
Hi @rishi-raj-jain! Would you mind filing [a bug report](https://github.com/withastro/starlight/issues/new?assignees=&labels=&projects=&template=---01-bug-report.yml) with context for what this fixes? Super appreciate the PR, but it’s hard for us to prioritise and understand the full context without that detail.

PR #496, opened 2023-08-11
Make `lastUpdated` position consistent

### LINKED ISSUE #417: Last update date position

0.5.6

2.9.3

pnpm

Mac

Chrome

### Describe the Bug

In some docs, we have the last update date feature enabled as it is relevant to the content documented. Most of the pages are a mix of manually created pages and auto-generated pages. The auto-generated pages have their `editUrl` set to `false` in the frontmatter.

When navigating between a manually created page and an auto-generated page, the last update position is switching from the right side to the left side of the page due to the edit link being hidden. I think this is not the best user experience to have identical elements being placed in different positions between pages.

For example, on a potentially manually created page with an edit link:

SCR-20230727-kfxr

And the next page which is potentially auto-generated and has the edit link hidden:

SCR-20230727-kfvk

If this is something judged as a potential issue, I guess a potential fix could be to use a grid (`1fr 1fr`) and always assign the edit link to the first column and the last update date to the second column.

__Note: the repro is a bit useless as I don't think it's possible to have last update date working on Stackblitz.__

https://stackblitz.com/edit/github-b3fbut?file=src%2Fcontent%2Fdocs%2Freference%2Fexample.md

### Participation

lorenzolewis:

- Changes to Starlight code

#### Description

Closes #417 

Before the "Last Edit Date" would be leading-aligned if there was no edit url set. This caused an inconsistency on the wide layouts:

Screenshot 2023-08-11 at 09 56 20

Now it will always be trailing-edge aligned if it's the only child:

Screenshot 2023-08-11 at 09 56 26

There's no change in behaviour for if the editUrl is present.

### DIFF
--- a/.changeset/smooth-ladybugs-yell.md
+++ b/.changeset/smooth-ladybugs-yell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fix `lastUpdated` date position to be consistent
--- a/packages/starlight/components/Footer.astro
+++ b/packages/starlight/components/Footer.astro
@@ -48,4 +48,7 @@ const prevNextLinks = getPrevNextLinks(sidebar, config.pagination, {
 		font-size: var(--sl-text-sm);
 		color: var(--sl-color-gray-3);
 	}
+	.meta > :global(p:only-child) {
+		margin-inline-start: auto;
+	}
 </style>


delucis (approved):
Thanks for following up on this one @lorenzolewis!

I don’t _love_ that this depends on knowing that the components render an `<a>` and a `<p>` element respectively (in fact, now I have doubts about _why_ they’re not both `<p>` 😅), but I can’t think of a great alternative — I think ideally this component could know “Do I have no edit link? OK, add a class, align children accordingly.” But our logic for this is split between this component and each child component.

Could be solved with a bigger refactor, but I think this is OK for now.

lorenzolewis:
@delucis I agree with you on that and thought the same when I was writing it. But like you said, it might not be worth that big of a refactor just for this low-hanging fruit and to get it cleaned up.

PR #501, opened 2023-08-12
i18n(fr): Updated `components.mdx`

dreyfus92:

#### Description

Updated `components.mdx` based on this commit 73eb5e6ac6511dc4a6f5c4ca6c0c60d521f1db3c

### DIFF
--- a/docs/src/content/docs/fr/guides/components.mdx
+++ b/docs/src/content/docs/fr/guides/components.mdx
@@ -129,6 +129,50 @@ Ajoutez l'attribut `stagger` pour décaler verticalement la deuxième colonne de
 
 :::
 
+### Link Cards
+
+Utilisez le composant `<LinkCard>` pour créer un lien vers différentes pages.
+
+Une `<LinkCard>` nécessite un `title` et un attribut [`href`](https://developer.mozilla.org/fr/docs/Web/HTML/Element/a#href). Vous pouvez optionellement inclure une courte `description` ou d'autres attributs de lien tels que `target`.
+
+Regroupez plusieurs composants `<LinkCard>` dans `<CardGrid>` pour afficher les cartes côte à côte lorsqu'il y a suffisamment d'espace.
+
+```mdx
+import { LinkCard, CardGrid } from '@astrojs/starlight/components';
+
+<LinkCard
+	title="Personnalisation de Starlight"
+	description="Apprenez à personnaliser votre site Starlight avec vos propres styles, vos polices et bien plus encore."
+	href="/fr/guides/customization/"
+/>
+
+<CardGrid>
+	<LinkCard
+		title="Création de contenu en Markdown"
+		href="/fr/guides/authoring-content/"
+	/>
+	<LinkCard title="Composants" href="/fr/guides/components/" />
+</CardGrid>
+```
+
+Le code ci-dessus génère ce qui suit sur la page :
+
+import { LinkCard } from '@astrojs/starlight/components';
+
+<LinkCard
+	title="Personnaliser Starlight"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/components.mdx
```diff
@@ -129,6 +129,50 @@ Ajoutez l'attribut `stagger` pour décaler verticalement la deuxième colonne de
 
 :::
 
+### Link Cards
+
+Utilisez le composant `<LinkCard>` pour créer un lien vers différentes pages.
```
delucis, line 134: In the English we say “link prominently” to distinguish this from just a regular link. Maybe something like this could accomplish the same here?

```suggestion
Utilisez le composant `<LinkCard>` pour créer un lien bien visible vers différentes pages.
```

docs/src/content/docs/fr/guides/components.mdx
```diff
@@ -129,6 +129,50 @@ Ajoutez l'attribut `stagger` pour décaler verticalement la deuxième colonne de
...
+
+Une `<LinkCard>` nécessite un `title` et un attribut [`href`](https://developer.mozilla.org/fr/docs/Web/HTML/Element/a#href). Vous pouvez optionellement inclure une courte `description` ou d'autres attributs de lien tels que `target`.
```
delucis, line 136: Here both `title` and `href` are attributes so maybe something like this would be more accurate?

```suggestion
Une `<LinkCard>` nécessite les attributs `title` et [`href`](https://developer.mozilla.org/fr/docs/Web/HTML/Element/a#href). Vous pouvez optionellement inclure une courte `description` ou d'autres attributs de lien tels que `target`.
```

docs/src/content/docs/fr/guides/components.mdx
```diff
@@ -129,6 +129,50 @@ Ajoutez l'attribut `stagger` pour décaler verticalement la deuxième colonne de
...
+
+Regroupez plusieurs composants `<LinkCard>` dans `<CardGrid>` pour afficher les cartes côte à côte lorsqu'il y a suffisamment d'espace.
```
delucis, line 138: Noticing this is the only place we refer to these as “cartes”. Should we also update the headings to use that too? Currently they use the English “Cards” and “Link Cards”.

I looked up the [French government design system](https://www.systeme-de-design.gouv.fr/elements-d-interface/composants/carte) and they also use “Carte” although on their index page they always include English and French terms side by side:
Three links from the French government design system components index showing titles including both French and English names for each component.

So maybe our headings should do something similar like “Cartes (Cards)” and “Cartes liens (Link Cards)”?

delucis (commented):
Thanks for the update @dreyfus92! Left a few comments.

docs/src/content/docs/fr/guides/components.mdx
```diff
@@ -129,6 +129,50 @@ Ajoutez l'attribut `stagger` pour décaler verticalement la deuxième colonne de
 
 :::
 
+### Cartes (LinkCard)
```
delucis, line 132: Ah, but this is a bit confusing isn’t it? Because the heading for the section above (for the `<Card />` component) is “Cards”, so now the table of contents is:

- Cards
- Cartes (LinkCard)

Which doesn’t seem right. We should translate both and translate “Link Cards” in a way that’s different from “Cards”. This would also match the translated headings for “Onglets” and “Icônes”.

dreyfus92: What if we translate Cards to Cartes and LinkCards to Cartes de Liaison?

delucis (approved):
Thanks @dreyfus92!

PR #513, opened 2023-08-14
Generated Index component

### LINKED ISSUE #480: Table of Contents Component

0.7.0

### What is your idea?

Create a user-facing component that behaves similarly to how sidebar objects work, but instead being listed in the content of a page.

The name "Table of Contents" could be confusing since technically that already exists on the trailing side of the page, so open to workshopping the name.

This would wrap the new [`<LinkCard>`](https://starlight.astro.build/guides/components/#link-cards) component within a [`<CardGrid>`](https://starlight.astro.build/guides/components/#cards) and dynamically fill in the contents based on page's frontmatter (`title` and `description`). 

This could potentially respect the new `sidebar.label`, `sidebar.order`. and `sidebar.hidden` properties in generation, although this could be confusing due to the values being under the `sidebar` name and this not being a sidebar (maybe this could be an opt-in/opt-out behaviour?)

The last bit to tackle would be _how_ you specify which content you'd like this component to render. I'm not 100% sure what I would expect this to be but I could think of these options:

- Pass an array of [`SidebarItem`](https://starlight.astro.build/reference/configuration/#sidebaritem) with what you'd like (autogeneration could be very very useful here)
- Pass a path to the path on disk you'd like to have built out (this could point to a folder) like `src/content/docs/foo` for example
- Derive from the sidebar defined in the config by specifying the slice of the sidebar you'd like to render

---

The approach I'd prefer is _very_ tightly coupled to the sidebar logic as it would allow us to just build on top of that. However it could be confusing DX to a user when using that outside of the context of the sidebar. So open to any thoughts on this as well.

I'm willing to work on a PR for this one depending on how willing we are to build on top of the sidebar logic. If we want a slightly different path of logic then I might need to step down depending on how complex it would be 😅 

Sometimes you just want a really nice looking landing page without writing it by hand. 

Docusaurus: https://docusaurus.io/docs/sidebar/items#generated-index-page and https://docusaurus.io/docs/sidebar/items#embedding-generated-index-in-doc-page

### Participation

lorenzolewis:

- Changes to Starlight code

#### Description

- Closes #480 
- Creates the `GeneratedIndex` component based on top of `LinkCard`, `CardGrid`, and some underlying sidebar logic.

Screenshot 2023-08-14 at 22 12 55

#### Remaining Items

- [ ] Only go the depth of the current directory (so that nested folders for a directory aren't rendered)
- [ ] Build the description logic
- [ ] Revise docs once logic is wrapped up
- [ ] Add changeset (minor I think?)


lorenzolewis:
**Question on the approach**: Right now this is able to run 100% on the logic exposed via `getSidebar`. However, that function returns `SidebarEntry[]` which does NOT contain the `description` frontmatter that I'd like to expose in this component. 

Here's the options I thought of:

1. Add a `description?: string` value to the `Link` type:

https://github.com/withastro/starlight/blob/78e0b1fcfd26ef59a618b2736c5fd94082f77b44/packages/starlight/utils/navigation.ts#L13-L18

2. Create a new type that extends `Link` with that additional attribute
3. Use [`getEntry()`](https://docs.astro.build/en/reference/api-reference/#getentry) (or a wrapper that accounts for locale if one exists in Starlight, haven't checked yet) constructing the slug/id from a `Link` item's `href` value.

Out of these options I don't like item 3 at all. It would be adding another hop of loading that entry and is also adding unneeded complexity to derive the id/slug from the href.

Option 2 just feels a bit more over-complicated to me, especially since these are all internal functions right now.

That leaves me with option 1, but it does feel a bit weird to add data like `description` to the `Link` type. But playing around with this approach it feels the most straightforward while letting us build on top of the already tried and tested logic these functions all use.

---

I think I'm going to do a bit more work on the first option to see if I hit any roadblocks, but welcome and appreciate any feedback and other thoughts on this.

lorenzolewis:
I'm going to close this one out. Unfortunately I haven't had time to re-visit it lately and I'd like to take a stab at re-building it once I'm back from vacation with all of the awesome changes happening around sidebar happening.

delucis:
Thanks for letting us know @lorenzolewis! Ans sorry I wasn’t able to get you more feedback on this earlier.

lorenzolewis:
No worries at all! I had some feedback from myself anyway on how I'd re-approach this and will hopefully be able to take another stab at it once I have a bit more capacity 💜

PR #515, opened 2023-08-14
Replace `pa11y-ci` with a custom runner for Pa11y

HiDeoo:

- Something else!

#### Description

This pull request is the follow-up of various conversations ([link 1](https://discord.com/channels/830184174198718474/1125482411094966292), [link 2](https://discord.com/channels/830184174198718474/1139150903304278017)) regarding Pa11y. It is a draft as I am not sure if we want to go down this route.

To quickly summarize some of the previous steps:

- The idea of using Pa11y against the examples was abandoned as it does not contain enough content compared to the documentation.
- https://github.com/withastro/starlight/commit/df24060936c18c9973afde4edd347a191edb6242 reduced the number of translation pages checked by Pa11y to speed up the process.
- Pa11y is only testing the light theme.

A potential idea to use Pa11y against both schemes was to use a JavaScript configuration file reading environment variables to know which scheme to test and provide to Pa11y a custom instance of a browser and page to use. This was a dead end:

- We cannot specify a path to a JavaScript config file to Pa11y including an extension as it will first try to read this file if it exists and parse it as JSON which will result in an error for a JavaScript file.
- If we omit the extension, it will try to load a JavaScript file by appending the `.js` extension altho this will fail to load in our case due to the `type: module` in our `package.json` file and we have no way to use a `.cjs` file instead.

As I could not see any viable alternative at this point, I decided to give up and rewrite a custom simplified runner for Pa11y and use it instead of Pa11y-CI.

- Pros
  - We get a lot more control over the process.
  - We can remove the `NO_GRADIENTS` workaround by injecting in the page custom CSS.
  - We can test both schemes.
- Cons
  - We need to maintain this code.
  - We have to use a version of Puppeteer compatible with Pa11y which is 10 ish versions behind.
  - When using `pa11y-ci`, we indirectly use `axe-core` version `4.7.2` but when using `pa11y` directly, we use `axe-core` version `4.2.1` which turns out has false postive for the `color-contrast` rules on some platform which means we basically have to override its version.
  - When passing down a custom browser and page instance to Pa11y, as mentioned in their doc, we have to create & close the page between each test.

Performance wise, it looks like 1 run for 1 color scheme is taking around 2m 10s which is pretty much the same as before or slightly slower depending on the run but we are testing 2 color schemes now with the GitHub Actions matrix feature.

UI wise, it's pretty much the same output as before:

demo

_(note: this is a juxtaposition of 2 screenshots so you can ignore the fact that it says that it is testing the light scheme and reports errors for the dark scheme ^^)_

Here is [a link to a failure](https://github.com/HiDeoo/starlight/actions/runs/5858716942/job/15883214666?pr=1) in a GitHub Actions run with some errors on my fork.

Potential improvements if we ever choose to go down this route:

- Take a second pass at the code as this is mostly a first draft to see if we want to use this approach.
- Add a npm script to run both schemes locally easily.
- We could remove the environment variables and pass the arguments directly to the script.
- I assumed `packages/*` is for public packages and folder at the root were for private packages. We can move the script if needed.

I think that pretty much covers everything. Personally I am not sure how I feel about this and I think my judgement is a bit clouded by the fact that I hit so many roadblocks with Pa11y and Pa11y-CI.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,5 +9,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["starlight-docs", "@example/*", "docs-i18n-tracker"]
+  "ignore": ["starlight-docs", "@example/*", "docs-i18n-tracker", "pa11y-runner"]
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,6 @@ jobs:
       - name: Build docs site
         working-directory: ./docs
         run: pnpm build
-        env:
-          NO_GRADIENTS: true
 
       - name: Run accessibility audit
         working-directory: ./docs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,9 @@ jobs:
   pa11y:
     name: Check for accessibility issues
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        scheme: [dark, light]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -58,6 +61,8 @@ jobs:
       - name: Run accessibility audit
         working-directory: ./docs
         run: pnpm t
+        env:
+          COLOR_SCHEME: ${{ matrix.scheme }}
 
   windows-smoke:
     name: Docs site builds on Windows
--- a/docs/.pa11yci
+++ b/docs/.pa11yci
@@ -1,7 +1,0 @@
-{
-  "defaults": {
-    "runners": [
-      "axe"
-    ]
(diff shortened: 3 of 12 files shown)


delucis:
Sorry I completely forgot about this @HiDeoo. I’m thinking through if we’ve had any big hurdles recently from our existing setup. While it’s true we aren’t testing both colour schemes, I don’t think we’ve had big colour contrast issues recently, so this change seems like probably more maintenance without a huge amount of gain. Going to close this as a helpful exploration — thanks again!

HiDeoo:
Oh, I totally forgot I even wrote that too 🤣 Happy with the decision, and if we ever need that, we can either come back to this, explore a Playwright approach or anything else, many options.

PR #476, opened 2023-08-09
[WIP] Make aside label canbe translate.

liruifengv:
#### What kind of changes does this PR include?

- Changes to Starlight code

#### Description

- Make aside label canbe translate.

PR #517, opened 2023-08-15
Support translating default aside titles

liruifengv:

- Changes to Starlight code

#### Description

- Make aside label canbe translate.

I added translations according to the astro docs, some are missing translations, en is used by default and needs to be updated

missing translations:
- cs: Czech
- da: Danish
- fa: Farsi
- gl: Galician
- he: Hebrew
- id: Bahasa Indonesia
- nb: Norwegian Bokmål
- nl: Dutch
- ru: Russian
- sv: Swedish
- tr: Turkish
- uk: Ukrainian
- vi: Vietnamese


liruifengv:
Sorry, I deleted the branch by mistake and caused https://github.com/withastro/starlight/pull/476 to be closed.

@delucis I have some questions, I don't know how to get the current language.

packages/starlight/integrations/asides.ts
```diff
@@ -96,18 +148,19 @@ function remarkAsides(): Plugin<[], Root> {
 	};
 
 	const transformer: Transformer<Root> = (tree) => {
```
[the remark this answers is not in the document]
liruifengv: Thank you very much for your help. I have completed the first version of the function, waiting for your review.

liruifengv:
@delucis Hi Chris, Can we keep pushing this PR?

delucis:
Hey @liruifengv — so sorry I haven’t had time to look at this yet. Promise I’ll get to this ASAP. Thank you again for pushing this — super helpful 💜

delucis:
Hey @liruifengv! Sorry to take so long to get back on this one, but I think we’re finally ready to try to do this. In #1003 I made it possible for plugins like the asides one to more easily access the same translations system as other Starlight code.

I’ve updated this branch and you should now find `useTranslations` in [`packages/starlight/index.ts`](https://github.com/liruifengv/starlight/blob/e82ad5355e10af10c63550643d672fd33cdf31a6/packages/starlight/index.ts#L30). You can pass this to the asides plugin and then in each locale do:

```js
const t = useTranslations(locale);
const label = t('aside.note'); // or tip, danger etc.
```

Are you still interested on working on this? Let me know if not and I’ll take a look at cleaning up your work. Thanks again for your patience on this.

By the way, all the checks are failing because the `aside` translations are missing in languages added since you started this PR. Let me know if you need help tracking down the missing translations.

liruifengv:
> Hey @liruifengv! Sorry to take so long to get back on this one, but I think we’re finally ready to try to do this. In #1003 I made it possible for plugins like the asides one to more easily access the same translations system as other Starlight code.
> 
> I’ve updated this branch and you should now find `useTranslations` in [`packages/starlight/index.ts`](https://github.com/liruifengv/starlight/blob/e82ad5355e10af10c63550643d672fd33cdf31a6/packages/starlight/index.ts#L30). You can pass this to the asides plugin and then in each locale do:
> 
> ```js
> const t = useTranslations(locale);
> const label = t('aside.note'); // or tip, danger etc.
> ```
> 
> Are you still interested on working on this? Let me know if not and I’ll take a look at cleaning up your work. Thanks again for your patience on this.
> 
> By the way, all the checks are failing because the `aside` translations are missing in languages added since you started this PR. Let me know if you need help tracking down the missing translations.

I'll take a look.

packages/starlight/integrations/asides.ts
```diff
@@ -5,6 +5,21 @@ import remarkDirective from 'remark-directive';
...
+  const formatPath = path.replace(/\\/g, '/');
+	const parts = formatPath.split('/');
+	const langIndex = parts.findIndex((part) => /^[a-z]{2}(-[a-z]{2})?$/.test(part));
+	if (langIndex !== -1) {
+		return parts[langIndex] || 'en';
+	} else {
+		const defaultLocale = config.defaultLocale?.lang || config.defaultLocale?.locale;
+		return defaultLocale || 'en';
+	}
+}
```
delucis, line 22: There are some differences here from our existing [slug utilities](https://github.com/withastro/starlight/blob/0100298b158f90cf2beed67e9a185f40a533465b/packages/starlight/utils/slugs.ts). Importantly, we support locales that don’t match the BCP `aa-AA` shape and also because `path` here is absolute (at least in my testing), we’d risk clashing with some other directory in the path.

I think the right approach is to use `root` and `srcDir` from the Astro config. With those we can strip the path down so we only get the path inside `content/docs/` and then apply the same logic we do for file IDs.

Something like this:

```js
function pathToLocale(slug: string, config: StarlightConfig): string | undefined {
	const locales = Object.keys(config.locales || {});
	const baseSegment = slug.split('/')[0];
	if (baseSegment && locales.includes(baseSegment)) return baseSegment;
	return undefined;
}

function localeToLang(locale: string | undefined, config: StarlightConfig): string {
	const lang = locale ? config.locales?.[locale]?.lang : config.locales?.root?.lang;
	const defaultLang = config.defaultLocale?.lang || config.defaultLocale?.locale;
	return lang || defaultLang || 'en';
}

/** get current lang from file full path */
function getLocaleFromPath(
	unformattedPath: string,
	starlightConfig: StarlightConfig,
	astroConfig: AstroConfig
): string {
	const srcDir = new URL(astroConfig.srcDir, astroConfig.root);
	const docsDir = new URL('content/docs/', srcDir);
	const path = unformattedPath
		// Format path to unix style path.
		.replace(/\\/g, '/')
		// Strip docs path leaving only content collection file ID.
		// Example: /Users/houston/repo/src/content/docs/en/guide.md => en/guide.md
		.replace(docsDir.pathname, '');
	const locale = pathToLocale(path, starlightConfig);
	return localeToLang(locale, starlightConfig);
}
```

To do that we’ll need to pass the Astro config object through from `index.ts` too.

packages/starlight/integrations/asides.ts
```diff
@@ -50,19 +65,11 @@ function s(el: string, attrs: Properties = {}, children: any[] = []): P {
  * </Aside>
  * ```
  */
-function remarkAsides(): Plugin<[], Root> {
+function remarkAsides(config: StarlightConfig, useTranslations: any): Plugin<[], Root> {
```
delucis, line 68: Instead of `any` we can type `useTranslations`.

1. Import the type:

    ```js
    import type { createTranslationSystemFromFs } from '../utils/translations-fs';
    ```
    
2. Use it like:

   ```suggestion
   function remarkAsides(
   	config: StarlightConfig,
   	useTranslations: ReturnType<typeof createTranslationSystemFromFs>
   ): Plugin<[], Root> {
   ```

packages/starlight/integrations/asides.ts
```diff
@@ -157,6 +166,6 @@ function remarkAsides(): Plugin<[], Root> {
 
 type RemarkPlugins = NonNullable<NonNullable<AstroUserConfig['markdown']>['remarkPlugins']>;
 
-export function starlightAsides(): RemarkPlugins {
-	return [remarkDirective, remarkAsides()];
+export function starlightAsides(opts: StarlightConfig, useTranslations: any): RemarkPlugins {
```
delucis, line 169: Same thing here:

```suggestion
export function starlightAsides(
	opts: StarlightConfig,
	useTranslations: ReturnType<typeof createTranslationSystemFromFs>
): RemarkPlugins {
```

.changeset/thick-boxes-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Make aside's title can be translated.
```
delucis, line 5: Because this adds a feature, we can use `minor` for the changeset and I might suggest this as a description:

```suggestion
---
"@astrojs/starlight": minor
---

Add i18n support for default aside labels
```

delucis (commented):
Working great @liruifengv! I think there’s just one area for improvement, which is how we extract the locale from the file path.

packages/starlight/integrations/asides.ts
liruifengv: Yeah. Maybe we don't need `localeToLang` because it's handled in `useTranslations`?

delucis: Good point! Yeah that just needs the locale I think.

delucis (approved):
Thanks again for the work and patience on this PR @liruifengv! I just made a last couple of changes and added some unit tests to make sure it’s working. LGTM!!! 🚀 

Super excited to merge this finally. Thank you for championing it 🙌

delucis:
>     1. Please note that some of the language was used the original English.

I think this is ok as ALL languages currently get the English version by default anyway. My plan is to ask for help from the community to update these as soon as we merge.


>     2. I changed getting-started in this PR. this will cause the translation status to be reset. Maybe it shouldn't be included in this PR

Ah, good catch! Thank you for flagging this. Yes, might be a good idea to remove that change and make it when updating the rest of the page.

liruifengv:
> > 1. Please note that some of the language was used the original English.
> 
> I think this is ok as ALL languages currently get the English version by default anyway. My plan is to ask for help from the community to update these as soon as we merge.
> 
> > 2. I changed getting-started in this PR. this will cause the translation status to be reset. Maybe it shouldn't be included in this PR
> 
> Ah, good catch! Thank you for flagging this. Yes, might be a good idea to remove that change and make it when updating the rest of the page.

Done.

PR #525, opened 2023-08-16
Improve inline code and code block support in RTL languages

delucis:

- Changes to Starlight code

#### Description

This PR applies a rehype plugin to Starlight Markdown and MDX content to improve appearance in RTL languages.

- Code blocks have `dir="ltr"` added to the `<pre>` tag.
- Inline code has `dir="auto"` added to the `<code>` tag.
- For code blocks this may be an interim solution until we migrate to Expressive Code.
- Research CodePen while working out the best set-up: https://codepen.io/delucis/pen/vYvBmKx

We don’t currently have an RTL language in our docs site to test with, so I added some Arabic content locally to test these changes.

| Before | After |
|---|---|
| RTL Starlight page where code content is misaligned to the right | RTL Starlight page where code content is correctly aligned to the left |

Note that in the second image, the code blocks are left-aligned even though page content is right-aligned and that the inline code reads correctly as `@astrojs/*` instead of `*/astrojs@`.

### DIFF
--- a/packages/starlight/integrations/code-rtl-support.ts
+++ b/packages/starlight/integrations/code-rtl-support.ts
@@ -1,0 +1,31 @@
+import type { Root } from 'hastscript/lib/core';
+import { CONTINUE, SKIP, visit } from 'unist-util-visit';
+
+/**
+ * rehype plugin that adds `dir` attributes to `<code>` and `<pre>`
+ * elements that don’t already have them.
+ *
+ * `<code>` will become `<code dir="auto">`
+ * `<pre>` will become `<pre dir="ltr">`
+ *
+ * `<code>` _inside_ `<pre>` is skipped, so respects the `ltr` on its parent.
+ *
+ * Reasoning:
+ * - `<pre>` is usually a code block and code should be LTR even in an RTL document
+ * - `<code>` is often LTR, but could also be RTL. `dir="auto"` ensures the bidirectional
+ *   algorithm treats the contents of `<code>` in isolation and gives its best guess.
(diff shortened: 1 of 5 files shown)


packages/starlight/integrations/code-rtl-support.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Reasoning:
+ * - `<pre>` is usually a code block and code should be LTR even in an RTL document
+ * - `<code>` is often LTR, but could also be RTL. `dir="auto"` ensures the bidirectional
+ *   algorithm treats the contents of `<code>` in isolation and gives its best guess.
+ */
+export function rehypeRtlCodeSupport() {
+	return () => (root: Root) => {
+		visit(root, 'element', (el) => {
+			if (el.tagName === 'pre' || el.tagName === 'code') {
+				el.properties = el.properties || {};
```
kevinzunigacuellar, line 22: I always wanted to suggest this new syntax
```suggestion
				el.properties ||= {};
```

delucis: 👀

Do you happen to know the Node version that was added in? We safe to use it in 16+?

HiDeoo: `??=` & `||=` are both 15.0.

dreyfus92: i'd go with `??=` which is more usual to see nullish coalescing, but i guess that it is also a matter of choice.

PR #528, opened 2023-08-18
i18n(ko-KR): add Korean language support

jsparkdev:

- Changes to Starlight code

#### Description

- add Korean language support


docs/src/content/docs/guides/i18n.mdx
```diff
@@ -37,6 +37,11 @@ Starlight provides built-in support for multilingual sites, including routing, f
    					label: 'العربية',
    					dir: 'rtl',
    				},
+   				// Korean docs in `src/content/docs/ko/`
+   				ko: {
+   					label: '한국어',
+   					lang: 'ko-KR',
+   				},
```
delucis, line 44: This code block is more to give an example of a few languages and doesn’t need to show every possible language, so I think we can remove Korean here:
```suggestion
```

docs/src/content/docs/guides/i18n.mdx
```diff
@@ -56,6 +61,7 @@ Starlight provides built-in support for multilingual sites, including routing, f
          - ar/
          - en/
          - zh/
+         - ko/
```
delucis, line 64: Similarly here, I think we can leave this example unchanged and people will still understand that they could use a `ko/` directory for Korean.
```suggestion
```

docs/src/content/docs/guides/i18n.mdx
```diff
@@ -143,7 +149,7 @@ If a translation is not yet available for a language, Starlight will show reader
 
 In addition to hosting translated content files, Starlight allows you to translate the default UI strings (e.g. the "On this page" heading in the table of contents) so that your readers can experience your site entirely in the selected language.
 
-English, Czech, French, German, Italian, Japanese, Portuguese, Dutch, Danish, Spanish, Turkish, Arabic, Norwegian, Farsi, and Simplified Chinese translated UI strings are provided out of the box, and we welcome [contributions to add more default languages](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md).
+English, Czech, French, German, Italian, Japanese, Portuguese, Dutch, Danish, Spanish, Turkish, Arabic, Norwegian, Farsi, Simplified Chinese and Korean translated UI strings are provided out of the box, and we welcome [contributions to add more default languages](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md).
```
delucis, line 152: We use a so-called [“Oxford Comma”](https://en.wikipedia.org/wiki/Serial_comma) in lists like this, so adding that before the “and” here:

```suggestion
English, Czech, French, German, Italian, Japanese, Portuguese, Dutch, Danish, Spanish, Turkish, Arabic, Norwegian, Farsi, Simplified Chinese, and Korean translated UI strings are provided out of the box, and we welcome [contributions to add more default languages](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md).
```

delucis (commented):
Thank you for this contribution @jsparkdev!

Are you planning on translating the documentation pages as well? If not, I’d suggest we remove those from this PR for now and make this PR focused on just the `ko.json` translations you added.

We don’t need to duplicate the English content into the `ko/` directory in our docs, because Starlight will automatically do this using its [fallback content system](https://starlight.astro.build/guides/i18n/#fallback-content).

docs/src/content/docs/guides/i18n.mdx
jsparkdev: I would like to translate all documents of starlight.

jsparkdev:
I would like to translate all documents of starlight.

delucis:
> I would like to translate all documents of starlight.

Amazing! That’s super exciting. In that case, I’d suggest in this PR starting by only translating `src/content/docs/ko/index.mdx`. Then we can merge this and translate the other pages one-by-one in follow-up PRs. This way reviewing changes can happen in smaller chunks. How does that sound?

jsparkdev:
What a great idea!
I will start translating the `src/content/docs/ko/index.mdx` file.

delucis (approved):
Thanks again @jsparkdev! I removed the untranslated pages to clean this up and it’s now good to go! 🚀

PR #559, opened 2023-08-23
Add Stack Overflow icon

delucis:

- Changes to Starlight code

#### Description

- Adds a Stack Overflow icon and includes it in the social icons options.


docs/src/content/docs/guides/customization.mdx
```diff
@@ -178,7 +178,7 @@ defineConfig({
 
 Starlight has built-in support for adding links to your social media accounts to the site header via the [`social`](/reference/configuration/#social) option in the Starlight integration.
 
-Currently, links to Bitbucket, Codeberg, CodePen, Discord, GitHub, GitLab, Gitter, Instagram, LinkedIn, Mastodon, Microsoft Teams, Threads, Twitch, Twitter, and Youtube are supported.
+Currently, links to Bitbucket, Codeberg, CodePen, Discord, GitHub, GitLab, Gitter, Instagram, LinkedIn, Mastodon, Microsoft Teams, Stack Overflow, Threads, Twitch, Twitter, and Youtube are supported.
```
kevinzunigacuellar, line 181: I think we are at the point where we could convert this into a list for better readability.

kevinzunigacuellar:
```suggestion
Currently, links include: 
- Bitbucket
- Codeberg
- CodePen
- Discord
- GitHub
- GitLab
- Gitter
- Instagram
- LinkedIn
- Mastodon
- Microsoft Teams
- Stack Overflow
- Threads
- Twitch
- Twitter
- Youtube

delucis: That takes up sooo much space though 😅

I’ve also been wondering if we can just drop this paragraph so we don’t have to constantly update it and just have the full list in reference.

kevinzunigacuellar: I see that we already have a link to the configuration reference. I think it would be safe to drop it

PR #564, opened 2023-08-24
Add announcement banner feature

delucis:

- Changes to Starlight code

#### Description

- Adds a `banner` feature to Starlight that makes it easy to display an announcement at the top of a page.
- Also uses this feature to add a banner to the Starlight docs homepage in preparation for Product Hunt launch next week.


dreyfus92 (commented):
image
Just a random thought. While checking the mobile version of the banner wouldn't this look better if the lines break completely in 2 separate lines, what do you think Chris?

HiDeoo:
Took a bit of time to research a bit more about the `banner` role and I am starting to think this may not be a good idea in the end:

- We saw Docusaurus using this role but this was added in the initial commit of the feature and not as an improvement which could contains a report / discussion regarding the choice of this role.
- The MDN documentation contains some infos that do not really fit an announcement bar:
  - The [description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) of the `banner` role does not really match the content of an announcement bar "global site header, which usually includes a logo, company name, search feature, and possibly the global navigation or a slogan".
  - It "should be reserved for globally repeating site-wide content" but the announcement bar is not global in this implementation.
  - "each page should generally be limited to a single element with the role of banner" and considering  the HTML's `<header>` element has an identical role, this would mean that pages with an announcement bar would have two element with the `banner` role.
- I researched a lot of websites that have this kind of announcement bar like Microsoft, Apple, etc. and could not find a single one using the `banner` role for this purpose.

Considering all this, I think we may not want to use the `banner` role for this purpose. `complementary` could be a better choice but only in some cases due to the "can stand alone when separated" requirement. This would be ok for a Product Hunt banner but not for a "Deprecation notice" banner for example. So personally, I think we should not use any specific role for this.

packages/starlight/components/Banner.astro
```diff
@@ -0,0 +1,22 @@
...
+<div class="sl-banner" role="banner" set:html={Astro.props.content} />
+
+<style>
+  .sl-banner {
+    --__sl-banner-text: var(--sl-color-banner-text, var(--sl-color-text-invert));
+    padding: var(--sl-nav-pad-y) var(--sl-nav-pad-x);
+    background-color: var(--sl-color-banner-bg, var(--sl-color-bg-accent));
+    color: var(--__sl-banner-text);
+    line-height: var(--sl-line-height-headings);
+    text-align: center;
```
lorenzolewis, line 16:
```suggestion
    text-align: center;
    text-wrap: balance;
```
If progressive CSS enhancements are something Starlight is open to then you could add this one: https://caniuse.com/css-text-wrap-balance, https://www.w3.org/TR/css-text-4/#text-wrap

delucis:
> I think we may not want to use the `banner` role for this purpose. `complementary` could be a better choice but only in some cases due to the "can stand alone when separated" requirement. 

Yeah, I also considered `<aside>` given that, in the current implementation, it's inside `<main>` but is not part of the main flow (notably positioned above the first `<h1>`). Need to think more if that would work or whether the concern for content that isn't strictly complementary means we should avoid it.

Definitely +1 on removing the `role=banner` though.

packages/starlight/components/Banner.astro
delucis: Definitely! And I think this is safe. Even if people can theoretically provide all kinds of content, I don't believe this would disrupt that.

delucis: Definitely! And I think this is safe. Even if people can theoretically provide all kinds of content, I don't believe this would disrupt that.

delucis:
> Just a random thought. While checking the mobile version of the banner wouldn't this look better if the lines break completely in 2 separate lines, what do you think Chris?

Good shout @dreyfus92! Spent some time making this match @doodlemarks’s designs, so should be good now.

dreyfus92 (approved):
Awesome job @delucis, LGTM ✅

sarah11918 (approved):
3/3 evil devices agree! This looks great on desktop and mobile for me! 🥳

kevinzunigacuellar (approved):
Passed the galaxy fold test 🚀

TheOtterlord (approved):
Looks good on tablet!

docs/src/assets/landing.css
```diff
@@ -25,3 +25,63 @@
 [data-has-hero] .hero > img {
 	filter: drop-shadow(0 0 3rem var(--overlay-blurple));
 }
+
+/* Product Hunt launch banner — can be removed on 30 August 2023. */
+[data-has-hero] .sl-banner {
+	--sl-color-banner-bg: #f5574a;
```
delucis, line 31: Realised the PH orange is too light for good color contrast here. Fixing to meet WCAG requirements:
```suggestion
	--sl-color-banner-bg: #d73f37;
```

PR #569, opened 2023-08-25
Use locale when sorting autogenerated sidebar entries

HiDeoo:

- Changes to Starlight code

#### Description

This pull request adds support for using the locale when sorting autogenerated sidebar entries.

At the moment, the current sorting algorithm, ignoring pages with an explicit order which is an early return, is as follows:

```ts
// If two pages have the same order value they will be sorted by their slug.
return a.slug < b.slug ? -1 : a.slug > b.slug ? 1 : 0;
```

If, for example, we have files with the slugs `route`, `radio` and `référence`, the following order will be used:

- `radio`
- `route`
- `référence`

Unfortunately, this is not the expected order for French users for example, who would expect the following order:

- `radio`
- `référence`
- `route`

When sorting directories, I am not sure we are actually doing any sorting at all as the `slug` would be `undefined` and having `undefined` in one of the comparisons will always return `0` and thus not changing the order.

To change this behavior, this PR uses an [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) (Node.js v13.0.0) to do language-sensitive string comparison.


dreyfus92 (approved):
This is a nice catch 🤯 @HiDeoo, LGTM ✅

packages/starlight/utils/navigation.ts
```diff
@@ -184,13 +184,17 @@ function getOrder(routeOrDir: Route | Dir): number {
 }
 
 /** Sort a directory’s entries by user-specified order or alphabetically if no order specified. */
-function sortDirEntries(dir: [string, Dir | Route][]): [string, Dir | Route][] {
-	return dir.sort(([, a], [, b]) => {
+function sortDirEntries(
+	dir: [string, Dir | Route][],
+	locale: string | undefined
+): [string, Dir | Route][] {
+	const collator = new Intl.Collator(localeToLang(locale));
```
delucis, line 191: Really nice! I didn’t know about `Intl.Collator` — so nice how much i18n stuff like this is built into Node these days.

delucis (approved):
This looks great! Thank you @HiDeoo 💖

packages/starlight/utils/navigation.ts
HiDeoo: Definitely, and super handy. I think that is what is used under the hood by the String [`localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) but they recommend creating an `Intl.Collator` object for repeated use.

Just to double-check for sanity, I can safely merge this one as it is a patch and it does not update the docs, right?

delucis: > I can safely merge this one as it is a patch and it does not update the docs, right?

Yes! Safe to merge.

PR #573, opened 2023-08-26
Add /ph rewrite

TheOtterlord:

- Something else!

#### Description

- What does this PR change? Adds a `/ph` rewrite to https://astro-houston-ph.pages.dev/

### DIFF
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -1,1 +1,2 @@
 /:lang/*    /:lang/404/   404
+/ph https://astro-houston-ph.pages.dev/ 200
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -1,2 +1,2 @@
+/ph/        https://astro-houston-ph.pages.dev/   200
 /:lang/*    /:lang/404/   404
-/ph/ https://astro-houston-ph.pages.dev/ 200
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -1,2 +1,2 @@
 /:lang/*    /:lang/404/   404
-/ph https://astro-houston-ph.pages.dev/ 200
+/ph/ https://astro-houston-ph.pages.dev/ 200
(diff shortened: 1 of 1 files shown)


TheOtterlord:
Looking at the [preview](https://deploy-preview-573--astro-starlight.netlify.app/ph), I'm not sure if this has worked

docs/public/_redirects
```diff
@@ -1 +1,2 @@
 /:lang/*    /:lang/404/   404
+/ph/ https://astro-houston-ph.pages.dev/ 200
```
delucis, line 2: I suspect we need the opposite order here. Netlify processes these top to bottom so the `/:lang/*` pattern is matching `/ph/` first.

```suggestion
/ph/        https://astro-houston-ph.pages.dev/   200
/:lang/*    /:lang/404/   404
```

TheOtterlord: Ahh, this makes a lot of sense :sweat_smile:

docs/public/_redirects
```diff
@@ -1 +1,2 @@
+/ph/        https://astro-houston-ph.pages.dev/   200
```
TheOtterlord, line 1: Looks like we need a wildcard to match all routes
```suggestion
/ph/*        https://astro-houston-ph.pages.dev/:path   200
```

TheOtterlord:
`/ph/` worked for the page, but didn't work for subroutes. the wildcard doesn't seem to work for any of them, but does prevent the normal 404 from rendering :thinking:

docs/public/_redirects
```diff
@@ -1 +1,2 @@
+/ph/*        https://astro-houston-ph.pages.dev/:path   200
```
TheOtterlord, line 1:
```suggestion
/ph/*        https://astro-houston-ph.pages.dev/:splat  200
```

TheOtterlord:
Looks like this works, but I'll need to update the site to use a `/ph` base route

docs/public/_redirects
```diff
@@ -1 +1,2 @@
+/ph/*        https://astro-houston-ph.pages.dev/:splat  200
```
TheOtterlord, line 1:
```suggestion
/ph/*        https://astro-houston-ph.pages.dev/ph/:splat  200
```

TheOtterlord:
This works now! [preview](https://deploy-preview-573--astro-starlight.netlify.app/ph/)

delucis (approved):
Great work! LGTM! 

(We can merge this now — no need to wait for next week)

PR #584, opened 2023-08-28
i18n(ko-KR): translate `i18n.mdx`

jsparkdev:

#### Description

translate `i18n.mdx`

### DIFF
--- a/docs/src/content/docs/ko/guides/i18n.mdx
+++ b/docs/src/content/docs/ko/guides/i18n.mdx
@@ -1,0 +1,218 @@
+---
+title: 국제화 (i18n)
+description: 여러 언어를 지원하도록 Starlight 사이트를 구성하는 방법을 알아보세요.
+---
+
+import FileTree from '../../../../components/file-tree.astro';
+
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/ko/guides/i18n.mdx
```diff
@@ -0,0 +1,218 @@
...
+import starlight from '@astrojs/starlight';
+
+export default defineConfig({
+	integrations: [
+		starlight({
+			title: '나의 문서',
+			locales: {
+				root: {
+					label: '简体中文',
+					lang: 'zh-CN',
```
delucis, line 126: While usually we want translations to match the English version 1:1, I think this is a good example where we might want to deviate to show Korean here (assuming many Korean readers who want a monolingual site, may want that to be Korean).

What do you think?

```suggestion
					label: '한국어',
					lang: 'ko',
```

delucis (commented):
This looks perfect! I have one suggestion, but the translation looks good 🙌

delucis (approved):
Thank you @jsparkdev! 🚀

PR #516, opened 2023-08-14
feat: add badge to sidebar link

kevinzunigacuellar:
#### What kind of changes does this PR include?

- Changes to Starlight code

#### Description

- Closes #478

| Light | Dark |
|-------|------|
|   image    | image |

PR #585, opened 2023-08-28
Use semantic names for badge variants

delucis:

- Changes to Starlight code

#### Description

- This PR proposes changes to #516 

- Moves away from using colors for the badge variants (`'purple'`, `'blue'` etc.) in favour of semantic variant names that align with the names we use for asides (`'tip'`, `'note'` etc.).

- Updates docs to match as well as doing a small restructure to docs to avoid some of the chunkier types getting hard to read.

Here’s what changed:

- `blue` => `note`
- `purple` => `tip`
- `yellow` => `caution`
- `pink` => `danger`
- `green` => `success`
- `default` is unchanged

I like how these feel. When updating a docs example, I noticed how natural it felt to do `{ text: 'Experimental', variant: 'caution' }`. I also noticed how the color names already didn’t align with how we name the colors internally — `pink` used `red` colors, `yellow` used `orange` colors — so that felt like it was pointing to potential confusion users could have about whether “pink” was really “pink” for them. Also guards us a bit against “Hey can you add a COLOR X variant” when these are semantic.

**For the future:** I’d suggest we migrate our colour custom property names internally to use these new semantic terms. This would make it easier to tell users: “Don’t like the `note` shade? Update the `note` color variables.” I’d also suggest adding a `success` aside as originally included in our mockups. That would bring badges and asides into full alignment.


sarah11918 (approved):
The semantic names make a lot of sense to me!

HiDeoo (approved):
I personally also prefer semantic names, easier to use imo and give a great visual intent.

I also agree with the suggestion of adding later the `success` aside and using these semantic names for the color custom properties.

packages/starlight/schema.ts
```diff
@@ -151,7 +151,7 @@ export function docsSchema() {
 					/**
 					 * Adds a badge to the sidebar link.
 					 * Can be a string or an object with a variant and text.
-					 * Variants include 'blue', 'pink', 'green', 'yellow' and 'purple'.
+					 * Variants include 'note', 'tip', 'caution', 'danger', 'success', and 'default'.
 					 * Passing only a string defaults to the 'blue' variant
```
kevinzunigacuellar, line 155:
```suggestion
					 * Passing only a string defaults to the 'default' variant which uses the site accent color
```

delucis: Ah yes — excellent catch @kevinzunigacuellar 🙌

delucis:
Yay! OK, will merge into the main PR. Thanks for all the reviews folks 🤩

PR #595, opened 2023-08-29
Fix pagefind_extended log message when building document

jenglamlow:

#### Description

Fix the following `pagefind_extended` message when building the docs, by updating the version for `pagefind` package use the latest version [0.12.0](https://github.com/CloudCannon/pagefind/releases/tag/v0.12.0) instead of alpha package
```
[Building search indexes]
⚠ Indexing Chinese in non-extended mode. 
In this mode, Pagefind will not segment words that are not whitespace separated. 
Running the extended Pagefind binary will include this segmentation. 
Either download the pagefind_extended binary, or run npx pagefind-extended.
⚠ Indexing Chinese in non-extended mode. 
In this mode, Pagefind will not segment words that are not whitespace separated. 
Running the extended Pagefind binary will include this segmentation. 
Either download the pagefind_extended binary, or run npx pagefind-extended.
```


delucis (commented):
Thanks for the contribution @jenglamlow. I’m not sure we want to downgrade like this. In fact, Pagefind will be releasing v1.0.0 soon, at which point we’ll want to upgrade to that.

Do you know why this only works for `0.12.0` and not `1.0.0-alpha` releases?

jenglamlow:
@delucis, you're right. Pagefind will be releasing v1.0.0 soon. Let's wait for that then

delucis:
Thanks anyway for the PR @jenglamlow! Pagefind 1.0 should be here very soon from what I’ve heard, so we can make a new PR when it lands 🥳

PR #614, opened 2023-08-30
docs: fix color of hovered icons in highlighted file tree items

julien-deramond:

- Minor CSS fix

#### Description

_(Issue originally spotted by @HiDeoo). I haven't created an issue first, but I can do so if needed._

This PR fixes a small rendering issue that can be seen at https://starlight.astro.build/manual-setup/#use-starlight-at-a-subpath (Firefox macOS at least) when the "guides" directory is hovered. We can see that the color of the icon changes and makes it invisible being the same as the background.

Different approaches could have been adopted; I've chosen to not apply the `summary:hover` rule when a `.highlight` is found in the DOM between a `summary` and a `.tree-icon`.

Obviously tested against `/manual-setup/#use-starlight-at-a-subpath` but also against `/guides/customization/#add-your-logo` for non-regression.

#### Live previews

- https://deploy-preview-614--astro-starlight.netlify.app/manual-setup/#use-starlight-at-a-subpath
- https://deploy-preview-614--astro-starlight.netlify.app//guides/customization/#add-your-logo (for non-regression testing with files)


docs/src/components/file-tree.astro
```diff
@@ -57,7 +57,7 @@ const processedContent = await fileTreeProcessor.process({
 		color: var(--sl-color-gray-3);
 	}
 
-	file-tree .directory > details > summary:hover .tree-icon,
+	file-tree .directory > details > summary:hover > :not(.highlight) > .tree-icon,
```
delucis, line 60: Looks like this is always disabling the hover styling for the tree icon because the DOM structure is not `summary > .highlight > .tree-icon` but more nested:

DOM structure showing additional nesting between summary, highlight and tree-icon

Maybe it’s simpler to just add an override style something like this?

```css
file-tree .directory > details > summary:hover .highlight .tree-icon {
	fill: var(--sl-color-text-invert);
}
```

delucis (commented):
Thank you for the PR @julien-deramond! Looks like your fix isn’t behaving quite as expected in my tests. Left a suggestion of what we could maybe do instead.

docs/src/components/file-tree.astro
julien-deramond: Good catch, I was focused on the hovered style and I missed the regression. I've changed the fix to use your version which was the alternative method I was thinking about initially. Tried to avoid it with a `:not` approach to gather together the rules, but it's prolly clearer to do it like this :)

delucis (approved):
Thanks @julien-deramond! Looks great now 🙌

PR #615, opened 2023-08-30
Support Astro v3

delucis:

- Changes to Starlight code

#### Description

- Updates Starlight for compatibility for v3
- Drops support for Astro v2
- This is a bit of a rough one. I’d hoped to continue supporting both Astro v2 and v3 for a little while, but because we install some Astro integrations for users, we don’t really have a way to support old and new versions (because v3-compatible integrations don’t support v2 and vice versa). The one path that could work would be to force each Starlight user to install `@astrojs/mdx` and `@astrojs/sitemap` themselves, but I think that will likely cause people more headaches than it’s worth, so I think doing this in one go is likely the best choice.


HiDeoo:
> I’d hoped to continue supporting both Astro v2 and v3 for a little while, but because we install some Astro integrations for users, we don’t really have a way to support old and new versions (because v3-compatible integrations don’t support v2 and vice versa). The one path that could work would be to force each Starlight user to install @astrojs/mdx and @astrojs/sitemap themselves, but I think that will likely cause people more headaches than it’s worth, so I think doing this in one go is likely the best choice.

I also think it may be the best choice on the long run.

For what it’s worth, I already planned on doing the same in all my integrations too and release a new minor version for each of them (as they all are major version zero) requiring at least Astro v3 and w/e Starlight version will include the changes from this PR.

packages/starlight/index.ts
```diff
@@ -45,12 +45,7 @@ export default function StarlightIntegration(opts: StarlightUserConfig): AstroIn
 							// Configure Shiki theme if the user is using the default github-dark theme.
 							config.markdown.shikiConfig.theme !== 'github-dark' ? {} : { theme: 'css-variables' },
 					},
-					build: { inlineStylesheets: 'auto' },
-					experimental: {
-						assets: true,
-						// @ts-ignore - Needed for older versions of Astro, but an error since astro@2.6.0
-						inlineStylesheets: 'auto',
-					},
+					scopedStyleStrategy: 'where',
```
delucis, line 48: For now I think the safest is to set this to `'where'`, which was the Astro v2 default. The new default breaks some Starlight styling due to shifts in relative CSS specificity between global and component CSS. Those same shifts likely make it harder for Starlight users to override components, so I think `'where'` is the right choice.

Long term, I think we’ll want to move towards using `@layer starlight {}` for our built-in styles to help make it easier for users to override things with their CSS without worrying about specificity. At that point, we’d probably be safe to switch back to the Astro default and refactor a few internal styles where things clash.

packages/tailwind/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "peerDependencies": {
     "@astrojs/starlight": ">=0.7.2",
```
delucis, line 31: This will need bumping manually to `>=0.9.0` as a major change in the Tailwind plugin. I think we need to do that in the Changesets release PR.

ElianCodes (approved):
This is awesome! Congratz Chris

PR #617, opened 2023-08-30
Fix link to athenaos.org

SudoVanilla:
The link shown for "athenaos.org" in the showcase section hangs, removing "www" fixes this.


HiDeoo (approved):
Nice catch. I wonder if they are aware of the issue, could be nice maybe to notify them of the issue.

delucis (approved):
Thanks @SudoVanilla!

PR #619, opened 2023-08-31
docs: add obytes starter to showcase

yjose:
#### What kind of changes does this PR include?

- Add [Obytes starter](https://starter.obytes.com/) to showcases pages

PR #625, opened 2023-09-01
Ensure showcase image aspect ratios

HiDeoo:

#### Description

Following #619, there are different aspect ratios for the images in the showcases. All the previous images were `800 × 450` and the last one is `1200 × 630` which is not `16 / 9`:

SCR-20230901-qghg

This PR adds CSS rules to make the images always `16 / 9`:

SCR-20230901-qgun

Another approach could be enforcing specific image sizes / aspect ratios for the showcase images.


delucis (approved):
Ah yeah missed this. This looks good! We should probably also go back and make sure the image itself is 16 / 9 too, but this a sensible thing to have in here anyway just to be safe.

PR #638, opened 2023-09-02
Add support for component overrides

### LINKED ISSUE #415: Ability to customise Starlight’s components

0.5.2

### What is your idea?

The header of my docs site looks pretty sparse. I would like to add some glanceable information to it, like the latest release, GitHub stars etc., similar to this feature in the
Mkdocs/Material theme.

Screenshot 2023-07-27 at 8 27 10 AM

The header is currently underutilised space. Top-level menus or custom widgets could make use of that space.

Material theme for Mkdocs https://squidfunk.github.io/mkdocs-material/

### Participation

TheOtterlord:

- Changes to Starlight code

#### Description

- Closes #415
- What does this PR change? Adds a user config option to provide components overriding built-in Starlight components

#### Todo

- [ ] Documentation
- [ ] Test/Fix importing internal Starlight components (`@astrojs/starlight/internal`?)
- [ ] Support custom layouts? (possibly out of scope/not something we want)


lorenzolewis:
(omg omg omg the excitement is real ❗)

delucis:
Hey @TheOtterlord! Wanted to thank you for kicking this off. We’ll probably have to go a slightly different direction to support all the features we want though. Overriding loading like this, similar to Vite’s [`resolve.alias`](https://vitejs.dev/config/shared-options.html#resolve-alias) config option I think could cause issues when:

- A user wants to import one of our built-in components and actually get the original implementation (e.g. to wrap it/extend it with some extra code)

- Supporting our concept of “layout slots” (not sure yet what we should call these), but basically pre-defined places in the layout that are usually empty where a user can slide things into without impacting default components. For that to work with an override approach, we’d need a separate empty component for each “slot”.

Pretty sure we also don’t want people to end up importing all these various component-specific prop types and have varies access to the route context depending on which component they’re in.

I’ve been doing some work on the https://github.com/withastro/starlight/tree/dx-326/component-customization branch to refactor our internal codebase and standardise all components to receive a single consistent props object. There are a couple of exceptions to this in our recursive list components for the sidebar and table of contents, but on the whole it works well.

Currently prototyping an override system locally and I think the best solution is to use a virtual module.

Basic flow:

1. Users configure component overrides in the Starlight config, e.g. `{ components: { Footer: './my-footer.astro' } }`. For any components they don’t define, we use a default value, e.g. `@astrojs/starlight/components/Footer.astro`.

2. In Starlight’s Vite plugin we inject a virtual module that re-exports each of these, e.g.
   ```js
   export { default as Footer } from '@astrojs/starlight/components/Footer.astro';
   // etc.
   ```
   (switching the module name for whatever a user provided instead if overriden)
   
3. In Starlight component code, instead of importing components directly we import from this virtual module.

This means that a user can still get the original implementation by importing it directly, while we can override components as needed.

TheOtterlord:
Yep, I like this. I had a similar plan for forwarding components, but I like the prop standardisation too. I'll close this PR, and the efforts can continue in https://github.com/withastro/starlight/tree/dx-326/component-customization

Issue #462, opened 2023-08-04
Synced tabs


HiDeoo:
### What version of `starlight` are you using?

0.6.1

### What is your idea?

This proposal is a follow-up on a [Discord discussion](https://discord.com/channels/830184174198718474/1070481941863878697/1136595999121489930) about the addition of synced tabs to Starlight.

### Why is this feature necessary?

This features is pretty useful to show tabular data where the user will benefit from having the tabs synced. e.g.:

- npm related commands for various package managers
- code examples in different languages / frameworks
- different versions of an API

### Do you have examples of this feature in other projects?

- A first good one is the Astro documentation, e.g. the [installation guide](https://docs.astro.build/en/install/auto/) where if a user select for example the `pnpm` package manager tab in the step 1, the step 2 command will be updated to show the `pnpm` command.

https://github.com/withastro/starlight/assets/494699/5b92c6b3-0d79-4958-ae27-68fbba16d6d8

- The Firebase documentation uses a similar approach, e.g. the [Cloud Firestore initialization steps](https://firebase.google.com/docs/firestore/query-data/get-data#initialize) where the user can select the language of the code example.

https://github.com/withastro/starlight/assets/494699/64a11c1c-3ba1-4262-8416-5aca5ca34f46

- I made a [small component](https://starlight-package-managers.vercel.app/demo/) to quickly display npm related commands for multiple package managers in my Starlight documentation sites and it uses synced tabs like the Astro docs to sync the package manager selection between the tabs.

https://github.com/withastro/starlight/assets/494699/bf44cbc3-7051-46fd-9a4a-16592c149610

### API

I like the Astro docs current approach of basically passing a basic string to the `<Tabs />` component when you want to sync the tabs. This is a very simple API and it is easy to understand what is going on imo.

Altho, I am personally not a huge fan of the `sharedStore` prop name currently used. It kinda feels like it is leaking some implementation details and not super obvious it's related to syncing the tabs. I would prefer something like `sync` or `syncKey` for example.

Any other ideas for the end user API?

### Implementation

The Astro current implementation is using a Preact component with a hook to handle the sync which relies on [Nano Stores](https://github.com/nanostores/nanostores) to sync some state between the tabs. I would assume this is an approach we would want to avoid in Starlight as it would add some dependencies and one [UI framework](https://docs.astro.build/en/core-concepts/framework-components/).

[...]

PR #640, opened 2023-09-04
Add support for synced tabs

HiDeoo:

- Changes to Starlight code

#### Description

- Closes ~~#462~~#953.

This pull request adds support for synced tabs.

https://github.com/withastro/starlight/assets/494699/ad1117b9-5862-465b-aea9-d8467e27b709

- Entirely opt-in.
- Relies on the `syncKey` property to enable syncing (can be renamed easily).
- It does not use an external store like the docs theme editor but everything is colocated in the `StarlightTabs` web component (we would only have 1 use case for such a store right now so I feel it's easier to follow if everything is scoped to this component).

Regarding testing, now that https://github.com/withastro/astro/pull/7979 has been merged, this PR introduces some E2E tests with an architecture a bit similar to the one in the Astro repo:

- It uses the new programmatic API to start a dev server for a specific fixture.
- We can use Playwright to interact with a page and assert that some tabs are synced, some are not, focus is behaving correctly, etc.

The approach is much more lightweight than the one in the Astro repo as we don't need at the moment so much features (edits, builds and preview, etc.). My idea was to make it as simple as possible to get started with E2E tests for what I needed in this PR and then, we can iterate on it depending on our needs. Maybe we will also want to add E2E tests for some other already existing features but I think it might be better to do it in a separate PR (or we could also only add tests for existing features when we get bug reports for them as regression tests).

I also edited the `CONTRIBUTING.md` file to add a section about E2E tests with an emphasis on the fact that they should only be used for parts that cannot be properly tested with unit tests (e.g. client code for user interactions).

Hope this is what you had in mind Chris when you mentioned the programmatic API in #462. Happy to discuss any part of this PR if needed.


packages/starlight/style/props.css
```diff
@@ -79,7 +79,7 @@
 	--sl-text-h4: var(--sl-text-xl);
 	--sl-text-h5: var(--sl-text-lg);
 
-	--sl-line-height: 1.8;
+	--sl-line-height: 1.75;
```
HiDeoo, line 82: This change is on purpose to avoid sub-pixel scroll position changes when switching between tabs with different content height.

In some browsers (Chrome (Blink) and Firefox), a line height with a fractional part (which is the case of Starlight before this change: `16 * 1.8 = 28.8`) seems to cause scripts accessing dimensions involving the line height to get an inconsistent rounded value.
In the specific case of tabs, when restoring the scroll position after switching between tabs and especially when the content height is different between tabs, this could cause a sub-pixel scroll position change.

https://github.com/withastro/starlight/assets/494699/26218f0c-79c8-416b-96b8-4783d2190755

This change brings the base line height to a whole number (`16 * 1.75 = 28`) which avoids the issue in all browsers I tested and all device pixel densities I could test.

https://github.com/withastro/starlight/assets/494699/c053b6e6-9511-499b-9c10-b7baf172e757

I also added a test which ensure the bounding box of a set of tabs is identical before and after switching tabs in a case where a tiny scroll position would have been applied before.

---

Additional question: I guess this should have its own changeset, right, as subtle as it is?

HiDeoo:
Just updated the pull request and also added a new behavior to avoid scroll position changes when alternating between tabs which could happen for example when switching between tabs with different content height.

CONTRIBUTING.md
```diff
@@ -187,6 +187,36 @@ pnpm test:coverage
 
 This will print a table to your terminal and also generate an HTML report you can load in a web browser by opening [`packages/starlight/__coverage__/index.html`](./packages/starlight/__coverage__/index.html).
 
+### E2E tests
+
+Starlight also includes E2E tests in [`packages/starlight/__e2e__/`](./packages/starlight/__e2e__/), which are run using [Playwright][playwright].
```
delucis, line 192: Suggesting spelling out the acronym when first used for anyone not familiar:

```suggestion
### End-to-end (E2E) tests

Starlight also includes E2E tests in [`packages/starlight/__e2e__/`](./packages/starlight/__e2e__/), which are run using [Playwright][playwright].
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -87,6 +87,48 @@ The code above generates the following tabs on the page:
 	</TabItem>
 </Tabs>
 
+#### Synced tabs
+
+Multiple sets of tabs on the same page can be synced so that switching tabs in one switches the others to a tab with the same label.
```
delucis, line 92: Suggested rewording here. I’m still not super happy — maybe @sarah11918 would have some good ideas for explaining this clearly.
```suggestion
Keep multiple sets of tabs on the same page in sync with the `syncKey` attribute. Switching to a synced tab will also switch to the tab with the same label in other `<Tabs>` that share the same `syncKey`.
```

packages/starlight/__e2e__/tabs.test.ts
```diff
@@ -0,0 +1,122 @@
+import { expect, testFactory, type Locator } from './test-utils';
+
+const test = await testFactory('./fixtures/basics/');
+
+test('syncs tabs with a click event', async ({ page, starlight }) => {
```
delucis, line 5: I updated branch and this test is failing now. I wonder if it’s due to #1723, which added `aria-select="false"` to inactive tabs? Might need to update the logic to correctly target the active tab.

packages/starlight/__e2e__/tabs.test.ts
```diff
@@ -0,0 +1,122 @@
...
+	// Get the bounding box of the tabs.
+	const initialBoundingBox = await tabs.boundingBox();
+
+	// Select the second tab which has a different height.
+	await selectedTab.press('ArrowRight');
+
+	// Ensure the tabs vertical position is exactly the same after selecting the second tab.
+	// Note that a small difference could be the result of the base line-height having a fractional part which can cause a
+	// sub-pixel difference in some browsers like Chrome or Firefox.
+	expect((await tabs.boundingBox())?.y).toBe(initialBoundingBox?.y);
```
delucis, line 116: Super impressed you managed to write a test for this! Galaxy brain stuff ✨

packages/starlight/style/props.css
delucis: Yes, I agree it might be a good idea to add a separate changeset documenting the line-height change, why it was made, and how to revert it if someone really wants to.

packages/starlight/user-components/Tabs.astro
```diff
@@ -117,9 +133,13 @@ const { html, panels } = processPanels(panelHtml);
 			});
 		}
 
-		switchTab(newTab: HTMLAnchorElement | null | undefined, index: number) {
+		switchTab(newTab: HTMLAnchorElement | null | undefined, index: number, shouldSync = true) {
 			if (!newTab) return;
 
+			// We store the current position so we can restore it after switching tabs to prevent the page from jumping when
+			// the new tab content is of a different height than the previous tab.
+			const previousTabsOffset = this.getBoundingClientRect().top;
```
delucis, line 141: Might it be worth only calculating this when `shouldSync` is true? Or is that an unnecessary optimisation?

delucis (commented):
So sorry to have taken so long to get through all this. Really amazing work as usual @HiDeoo!

Left a few comments, but on the whole I think we can get this ready to ship pretty soon 🚀

packages/starlight/user-components/Tabs.astro
HiDeoo: Definitely, I don't think this would lead to a performance issue in this case but it's well known that `getBoundingClientRect()` can be a performance bottleneck in more intensive scenarios so we may as well be a good citizen and avoid it when totally unnecessary. Great catch!

packages/starlight/__e2e__/tabs.test.ts
HiDeoo: Turns out, this was a flacky test and it took me a bit to figure out what was going on. After trying to reproduce locally and couldn't, I tried a Linux machine, a Windows machine but all worked fine.
I had to enable [traces](https://playwright.dev/docs/trace-viewer) of the test on GitHub Actions and replay it locally to figure it out (amazing feature btw) what was going on.
During the early stage of the PR, I used the dev server to setup the E2E tests to iterate faster and never thought of switching to a production build (which is what we ultimately want to test). According to the traces, during the huge amount of requests made by Vite, the page could end up being reloaded entirely sometimes on top of being slow to load the first time (I assume due to the limited resources of the CI).
After noticing this pattern in failing traces, I switched to a production build as expected and this also fixed the flacky test.

docs/src/content/docs/guides/components.mdx
```diff
@@ -87,6 +87,49 @@ The code above generates the following tabs on the page:
 	</TabItem>
 </Tabs>
 
+#### Synced tabs
+
+Keep multiple sets of tabs on the same page in sync with the `syncKey` attribute.
+Switching to a synced tab will also switch to the tab with the same label in other `<Tabs>` that share the same `syncKey`.
+
+To synchronize tabs, add an identical `syncKey` property to each related `<Tabs>` component.
```
sarah11918, line 95:
```suggestion
Update the display of multiple tab groups at the same time by adding the `syncKey` attribute.

All `<Tabs>` on a page with the same `syncKey` value will display the same active label. This allows your reader to choose once (e.g. their operating system or package manager), and and see their choice reflected throughout the page.

To synchronize related tabs, add an identical `syncKey` property to each `<Tabs>` component and ensure that they all use the same `<TabItem>` labels:
```
Here's my attempt? Spark any better idea?

favna (changes requested):

packages/starlight/user-components/Tabs.astro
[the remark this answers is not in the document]
delucis: Thanks for the feedback! I think it makes sense to ship stuff incrementally, so should be fine to add this initial support for synced tabs and then consider whether to extend it to use `localStorage` in a subsequent PR.

docs/src/content/docs/guides/components.mdx
delucis: I like this!

I also still quite like “Keep […] in sync” because it’s not so much the author who will “update the display,” more how they are managing the user’s experience.

Maybe we can keep it with something like this?

```suggestion
Keep multiple tab groups synchronized by adding the `syncKey` attribute.

All `<Tabs>` on a page with the same `syncKey` value will display the same active label. This allows your reader to choose once (e.g. their operating system or package manager), and see their choice reflected throughout the page.

To synchronize related tabs, add an identical `syncKey` property to each `<Tabs>` component and ensure that they all use the same `<TabItem>` labels:
```

delucis (approved):
Approving for release! Still working my way through our PR backlog, but this should be able to go out later today.

Thank you for the work on this @HiDeoo — probably taking the award for longest lived PR so far 😅

packages/starlight/user-components/Tabs.astro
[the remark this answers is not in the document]
favna: That's fine by me

PR #647, opened 2023-09-04
add option to docs schema to exclude a page from pagefind

### LINKED ISSUE #646: Option to exclude a page from the pagefind index

0.9.0

### What is your idea?

I want to be able to exclude a page (published, not a draft) from the pagefind index. It seems like an easy change, so I'll include a pull request.

There could be a few reasons for excluding a page from search – perhaps there are multiple pages  with similar content, or pages that are not intended for a site's general audience.

_No response_

### Participation

bgmort:

- adds ability to exclude a page from the pagefind index by setting `pagefind: false` in the frontmatter
- Changes to Starlight code

#### Description

- Closes #646 
- adds ability to exclude a page from the pagefind index by setting `pagefind: false` in the frontmatter


packages/starlight/schema.ts
```diff
@@ -165,5 +165,8 @@ export function docsSchema() {
 					content: z.string(),
 				})
 				.optional(),
+
+			/** Pagefind indexing for this page - set to false to disable. */
+			pagefind: z.boolean().optional().default(true),
```
delucis, line 170: `.default()` here implies `.optional()`, so we can simplify that:

```suggestion
			pagefind: z.boolean().default(true),
```

packages/starlight/layout/Page.astro
```diff
@@ -45,6 +45,7 @@ const tocConfig = !hasSidebar
 	: config.tableOfContents;
 const hasToC = Boolean(tocConfig);
 const hasHero = Boolean(entry.data.hero);
+const pagefindEnabled = entry.slug !== '404' && entry.data.pagefind !== false;
```
delucis, line 48: Revisiting this code reminded me we never accounted for translated 404s with this code. We can fix that by adding a test for slugs ending in `/404` like `es/404` etc.

```suggestion
const pagefindEnabled = entry.slug !== '404' && !entry.slug.endsWith('/404') && entry.data.pagefind !== false;
```

delucis (commented):
Thank you for this @bgmort!

We’ve talked about adding more complex options behind a `pagefind` key in frontmatter, but I don’t think supporting `pagefind: false` will block that. Later we can keep the `false` shorthand to disable indexing even if we add more options.

Can you add docs for this to https://starlight.astro.build/reference/frontmatter/ ?

bgmort:
> Can you add docs for this to https://starlight.astro.build/reference/frontmatter/ ?

Done!

delucis (approved):
Thank you again for this @bgmort! I’ve added some changesets describing what this PR does and will get it released now. Welcome to Starlight’s contributors 🤩

PR #656, opened 2023-09-05
Remove sharp image processor config

sibiraj-s:

- Updates to templates

#### Description

Removed sharp image processor from templates/examples. Sharp is the default image processor for Astro since v3.


delucis (approved):
Thanks for updating these @sibiraj-s!

PR #659, opened 2023-09-05
Fix sidebar guide a11y and type issue

HiDeoo:

#### Description

This PR fixes 2 issues with the newly added sidebar guide:

- Remove the ID in the `sidebar-preview` component to avoid duplicated IDs and we don't need it here anyway.
- Fix a type issue in the `sidebar-preview` component (I assume the import was auto-generated by VSC when the project was still using Astro v2 and somehow fine but now it's not as expected as they are not exported types and was not catch as we do not run `astro check`).

Edit: I tested locally before pushing but pa11y also properly run on the sidebar guide in this PR:

> \> http://localhost:4321/guides/sidebar/ - 0 errors

This PR does not investigate why `pa11y` was not run in the sidebar guide PR, I'll do that in a separate PR.

### DIFF
--- a/docs/src/components/sidebar-preview.astro
+++ b/docs/src/components/sidebar-preview.astro
@@ -1,6 +1,5 @@
 ---
-import type { AutoSidebarGroup, SidebarItem } from '@astrojs/starlight/utils/user-config';
-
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! Thanks for the quick fix @HiDeoo ✅

PR #660, opened 2023-09-06
Update `pa11y-ci` sitemap URL

HiDeoo:

- Something else!

#### Description

This PR updates the sitemap URL used by `pa11y-ci` during tests to no longer use the sitemap index file as the `--sitemap-find` and `--sitemap-replace` options are not used for a sitemap index. This would mean that new pages were not being tested before hitting production.

I tested this locally by adding a new guide - on the left, a run before the change, on the right, a run after the change:

pa11y


delucis (approved):
Nice. Thanks for spotting this yesterday and for the fix!

PR #565, opened 2023-08-24
Add sidebar guide

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content

#### Description

This pull request is an attempt to add a sidebar guide to the Starlight docs site as mentioned by @sarah11918 in the last Talking and Doc'ing session.

I am usually not the best at guides, even more in English, but I decided to give it a try anyway. The main ideas are:

- Provide a progression: no sidebar config at all -> basic sidebar config -> various advanced use cases
- For each step, start from either a filesystem structure or a configuration example and then show the resulting generated sidebar with a small "widget".

This started as a new section in the "Customizing Starlight" page but at some point, it felt too long and I decided to make it a separate page.

This PR is a draft for multiple reasons:

- Is this any good? Does this provide any value compared to the current docs?
- Is this too long?
- If this is at some point merged, should some changes be made to the current sidebar docs?
- The custom component used to display the sidebar examples was quickly written as a PoC and can probably be improved but I wanted to get some feedback on the idea before spending too much time on it.

I am open to any feedback and suggestions or even starting from scratch if this is not the right approach for a sidebar guide.

PR #672, opened 2023-09-06
i18n(fr): add `guides/sidebar.mdx`

HiDeoo:

#### Description

This PR adds the french translation of `guides/sidebar.mdx` added in #565.

One interesting question about this one:

- The guide contains sidebar examples which often redirects to existing pages in the docs.
- We usually translate doc examples.
- I tried to follow the same process here and links redirect to the french version of the docs.

Altho, this is a bit of a problem for autogenerated groups as if we want to redirect to the french version of the docs, we would need to add a `fr` folder in the file tree examples which adds a bit of unecessary complexity I think.

So I decided to keep linking to the english version of the docs for these examples (even tho I translated the labels) but I am not sure if this is the best way to do it. I am open to suggestions.


docs/src/content/docs/fr/guides/sidebar.mdx
```diff
@@ -0,0 +1,488 @@
...
+
+<SidebarPreview
+	config={[
+		{
+			label: 'Guides',
+			items: [
+				{ label: 'Composants', link: '/guides/components/' },
+				{ label: 'Internationalisation (i18n)', link: '/guides/i18n/' },
+				{
+					label: 'advanced',
```
julien-deramond, line 187: This label should probably be translated as well here, shouldn't it?

```suggestion
					label: 'Avancé',
```

HiDeoo: > This label should probably be translated as well here to avoid the following rendering, shouldn't it?

It should not in this case. The entire "Guides" group is auto generated meaning pages and groups are based on the file structure:

- Pages like `components.md`, `i18n.md`, `project-structure.md` have their label picked from the page frontmatter titles or `sidebar.label` property meaning it can be translated in the page.
- Subdirectories like `advanced/` are displayed as-is meaning that to get "Avancé" the directory on disk should actually be `Avancé/`.

Maybe directory names could be translated but it feels a bit weird to me considering that some won't be translated like `src`, `content`, etc. and even more with an `é` in the directory name.

julien-deramond: > Subdirectories like advanced/ are displayed as-is meaning that to get "Avancé" the directory on disk should actually be Avancé/.

Oh OK, I understand now the subtlety here. Found it so weird while reading this page. Thanks for the clarification.

delucis (approved):
Are there still any open questions here? I think this looks good!

HiDeoo:
> Are there still any open questions here?

Almost forgot about this one ^^ I think we are good to go here now that it has been reviewed. I'll integrate the changes from #696 and then merge this.

Issue #674, opened 2023-09-06
Pagefind incorrect urls when built


Tanhe:
### What version of `starlight` are you using?

0.9.0

### What version of `astro` are you using?

3.0.6

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Chromium (Edge)

### Describe the Bug

Starting from a brand new project. 

`npm run build`

`npm run preview`

When you select the search box the page fails to load the pagefind scripts as it's looking for `_pagefind/` but the dist folder has `pagefind/`.

![image](https://github.com/withastro/starlight/assets/5613534/2bc6f70c-83e1-426f-9f56-ae7e737a5c83)

![image](https://github.com/withastro/starlight/assets/5613534/b5cafafc-6c8e-491c-9d11-b871bcb7bc3c)

Looking at this it might be this hardcoded bundlePath in `Search.astro`:

![image](https://github.com/withastro/starlight/assets/5613534/ae99e518-fb7c-4bf9-87f3-8a42cecb2dbf)

As a separate note, the build issues this warning:

![image](https://github.com/withastro/starlight/assets/5613534/fbe07c7f-b525-410e-bb5c-dc359d665d74)

There is a required change necessary here in index.ts:

![image](https://github.com/withastro/starlight/assets/5613534/2b80e31a-185f-405e-81e9-819e32e7075e)
^^ --source changed to --site makes the warning go away

### Participation

- [ ] I am willing to submit a pull request for this issue.

delucis:
Closed in #676!

PR #676, opened 2023-09-06
fixes issue #674

vedmalex:
this PR fixes issue with pagefind configuration #674


vedmalex:
@HiDeoo I've fixed deps and will examine search changes

bglw:
Hey @HiDeoo & @vedmalex 👋 

Just popping into this PR to mention that Pagefind `v1.0.0-beta.3` includes improvements around compound words and excerpts, thanks for the feedback! ❤️ 

That excerpt _does_ have a lot of mentions of `component`, so it might still favor the code, but in most cases compound words will be less likely to be the focus of an excerpt.

Additionally, `beta.3` adds the ability to lower the weight of sections, and this weighting is used for both search ranking and excerpt generation, which means you can tailor this further by wrapping your code blocks in `data-pagefind-weight="0.5"` (or any number between 0 and 1) — which would make code even less likely to be the focus.

HiDeoo:
Damn @bglw, this looks super great and this was so fast 😮 Thanks for the amazing work! Time to play with all the new changes ^^

bglw:
Follow-up to ☝️ 

There was a wee issue with that initial fix that meant some compound words actually ranked _higher_ 😅 

That's fixed on `v1.0.0-beta.5`, if you want to try it out in the next... 22 hours or so 😄

bglw:
Final follow-up, `v1.0.0-beta.6` sneaks in some last minute quick-wins on ranking.

Before / after:
Pagefind MDN Search Demo

HiDeoo:
> Final follow-up, `v1.0.0-beta.6` sneaks in some last minute quick-wins on ranking.

Wow, this looks super neat! I have to say that I am super impatient, amazing work once again.

PR #679, opened 2023-09-07
i18n(pt-BR): Update `reference/frontmatter.md`, `guides/customization.mdx` & `showcase.mdx`

Alynva:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content

#### Description

- What does this PR change? Give us a brief description.
  - Update translating based on changes in the original files.

PR #688, opened 2023-09-08
Extract showcase sites to a new component

HiDeoo:

#### Description

As suggested [on Discord](https://discord.com/channels/830184174198718474/1070481941863878697/1149638074037452870), this PR moves all the showcase sites to a new component to avoid having to update it in different languages when a new site is added considering there is no language-specific content.

> I guess you could extract the data to component frontmatter if you like?

Tried that, but it does not have many advantages imo over the current approach right now even diff wise.

Notes:

- We are not doing the same for the community plugins as they all include a language-specific description.
- It looks like (at the moment of creating this PR) #679 is the only opened i18n PR updating the showcase, I guess it should be merged before this one so they get credit for the work and to not lose track of this in the i18n tracker.

### DIFF
--- a/docs/src/components/showcase-sites.astro
+++ b/docs/src/components/showcase-sites.astro
@@ -1,0 +1,55 @@
+---
+import Card from './showcase-card.astro';
+import FluidGrid from './fluid-grid.astro';
+---
+
+<FluidGrid>
+	<Card title="Athena OS" href="https://athenaos.org" thumbnail="www.athenaos.org.png" />
+	<Card
+		title="PubIndexAPI Docs"
+		href="https://docs.pubindexapi.com/"
+		thumbnail="docs.pubindexapi.com.png"
+	/>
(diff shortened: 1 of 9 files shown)


dreyfus92 (approved):
Nice work @HiDeoo 😁, LGTM ✅

delucis (commented):
Thank you @HiDeoo — this looks great! I think we can safely go through updating the translated showcase pages to use this as part of this PR as well. I just merged the two open PRs that updated Japanese and Portuguese translations, so the only remaining out-of-date translation is the French one, but only due to changes that will be updated by moving to this component.

HiDeoo:
Perfect, I just pushed an update to use the new component in all translated pages.

I guess it will need the i18n-skip-thingy label but I don't remember the name and don't have my notes with me. After that, I guess the only remaining task will be to update #711 to use this (and maybe update their screenshot, it looks weird when I quickly checked it).

delucis (approved):
Fantastic — LGTM! No need to `i18nIgnore` on this one as we want to mark all the impacted pages as up-to-date and also include this commit in the update history for the German page whenever it is added.

PR #694, opened 2023-09-09
feat: ViewTransition Astro component

abosch19:
Allow user to set a new user configuration: `viewTransitions` (boolean)
This config sets Astro ViewTransitions component on Page layout if true

- Changes to Starlight code

#### Description

- What does this PR change? User configuration and Page layout

### DIFF
--- a/packages/starlight/layout/Page.astro
+++ b/packages/starlight/layout/Page.astro
@@ -1,6 +1,8 @@
 ---
 import config from 'virtual:starlight/user-config';
 import type { MarkdownHeading } from 'astro';
+import { ViewTransitions } from 'astro:transitions';
+
 import { getSidebar } from '../utils/navigation';
 import type { Route } from '../utils/routing';
 
@@ -84,6 +86,7 @@ const pagefindEnabled =
 				}
 			}
 		</style>
+		{ entry.data.viewTransitions && <ViewTransitions /> }
 	</head>
 	<body>
 		<ThemeProvider />
--- a/packages/starlight/layout/Page.astro
+++ b/packages/starlight/layout/Page.astro
@@ -86,7 +86,7 @@ const pagefindEnabled =
 				}
 			}
 		</style>
-		{ entry.data.viewTransitions && <ViewTransitions /> }
+		{ config.viewTransitions && <ViewTransitions /> }
 	</head>
 	<body>
 		<ThemeProvider />
--- a/packages/starlight/schema.ts
+++ b/packages/starlight/schema.ts
@@ -59,6 +59,12 @@ export function docsSchema() {
 			 */
 			template: z.enum(['doc', 'splash']).default('doc'),
 
+			/**
+			 * Set ViewTransitions Astro component for this page.
+			 * 
+			 */
+			viewTransitions: z.boolean().optional().default(false),
+
 			/** Display a hero section on this page. */
 			hero: z
 				.object({
--- a/packages/starlight/schema.ts
+++ b/packages/starlight/schema.ts
@@ -59,12 +59,6 @@ export function docsSchema() {
 			 */
 			template: z.enum(['doc', 'splash']).default('doc'),
 
-			/**
-			 * Set ViewTransitions Astro component for this page.
-			 * 
-			 */
-			viewTransitions: z.boolean().optional().default(false),
-
 			/** Display a hero section on this page. */
 			hero: z
 				.object({
--- a/packages/starlight/utils/user-config.ts
+++ b/packages/starlight/utils/user-config.ts
@@ -223,6 +223,12 @@ const UserConfigSchema = z.object({
 	 */
 	defaultLocale: z.string().optional(),
 
+	/**
+	 * Set ViewTransitions Astro component for this page.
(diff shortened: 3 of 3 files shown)


turbek:
It looks like this currently breaks the pagefind/search functionality and also needs an astro version update.

delucis:
> It looks like this currently breaks the pagefind/search functionality and also needs an astro version update. 

Thanks for testing that @turbek! Yes, View Transitions require all client-side scripts to include special handling, so it may be more involved than the changes currently in this PR. We _might_ be able to solve this with `transition:persist`, but I’m not sure.

turbek:
`transition:persist="site-search"` and a littlebit of extra code for the initialization solves the problem:
```
document.addEventListener('astro:page-load', () => {
	if (import.meta.env.DEV) return;

	const pageFind = document.querySelector('.pagefind-ui__search-input');

	if (pageFind) return;

	const onIdle = window.requestIdleCallback || ((cb) => setTimeout(cb, 1));
	onIdle(async () => {
		const { PagefindUI } = await import('@pagefind/default-ui');
		new PagefindUI({
			element: '#starlight__search',
			baseUrl: import.meta.env.BASE_URL,
			bundlePath: import.meta.env.BASE_URL.replace(/\/$/, '') + '/pagefind/',
			showImages: false,
			translations,
			showSubResults: true,
		});
	});
}, { once: true });
```

Although I'm not sure how ideal it is

turbek:
Actually we removed the language switching because we currently support 1 language and since our website doesn't support dark theme atm, we removed theme switching as well. So we never tested theme/language switching features with the SPA mode.

ToC highlighting seems to work as expected though.

onemen:
I was able to activate `View Transitions` by overrides Starlight `Head.astro` component.

I omitted translations since i don't have any. 

```
 npm install @pagefind/default-ui  
```

Add this custom `Head.astro` to your custom component in the [components](https://starlight.astro.build/reference/configuration/#components) configuration option in `astro.config.mjs`:
```astro

---
import Default from '@starlight/components/Head.astro';
import type { Props } from '@starlight/props';
import { ViewTransitions } from 'astro:transitions';
---

<Default {...Astro.props}><slot /></Default>
<ViewTransitions />

<script>
  // Skip PagefindUI on initial page load since it's already run by Search.astro.

  // Flag to track initial page load
  let initialLoad = true;

  document.addEventListener('astro:page-load', () => {
    // Skip PagefindUI on initial page load (and only execute on subsequent loads)
    if (initialLoad) {
      initialLoad = false;
      return;
    }

    // Code to initialize PagefindUI on subsequent loads (if needed)
    if (import.meta.env.DEV) return;

    const pageFind = document.querySelector('.pagefind-ui__search-input');
    if (pageFind) return;

    const onIdle = window.requestIdleCallback || (cb => setTimeout(cb, 1));
    onIdle(async () => {
      // @ts-expect-error — Missing types for @pagefind/default-ui package.
      const { PagefindUI } = await import('@pagefind/default-ui');
      new PagefindUI({
        element: '#starlight__search',
        baseUrl: import.meta.env.BASE_URL,
        bundlePath: import.meta.env.BASE_URL.replace(/\/$/, '') + '/pagefind/',
        showImages: false,
        translations: {},
        showSubResults: true,
      });
    });
  });
</script>

```

lukethacoder:
Super easy to add `ViewTransitions` to starlight in a similar way to @onemen, just needed to adjust the import paths.

Using `@astrojs/starlight/*` instead of `@starlight/*`

```astro
---
import Default from '@astrojs/starlight/components/Head.astro'
import type { Props } from '@astrojs/starlight/props'
import { ViewTransitions } from 'astro:transitions'
---

<Default {...Astro.props}><slot /></Default>
<ViewTransitions />
```

```js
import { defineConfig } from 'astro/config'
import starlight from '@astrojs/starlight'

// https://astro.build/config
export default defineConfig({
  integrations: [
    starlight({
      components: {
        Head: './src/components/Head.astro',
      }
    }),
  ],
})
```

martrapp:
Adding view transition to a starlight side might involve more than enabling the basic transitions.
Beside all the things Fryuni mentioned that must work if you use them, there is a brave new world of extensions you might want to have, one view transitions are enabled:
- Sidebar behavior that auto-opens categories if closed, e.g. on back navigation or that scrolls the current entry into view if it isn't
- Reasonable animations when you navigate after you scrolled a page down (this typically results in a large morph animations that looks like wild scrolling
- Animation direction dependent on the order of your pages 

It might be possible to add all this and a lot of options to control that to starlight itself. 
But this would also bring some complexity to the project.

For all being interested in a community solution for adding view transitions to starlight, have a look at the guide in the 👜 Bag of Tricks ✨:
https://events-3bg.pages.dev/jotter/starlight/guide/

delucis:
Thanks again for this PR and apologies for not getting back to you sooner.

Given https://events-3bg.pages.dev/jotter/starlight/guide/ now provides guidance on enabling view transitions in Starlight for those who want them, and given the gradual browser adoption of transitions _without_ the need for any client-side router, I think we can close this.

PR #712, opened 2023-09-12
Feat: Add badge to LinkCard component

lorenzolewis:

- Changes to Starlight code

#### Description

- Exposes an optional `badge` prop to the `LinkCard` component
Screenshot 2023-09-12 at 10 04 20

- I'm not 100% sure if I used `BadgeConfigSchema().parse()` correctly in `LinkCard.astro` as this is my first time working with that bit of Zod, so any review/feedback on that would be much appreciated.


packages/starlight/user-components/LinkCard.astro
```diff
@@ -1,20 +1,36 @@
 ---
 import Icon from './Icon.astro';
+import Badge from '../components/Badge.astro';
+import type { Badge as BadgeType } from '../schemas/badge';
+import { BadgeConfigSchema } from '../schemas/badge';
```
HiDeoo, line 5:
```suggestion
import { BadgeConfigSchema, type BadgeUserConfig } from '../schemas/badge';
import { errorMap } from '../utils/error-map';
```

We can consolidate both `import` here and I explain `BadgeUserConfig` and the `errorMap` a little bit below.

packages/starlight/user-components/LinkCard.astro
```diff
@@ -1,20 +1,36 @@
...
 import type { HTMLAttributes } from 'astro/types';
 
 interface Props extends Omit<HTMLAttributes<'a'>, 'title'> {
 	title: string;
+	badge: BadgeType | undefined;
```
HiDeoo, line 10:
```suggestion
	badge?: BadgeUserConfig;
```

This needs to be optional otherwise TypeScript would complain if a `<LinkCard>` is called without a `badge` property.

Regarding `BadgeUserConfig`, you cannot use the `Badge` type exported from `packages/starlight/schemas/badge.ts` as it only contains the object definition and not the string shorthand.

To get a type accepting both, you would need to export a new type in `packages/starlight/schemas/badge.ts`:

```ts
export type BadgeUserConfig = z.input<ReturnType<typeof BadgeConfigSchema>>;
```

packages/starlight/user-components/LinkCard.astro
```diff
@@ -1,20 +1,36 @@
...
 	description?: string;
 }
 
-const { title, description, ...attributes } = Astro.props;
+const { title, badge, description, ...attributes } = Astro.props;
+
+const badgeData = BadgeConfigSchema().parse(badge);
```
HiDeoo, line 16:
```suggestion
const badgeData = BadgeConfigSchema().parse(badge, { errorMap });
```

If an invalid prop is passed down, without using the Starlight error map, the user will see:

SCR-20230912-qylk

When using the error map, the error becomes:

SCR-20230912-qypb

Altho, I wonder if we should use `parse` here and not `safeParse` and re-throw a better error considering that with this setup, the error would be reported in `lib/index.mjs:537:31` or something which not super helpful.

Any thoughts? (an example on how to use `safeParse` is available in `packages/starlight/index.ts`)

HiDeoo (commented):
I really like the idea of this PR personally. Left a few comments mostly regarding the technical aspect. 

UI-wise, I am not sure if it's just me or not but the label does not feel vertically centered but I'll have to take a closer look when I have more time.

packages/starlight/user-components/LinkCard.astro
lorenzolewis: Ah thank you. I assumed there was something I was missing here and this makes sense

packages/starlight/user-components/LinkCard.astro
lorenzolewis: Awesome! Thanks for the suggestion! If we use `safeParse` then that would enable better error surfacing if I'm understanding your comment correctly, right? If so then I'd get behind that.

lorenzolewis:
> I really like the idea of this PR personally. Left a few comments mostly regarding the technical aspect.

Thanks! Appreciate it!

> UI-wise, I am not sure if it's just me or not but the label does not feel vertically centered but I'll have to take a closer look when I have more time.

I agree. I think the sidebar has the same sort of issue and this is the same CSS as that. I think it's aligned to the text baseline here, but I wonder if it would make sense to centre it overall vertically instead via `flex` or similar?

packages/starlight/user-components/LinkCard.astro
HiDeoo: Yeah, we could do something like this technically:

```ts
import { AstroError } from 'astro/errors';

// …

const badgeData = BadgeConfigSchema().safeParse(badge, { errorMap });

if (!badgeData.success) {
	throw new AstroError(
		'Invalid `badge` prop passed down to `<LinkCard>`.',
		'Find out more in our documentation at https://starlight.astro.build/reference/frontmatter/#sidebar'
	);
}

// …

// Here, we know the data is valid.
badgeData.data;
```

Which would give us maybe a better error:

image

Is this any better? I think this would be the first time we would use zod in a user component, so not sure what would be the best way to surface this kind of error.

packages/starlight/user-components/LinkCard.astro
HiDeoo: Unresolving this to not forget it but `export type BadgeUserConfig = z.input<ReturnType<typeof BadgeConfigSchema>>;` would still need to be added to `packages/starlight/schemas/badge.ts`.

lorenzolewis: Thanks for that, just added it to the latest commit

packages/starlight/user-components/LinkCard.astro
lorenzolewis: Thanks! I tweaked the error message slightly from your suggestion to hopefully make it a little more clear:

Screenshot 2023-09-12 at 14 00 00

simonhyll:
Gonna throw my idea in here:
Automatic application of the "New" badge to the sidebar and LinkCards based on the pages `created` age according to Git, with a configuration for how old you consider to still be new. Could also have an "Updated" badge based on `changed` in Git. Similar to how the last changed footer feature works.

HiDeoo (commented):
> I tweaked the error message slightly from your suggestion to hopefully make it a little more clear:

Good call, I didn't really think too much about the text when typing up the example 👍

HiDeoo:
> based on the pages `created` age according to Git

The issue with this kind of features is that certain hosting providers don't allow for deep clones of their repos so you don't get any of the git history data when building there.

Still an interesting idea tho, I also think later this could be done by a custom integration/plugin through component customization.

delucis:
> > UI-wise, I am not sure if it's just me or not but the label does not feel vertically centered but I'll have to take a closer look when I have more time.
> 
> I agree. I think the sidebar has the same sort of issue and this is the same CSS as that. I think it's aligned to the text baseline here, but I wonder if it would make sense to centre it overall vertically instead via `flex` or similar?

Yeah, this is due to us relying on inline styling in the sidebar. Can be fixed with a `flex` layout if that works for this case. In the case of the sidebar we couldn’t use flex because of our layout requirements and how we’d want lines to wrap. I actually spent a good afternoon researching this, but there’s no silver bullet for the sidebar case because we don’t know what font is used and font metrics need to be known with current CSS options (see e.g. https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align). There are some hopes that `vertical-align: central` will improve this when available, but for now, `middle` means align to the center of the `x` height, not cap height resulting in these off-center appearance when inline flow is required.

lorenzolewis:
How does this thing still not have any conflicts? 😅 

How are we feeling now that Badges are merged in? Do we want to update this to use them or toss this out? I would personally still like to have it for the sake of having feature parity with sidebar entries (this was really some pre-work to build a "Table of Contents" component for a section of a sidebar).

sergeysova:
20 minutes ago I removed badge from my content collections, just because LinkCard doesn't support it. Good news, PR exists! Bad news, last update 6 mo ago.

Issue #398, opened 2023-07-25
Config to enable/disable theming 


lorenzolewis:
### What version of `starlight` are you using?

0.5.6

### What is your idea?

A way to set the Starlight site to only be dark or light (or auto) via the config

### Why is this feature necessary?

Sometimes branding only has light or dark mode colors (or a developer only wants a specific theme). This would expose an option in the Starlight config for a user to only create a dark or light website. This could disable the theme dropdown in the navbar as well as reduce the resulting CSS.

### Do you have examples of this feature in other projects?

Docusaurus theme config for color: https://docusaurus.io/docs/api/themes/configuration#color-mode---dark-mode

I like how they also let the developer set a "preferred" mode (`defaultMode`) so that the dev could specify a specific theme but then the user could also override it if they'd like. I remember some conversations over there around all the different permutations and states this could result in so it would probably be good to have a solid think through what the API and UX would be around this.

### Participation

- [X] I am willing to submit a pull request for this issue.

delucis:
I’m pretty strongly in favour of letting site users pick dark/light mode — forcing someone to use colours different from their preference is an accessibility concern.

If people want it enough, I think we could make the default theme configurable though — i.e. `disrespectUserThemePreference: 'light'` 😛 — so that pages load in light/dark instead of auto while still letting users switch to their preference. There are situations where this could make sense, like https://terrateam.io/ where the transition from their dark marketing homepage to the docs in light mode (when that’s the device preference) can be quite jarring.

lorenzolewis:
Love that suggested key name ;) 

Get what you're saying around accessibility (and even my personal experience with being blinded by the light[-theme]). 

Maybe a config option could be: `preferredTheme: 'userPreference' | 'dark' | 'light';`

delucis:
Yeah! I might suggest `defaultTheme` or `initialTheme` perhaps? And the values would probably be clearest if they match how we label them in the UI `'auto' | 'dark' | 'light'`. We could make it clear in docs that setting this config option _overrides_ user preference.

dhruvkb:
Personally, I would prefer sites have 'auto' theme be the default, as it allows me to set my light/dark preference once (at the device level) and have my web browsing experience consistently match my preference.

delucis:
> Personally, I would prefer sites have 'auto' theme be the default

Yeah, absolutely, we won’t change that default.

PR #713, opened 2023-09-12
Provide theme select configuration

tomas-padrieza:

- Changes to Starlight code

#### Description

- This PR introduces new configuration settings to control the display of theme selection dropdown and default theme, that can be preconfigured by the developer.
- Also provided documentation changes for the new options.

Addresses #398


delucis:
Thanks for the PR @tomas-padrieza!

I’m not sure we want to offer an API to disable user preferences for theme selection. As discussed in #398, the proposal was to maybe allow a setting that supported setting a _default_, but not also a way to disable this entirely.

#709 should soon bring component customisation support, so I think for people who really need this, my recommended path would be to override the built-in `<ThemeSelect>` component via that API rather than to support this with a configuration option.

Happy to hear arguments for why this must be something set at the configuration level, but sometimes we do have to make calls about what is high priority enough to get a built-in option and what is left to users to implement. In this case, because disabling theme selection reduces user control, potentially impacting accessibility, I feel like it’s something that we should leave to individual people to override rather than making it an “official” option.

tomas-padrieza:
https://github.com/withastro/starlight/pull/709 looks promising, i'm closing this PR, definitely a better approach to customize the overall layout.

PR #715, opened 2023-09-14
Prevent scroll on `body` when search is open

itsmatteomanf:

- Changes to Starlight code

#### Description

This PR resolves the scroll on `body` when the search is open.

From my tests, it works fine even on mobile, with the menu open.

### DIFF
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -87,6 +87,12 @@ const pagefindTranslations = {
 			openBtn.addEventListener('click', openModal);
 			openBtn.disabled = false;
 			closeBtn.addEventListener('click', closeModal);
+			window.addEventListener('keypress', (e) => {
+				if (e.key === 'Escape' && dialog.open) {
+					e.preventDefault();
+					closeModal();
+				}
+			});
 
 			// Listen for `/` and `cmd + k` keyboard shortcuts.
 			window.addEventListener('keydown', (e) => {
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -87,7 +87,7 @@ const pagefindTranslations = {
 			openBtn.addEventListener('click', openModal);
 			openBtn.disabled = false;
 			closeBtn.addEventListener('click', closeModal);
-			window.addEventListener('keypress', (e) => {
+			window.addEventListener('keydown', (e) => {
 				if (e.key === 'Escape' && dialog.open) {
 					e.preventDefault();
 					closeModal();
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -72,6 +72,7 @@ const pagefindTranslations = {
 
 			const openModal = (event?: MouseEvent) => {
 				dialog.showModal();
+				document.body.toggleAttribute('data-search-modal-open', true);
 				this.querySelector('input')?.focus();
 				event?.stopPropagation();
 				window.addEventListener('click', onClick);
@@ -79,6 +80,7 @@ const pagefindTranslations = {
 
 			const closeModal = () => {
 				dialog.close();
+				document.body.toggleAttribute('data-search-modal-open', false);
 				window.removeEventListener('click', onClick);
 			};
 
@@ -230,6 +232,10 @@ const pagefindTranslations = {
 </style>
 
 <style is:global>
+	[data-search-modal-open] {
+		overflow: hidden;
+	}
+
 	#starlight__search {
 		--sl-search-result-spacing: calc(1.25rem * var(--pagefind-ui-scale));
 		--sl-search-result-pad-inline-start: calc(3.75rem * var(--pagefind-ui-scale));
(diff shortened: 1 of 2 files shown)


packages/starlight/components/Search.astro
```diff
@@ -72,13 +72,15 @@ const pagefindTranslations = {
 
 			const openModal = (event?: MouseEvent) => {
 				dialog.showModal();
+				document.body.toggleAttribute('data-search-modal-open', true);
```
itsmatteomanf, line 75: I added the force option here, just to *never* get in a situation where it's the wrong way around for any weird reason.

This could work, too.

```suggestion
				document.body.toggleAttribute('data-search-modal-open');
```

kevinzunigacuellar:
Hey @itsmatteomanf thanks a lot for the PR. I was testing the deploy preview and It seems that closing the modal using the "esc" key doesn't remove the attribute `data-search-modal-open`.

dreyfus92:
> Hey @itsmatteomanf thanks a lot for the PR. I was testing the deploy preview and It seems that closing the modal using the "esc" key doesn't remove the attribute `data-search-modal-open`.

I think that it would be worth adding a `KeyboardEvent` as well to the arguments of both functions. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

itsmatteomanf:
Thanks for the review, @kevinzunigacuellar, @dreyfus92.

> I think that it would be worth adding a `KeyboardEvent` as well to the arguments of both functions.

I didn't want to change the actual behaviour of the search modal, but I see now that it actually does need to be intercepted as it actually sticks as locked. Gonna do the change immediately.

dreyfus92 (approved):
LGTM ✅

HiDeoo (commented):
I tested the Escape key workaround in Chrome and Firefox and it does not seem to work for me. Did not really investigate why but maybe it could be because [`keypress`](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) is now deprecated and no longer recommended?

I guess this could be handled in the existing `keydown` event listener just below the `keypress` one.

One alternative solution could be the the [`HTMLDialogElement` `close` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/close_event):

> The close event is fired on an HTMLDialogElement object when the <dialog> it represents has been closed.

We could handle everything in a code like this I assume:

```ts
dialog.addEventListener('close', () => {
  document.body.toggleAttribute('data-search-modal-open', false);
});
```

This would be perfect in this case but maybe the [browser support](https://caniuse.com/mdn-api_htmldialogelement_close_event) is not good enough yet?

itsmatteomanf:
> I guess this could be handled in the existing `keydown` event listener just below the `keypress` one.

You could actually just replace it, it does the same thing, as far as I know.
Replacing it, seems to work fine.

> One alternative solution could be the the [`HTMLDialogElement` `close` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/close_event):

Not sure if we can classify it as supported enough, this allows you to support basically any browser. Can be added, too. But not sure it's worth doing that twice.

kevinzunigacuellar:
> This would be perfect in this case but maybe the [browser support](https://caniuse.com/mdn-api_htmldialogelement_close_event) is not good enough yet?

I am not sure if its the same case but we already use the dialog element and its [browser support](https://caniuse.com/dialog) is similar to the close event. Hence I think it should be safe to use.

HiDeoo:
> I am not sure if its the same case but we already use the dialog element and its [browser support](https://caniuse.com/dialog) is similar to the close event. Hence I think it should be safe to use.

Indeed, great catch, I imagined `<dialog/>` being usable before Safari 15.4 but I guess I dreamed about that ^^

In this case, I guess it would make sense to use the `HTMLDialogElement` `close` event for all `document.body.toggleAttribute('data-search-modal-open', false);`.

packages/starlight/components/Search.astro
```diff
@@ -86,6 +87,10 @@ const pagefindTranslations = {
 			openBtn.disabled = false;
 			closeBtn.addEventListener('click', closeModal);
 
+			dialog.addEventListener('close', () => {
+				document.body.toggleAttribute('data-search-modal-open', false);
```
delucis, line 91: What do you think about moving all the teardown code in here?

```suggestion
				document.body.toggleAttribute('data-search-modal-open', false);
				window.removeEventListener('click', onClick);
```

And then `closeModal` gets simplified down to:

```js
			const closeModal = () => dialog.close();
```

Feels right to me to keep that code together to help maintain it all running together in the future.

delucis (commented):
This looks great to me @itsmatteomanf! Thanks for tackling and thank you to everyone reviewing along the way to polish this. Arrived at a nice solution I think.

I left one suggestion for a tiny refactor here, but overall this looks great 🙌

packages/starlight/components/Search.astro
itsmatteomanf: Makes sense, did both in one commit.

delucis (approved):
Thanks again @itsmatteomanf! LGTM 🚀

PR #726, opened 2023-09-18
Fix ToC bug & add tests

delucis:

- Changes to Starlight code

#### Description

- Adds tests for our `generateToC()` utility
- Rewrites the `generateToC()` utility to fix bugs uncovered by testing. The current implementation doesn’t handle situations where the heading depth jumps, e.g. h2 => h4 on a page. (With Starlight’s defaults, you don’t hit this bug, because we only consider h2 & h3, so I’m guessing not many people have run into this.) While headings should ideally be sequential for a11y reasons, I don’t think bugs in the ToC code are the right punishment for people ignoring that :-D
- Has the benefit of simplifying `generateToC()` quite a bit from the code I copy/pasted from Astro’s docs site.


HiDeoo (approved):
Clean refactor :thumbsup: The new behavior looks good to me.

delucis:
Thank you for the review @HiDeoo!

PR #735, opened 2023-09-20
Small search modal improvements

delucis:

- Changes to Starlight code

#### Description

- Use Starlight’s font variable inside the Pagefind modal for a consistent experience (will also respect user-configured `--sl-font` if set).
- Fixes the new styles used for subheadings to flip the iconography in RTL pages:
   Search result UI showing some styling that highlights the relationship between a page title and page subheadings. This UI is now flipped to read from right to left as expected.

### DIFF
--- a/.changeset/great-tomatoes-tie.md
+++ b/.changeset/great-tomatoes-tie.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Use Starlight font custom property in Pagefind modal
--- a/.changeset/lemon-cameras-tell.md
+++ b/.changeset/lemon-cameras-tell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fix RTL styling in Pagefind modal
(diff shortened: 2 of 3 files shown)


HiDeoo (commented):
These changes looks good to me. One small extra thing I noticed while testing it out, on `rtl` and on mobile, the magnifying glass icon overlaps the Cancel button.

SCR-20230921-jozz

Maybe this could be fixed in the same PR? Considering the current rule is `left: calc(20px * var(--pagefind-ui-scale));`, I think adding something like this could do the trick:

```css
[dir='rtl'] #starlight__search .pagefind-ui__form::before {
  left: calc(20px * var(--pagefind-ui-scale) + var(--sl-search-cancel-space));
}
```

delucis:
Ah yeah, I forgot about that. I have an item on my to-do list to actually implement better RTL support upstream in Pagefind that I keep not getting to. This feels a bit like that’s where it should be fixed, but perhaps we should monkey patch it here for people for now 👍

HiDeoo:
> This feels a bit like that’s where it should be fixed

Definitely :100:

> perhaps we should monkey patch it here for people for now

Not sure, I think this would have been my first choice but at the same time it has not been reported so far and if the plan is to fix it in Pagefind anyway, it might be better to wait for it?

Issue #734, opened 2023-09-20
LinkedCard doesn't support icon


bugzpodder:
### What version of `starlight` are you using?

latest

### What is your idea?

It seems Card supports icon prop but LinkCard doesn't?

### Why is this feature necessary?

I think it's common to have a LinkedCard with an icon?

### Do you have examples of this feature in other projects?

https://bugzpodder.github.io/pysandbox/ (I have a LinkedCard but no icon)

### Participation

- [X] I am willing to submit a pull request for this issue.

jsparkdev:
The LinkCard component doesn't seem to support icon insertion.

How about using the <a> tag within the Card component instead of LinkCard?

lschierer:
You can do that, but the link target ends up only being the elements inside the card, whereas with LinkCard, the entire card acts like a button.  The user experience is very different.

PR #736, opened 2023-09-21
feat: add icon support for LinkCard

jsparkdev:

- Changes to Starlight code

#### Description

- PR for resolving issue #734 
- You can now add an icon before the title of the LinkCard component.
- The previous LinkCard component could not receive icon props, so it did not render the icon even if icon was passed as a prop. However, in the new code, you can render an icon in front of the title by receiving icon props like a Card component.

linkcard


jsparkdev:
This is bad PR so you can ignore it.

PR #740, opened 2023-09-22
Remove CommandKit from showcase

notunderctrl:

#### Description

- What does this PR change? Remove the CommandKit docs site as it no longer uses Astro Starlight.


notunderctrl:
First of all, I don't hate Starlight. It's one of the best tools I've ever used to setup docs, so good job there!

The reason I've moved to Nextra was really just mostly about the docs look I was going for + SPA feel I wanted. I did use view transitions but unfortunately that's not fully adopted by browsers right now.

With that said I'll definitely use Starlight in the future considering how easy it was to work with it.

Thanks :)

PR #742, opened 2023-09-22
Add Expressive Code to Starlight

### LINKED ISSUE #461: Copy button for Code Blocks

0.6.1

### What is your idea?

Add a `copy` button to code blocks so people can easily copy-paste example code.

Already a thread in Discord on this but figured it would be nice to turn it into an issue (for selfish reasons so I can track it downstream 😅): https://discord.com/channels/830184174198718474/1070481941863878697/1120347723128909914

Make it easier for developers to follow code examples

The amazing Astro Docs: https://docs.astro.build/en/getting-started/#start-your-first-project

### Participation

hippotastic:

- Changes to Starlight code

#### Description

- Integrates `astro-expressive-code` into Starlight
- Adds an `expressiveCode` config option & schema that allows configuring the integration.
- Adds `expressiveCodeI18nSchema` including all translatable strings, similar to the Pagefind schema.
- Adds information about the configuration options and i18n to the docs.
- Closes #461


HiDeoo:
This is looking super great :tada: Amazing work!

I'll play more with it later and make a more thorough review but I have a few small questions:

- Is the font change expected, shouldn't it respect `--sl-font-system-mono`? Same question for the font size.
- I see you are using `color-mix()` which [seems](https://caniuse.com/?search=color-mix) to have landed fairly recently on some browsers, e.g. March 2023 on Chrome, May 2023 on Firefox, and 10 days ago on Chrome for Android if I read that correctly. Could it be a concern? Wouldn't it be a bit too early to use it?
- "file names/titles" in code block are not selectable, is that expected? I have to say, even more for new files, it's super handy to be able to quickly select it to copy it. At least that would be my expectation.

Again, great work! :clap:

hippotastic:
Thank you for the valuable feedback, @HiDeoo!

- Great catch regarding the font family & font size! I have now referenced the CSS variables `var(--__sl-font-mono)` and `var(--sl-text-code)` that Starlight also uses for inline code.
- The file names / titles were not selectable due to a pseudo-element blocking the inputs. I have also fixed this. Thank you for testing this!
- I intentionally only used `color-mix()` for a single border style that looked too prominent otherwise. If the browser does not support this function, the border will simply become transparent, which does not impact the overall look and usability in my opininon.

HiDeoo:
Wow, thanks for the quick and efficient updates and answers! Really appreciate it.

kevinzunigacuellar:
One small nits, 

- code blocks were slightly updated from sharp to rounded corners. 
- the syntax highlight seems to be different from the previous default

| Before                                                                                                                          | After                                                                                                                           |
|---------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
| image | image |

hippotastic:
Both correct, good observations!

The slightly rounded corners could be turned off by providing a styleOverrides setting of `{ borderRadius: '0' }`. However, I honestly like the rounded ones better - especially in combination with the tabs. Starlight also has rounded corners in other places like the next/previous page navigation buttons at the end of pages, so I hoped that this is close enough. 😄 If it's a hard requirement to keep the previous style, we could of course change the defaults.

The different syntax highlighting colors are a result of different default themes. Starlight only supported a single theme before and used custom CSS variables to make this one theme work in both dark and light mode, while Expressive Code supports freely definable separate dark and light themes and uses the GitHub dark and light themes as the default. If you have another VS code / Shiki theme in mind that you'd like to use instead of `github-dark`, please let me know. We could even develop our own themes, which might be interesting from a branding perspective! I just picked these ones for now because they are already bundled with Shiki and seemed to be a popular choice.

HiDeoo:
I have a bit more high-level question (and I am also curious ^^).

With these changes, we would be shifting away from the `css-variables` theme to the good old shiki approach of rendering & shipping every code blocks twice. According to [this comment](https://github.com/shikijs/shiki/issues/33#issuecomment-905953016) from Fred, there was a bit of a pushback from the Astro community regarding that last approach, which is what actually led to the `css-variables` theme. I am wondering if with Starlight's [emphasis on page weight](https://starlight.astro.build/environmental-impact/#page-weight), this could also be a concern?

Was any thought or experimentation done regarding using [`shikiji`](https://github.com/antfu/shikiji) (Anthony Fu ESM-focused rewrite of `shiki`) which replaces the `css-variables` theme by a new [interesting approach](https://github.com/antfu/shikiji#lightdark-dual-themes) to support light/dark themes? Or maybe the idea would be to wait until this is ported back to `shiki` itself, if that ever happens, as this [seems](https://github.com/antfu/shikiji#whats-next) to be the plan?

hippotastic:
I'm honestly not a big fan of the CSS variables theme as it's making it impossible to ensure an accessible color contrast ratio when adding annotations to code. Expressive Code puts an emphasis on accessibility by always checking the color contrast and tweaking the syntax highlighting colors when necessary. With CSS variables, it cannot know what the actual syntax token color is going to be, so we might end up with bad accessibility. Due to this, I even intentionally disabled the CSS variables theme in Expressive Code.

hippotastic:
Just had a look at Shikiji, and an approach like this definitely looks more promising than the CSS variables theme to me!

However, themes in Expressive Code are more than just syntax highlighting. The themes provide a full spectrum of UI colors as well, and these will be used by plugins like frames to render editor tabs, terminal windows etc., all of which Shikiji doesn't address because it seems to be focused on syntax highlighting only.

I think we can (and I will) explore approaches for more efficient multi-theming for sure, but this would take a considerable amount of time and should not be done in the scope of this PR in my opinion.

HiDeoo:
I totally agree with you regarding the `css-variables` theme approach to use CSS variables to read the colors for the tokens.

The `shikiji` new approach to use CSS variables to actually store the colors on each token is a very interesting one indeed and I was curious to get your opinion on it (hence the "more high-level question" comment). Thanks for taking the time to detail how theming works in Expressive Code as this is not something I'm in-depth familiar with and also your plans to explore more ideas. 🙌 Very interesting!

PS: Definitely outside the scope of this PR to experiment with this, I agree.

hippotastic:
Sure thing! I love both Orta's and Anthony's work on this and think that a Shikiji-like approach might work even for UI theming in Expressive Code. Finding a way to get efficient multi-theme output is on my roadmap as well. Thank you for all the valuable input and context!

hippotastic:
Ok, so I couldn't get this out of my head and experimented with CSS variable-based approaches to multi-theming inspired by Shikiji - and it seems to work!

One of the next Expressive Code releases will always generate a single output per code block, no matter how many themes you define. The theme-based parts of the UI (everything you can see in `styleOverrides`) will all be migrated to CSS variables, too.

This migration will take me a while though, so I think it's your decision whether the current version is good enough for Starlight and we'll integrate it now and just upgrade the dependency later when the new rendering is available, or if you'd rather want to wait for the more efficient multi-theme rendering. I know how much we are focused on performance and minimal footprint (and I care a lot about that myself), but I also know how many devs are already waiting for Expressive Code features like text markers or the copy button. 😄 Probably not easy to decide - I'm fine with both options!

HiDeoo:
> Ok, so I couldn't get this out of my head and experimented with CSS variable-based approaches to multi-theming inspired by Shikiji - and it seems to work!

Oh boy, this is all my fault. I'm so sorry. Altho, I find it excellent news to know that the discussion could lead to some improvements in that area and that their approach could potentially work in a way more complex setup like the one from Expressive Code. :tada:

Personally, and it's only my own opinion of course, I think it would be fine to pursue with the proposed changes in this PR and later, when it's ready, upgrade the dependency. This would mean people can benefit from the amazing Expressive Code features as soon as possible. When I mentioned the double rendering, I did that asking in a question if it could be a concern because I have no idea personally so maybe some other people will have a better idea about it.

hippotastic:
No worries at all! It's not your fault if you provide great input to successfully nerd-snipe me! 😄

HiDeoo:
A few more high-level comments:

- I think the [code blocks section](https://starlight.astro.build/guides/authoring-content/#code-blocks) in the Authoring Content in Markdown guide could be updated to describe some new syntaxes added by Expressive Code? It may also be worth linking to the Expressive Code docs for Markdown/MDX usage for the [frames](https://github.com/expressive-code/expressive-code/blob/main/packages/@expressive-code/plugin-frames/README.md#usage-in-markdown--mdx-documents) and [text markers](https://github.com/expressive-code/expressive-code/blob/main/packages/@expressive-code/plugin-text-markers/README.md#usage-in-markdown--mdx-documents) plugins?
- There are many places in the documentation where we could use some Expressive Code text markers, but I guess it may be easier to add them in a follow-up PR. I'll try to prepare a list of places where I think we could use them.

lorenzolewis:
A bit late to the party so feel free to disregard, but when we were doing some work on other UI components the Starlight design system has a pattern of interactive components having rounded corners and non-interactive having square corners. Here's a link to the Figma file in case you wanted to take a peek: https://www.figma.com/file/GrDkcguAR7tSWPFuLVDixq/Theme---Starlight-Docs?type=design&node-id=1%3A2&mode=design&t=ypLDqrL9TDbNrovJ-1

Regardless, I'm super excited to see this and I know myself (as well as some of our readers over at Tauri) are excited for it! 🥳

packages/starlight/integrations/expressive-code.ts
```diff
@@ -0,0 +1,265 @@
...
+					theme = customizeTheme(theme) ?? theme;
+				}
+				if (minSyntaxHighlightingColorContrast > 0) {
+					ensureMinTokenColorContrast(theme, minSyntaxHighlightingColorContrast);
+				}
+				return theme;
+			},
+			useThemedSelectionColors,
+			styleOverrides: {
+				codeFontFamily: 'var(--__sl-font-mono)',
```
delucis, line 162: I believe APIs might change slightly but documenting here some tweaks I made to align the design closer to our Figma mock-ups:

```suggestion
				borderRadius: '0',
				borderWidth: '1px',
				codePaddingBlock: '0.75rem',
				codePaddingInline: '1rem',
				codeFontFamily: 'var(--__sl-font-mono)',
```

delucis (commented):
Super excited for this. Making some notes although I know @hippotastic is working on some upstream changes that will need to be integrated here before we merge this, so appreciate everyone’s patience while waiting for this amazing PR!

packages/starlight/integrations/expressive-code.ts
trueberryless: I'm not sure if I speak for everybody - I now opinions are different -, but personally I liked the rounded corners a lot more actually... I used them on my website already and the look fantastic, especially sh and bash with the terminal look. But I guess if I want rounded corners, I have to edit the settings locally...

delucis: Yes, this change would just align the default design more with all our other defaults but you could still set options to configure this to your taste and customise how things look.

packages/starlight/integrations/expressive-code.ts
```diff
@@ -0,0 +1,265 @@
...
+				codeFontSize: 'var(--sl-text-code)',
+				codeLineHeight: 'var(--sl-line-height)',
+				uiFontFamily: 'var(--__sl-font)',
+				textMarkers: {
+					defaultChroma: '45',
+					backgroundOpacity: '60%',
+					...textMarkersStyleOverrides,
+				},
```
delucis, line 170: One more comment here to note that we want to (with the future API coming soon™️) make the default highlight better match the user-configured colour scheme. Currently the default highlight blue and Starlight UI blue don’t quite match:

UI elements on a Starlight page alongside the code highlight showing different shades of blue

And this could be even more noticeable with a custom colour scheme. While the green/red highlights for insertions and deletions have a semantic purpose, I’m proposing using a highlight colour that is simply a lighter version of the background colour so that it remains consistent no matter the user-configured colour scheme. Here’s a small example thrown together (exact brightness TBD — should maintain accessible contrast with the text):

| With default colour scheme | With custom colour scheme |
|---|---|
| image | image |

Obviously this would just be the default we set, and users who prefer a specific highlight colour could choose to configure one.

The other option would have been to use the user’s theme accent colour for these highlight’s but I think that risks clashing with the insertion and deletion styles if someone has a red or green theme accent, so this feels like a good neutral default where we only rely on a brightness contrast and not also a hue difference.

packages/starlight/integrations/expressive-code.ts
```diff
@@ -0,0 +1,265 @@
...
+	lang: string,
+	target: T,
+	ids: (keyof (typeof translations)['en'])[]
+) {
+	const builtinTexts = translations[lang];
+	let userTexts: Partial<(typeof translations)['en']> = {};
+	try {
+		// As Astro has not been initialized yet, we cannot use content collections directly
+		// and have to manually load any user-provided translations from the file system
+		userTexts = JSON.parse(fs.readFileSync(`src/content/i18n/${lang}.json`, 'utf8'));
```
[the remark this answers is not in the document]
delucis: Refactored the translation system in #1003, which is now merged, so I added a commit to this PR which addresses this comment — EC now uses the same translations system we already use internally in Starlight.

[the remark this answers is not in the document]
hippotastic: That's awesome, thank you so much! My architectural changes to the new efficient multi-theme rendering based on CSS variables are nearing completion as well. I just got all my existing tests green today and now only need to add a few new tests and perform finalization & cleanup work. Then the biggest change to EC since its first release on NPM is done and I can finally get this PR ready to merge! Looking really forward to that. Thanks everyone for your patience!

packages/starlight/integrations/expressive-code.ts
hippotastic: I like the idea of just changing the lightness of the background color. I experimented with this a bit and found that setting a semi-transparent white or black color as the `mark` background and border seems to already do the trick. You can see the result in the deploy preview, e.g. [here](https://deploy-preview-742--astro-starlight.netlify.app/manual-setup/#configure-the-integration). I like it!

packages/starlight/integrations/expressive-code.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	theme.styleOverrides.frames = {
+		// Use the same color for editor background, terminal background and active tab background
+		editorBackground: editorBackgroundColor,
+		terminalBackground: editorBackgroundColor,
+		editorActiveTabBackground: editorBackgroundColor,
+		terminalTitlebarDotsForeground: borderColor,
+		terminalTitlebarDotsOpacity: '0.75',
+		inlineButtonForeground: 'var(--sl-color-text)',
+		frameBoxShadowCssValue: 'var(--sl-shadow-sm)',
```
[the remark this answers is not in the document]
hippotastic: Done.

packages/starlight/integrations/expressive-code.ts
hippotastic: I added the proposed styles. However, I'd like to note that your reduced inline padding does not leave much space for the `+`/`-` diff indicators anymore. I think it's still okay in this state, but any less and it wouldn't fit anymore. :)

hippotastic:
The migration to the new efficient multi-theme rendering is complete now. :) I also implemented all review suggestions. Please have a look if there is anything else you'd like to change!

Genteure:
Thank you for the amazing work!

Would it make sense to mention expressive-code in the "Authoring Content in Markdown" page and maybe show a simple example?

hippotastic:
> Would it make sense to mention expressive-code in the "Authoring Content in Markdown" page and maybe show a simple example?

That's a great idea! I'll have a look and add some examples that I already wrote for the Expressive Code docs. :)

packages/starlight/package.json
```diff
@@ -157,6 +157,7 @@
     "./props": "./props.ts",
     "./schema": "./schema.ts",
     "./types": "./types.ts",
+    "./expressive-code": "./integrations/expressive-code.ts",
```
HiDeoo, line 160: Do we really need to export this? What would be the use case?

If this is on purpose and I am missing something, I guess it should be documented as it exposes the entirety of `astro-expressive-code` and a few other functions?

packages/starlight/schemas/expressiveCode.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'astro/zod';
+import type { StarlightExpressiveCodeOptions } from '../integrations/expressive-code';
+
+export const ExpressiveCodeSchema = () =>
+	z
+		.union([
+			z.custom<StarlightExpressiveCodeOptions>((value) => typeof value === 'object' && value),
```
HiDeoo, line 7: Is the configuration passed down to `astro-expressive-code` validated using Zod? 

If yes, could the schema be exported from it so we can build upon it here instead of using `z.custom` and then infer `StarlightExpressiveCodeOptions` from it in order to provide a consistent experience regarding configuration validation?

HiDeoo (commented):
Such an amazing update! :clap: The new rendering method is so great :star_struck: 

Left a few questions but this is looking so good!

packages/starlight/schemas/expressiveCode.ts
hippotastic: No, I couldn't justify the (unfortunately) massive footprint of Zod, so I'm not using it in Expressive Code although I would have loved to.

packages/starlight/package.json
hippotastic: Well spotted! The use case is to provide users with all the classes and functions required for customization, without having to install `astro-expressive-code` as an additional dependency (and thereby risiking version conflicts). What users can do with the exports can be as simple as instantiating their own themes (`ExpressiveCodeTheme`), but can also become more involved to e.g. create own renderers (`createRenderer`) to be used for the `customCreateRenderer` option, or access and modify the texts of plugin-frames (`pluginFramesTexts`).

packages/starlight/schemas/expressiveCode.ts
hippotastic: My problem here is that Expressive Code is not just for Astro, all the layers below `astro-expressive-code` (where the majority of the config actually comes from) have no Astro dependency at all.

hippotastic:
I added some examples to the [Authoring Content in Markdown](https://deploy-preview-742--astro-starlight.netlify.app/guides/authoring-content/#text-markers) page to highlight the features provided by Expressive Code.

Thank you @Genteure and @HiDeoo for this suggestion!

packages/starlight/package.json
hippotastic: Regarding your idea of documenting the export, do you have any specific location in mind where you would most likely search for something like this? I just had a look around the codebase and couldn't find any documentation on the existing exports, so we might need to add an entirely new section somewhere if we really consider this to be worth documenting. What do you think?

HiDeoo: I personally would say maybe adding an Expressive Code section in the [customization guide](https://starlight.astro.build/guides/customization/), kinda like edit links which have an [entry](https://starlight.astro.build/guides/customization/#edit-links) in this guide and in the configuration reference [too](https://starlight.astro.build/reference/configuration/#editlink). I also guess we would not need to detail everything as we can link to the Expressive Code documentation from there and only give an overview of what it is possible to do with it.

But maybe it's worth gathering more opinions on this before pulling the trigger.

hippotastic:
I'm not sure if it was visible enough from my previous commit message: EC is now no longer inlining its CSS, but creating an immutable stylesheet with a hashed file name instead that can be cached indefinitely by browsers. The HTML size difference should be negligible now.

delucis:
Thank you for that performance improvement @hippotastic!

Here’s my table again to show the impact. Since the last comparison, we migrated to Vercel, so these are now Brotli compressed HTML responses instead of GZip:

| route                       | Prod   | EC PR  | change        |
| --------------------------- | ------ | ------ | ------------- |
| `/getting-started/`         | 12.97k | 13.27k | +0.30k  (+2%) |
| `/manual-setup/`            | 11.32k | 11.79k | +0.47k  (+4%) |
| `/guides/css-and-tailwind/` | 13.36k | 14.24k | +0.88k  (+7%) |
| `/reference/configuration/` | 14.17k | 16.10k | +1.93k (+14%) |

(The `/reference/configuration/` one seems bigger, but bear in mind that this PR also adds a new section the reference page.)

hippotastic:
Nice! And we should also keep in mind that this size already includes two independently configurable themes with full syntax support, whereas the current Starlight code blocks only have a single limited CSS variables theme.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -65,6 +65,33 @@ type LogoConfig = { alt?: string; replacesTitle?: boolean } & (
...
+  - any theme object compatible with VS Code or Shiki (e.g. imported from an NPM theme package)
+  - any ExpressiveCodeTheme instance (e.g. using `ExpressiveCodeTheme.fromJSONString(...)` to load a custom JSON/JSONC theme file yourself)
+
+  Defaults to the dark and light variants of the "Night Owl" theme by Sarah Drasner, which are also exported from `@astrojs/starlight/expressive-code` as `getStarlightDefaultThemes`.
+
+  **Note**: If you pass at least one dark and one light theme to this option, Starlight will automatically ensure that the displayed theme matches the current dark mode switch state. See the `useStarlightDarkModeSwitch` option to configure this behavior. If you turn it off, you will need to add your own CSS code to the site to allow manual switching between themes.
+
+- `useStarlightDarkModeSwitch` — determines if CSS code should be added that automatically displays dark or light code blocks depending on Starlight's dark mode switch state.
+
+  Defaults to `true` if the `themes` option is not set or contains at least one dark and one light theme, and `false` otherwise.
```
delucis, line 87: Wondering what the use case for this option is? I think it’s potentially confusing because `false` does not say “I want always dark” or “I want always light”, so I guess it only really works well when you also set `themes` to a single theme?

My instinct would be to suggest removing this in favour of respecting `themes`: single theme persists across dark and light modes, two themes toggles with the dark mode switch. Unless we have a clear use case this solves for?

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -65,6 +65,33 @@ type LogoConfig = { alt?: string; replacesTitle?: boolean } & (
...
+**type:** `StarlightExpressiveCodeOptions | boolean`
+
+Starlight uses [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code) to render code blocks on your site. The integration is enabled by default and can be disabled by passing `false` to this option.
+
+You can customize your code blocks by passing an object with any of the following properties:
+
+- `themes` — an array of themes that should be used to render code blocks. The following item types are supported in this array:
+  - any theme name bundled with Shiki (e.g. `dracula`)
+  - any theme object compatible with VS Code or Shiki (e.g. imported from an NPM theme package)
+  - any ExpressiveCodeTheme instance (e.g. using `ExpressiveCodeTheme.fromJSONString(...)` to load a custom JSON/JSONC theme file yourself)
```
delucis, line 79: Are you planning some kind of special Expressive Code support for more than 2 themes? I imagine the scenarios where this makes sense require some custom UI for picking the user-preferred colour theme?

Would there be an argument for making the Starlight option a little simpler and just have two `lightTheme` and `darkTheme` options? Or would the current array also automatically support e.g. light/dark/light high contrast/dark high contrast?

I’m also curious what actually happens if you provide more than two themes? Does only e.g. the first dark theme get used?

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -65,6 +65,33 @@ type LogoConfig = { alt?: string; replacesTitle?: boolean } & (
...
+
+  Defaults to the dark and light variants of the "Night Owl" theme by Sarah Drasner, which are also exported from `@astrojs/starlight/expressive-code` as `getStarlightDefaultThemes`.
```
delucis, line 81: Instead of exporting this method (I’m assuming for cases where people want to use one of the two default themes?), what do you think of giving these string values people can use?

e.g. `themes: ['starlight-light', 'starlight-dark']` is the default and we resolve those two strings under the hood for people? Seems nicer than exposing the internal function which returns an array it is hard to pick one theme out of.

I think this is also the only place exposing `@astrojs/starlight/expressive-code` is necessary? In which case I think it would be a benefit of this change to avoid exposing that in package `exports`.

packages/starlight/schemas/i18n.ts
```diff
@@ -162,3 +166,21 @@ function pagefindI18nSchema() {
...
+				.string()
+				.describe('Expressive Code UI translation. English default value: `"Copied!"`'),
+
+			'expressiveCode.copyButtonTooltip': z
+				.string()
+				.describe('Expressive Code UI translation. English default value: `"Copy to clipboard"`'),
+
+			'expressiveCode.terminalWindowFallbackTitle': z
+				.string()
+				.describe('Expressive Code UI translation. English default value: `"Terminal window"`'),
```
delucis, line 183: Unless I’m mistaken Expressive Code only provides these in English and German currently. Should Starlight look to add these in their built-in translation `.json` files given we’re likely supporting more languages?

delucis (commented):
Thanks again for this amazing PR @hippotastic. Here I am finally with a code review!

I think the internal code looks great for the most part. I opened https://github.com/hippotastic/starlight/pull/1 to refactor the logic for getting a file’s locale to reuse some logic we already built for #517.

I’ve left some big picture questions around the configuration options we give to Starlight users. I think we can improve slightly here to give a simpler set of more intuitive options, but there may be use cases I’m not aware of, so would love to discuss and hear more about the current choices.

packages/starlight/schemas/i18n.ts
hippotastic: I was just following the example set by the pagefind schema here. The idea was to have these strings translatable through the Starlight `.json` translation files, and my understanding was that this is achieved by the code in this PR. I also tested this locally, however this was before your changes to the i18n logic. :)

hippotastic: I just tested it again and it still works. I added an entry `"expressiveCode.copyButtonCopied": "That worked!"` to `packages\starlight\translations\en.json` and it gets displayed. I also tested it with the Spanish version and it's also possible to customize the string there. Seems to work as intended!

delucis: Ah sorry, I think my comment wasn’t clear — I also tested this and it was working great!

My question was more about built-in support for more languages. From what I could see in the EC repo, only German and English strings were provided there, so I was asking whether we should be asking translators to add translations to Starlight’s built-in JSON files or upstream to EC — all other parts of the UI are available in 20+ languages, so would be good to work towards (once this PR is merged) also having EC strings covered in those languages.

docs/src/content/docs/reference/configuration.mdx
hippotastic: Yes, I exported these to allow people to reuse our special Starlight versions of the the Night Owl themes in their own set of themes. I like the idea of making them accessible through a string value, and will add that in favor of the `getStarlightDefaultThemes` export.

Exposing `@astrojs/starlight/expressive-code` will still be necessary to allow users deeper customization without risking to run into package version conflicts by having to install their own copy of astro-expressive-code. I mentioned in [this earlier comment](https://github.com/withastro/starlight/pull/742#discussion_r1385148195) what the potential customization use cases are.

packages/starlight/schemas/i18n.ts
hippotastic: Ahh, thanks a lot for clarifying! I would love to have translations for the EC strings in all these beautiful languages, but I didn't want to add the burden on Starlight translators to also submit PRs to my upstream EC repository. :)

Therefore, my current plan was to let translators keep following the Starlight-focused workflow they are used to, and perform the work of copying the translations to EC at some point myself. I was hoping that this is the easiest process for the translators.

docs/src/content/docs/reference/configuration.mdx
hippotastic: Expressive Code supports any number of themes out of the box (not just "dark" and "light"), and I've already seen sites using this to provide additional theming options (non-Starlight if I remember correctly).

I've seen one site providing a three-state theme selector with options like "day", "sunset" and "night", and another adding a custom "high contrast" toggle that selects special versions of the dark and light themes when enabled (so effectively providing 4 themes in total, just like you mentioned in your comment).

I didn't want to limit this functionality to just two themes when EC is being used inside Starlight, so I just kept the same `themes` array option as provided by the upstream `astro-expressive-code` package.

If you provide more than two themes, the Starlight integration's `useStarlightDarkModeSwitch` option (which is enabled by default) will create CSS selectors that integrate with Starlight's default theme switcher by making the first dark theme available through `data-theme="dark"` and the first light theme through `data-theme="light"`, and all other themes through their individual names, e.g. `data-theme="my-custom-high-contrast-dark-theme"`.

packages/starlight/schemas/i18n.ts
delucis: Sounds good. We can also encourage our translators to contribute to EC. I might recommend moving your strings out of the code like they are now into JSON or similar files that are easy to find and edit to help people contribute there.

docs/src/content/docs/reference/configuration.mdx
delucis: Aha, got it. Maybe we should make a exports-specific module for the public API? So users can’t accidentally import some of the internal functions that are exported to be used in Starlight’s `index.ts`?

docs/src/content/docs/reference/configuration.mdx
hippotastic: If you leave this option undefined, the default behavior is meant to be just like you described and automatically respect what you configured in `themes`.

The use case I was thinking of when adding this option was to allow people to provide their own logic for code block theme switching (e.g. by adding custom UI to select a preferred syntax highlighting theme), instead of the code blocks reacting automatically to the Starlight theme switcher.

I can understand if you'd prefer leaving this option out until there are actual users requesting it to be added. Just let me know and I can remove it! :)

packages/starlight/schemas/i18n.ts
hippotastic: I agree! After seeing how i18n is done in Starlight, I was also thinking about moving the translations into separate files in EC as well. That's maybe another reason why it's better to have translators focus on the Starlight JSON files for now until I've made the translation process in my upstream repo easier.

docs/src/content/docs/reference/configuration.mdx
hippotastic: Great point! I'll do that and create a separate file for the export.

docs/src/content/docs/reference/configuration.mdx
delucis: Thanks for the explanation!

I’m thinking a bit about that last case using `data-theme`. We don’t necessarily have to solve this in this PR — we have a chance to break the API in a future minor before Starlight v1 if needed — but because you can’t have `data-theme` set twice on an element, setting `<html data-theme="my-custom-high-contrast-dark-theme">` would presumably override the general Starlight theme (switching it to the dark theme, which is our default when `data-theme="light"` isn’t set). So basically Starlight theme and EC theme are tied together. Obviously in the general case dark/light switching with Starlight makes sense though.

I guess this the case that the `useStarlightDarkModeSwitch` option solves? You set that to `false` and decouple the switching?

hippotastic: Yes, you set this option to false and decouple the switching, and then you can either just define your own CSS, or you use the `themeCssSelector` option to switch syntax highlighting themes based on something other than `data-theme`. You can use classes there, other data attributes etc. - allowing them to coexist with Starlight's dark/light mode switch without relying on the same attribute.

delucis: Got it. OK! I think I’m happy to leave these two options as they are. I don’t think they’re super easy to understand or use for the advanced cases, but hopefully that’s a niche concern, and I don’t off the top of my head have any ideas on how to improve it 😁

docs/src/content/docs/reference/configuration.mdx
delucis: Thank again for helping me understand everything. I think this would then be the last outstanding point to tackle. In the meantime, I’ll also open another PR to your branch to tweak the docs a bit to match our general style a little closer.

docs/src/content/docs/reference/configuration.mdx
hippotastic: Ok, let's keep them like this for now then. If we or our users come up with better ideas in the future, I'm always happy to improve the options further. :)

Thank you for the thorough review and great discussions about the use cases!

packages/starlight/integrations/expressive-code/index.ts
```diff
@@ -0,0 +1,250 @@
...
+	if (expressiveCode === false) return [];
+	const config: StarlightExpressiveCodeOptions =
+		typeof expressiveCode === 'object' ? expressiveCode : {};
+
+	const {
+		themes = [],
+		customizeTheme,
+		styleOverrides: { textMarkers: textMarkersStyleOverrides, ...otherStyleOverrides } = {},
+		useStarlightDarkModeSwitch,
+		useStarlightUiThemeColors = config.themes === undefined,
```
delucis, line 68: Coming back with one more bit of feedback as I work on the docs.

While I understand the idea behind making the `useStarlightDarkModeSwitch` and `useStarlightUiThemeColors` options responsive to how `themes` is configured — basically trying to guess what a user wants based on `themes` — I actually think introducing this interdependence between configuration options can be quite confusing. Instead of having a static default, a user has to work out what is going on in another config key and then understand the mapping.

I think I’d like to propose making `useStarlightUiThemeColors` always default to `true`, letting people override it with `false` when that’s not what they want:

```suggestion
		useStarlightDarkModeSwitch,
		useStarlightUiThemeColors = true,
```

`useStarlightDarkModeSwitch` is obviously more heavily dependent on `themes` — if there’s only one theme, or no theme for one of light/dark, we obviously can’t sync with the dark mode. So I think the current behaviour is good. I think I can make the docs clear in this case by saying the default is `true` and then note the exception that if you set a custom theme and don’t provide both dark and light the setting can’t apply. Still drafting this, but something like:

> #### `useStarlightDarkModeSwitch`
>
> **type:** `boolean`  
> **default:** `true`
>
> When `true`, code blocks automatically switch between light and dark themes when the site theme changes.
>
> :::note
> When setting `themes`, you must provide at least one dark and one light theme for the Starlight dark mode switch to work.
> :::

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -206,6 +206,132 @@ var fun = function lang(l) {
...
+      // and we can still add diff markers to it!
+  -   console.log('Old code to be removed')
+  +   console.log('New and shiny code!')
+    }
+  ```
+  ````
+
+### Window frames
+
+By default, every code block that has a language identifier is rendered inside a window frame. This is done by the frames plugin of Expressive Code, which is enabled by default in Starlight. Depending on the code's language identifier, the frame can look like a code editor (similar to VS Code), or like a terminal window.
```
delucis, line 293: This first sentence here is not strictly speaking true from what I see. Every shell language (`sh`, `bash` etc.) will be rendered in a Terminal-like frame, but other languages only get the tab window if they have a title, I think? Is that expected? I can update this in my docs PR if so.

delucis: And I guess the follow-up question would be when the `frame="none"` attribute becomes necessary? I guess mainly if a code comment would otherwise be used as the title?

packages/starlight/integrations/expressive-code/index.ts
hippotastic: It's certainly a valid point that interdependent config options are more difficult to understand. My idea was that if users explicitly specify their own set of themes instead of just keeping the defaults, they would likely expect the themes to look like they do in the VS Code theme gallery, and not like a hybrid between the theme and Starlight.

I could actually imagine dropping this option altogether if we add the `starlight-light` and `starlight-dark` string options as discussed. That the theme respects the Starlight colors would then be a feature of the Starlight themes. What do you think?

docs/src/content/docs/guides/authoring-content.md
hippotastic: Oh, you're right! The code behaves as expected (and covered by many tests), I just wrote an incorrect statement in the accompanying documentation. Great that you found that! If you have the time to update this in the docs PR, that would be much appreciated!

packages/starlight/integrations/expressive-code/index.ts
delucis: I think the option has very valid use cases so I wouldn’t drop it. It lets you choose a syntax highlighting theme without also opting out of the UI frame fitting with your wider site theme.

For example, here’s a quick example of `github-dark` with its built-in colours and with the site colours:

| `useStarlightUiThemeColors` unset | `useStarlightUiThemeColors: true` |
|---|---|
| image | image |

Not super clear without the full site context, but I think having the UI harmonious with the broader site is actually a pretty good default and opting into using the theme for the full UI would be the add-on.

Can be shown even with the default theme where I find disabling the Starlight colours pretty odd (again maybe more out of place with the full page context, but keeping the screenshots small for GitHub):

| `useStarlightUiThemeColors` unset | `useStarlightUiThemeColors: false` |
|---|---|
| image | image |

So I’d definitely keep the option, but I think making it a static default of `true` makes most sense to me.

hippotastic: Ok, I'll make it default to `true` then. I guess if users notice that their chosen themes look different in Starlight than what they expected, they will probably find this option. :)

docs/src/content/docs/guides/authoring-content.md
delucis: Yes, done in https://github.com/hippotastic/starlight/pull/2!

packages/starlight/integrations/expressive-code/index.ts
hippotastic: After making the suggested change to the code and always defaulting the option to `true`, I remembered why I made it dependent on the themes in the first place: If you set `themes` to a single theme like `['github-dark']`, and then switch your Starlight theme to light, Starlight's CSS variables will switch from dark to light without switching the syntax highlighting theme (because there is only one), which results in a dark theme displayed with a light background color.

My automatic color contrast correction cannot fix this scenario because it works per theme and has to rely on the background having a certain lightness. If this lightness changes dramatically at runtime, there is no way to preserve color contrast anymore.

hippotastic: Still thinking about this issue and I don't see any sensible solutions that do not depend on the `themes` option.

For the sake of finalizing this PR, I'd suggest keeping my initial proposal of the option automatically defaulting to false when `themes` has been customized. I would not want to release default settings that can lead to inaccessible color contrast.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -346,6 +346,68 @@ starlight({
...
+
+If you want to disable Expressive Code, set `expressiveCode: false` in your Starlight config:
+
+```js ins={2}
+starlight({
+	expressiveCode: false,
+});
+```
+
+In addition to the standard Expressive Code options, you can customize your code blocks with the following Starlight-specific properties:
```
sarah11918, line 376:
```suggestion
In addition to the standard Expressive Code options, you can also set the following Starlight-specific properties in your `expressiveCode` config to further customize theme behavior for your code blocks :
```
This might be a bit wordy, but especially with no example below showing these configured inside `expressiveCode`, I think a bit of extra context, and especially when these are all theme-related, is nice for reinforcing that even though it's called "expressiveCode", this is where you configure EVERYTHING for your code samples.

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
-```md
-Long, single-line code blocks should not wrap. They should horizontally scroll if they are too long. This line should be long enough to demonstrate this.
-```
+### Expressive Code features
+
+Starlight uses [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code) to extend formatting possibilities for code blocks.
+Expressive Code’s text markers and window frames plugins are enabled by default.
+Code block rendering can be configured using Starlight’s [`expressiveCode` configuration option](/reference/configuration/#expressivecode).
+
+#### Text markers
```
sarah11918, line 211: How standard is the phrase "text markers" for what seems to be describing "highlighting"? I don't think I use this phrase myself at all, so I might be tempted to rename?

```suggestion
#### Line and Text highlighting
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+
+You can highlight specific lines or parts of your code blocks using text markers.
```
sarah11918, line 213:
```suggestion
You can highlight specific lines or parts of your code blocks using [Expressive Code text markers](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#usage-in-markdown--mdx-documents) on the opening line of your code block.
Use curly braces (`{ }`)  to highlight entire lines, and quotation marks to highlight strings of text.
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+There are three styles of marker: a default neutral colored highlight, a green highlight indicating code that was inserted, and a red highlight indicating code that was deleted.
```
sarah11918, line 214:
```suggestion

There are three highlighting styles: neutral for calling attention to code, green for indicating inserted code, and red for indicating deleted code.
Both text and entire lines can be marked using the default marker, or in combination with `ins=` and `del=` to produce the desired highlighting.
```
I've put each sentence on its own line in this section, but I doubt this should all be one paragraph, so there's a line break at the top. Obvs, do what makes visual sense here.

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+
+- [Mark entire lines & line ranges](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#marking-entire-lines--line-ranges):
```
sarah11918, line 216:
```suggestion
Expressive Code provides several options for customizing the visual appearance of your code samples. 
Many of these can be combined, for highly illustrative code samples.
Please explore the [Expressive Code documentation](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md) for the extensive options available.
Some of the most common examples are shown below: 

- [Mark entire lines & line ranges using the `{ }` marker](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#marking-entire-lines--line-ranges):
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+
+  ```js {2-3}
```
sarah11918, line 218: I have a medium-strong preference for swapping all these examples, and showing the code you need to write first, and *then* what it looks like.

Feels more in keeping with helping people write something: see the code, THEN see the results it produces.

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+  ````md
+  ```js {2-3}
+  function demo() {
+    // This line (#2) and the next one are highlighted
+    return 'This is line #3 of this snippet';
+  }
+  ```
+  ````
+
+- [Mark selections of text](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#marking-entire-lines--line-ranges):
```
sarah11918, line 234:
```suggestion
- [Mark selections of text using the `" "` marker or regular expressions](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#marking-entire-lines--line-ranges):
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+  ````md
+  ```js "Individual terms" /Even.*supported/
+  // Individual terms can be highlighted, too
+  function demo() {
+    return 'Even regular expressions are supported';
+  }
+  ```
+  ````
+
+- [Mark text as inserted or deleted with `ins` or `del`](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#selecting-marker-types-mark-ins-del):
```
sarah11918, line 252:
```suggestion
- [Mark text or lines as inserted or deleted with `ins` or `del`](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#selecting-marker-types-mark-ins-del):
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -202,9 +202,140 @@ var fun = function lang(l) {
...
+  ```js "return true;" ins="inserted" del="deleted"
+  function demo() {
+    console.log('These are inserted and deleted marker types');
+    // The return statement uses the default marker type
+    return true;
+  }
+  ```
+  ````
+
+- [Combine syntax highlighting with `diff`-like syntax](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#combining-syntax-highlighting-with-diff-like-syntax):
```
sarah11918, line 272:
```suggestion
- [Combine language-based syntax highlighting with `diff`-like syntax](https://github.com/expressive-code/expressive-code/blob/main/packages/%40expressive-code/plugin-text-markers/README.md#combining-syntax-highlighting-with-diff-like-syntax):
```

OOh, question, does this mean that for Starlight, we do NOT do ins/del for individual lines? I know this is the solution we use for standard Markdown, but our normal `ins={2-4}` should work in Starlight, right? Does this mean you are suggesting this notation (for universal compatibility?)

If so, I might further explain this not as the first resort, but as a way to write code samples in such a way that they'll render in non-Expressive Code environments as diffs, but in your Starlight site, they can take advantage additionally of language highlighting.  (In part, because our code samples in Docs are typically going to use the `ins={2-4}` notation, and so the docs here should maybe have an example of that?)

sarah11918 (commented):
This is so exciting and I can't wait for this! Left some comments on the docs here!

packages/starlight/integrations/expressive-code/index.ts
delucis: Ahhhhhhh, right. One hypothetical option would I guess to be to use the Starlight background color but pin it to the syntax theme’s light/dark type. But that’s not straightforward because you only have access to the global CSS variable which will change.

Yes, I think we should probably revert back to your original choice. Sorry for the thrash on that — I didn’t test this case while trying things out yesterday.

docs/src/content/docs/guides/authoring-content.md
delucis: I also considered this — and “highlight” is used as a verb in this section. I think the slight difficulty is that “syntax highlighting” has such a specific meaning that I wasn’t sure if “highlighting” in this context might imply _that_, rather than what we might think of as “highlighting” in a word processing or handwritten context. Do you think that could be an issue?

docs/src/content/docs/guides/authoring-content.md
delucis: I also do! The only issue is that the entire page currently uses this ordering so I felt unsure about changing this without also revisiting the whole page, which felt like something that should maybe be done in a dedicated PR?

docs/src/content/docs/guides/authoring-content.md
delucis: I believe both diff-style and `ins={...}` should work! So it’s a matter of taste which you choose (although I think the diff-style is probably preferable in most cases for a couple of reasons:

1. Diff highlighting works everywhere (same reason we were using it in Astro integration READMEs).
2. You don’t need to count and maintain line numbers. If you use `+`/`-` at the start of lines, adding or moving parts of the code doesn’t require you to manually update the `ins`/`del` ranges.

packages/starlight/integrations/expressive-code/index.ts
hippotastic: No worries at all! We're both striving for simplicity and ease of use, and it's absolutely valid to consider alternatives. Discussing them helps us all develop a better understanding of the options. I'll keep this behavior unchanged then.

Do you want to adapt the description in the docs or should I try to fit your tone in the updated version?

delucis: Up to you. Maybe as a first step: do YOU want to go through accepting Sarah’s suggestions or would you prefer I do that?

docs/src/content/docs/guides/authoring-content.md
hippotastic: They do both still work, yes. Nothing has been removed. I just tried to limit the amount of EC-focused examples included in the Starlight docs to a reasonable amount and therefore decided to just show an example of the diff-like syntax for the reasons Chris mentioned.

delucis: Yes, I think once Expressive Code has its own docs to show these off, we may even be able to reduce the examples in favour of linking there.

packages/starlight/integrations/expressive-code/index.ts
hippotastic: I think you're better suited to get the language details right as a native speaker, and as the Starlight lead you bear the burden of making all contributions fit together nicely, so I'd gladly leave this to you if you have the time. :)

docs/src/content/docs/reference/configuration.mdx
hippotastic: I have moved the exports into a separate file now as discussed. I also added an explanatory comment to the file to clarify its purpose, even including a simple usage example.

.changeset/sweet-berries-begin.md
```diff
@@ -0,0 +1,44 @@
...
+
+⚠️ **Potentially breaking change:**
+This addition changes how Markdown code blocks are rendered to use [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code).
+If you were already customizing how code blocks are rendered, you may wish to disable Expressive Code to preserve previous behavior:
+
+```js
+starlight({
+  expressiveCode: false,
+})
+```
```
hippotastic, line 15:
```suggestion
If you were already customizing how code blocks are rendered and don't want to use the [features provided by Expressive Code](https://starlight.astro.build/guides/authoring-content/#expressive-code-features), you can preserve the previous behavior by setting the new config option `expressiveCode` to `false`.
```

Reading the current text gave me the impression that we're subtly recommending disabling Expressive Code ("you may wish to", including a prominent example of the code to disable it), so I tried to make this more neutral and a bit less prominent by converting the example block to inline text. Of course I'm biased though, so I'm happy to hear your thoughts on this. :)

docs/src/content/docs/guides/authoring-content.md
hippotastic: I agree with both of you! I also see value in swapping the examples, but we should do it in a separate PR after getting this one over the finish line. :)

docs/src/content/docs/guides/authoring-content.md
hippotastic: I went through the same struggle when naming this feature and the plugin providing it. The unwanted proximity to the widely known and used term "syntax highlighting" was the deciding factor for me that I had to find another name, so I ended up with "text markers".

.changeset/sweet-berries-begin.md
delucis: Hehe. Sure, your version works too. I think technically we ARE recommending this for people using some other custom code block handling, but making it smaller is fine by me.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -346,6 +346,68 @@ starlight({
...
+### `expressiveCode`
+
+**type:** `StarlightExpressiveCodeOptions | boolean`  
+**default:** `true`
+
+Starlight uses [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code) to render code blocks and add support for highlighting parts of code examples, adding filenames to code blocks, and more.
+See the [“Code blocks” guide](/guides/authoring-content/#code-blocks) to learn how to use Expressive Code syntax in your Markdown and MDX content.
+
+You can use any of the standard [Expressive Code configuration options](https://github.com/expressive-code/expressive-code/blob/main/packages/astro-expressive-code/README.md#configuration) by setting them in Starlight’s `expressiveCode` option.
+For example, to give your code blocks rounded corners, set the `styleOverrides.borderRadius` option:
```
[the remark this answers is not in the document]
delucis: I don’t love how this suggestion makes it sound like `styleOverrides` applies rounded corners specifically. Maybe something like this would work?

```suggestion
For example, set Expressive Code’s `styleOverrides` option to override the default CSS. This enables customizations like giving your code blocks rounded corners:
```

packages/starlight/package.json
delucis: Resolved! Now exporting a dedicated `exports` file instead.

.changeset/sweet-berries-begin.md
```diff
@@ -0,0 +1,38 @@
+---
+'@astrojs/starlight': minor
+---
+
+Adds Expressive Code as Starlight’s default code block renderer
+
+⚠️ **Potentially breaking change:**
+This addition changes how Markdown code blocks are rendered to use [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code).
```
sarah11918, line 8:
```suggestion
This addition changes how Markdown code blocks are rendered. By default, Starlight will now use [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code).
```

.changeset/sweet-berries-begin.md
```diff
@@ -0,0 +1,38 @@
...
+If you were already customizing how code blocks are rendered and don't want to use the [features provided by Expressive Code](https://starlight.astro.build/guides/authoring-content/#expressive-code-features), you can preserve the previous behavior by setting the new config option `expressiveCode` to `false`.
+
+If you had previously added Expressive Code manually to your Starlight project, you can now remove the manual set-up in `astro.config.mjs`:
```
sarah11918, line 11:
```suggestion
If you had previously added Expressive Code manually to your Starlight project, you must now remove the manual set-up in `astro.config.mjs`:
```

delucis: Not necessarily! If someone adds Expressive Code manually, Starlight will skip adding it on its end.

hippotastic: Just to provide a bit of context: If you leave your manually added version of `astro-expressive-code` in Astro's `integrations` array, Starlight will detect this and skip adding its own version of the integration. So it's not a "must" per se, but if you want the full Starlight-tailored experience, you should remove your manual one.

hippotastic:
Btw, I just updated to the latest version of `astro-expressive-code` to add the automatic frontmatter block cleanup logic we discussed in the context of Astro Docs. It's also affecting one location in the Starlight docs:

[**Before:**](https://starlight-pfrecczus-astrodotbuild.vercel.app/guides/components/#compatibility-with-starlights-styles)

image

[**After:**](https://starlight-m512hmlgl-astrodotbuild.vercel.app/guides/components/#compatibility-with-starlights-styles)

image

.changeset/sweet-berries-begin.md
delucis: I think I’ll leave this is as “can” if that’s OK. The Starlight version of EC does differ slightly in its defaults, so some people may wish to keep their current config if they’re happy with it (and there’s no real harm in doing so except potentially some extra dependencies).

kevinzunigacuellar: if both of them (EC and Starlight) use the same config options, we could suggest to merge configs to override starlight defaults and keep the user custom config.

hippotastic: > if both of them (EC and Starlight) use the same config options, we could suggest to merge configs to override starlight defaults and keep the user custom config.

I think this is what we are suggesting in the following sentence and showing in the example. By moving your previous manual config into the new `expressiveCode` option, you are overriding the Starlight default options with your changes. This way, you're able to keep your customizations while migrating to the new EC integration built into Starlight.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -346,6 +346,72 @@ starlight({
...
+
+**type:** `boolean`  
+**default:** `true` if the `themes` option is not set (= you are using Starlight's default themes), and `false` otherwise
+
+When `true`, Starlight's CSS variables are used for the colors of code block UI elements (backgrounds, buttons, shadows etc.), matching the [site color theme](/guides/css-and-tailwind/#theming).
+When `false`, the colors provided by the active syntax highlighting theme are used for these elements.
+
+:::note
+When manually setting this to `true` with your custom set of `themes`, you must provide at least one dark and one light theme to ensure proper color contrast.
+:::
```
delucis, line 413: Small tweak and then should be good to go!

```suggestion
#### `useStarlightUiThemeColors`

**type:** `boolean`  
**default:** `true` if `themes` is not set, otherwise `false`

When `true`, Starlight's CSS variables are used for the colors of code block UI elements (backgrounds, buttons, shadows etc.), matching the [site color theme](/guides/css-and-tailwind/#theming).
When `false`, the colors provided by the active syntax highlighting theme are used for these elements.

:::note
When using custom themes and setting this to `true`, you must provide at least one dark and one light theme to ensure proper color contrast.
:::
```

delucis (commented):
Last two comments and then I think I’ll… merge???

packages/starlight/integrations/expressive-code/index.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	// Handle the `themes` option
+	const themes = preprocessThemes(themesInput);
+	if (useStarlightUiThemeColors === true && themes.length < 2) {
+		console.warn(
+			[
+				`*** Warning: Using the config option "useStarlightUiThemeColors: true" `,
+				`with a single theme is not recommended. For better color contrast, `,
+				`please provide at least one dark and one light theme.\n`,
+			].join('')
```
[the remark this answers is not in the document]
hippotastic: Prettier kept reformatting this on me, that's why I went with the array style. :) But you're right, this would be nicer! Let's just give your version a try. I didn't recheck after recently updating Prettier to see if it still reformats this.

docs/src/content/docs/reference/configuration.mdx
hippotastic: Nice simplification, I like it!

packages/starlight/integrations/expressive-code/index.ts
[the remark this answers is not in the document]
hippotastic: Yep, it works now! Nice!

delucis (approved):
Ladies, gentlemen, hippopotamuses…

I HEREBY DECLARE THIS PULL REQUEST READY TO MERGE 📣

Thank you for the hard work @hippotastic and thanks to everyone who reviewed and gave feedback!

PR #743, opened 2023-09-22
Fix pa11y tests on Windows

hippotastic:

- Changes to Starlight code

#### Description

- Fixes an issue that made it impossible to run accessibility tests on Windows. In the `docs` folder, neither `pnpm test` nor `pnpm pa11y` were working due to OS differences.
- To solve the differences, I enabled the `shell-emulator` flag in `.npmrc` and changed the `test` script to ensure it actually uses `pnpm` instead of its default `npm`.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 prefer-workspace-packages=true
 link-workspace-packages=true
+shell-emulator=true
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thank you @hippotastic! Super helpful — can always rely on you to uncover — AND FIX — these kinds of issues. Appreciate it 🙌

PR #739, opened 2023-09-22
feat: add Indonesian UI translation

radenpioneer:
#### What kind of changes does this PR include?

- Changes or translations of Starlight UI

#### Description

Added Indonesian to Starlight UI translations.

PR #750, opened 2023-09-24
feat(translations): add id.json (Bahasa Indonesia)

ImBIOS:


HiDeoo (commented):
Thanks for your contribution!

It looks like this may be a duplicate of #739 which is being worked on. If this is the case, it might be worth concentrating efforts there and as it looks like there are some differences between the two translations, perhaps reviewing and suggesting some changes if necessary to the proposed translation?

The different translations that I can see between the two PRs are:

- `search.label`
- `search.devWarning`
- `tableOfContents.onThisPage`
- `tableOfContents.overview`
- `page.lastUpdated`
- `404.text`

And also a few capitalisation differences in various places.

delucis:
Thanks again for the PR @ImBIOS! As @HiDeoo we already had an open PR for this which I just merged, so will close this. If you spot any issues with the newly added Indonesian language support, please do open a PR to fix them!

radenpioneer:
Hey @ImBIOS, just noticed you also doing translation on this project! I've started to translate the docs too, which recently merged (#765). There are lot to translate, so lets join to translate all the docs to Bahasa Indonesia!

The list of untranslated page is in https://i18n.starlight.astro.build/ , 

_Kuy lah bantuin, sekalian ajak-ajakin yang lain!_

PR #752, opened 2023-09-24
add twitter new logo

apinet:
- Add an icon

#### Description
- Add the new twitter svg logo next to the older one.


HiDeoo:
Thanks for your contribution. :raised_hands:

This changes makes me wonder a few things:

- Should this be added as a new logo or should it replace the current `twitter` one?
- Should this have an impact on the [social media icons](https://starlight.astro.build/reference/configuration/#social) displayed in the site header?
- If yes, should `twitter` be renamed to `x` or both be kept?

Not sure what to think as `x.com` links redirects to `twitter.com` but `twitter.com` uses the new `x` logo on the official Twitter website, this is a bit confusing ^^ :thinking:

apinet:
If we replace `twitter` for `x`, we risk to break references to the old icon if some dev buddies used it directly. Maybe we could keep both for a few months and remove twitter icon after that?

Regarding [social media icons](https://starlight.astro.build/reference/configuration/#social) I will create a new pull request for supporting X as social media.

delucis (commented):
Thanks for this @apinet! I think where you and @HiDeoo ended up makes sense for now, even if it is a bit odd having both, it lets people decide how much they want to humour the rebranding shenanigans 😅 

One thing: I wonder if we should use `x.com` as the key here? We’ve discussed expanding our built-in icon support to include the full set of Unicons (the icon set Starlight uses currently) and it already includes a generic `x` icon that could clash. What do you think?

HiDeoo:
> One thing: I wonder if we should use `x.com` as the key here? We’ve discussed expanding our built-in icon support to include the full set of Unicons (the icon set Starlight uses currently) and it already includes a generic `x` icon that could clash. What do you think?

Sounds fine by me.

I guess the remaining missing changes after updating the key would be updating `docs/src/content/docs/guides/customization.mdx` & `docs/src/content/docs/reference/configuration.md` to document it and adding a changeset.

apinet:
Hi @HiDeoo. I have updated customization & configuration files in all available languages. Let me know if I miss something but I think w're good :)

HiDeoo (commented):
Thanks for the update, this PR may requires a few changes:

- Following Chris' [comment](https://github.com/withastro/starlight/pull/752#pullrequestreview-1642748574), we should use `x.com` for the icon key to avoid a potential clash with the `x` icon already existing in the Unicons set.
- Would you mind reverting the changes in [d715b5f](https://github.com/withastro/starlight/pull/752/commits/d715b5f8a3148e87d7bac074200a3e12ee5e5713) and only keep the changes for the English version (`docs/src/content/docs/guides/customization.mdx` & `docs/src/content/docs/reference/configuration.md`)? We usually only update the English pages and leave changes to the translations to our translators as this makes it easier for the [Starlight translation tracker](https://i18n.starlight.astro.build/) to keep track of the changes.

delucis (approved):
Thanks again for your contribution @apinet! Hope you don’t mind I took the liberty of updating this PR with the last few changes so we can get this out in our next release.

PR #756, opened 2023-09-25
Prevent text from overflowing in various cases

### LINKED ISSUE #754: bug(minor): Sidebar Labels breaking / not properly displaying

0.10.1

3.1.2

npm, yarn

Windows

Chrome, Firefox

### Describe the Bug

The Sidebar labels do not wrap when the given word is too long, however this works completely fine for longer sentences.

This is rather a minor issue.

Explanation with pictures:

| ❌ Long Words | ✅ Long Sentences|
| :-----------: | :-----------: |
| The word does not wrap as long sentences do | The word/sentence wraps as it should |
| image | image  |

https://stackblitz.com/github-7maapr

### Participation

julien-deramond:

- Changes to Starlight code

#### Description

Closes #754

⚠️ **This PR is a draft**. It has been opened early on in order to add some screenshots, a task list, and maybe to discuss some technical choices. Feel free to close it if you receive too many notifications :)

The main idea here, for now, is to add some `overflow-wrap: break-word`. 

`word-break: break-word` is already used in this project but I'd suggest to replace it by `overflow-wrap: break-word` since `break-word` seems to be deprecated.

On top of that, some boxes don't seem to have the right width (or any width defined), so the PR suggests setting it to either 100% or 90% for the accordion so that the caret remains visible.

#### TODO

- [ ] Double-check that `break-word` is indeed well deprecated (see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values).
- [ ] Double-check that the various fixes are the only options
- [ ] Non-regression testing
- [ ] Provide screenshots for the different use cases and add more detailed descriptions for each use case
- [ ] Add changeset
- [ ] **Before merging**, remove the commit that adds some long words to the documentation


packages/starlight/components/SidebarSublist.astro
```diff
@@ -57,6 +57,11 @@ interface Props {
 		margin-inline-start: var(--sl-sidebar-item-padding-inline);
 		border-inline-start: 1px solid var(--sl-color-hairline-light);
 		padding-inline-start: var(--sl-sidebar-item-padding-inline);
+		overflow-wrap: break-word;
```
HiDeoo, line 60: This selector seems to be a bit too specific and does not cover top-level links:

SCR-20230926-jczb

To test this use-case, you can locally add a new entry to the sidebar configuration:, e.g.:

```ts
sidebar: [
  {
    label: "ThisIsAVeryLongLabelToTestTheSidebarIssues",
    link: "test",
  },
  // Existing sidebar configuration
];
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -57,6 +57,11 @@ interface Props {
...
+	}
+
+	ul :global(.sl-badge) {
+		max-width: 100%;
 	}
```
HiDeoo, line 65: I think an alternative solution could be to not do that and use `overflow-wrap: anywhere;` instead of `overflow-wrap: break-word;` in the `<Badge/>` component. This would mean that we would not have to do that every time we use a badge in a similar context.

Any thoughts?

packages/starlight/components/SidebarSublist.astro
```diff
@@ -82,6 +87,10 @@ interface Props {
 		display: none;
 	}
 
+	summary > h2 {
+		width: calc(100% - 1.25rem);
+	}
+
```
HiDeoo, line 93: > Tried some flex-shriking on the caret without success.

What issues did you encounter? Disabling shrinking using `flex-shrink: 0;` on the caret icon and using `overflow-wrap: anywhere;` on the title seems to do the trick for me but I may be missing something.

packages/starlight/components/PrevNextLinks.astro
```diff
@@ -77,4 +78,12 @@ const t = useTranslations(locale);
...
 		line-height: var(--sl-line-height-headings);
 	}
+
+	svg {
+		flex-shrink: 0;
+	}
+
+	svg + span {
+		width: calc(100% - 1.5rem);
+	}
```
HiDeoo, line 88: If I am not missing something, I guess like in the `<SidebarSublist/>` component, this could be avoided by using `overflow-wrap: anywhere;` instead of `overflow-wrap: break-word;` for the link.

Any thoughts?

packages/starlight/components/PrevNextLinks.astro
HiDeoo, line 1: This one is a bit trickier indeed.

The current approach will break the existing wrapping behavior on small width screens:

| Before          | After           |
| --------------- | --------------- |
| SCR-20230926-jkeq | SCR-20230926-jkcm |

It also looks like this would not work when you are on a page where both the previous and next links are fairly long:

SCR-20230926-jkph

Not sure yet what would be the best approach here, might need to think about it a bit more :thinking:

HiDeoo (commented):

I noticed a few edge cases and left a few comments/questions but the changes seems to work relatively well overall and be a reall improvement so far.

packages/starlight/components/SidebarSublist.astro
julien-deramond: Neat, it works pretty well! So obviously, this is way better in terms of usage since there's nothing to do while using the component.

I was a bit worried about [`overflow-wrap: anywhere` browser support](https://caniuse.com/?search=overflow-wrap%3A%20anywhere) for Safari which is only from 15.4.



But the team confirmed that other features used in Astro and/or Starlight are only working from Safari 15.4.
I ran some tests with BrowserStack with Safari 15.3 just to see how's the fallback in our specific use case, but it's impossible to say what's happening:

Screenshot 2023-09-26 at 18 43 49

I don't have access to any Safari 15.4, but from 16, everything's rendered as what you can see in your browser right now deployed by Netlify. Double-checked with various versions of Chrome and Firefox and seems to be good to go with this solution.

packages/starlight/components/SidebarSublist.astro
julien-deramond: Ohhh good catch! I've added this use case in the commit so that we can test it.
So I've dropped this rule and added instead the following which is less specific:
```css
.top-level > li {
  overflow-wrap: anywhere;
}
```

packages/starlight/components/PrevNextLinks.astro
julien-deramond: Based on the previous comments and fixes, this has been removed.

packages/starlight/components/PrevNextLinks.astro
julien-deramond: Good catch, I didn't test at the right breakpoints for mobile. So I've tried something with a `@media` rule to handle the 2 types of rendering: 1 row, 2 rows:

```css
a {
  /* ... */
  width: 100%;
  flex-basis: calc(100% - 0.5rem);
  /* ... */
}

@media (min-width: 29.6em) {
  a {
    width: 50%;
      flex-basis: calc(50% - 1rem);
    }
}
```

Note that `29.6em` might be a lil' bit magical here. I've found this value by comparing the rendering between this branch and the _main_ branch. The switch between 1 row to 2 rows was between 473px and 474px which led me to that `29.6em`. Obviously, can be changed to something like `30em` or other values.

Being still a bit tricky to test, I hope that I've covered most of the use cases 😅

julien-deramond:
Please note that I haven't yet updated the description of the PR with the new modifications. I'm waiting for an agreement on the final solution with the first code review since the description is a bit heavy to update. Hope you don't mind.

HiDeoo (commented):
From all the cases/pages I tested, the new changes look great. Let's see if some other people find more edge cases or not.

I'm not particularly a huge fan of the changes for the next/prev links but I have yet to find a more elegant solution not breaking anything.

Btw, did you investigate the `CI / Check for accessibility issues` failure or do you want me to take a look at it? (note: to run them locally you can `NO_GRADIENTS=true pnpm build && pnpm test`)

julien-deramond:
> I'm not particularly a huge fan of the changes for the next/prev links but I have yet to find a more elegant solution not breaking anything.

Neither do I.

> Btw, did you investigate the CI / Check for accessibility issues failure or do you want me to take a look at it? (note: to run them locally you can NO_GRADIENTS=true pnpm build && pnpm test)

Not yet, I'll check this out ASAP.

julien-deramond:
I haven't the precise answer yet but there are two errors about color contrast ratio thresholds located at:
```
<span class="astro-sn2en3ud">Configuration Reference</span>
<span class="astro-sn2en3ud">Frontmatter Reference</span>
```
Those elements are in the sidemenu and their rendering has not been changed.
When I remove my extra-long wordings, pa11y-ci is happy again.
So I'd say that the errors start to happen when there's too much content in the sidebar and that it's not especially related to this PR (where the extra-long labels will be removed before merging).

I can dig the issue to have more details if necessary.

HiDeoo:
Thanks for the investigation.

Interesting. Even more interesting is that if I switch locally to this branch and run the tests on my machine, they pass. I guess it depends on the window size which depends on the screen resolution. I assumed that the window size would be fixed, but it seems that it is not or something else is going on.

Nonetheless, I guess this is indeed something that should be fixed in another PR as when reverting your temporary long texts, the tests should pass again.

julien-deramond:
FYI I've updated the description so that it reflects the final changes, squashed the commits, and split them so that it is easier for you to remove the long text tests when the PR is considered ready to merge (you'll just have to drop https://github.com/withastro/starlight/pull/756/commits/9df0e55b6bd23fc492673c0fcbf31667e61ac26e).

packages/starlight/components/PrevNextLinks.astro
delucis: Currently inside Starlight we only use `(min-width: 50rem)` and `(min-width: 72rem)`. Would love to keep us to that consistency if possible.

This is definitely one of the trickier ones. I also think it’s one of the less crucial ones, so we could also decide to split out fixing this one into a follow-up PR if figuring it out will otherwise block this PR.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -65,6 +65,10 @@ interface Props {
 		color: var(--sl-color-white);
 	}
 
+	.top-level > li {
+		overflow-wrap: anywhere;
+	}
```
delucis, line 70: Should this really only apply to the top-level list, or any `li`?

delucis (commented):
Thanks for tackling all of these edge cases @julien-deramond! Really nice work. Left a couple of notes to try to help get this PR over the line! 🙌

packages/starlight/components/SidebarSublist.astro
julien-deramond: The components being black boxes, it's safe to apply the rule directly to `li` as defensive CSS. I've changed it to:

```diff
diff --git a/packages/starlight/components/SidebarSublist.astro b/packages/starlight/components/SidebarSublist.astro
index c4c5882..63308f6 100644
--- a/packages/starlight/components/SidebarSublist.astro
+++ b/packages/starlight/components/SidebarSublist.astro
@@ -53,6 +53,10 @@ interface Props {
                padding: 0;
        }
 
+       li {
+               overflow-wrap: anywhere;
+       }
+
        ul ul li {
                margin-inline-start: var(--sl-sidebar-item-padding-inline);
                border-inline-start: 1px solid var(--sl-color-hairline-light);
@@ -65,10 +69,6 @@ interface Props {
                color: var(--sl-color-white);
        }
 
-       .top-level > li {
-               overflow-wrap: anywhere;
-       }
-
        .top-level > li + li {
                margin-top: 0.75rem;
        }
```

packages/starlight/components/PrevNextLinks.astro
julien-deramond: I definitely can understand that we prefer to stick with only `(min-width: 50rem)` and `(min-width: 72rem)`. I've made some tests with container queries that could help but they are not available before Safari 16.
So I've dropped this modification (kept the `flex-shrink: 0` for the `svg`) and I'll try to tackle it in a separate PR.

delucis (changes requested):
Thanks for splitting out that one fix to #814 @julien-deramond!

I think this is looking good if you’d like to go ahead and remove the text strings being used for testing.

julien-deramond:
https://github.com/withastro/starlight/pull/756/commits/55be3549896799f0eab604df2c5a4a63beac9a16 has been removed, and so the a11y tests are back to normal.

HiDeoo (dismissed):

HiDeoo (commented):
Just noticed we still need a changeset for the changes.

julien-deramond:
Added a changeset via https://github.com/withastro/starlight/pull/756/commits/47526e255acd80624109b3f9193599d8d95f7396. I supposed it was a "patch" and used a really simple generic sentence. Don't hesitate to update it to change the type of changeset or to enhance the description with more detail if needed.

delucis (approved):
Thanks again for all the hard work chasing down these issues and figuring out the fixes @julien-deramond — and for helping @HiDeoo!

PR #760, opened 2023-09-25
Escape argument in examples build script

delucis:

- Something else!

#### Description

- #743 introduced `shell-emulator=true` in `.npmrc`
- This caused the Size Limit check to fail in #709 because the example build script included `@examples/*` unescaped, which with shell emulation enabled was parsed as a glob and errored when it failed to expand the `*`.
- This PR wraps the `--filter` pattern in `'` to escape it.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 prefer-workspace-packages=true
 link-workspace-packages=true
+shell-emulator=true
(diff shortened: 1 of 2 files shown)


delucis:
Hmm, looks like that wasn’t it 🤔 

Will need to dig further to work out why it’s hitting this error:
```
> pnpm --filter @example/* build

No matches found: "@example/*"
```

PR #774, opened 2023-09-27
Add support for custom HTML attributes to sidebar links

### LINKED ISSUE #395: Ability to add arbitrary attributes to sidebar items

0.5.5

### What is your idea?

It would be helpful if we were able to have additional attributes to sidebar items such that we can add tags, e.g. "deprecated" tags, to individual sidebar items.

This feature is relevant for docs sites where we have multiple versions of the product and certain features have been deprecated. We do not want to remove the documentation for the deprecated features altogether, but make it clear these features are deprecated both on the markdown itself, and also in the sidebar, as a convenient visual indicator.

- MDN reference
    Screenshot_2023-07-24_at_10 39 11_AM

- Docusaurus allows the `className` attribute which allows this to be done with CSS (see https://docusaurus.io/docs/sidebar/items).

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #395.

This pull request adds support for custom HTML attributes to sidebar links from config and frontmatter.

An important point regarding this is how to validate/type this? There are several options:

- Add, use and maintain a giant but complete zod schema for all possible HTML attributes (_I don't think this is a good idea_).
- Add and use a smaller schema to validate the attributes as `Record<string, string | number | boolean | undefined>`.
- Add and use a smaller schema to validate the attributes as `Record<string, string | number | boolean | undefined>` but type the attributes as a variant of `HTMLAttributes<'a'>` for user convenience so they would get autocompletion in their config.

I went with the last option as it seemed like a good balance between validation and user experience and it's also the approach (ignoring the first one) with the most amount of code so it'll be easy to remove if we decide to go with a different approach.

A few implementation notes:

- Used `attrs` for consistency with [`HeadConfig`](https://starlight.astro.build/reference/configuration/#headconfig).
- Moved sidebar related schemas to `schemas/sidebar.ts` as some piece are now shared between the config and frontmatter (and it also started to get a bit messy imo).
- Added a [new section](https://deploy-preview-774--astro-starlight.netlify.app/guides/sidebar/#custom-html-attributes) to the sidebar guide with an example (picked the external link with icon as it may be a common use case).


packages/starlight/schemas/sidebar.ts
```diff
@@ -0,0 +1,89 @@
...
+
+// HTML attributes that can be added to an anchor element, validated as
+// `Record<string, string | number | boolean | undefined>` but typed as `HTMLAttributes<'a'>`
+// for user convenience.
+const linkHTMLAttributesSchema = z.record(
+	z.union([z.string(), z.number(), z.boolean(), z.undefined()])
+) as z.Schema<Omit<HTMLAttributes<'a'>, keyof AstroBuiltinAttributes | 'children'>>;
+export type LinkHTMLAttributes = z.infer<typeof linkHTMLAttributesSchema>;
+
+export const SidebarLinkItemHTMLAttributesSchema = () => linkHTMLAttributesSchema.optional();
```
delucis, line 26: Would a default value here also make sense? Would avoid us needing to do any optional chaining when reading from it later and I don’t think we’re relying on this being undefined?

```suggestion
export const SidebarLinkItemHTMLAttributesSchema = () => linkHTMLAttributesSchema.default({});
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -17,7 +17,8 @@ interface Props {
 					<a
 						href={entry.href}
 						aria-current={entry.isCurrent && 'page'}
-						class:list={{ large: !Astro.props.nested }}
+						class:list={[{ large: !Astro.props.nested }, entry.attrs?.['class']]}
```
delucis, line 20: For example here we’d be able to do this:
```suggestion
						class:list={[{ large: !Astro.props.nested }, entry.attrs.class]}
```

packages/starlight/schemas/sidebar.ts
```diff
@@ -0,0 +1,89 @@
...
+	/** Whether this item should be collapsed by default. */
+	collapsed: z.boolean().default(false),
+});
+
+// HTML attributes that can be added to an anchor element, validated as
+// `Record<string, string | number | boolean | undefined>` but typed as `HTMLAttributes<'a'>`
+// for user convenience.
+const linkHTMLAttributesSchema = z.record(
+	z.union([z.string(), z.number(), z.boolean(), z.undefined()])
+) as z.Schema<Omit<HTMLAttributes<'a'>, keyof AstroBuiltinAttributes | 'children'>>;
```
delucis, line 23: Love this solution! Nice that we can leverage Astro’s built-in type and still keep this pretty lightweight.

packages/starlight/utils/navigation.ts
```diff
@@ -17,6 +22,7 @@ export interface Link {
 	href: string;
 	isCurrent: boolean;
 	badge: Badge | undefined;
+	attrs: LinkHTMLAttributes | undefined;
```
delucis, line 25: If we have that default value in the schema, I guess this also would need updating:

```suggestion
	attrs: LinkHTMLAttributes;
```

packages/starlight/utils/navigation.ts
```diff
@@ -308,6 +321,8 @@ function applyPrevNextLinkConfig(
 				...link,
 				label: config.label ?? link.label,
 				href: config.link ?? link.href,
+				// Explicitly remove sidebar link attributes for prev/next links.
+				attrs: undefined,
```
delucis, line 325: Ah, although I guess this would diverge here then. Maybe resetting to `attrs: {}` could make sense? What are your thoughts?

docs/src/content/docs/guides/sidebar.mdx
delucis, line 1: The example in docs is very pretty with its icon, but it may be slightly confusing as it’s kind of a “Draw the rest of the owl” example because we show “add class!” and then 🎉 magic icon appears, without showing how exactly that’s created.

Given the focus here is probably not to go in depth into different techniques for adding icons with CSS, I might suggest a simpler example e.g. with `style: 'color: red'` or something that shows the basic mechanism without needing to guess at some hidden code.

delucis (commented):
Awesome work as always @HiDeoo! Left a few comments but really thorough and well thought through 👏

packages/starlight/schemas/sidebar.ts
HiDeoo: I think so, I actually started to implement it with a default value at first and I refactored to an optional value later on just to match the badges implementation but it's not really necessary.

packages/starlight/components/SidebarSublist.astro
HiDeoo: Oh yeah, I'm way to used to `noPropertyAccessFromIndexSignature` ^^

packages/starlight/utils/navigation.ts
HiDeoo: Makes sense to me 👍

docs/src/content/docs/guides/sidebar.mdx
HiDeoo: I went for `font-style: italic` after trying for way too long to find a color matching the WCAG 2.0 level AA requirements for both the light & dark theme ^^ Hope it is obvious enough.

delucis: Looks good! We only need to demonstrate the concept here and this should do that nicely.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -316,6 +316,50 @@ The configuration above generates the following sidebar:
 	]}
 />
 
+## Custom HTML attributes
+
+Links can also include an `attrs` property to add custom HTML attributes to the link element.
+
+An example use case could be to add a `target="_blank"` attribute to external links to open them in a new tab and also apply a custom `style` attribute to italicize the link label.
```
delucis, line 323: Suggesting a small rewording here inspired by recent guidance from @sarah11918 about introducing code examples:

```suggestion
In the following example, `attrs` is used to add a `target="_blank"` attribute, so that the link opens in a new tab, and to apply a custom `style` attribute to italicize the link label:
```

delucis (commented):
Thanks @HiDeoo — I think the new style in the sidebar guide looks good.

Left one suggestion but then this should be good for when we do our next minor release. Thanks again for tackling it!

PR #786, opened 2023-10-02
Add browserslist query to `CONTRIBUTING.md`

HiDeoo:

- Something else!

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1156269451730235392) on Discord, this PR adds a browserslist query to the `CONTRIBUTING.md` file to make it easier for contributors to know which browsers are targeted by Starlight when adding changes involving CSS.

I explicitely made sure the sentence contains the word `browserslist` so people looking for it can find it easily considering we choose to not use a `.browserslistrc` file.

The discriminators in the Starlight codebase seems to currently be support for the [`:where()` CSS selector](https://caniuse.com/mdn-css_selectors_where) and the [HTML `dialog` element](https://caniuse.com/dialog) so I ended up with the following query:

```
> 0.5%, not dead, Chrome >= 88, Edge >= 88, Firefox >= 98, Safari >= 15.4, iOS >= 15.4, not op_mini all
```

There is an alternative syntax for browserslist query to target specific features like `fully supports dialog` but unfortunately it looks like we cannot use it to target the `:where()` selector as the feature is not tracked by Can I use directly but using MDN browser-compat-data instead (I tried `fully supports where`, `fully supports css_selectors_where` and `fully supports mdn-css_selectors_where`).

Considering this, I decided to stick to manually hand picking the browsers that support the features we use:

- `> 0.5%`: Removes browsers with less than 0.5% of market share
- `not dead`: Removes browsers that are no longer maintained
- `Chrome >= 88`: First version of Chrome to support `:where()`
- `Edge >= 88`: First version of Edge to support `:where()`
- `Firefox >= 98`: First version of Firefox to support `dialog`
- `Safari >= 15.4`: First version of Safari to support `dialog`
- `iOS >= 15.4`: First version of Safari for iOS to support `dialog`
- `not op_mini all`: Removes Opera Mini which does not support CSS custom properties


HiDeoo:
Hmmm, good question indeed. We could include "client-side JavaScript" in the new bullet point so that it reads:

```md
- Components that require client-side JavaScript or CSS for their styling should use features that are well-supported by browsers.

  You can find a list of supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+88%2C+Edge+%3E%3D+88%2C+Firefox+%3E%3D+98%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all). To check whether or not a JavaScript/CSS feature is supported, you can visit the [Can I use](https://caniuse.com) website and search for the feature.
```

Altho, before doing that, even tho I think it would not change the query, I'd like to make a pass at our web components to double check that.

HiDeoo:
As expected, after checking our client-side JS usage, the browserslist query does not need any update if I did not miss anything so I tweaked the sentence to include client-side JavaScript.

delucis (approved):
LGTM! Thanks @HiDeoo 💜 

Noting that https://caniuse.com doesn’t support quite the same syntax, so setting up a custom browser set in settings needs a tweak:

```
> 0.5% in primary
chrome >= 88
edge >= 88
firefox >= 98
safari >= 15.4
ios_saf >= 15.4
not op_mini > 0
```

Don’t think we need to include that in this text, but thought I’d mention it after just setting this up.

PR #789, opened 2023-10-02
Bump astro and @astrojs/mdx dependencies

delucis:

- Something else!

#### Description

- Upgrades dependencies
- `astro`: 3.0.6 => 3.2.1
- `@astrojs/mdx`: 1.0.0 => 1.1.0

### DIFF
--- a/.changeset/fifty-meals-explain.md
+++ b/.changeset/fifty-meals-explain.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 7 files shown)


delucis:
Noting that the 14% CSS size increase `size-limit` is reporting is a little misleading. Due to https://github.com/withastro/astro/issues/8567 Astro is now outputting an extra (unused) file for each of our user-exposed components (`Tabs`, `Icon` etc.) That said, given we test the `basics` example, which does not use any of those components, we’re technically tracking something we weren’t before, so it’s not all bad.

PR #796, opened 2023-10-02
Add built-in integrations conditionally

### LINKED ISSUE #717: Starlight with an existing sitemap integration

0.9.1

3.0.10

npm

Windows

Chrome

### Describe the Bug

We'd like to use starlight with an existing astro setup. 

Starlight forces you to use the `@astrojs/sitemap` integration. There is no way to opt-out (only from injecting a `<link>` to the head) nor configure it (add filters).

Currently there is no way to avoid this issue if you'd like to use a configured sitemap integration:

```
 error   Found more than one instance of @astrojs/sitemap.
  Starlight includes @astrojs/sitemap by default.
  Please remove it from your integrations array in astro.config.mjs
```

it is easy to check, the sitemap integration is not configurable:
https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/sitemap.ts

I think it would be preferable to have an option to opt-out entirely from all of the sitemap generation in starlight in case of an existing astro setup.

_No response_

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #717.

This PR is an attempt to refactor how Starlight adds its built-in integrations (`@astrojs/sitemap` & `@astrojs/mdx`).

It relies on [the changes](https://github.com/withastro/astro/pull/8672) introduced in [Astro v3.2.0](https://github.com/withastro/astro/releases/tag/astro%403.2.0) to add the integrations dynamically using `updateConfig()` in the `astro:config:setup` hook of the Starlight integration.

A few notes:

- The destructuring is needed for the inline syntax as the `condition ? integration() : null` syntax [supported](https://github.com/withastro/astro/blob/78fda5c3ecebe8e15a0ce2941c80f3742665b9f7/packages/astro/src/integrations/index.ts#L88-L91) in the Astro config is not supported when using `updateConfig()` as commented [here](https://github.com/withastro/astro/pull/8672#pullrequestreview-1646575107). This would result in the `Cannot read properties of null (reading 'hooks')` error [here](https://github.com/withastro/astro/blob/78fda5c3ecebe8e15a0ce2941c80f3742665b9f7/packages/astro/src/integrations/index.ts#L94).
  - We could refactor this to a more readable syntax with an array declared earlier and conditionally push to it if needed with some explicit `if` statements.
- I kept the `astro:config:done` check, even though it should be rarely hit but I updated the error message (not sure if it's explicit enough tho).
- This is a breaking change as it requires Astro v3.2.0. Should we link to [Upgrade Astro steps](https://docs.astro.build/en/guides/upgrade-to/v3/#upgrade-astro) on the Astro v3 upgrade guide from the changeset like we did for the Astro v3.0.0 upgrade or is there a more generic page regarding Astro upgrades?


packages/starlight/index.ts
```diff
@@ -35,6 +35,12 @@ export default function StarlightIntegration(opts: StarlightUserConfig): AstroIn
 					entryPoint: '@astrojs/starlight/index.astro',
 				});
 				const newConfig: AstroUserConfig = {
+					integrations: [
+						...(!config.integrations.find(({ name }) => name === '@astrojs/sitemap')
+							? [starlightSitemap(userConfig)]
+							: []),
+						...(!config.integrations.find(({ name }) => name === '@astrojs/mdx') ? [mdx()] : []),
+					],
```
delucis, line 43: > We could refactor this to a more readable syntax with an array declared earlier and conditionally push to it if needed with some explicit if statements.

Absolutely — let’s do it!

```suggestion
				const integrations: AstroIntegration[] = [];
				if (!config.integrations.find(({ name }) => name === '@astrojs/sitemap') {
					integrations.push(starlightSitemap(userConfig));
				}
				if (!config.integrations.find(({ name }) => name === '@astrojs/mdx') {
					integrations.push(mdx());
				}
				const newConfig: AstroUserConfig = {
					integrations,
```

.changeset/happy-tips-burn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/starlight': minor
+---
+
+Add the `@astrojs/sitemap` and `@astrojs/mdx` integrations only if they are not detected in the Astro configuration.
+
+⚠️ **BREAKING CHANGE** The minimum supported version of Astro is now v3.2.0. Make sure you update Astro at the same time as updating Starlight.
```
delucis, line 7: > This is a breaking change as it requires Astro v3.2.0. Should we link to [Upgrade Astro steps](https://docs.astro.build/en/guides/upgrade-to/v3/#upgrade-astro) on the Astro v3 upgrade guide from the changeset like we did for the Astro v3.0.0 upgrade or is there a more generic page regarding Astro upgrades?

We don’t have generic guidance — although perhaps we should. One option could just be to include that guidance directly here?

````suggestion
⚠️ **BREAKING CHANGE** The minimum supported version of Astro is now v3.2.0. Make sure you update Astro at the same time as updating Starlight:

```sh
npm install astro@latest
```
````

delucis (commented):
Fantastic. Thanks for tackling that @HiDeoo!

> I kept the `astro:config:done` check

I think actually it could make sense to just remove it for now and see if people ever actually hit this. Technically we could know the candidates for causing this by storing `config.integrations` in the setup hook. The only integrations that could dynamically add something _after_ Starlight’s setup hook, would be integrations after Starlight in the array, i.e. given an array of `[mdx(), starlight(), mystery()]`, we could tell people `mystery` was responsible.

Possible, but I feel like rare enough that removing the maintenance of this code for now makes sense.

.changeset/happy-tips-burn.md
HiDeoo: Great idea, very clever, did not think of that at all 👍

HiDeoo:
Thanks for the review 🙌 

> Possible, but I feel like rare enough that removing the maintenance of this code for now makes sense.

Seems perfectly fine to me.

And if someone hits this issue, from their config it'll be easy to pinpoint the faulty integration and we could link them your [section](https://astro.build/blog/astro-320/#dynamically-added-integrations) in the Astro 3.2 blog post.

delucis (approved):
Thanks @HiDeoo! I think we can release this in the next minor. Let's hold off from merging immediately, but we can aim to get a minor release out this week.

PR #801, opened 2023-10-03
[Chore] Update Docs For subpath

guidiego:
> Expected to use it in hacktoberfest! 

The PR explains a smart way to lead with a subpath without a new directory-level

**Problems:** When using a subpath people may expect `/lang/subpath` which did not happen in this case.

#### Description
Related with:
Discord Talks: 

### ToDo
This is a smart/fast way to introduce the subpath, we keep need to make some changes to support lang before the subpath!

### DIFF
--- a/docs/src/content/docs/manual-setup.mdx
+++ b/docs/src/content/docs/manual-setup.mdx
@@ -102,26 +102,17 @@ If you have an existing Astro project, you can use Starlight to quickly add a do
 
 ### Use Starlight at a subpath
 
-To add all Starlight pages at a subpath, place all your docs content inside a subdirectory of `src/content/docs/`.
+To add all Starlight pages at a subpath, you could relay on top of Astro.js configurations. `base` and `assetPrefix`
 
-For example, if Starlight pages should all start with `/guides/`, add your content in the `src/content/docs/guides/` directory:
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/manual-setup.mdx
```diff
@@ -102,26 +102,34 @@ If you have an existing Astro project, you can use Starlight to quickly add a do
 
 ### Use Starlight at a subpath
 
-To add all Starlight pages at a subpath, place all your docs content inside a subdirectory of `src/content/docs/`.
+To add all Starlight pages at a subpath, you could relay on top of Astro.js configurations. `base` and `assetsPrefix`
```
ringods, line 105: Does this mean you can use Starlight only on a single subpath?

I want to document more than one distinct "project", each on a different subpath, having a separate ToC, but adding up to a single shared search index. E.g.:

/site (docs describing my own website setup)
/homelab (docs describing my homelab)
/sports (docs describing my cycling gear, associated electronics and configuration)

delucis:
Sorry we weren’t able to provide timely feedback here @guidiego. Closing as it’s now a bit out of date and I’m not yet confident we want to document all these details in Starlight’s docs.

PR #814, opened 2023-10-06
fix: prevent text from overflowing prev/next links

julien-deramond:

- Changes to Starlight code

#### Before the final merge if/when approved

- [x] Remove https://github.com/withastro/starlight/pull/814/commits/2cd86e9d6c40e2f4ddf6323fee3fd3874a17d65f which gathers the extra long texts added for testing purposes
  - ⚠️ This commit breaks pa11y-ci job temporarily

#### Description

This PR is a follow-up of https://github.com/withastro/starlight/pull/756 where we decided to extract this fix for prev/next links in a separate PR. We agreed at the time that the suggested fix wasn't good enough because relying on media queries and values that don't exist already as media queries in Starlight (see discussion from https://github.com/withastro/starlight/pull/756#discussion_r1337584797).

A possible simple fix would have been to use media containers because what we really want is a specific rendering based on the size of the container but the browsers' support is not yet good enough.

So I had to rely on some advanced features of `display: grid` often used to create complex layouts.

The following code already existed and worked well for the rendering of each pagination item.
```css
a {
  flex-basis: calc(50% - 0.5rem);
  flex-grow: 1;
}
```

Since we are playing with auto layouts and sizes, the icon size can sometimes be reduced (or the icon can "disappear"), so the following allows us to keep the same size for the icon:

```css
svg {
  flex-shrink: 0
}
```

The most complex part is definitely the following:

```css
.pagination-links {
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
  grid-gap: 1rem;
}
```

The `grid-template-columns` property defines the columns of the grid. The `repeat(auto-fit, minmax(200px, 1fr))` value means that the grid will automatically create as many columns as it can fit in the available space, with a minimum width of 200px and a maximum width of 1fr (which means the columns will take up equal space).
If I say it differently, the pagination item will take the whole space until there's enough space to render two 200px pagination items on the same row.

The `grid-gap` property sets the gap between the columns.

#### Tested environments

In terms of browsers' support, it seems to be safe to use:
- [`display: grid`](https://caniuse.com/?search=display%3A%20grid)
- [`grid-template-columns`](https://caniuse.com/?search=grid-template-columns)
- [`repeat()`](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#browser_compatibility)
- [ `minmax()`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax#browser_compatibility)

So I tested only the oldest combinations of browsers/versions listed in [our compatibility browsers list](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+88%2C+Edge+%3E%3D+88%2C+Firefox+%3E%3D+98%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all)):

- Firefox 118.0.1 / macOS 14.1 (my env by default)
- Chrome 88 / Windows 8
- Edge 88 / Windows 8
- iPhone 12 Pro Max / Safari 16
  - ⚠️ I don't have any 15.4 env to test on BrowserStack, before 15.4, the layout is completely broken so untestable
- Firefox 98 / macOS 12.1

### DIFF
--- a/packages/starlight/components/Pagination.astro
+++ b/packages/starlight/components/Pagination.astro
@@ -9,7 +9,7 @@ const isRtl = dir === 'rtl';
 const t = useTranslations(locale);
 ---
 
-<div class="pagination-links sl-flex" dir={dir}>
+<div class="pagination-links" dir={dir}>
 	{
 		prev && (
 			<a href={prev.href} rel="prev">
@@ -38,7 +38,8 @@ const t = useTranslations(locale);
 
 <style>
 	.pagination-links {
-		flex-wrap: wrap;
+		display: grid;
+		grid-template-columns: repeat(auto-fit, minmax(min(18rem, 100%), 1fr));
 		gap: 1rem;
 	}
 
@@ -56,6 +57,7 @@ const t = useTranslations(locale);
 		text-decoration: none;
 		color: var(--sl-color-gray-2);
 		box-shadow: var(--sl-shadow-md);
+		overflow-wrap: anywhere;
 	}
 	[rel='next'] {
 		justify-content: end;
@@ -70,5 +72,9 @@ const t = useTranslations(locale);
 		color: var(--sl-color-white);
 		font-size: var(--sl-text-2xl);
 		line-height: var(--sl-line-height-headings);
+	}
+
+	svg {
+		flex-shrink: 0;
 	}
 </style>
--- a/.changeset/tasty-garlics-dress.md
+++ b/.changeset/tasty-garlics-dress.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Prevent text from overflowing pagination items


julien-deramond:
Tried another approach with https://github.com/withastro/starlight/pull/814/files#diff-a68adf20b69e527e32f37849b62cef1c5593b5f424d7e87d506ffa015dbca52c. Still need to check if it can work well cross browsers, versions, use cases, etc. But it doesn't use media queries, media containers, or JS so 🤞

packages/starlight/components/Pagination.astro
```diff
@@ -38,8 +38,9 @@ const t = useTranslations(locale);
 
 <style>
 	.pagination-links {
-		flex-wrap: wrap;
-		gap: 1rem;
+		display: grid;
+		grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
```
delucis, line 42: Played around a bit and I think a slightly higher value here could make sense (`18rem` is `288px`). I wondered if this should be linked to other variables (e.g. `calc(var(--sl-content-width) / 3)` or something), but I think the target size is independent of anything else — if someone sets a larger content width, they probably still want these links to shrink at smaller viewports and not break earlier. So I think setting a value here is fine. (Could also be pulled out into a variable in `props.css`, e.g. `--sl-min-pagination-link-width` or something.)

We should also wrap the target value in `min()` so that it never exceeds 100% of the container. (One quirk of grid layout is that `1fr` can be MORE than `100%` because in the case where there’s just one grid item, the “free” space is calculated based on our minimum size.)

```suggestion
		grid-template-columns: repeat(auto-fit, minmax(min(18rem, 100%), 1fr));
```

delucis (commented):
Love this solution! Thank you for cracking it @julien-deramond 💖 

Left a couple of small notes, but I think this is 100% the right solution.

packages/starlight/components/Pagination.astro
julien-deramond: > I wondered if this should be linked to other variables (e.g. calc(var(--sl-content-width) / 3) or something), but I think the target size is independent of anything else

Yep, I tried to find some logic based on other custom properties as well but couldn't find anything related.

> We should also wrap the target value in min() so that it never exceeds 100% of the container. (One quirk of grid layout is that 1fr can be MORE than 100% because in the case where there’s just one grid item, the “free” space is calculated based on our minimum size.)

Oh yeah, good catch! Let's do that!

packages/starlight/components/Pagination.astro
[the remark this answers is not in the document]
julien-deramond: Done via https://github.com/withastro/starlight/pull/814/commits/69d6f17be7aa829f4082ff0a6f152c5862454f14

julien-deramond:
* 🟢 Fork synced with the _main_ branch
* 🟢 Branch rebased onto _main_ branch, squashed, and is now without the extra commit
* 🤞 Waiting for the deployment to check the rendering, but should be good to go

delucis (approved):
Yay! Thanks again for your persistence figuring this out. Really appreciate it 🤩

PR #845, opened 2023-10-08
Adding showcase for "Some drops of PHP" HTML eBook

roberto-butti:

Adding showcase for "Some drops of PHP" HTML eBook:
- added Showcase Card
- added the thumbnail

#### Description

Adding showcase for "Some drops of PHP" HTML eBook

I joined also Hacktoberfest


HiDeoo (commented):
Great addition to the showcase, we actually showcased it during the last Astro Community Call too, thanks a lot for this! 

Would you be able to update the thumbnail image to be 800 × 450 to match the dimensions of the other thumbnails in the showcase?

roberto-butti:
Thank you @HiDeoo .
I updated the thumbnail dimensions.

delucis (approved):
Thanks for adding this @roberto-butti!

PR #855, opened 2023-10-10
[i18nIgnore] i18n(zh-cn): Fix Chinese subtag

liruifengv:

- Changes to Starlight code

#### Description

- Fix Chinese subtag
- Url: zh -> zh-cn
- Built-in Translations: zh -> zh-CN
- All link /zh/ -> /zh-cn/
- Add redirect: /zh/*    /zh-cn/:splat


liruifengv:
related PR: https://github.com/withastro/starlight/pull/475
discuss: https://discord.com/channels/830184174198718474/1160994417470079138

Genteure:
Some thoughts:

- This change would be better to make after [existing i18n PRs to docs](https://github.com/withastro/starlight/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+i18n+zh) are merged.
- Links can be mass search-replaced in the same PR (`/zh/` -> `/zh-cn/`).
- A changeset is needed as this would be a breaking change.

docs/public/_redirects
```diff
@@ -1,2 +1,3 @@
 /ph/*        https://astro-houston-ph.pages.dev/ph/:splat  200
 /:lang/*    /:lang/404/   404
+/zh/*       /zh-cn/:splat  200
```
liruifengv, line 3: How to test the redirect?

Genteure: Redirect not working, getting HTTP 404 on this page: https://deploy-preview-855--astro-starlight.netlify.app/zh/getting-started/

liruifengv: I got it. It's order!

packages/starlight/translations/index.ts
```diff
@@ -15,7 +15,7 @@ import da from './da.json';
 import tr from './tr.json';
 import ar from './ar.json';
 import nb from './nb.json';
-import zh from './zh.json';
+import zhCN from './zh-CN.json';
```
delucis, line 18: As @Genteure suggested, we can probably stick with this being `zh` here. That will mean people already using `lang: 'zh'` can keep using this. And our system is able to pick this up anyway, so sites using `lang: 'zh-CN'` will also get this file.

```suggestion
import zh from './zh-CN.json';
```

packages/starlight/translations/index.ts
```diff
@@ -41,7 +41,7 @@ export default Object.fromEntries(
 		tr,
 		ar,
 		nb,
-		zh,
+		"zh-CN": zhCN,
```
delucis, line 44:
```suggestion
		zh,
```

delucis (commented):
Thanks for updating this @liruifengv!

Just one comment on what we do with our built-in translation dictionary.

Genteure:
@delucis Do you think the changeset is still needed, or it can be removed since this ended up mostly changing starlight's own docs? I suggested adding a changeset because I originally thought this would be a breaking change for existing users with Chinese docs in their project.

packages/starlight/translations/zh-CN.json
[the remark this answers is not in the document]
liruifengv: I think it's great, too.

delucis:
Great point @Genteure. Yes, this is now a 100% transparent change that doesn’t impact Starlight users really, so we can remove the changeset.

liruifengv:
> Great point @Genteure. Yes, this is now a 100% transparent change that doesn’t impact Starlight users really, so we can remove the changeset.

I removed it. It doesn’t impact existing users because we kept the built-in UI as zh

delucis (approved):
Great — I think this PR is ready to merge then. Thanks for all the work on this!

PR #862, opened 2023-10-11
docs: update wrong description

jsparkdev:

#### Description

- Among the built-in components, `SocialLinks` do not exist, but `SocialIcons` do.
- Also, looking at line 52, 69 of the code, it looks like it's correct to use `SocialIcons`.
- I am participating in Hacktoberfest, please label it (hacktoberfest-accepted)


HiDeoo (approved):

delucis (commented):
Ah, great catch — thank you @jsparkdev!

Although in general we don’t recommend it, I think this is a case where updating the other languages that translated this page would make sense to avoid needing multiple PRs just to fix this typo. Would you like to make this change in those files too?

- https://github.com/withastro/starlight/blob/main/docs/src/content/docs/es/guides/overriding-components.md
- https://github.com/withastro/starlight/blob/main/docs/src/content/docs/fr/guides/overriding-components.md
- https://github.com/withastro/starlight/blob/main/docs/src/content/docs/ko/guides/overriding-components.md

delucis (approved):
Thanks @jsparkdev!

PR #869, opened 2023-10-11
docs: refactor social links configuration reference

HiDeoo:

#### Description

As [discussed](https://discord.com/channels/830184174198718474/1070481941863878697/1161674105729319062) on Discord, this pull request refactors the social links section of the Customizing Starlight guide by removing the complete list of social links and replacing it with a link to the configuration reference where the type is now dynamically generated.

This means that PRs adding new social links will no longer require follow-up i18n PRs.

Notes: 

- We cannot apply the same strategy used in #688 to apply this change to all translated pages because this requires a new sentence to be added to `guides/customization.mdx`.
- I guess it may be better to merge this after #854.


docs/src/content/docs/reference/configuration.mdx
```diff
@@ -208,7 +208,7 @@ interface BadgeConfig {
 
 ### `locales`
 
-**type:** <code>{ \[dir: string\]: [LocaleConfig](#localeconfig) }</code>
+**type:** <code>\{ \[dir: string\]: [LocaleConfig](#localeconfig) \}</code>
```
HiDeoo, line 211: Change required due to the MDX conversion to avoid interpreting this as an expression.

docs/src/components/social-links-type.astro
```diff
@@ -0,0 +1,7 @@
+---
+import { socialLinks } from '../../../packages/starlight/schemas/social';
+
+const socials = [...socialLinks].sort((a, b) => a.localeCompare(b, 'en'));
```
HiDeoo, line 4: Relying on `localeCompare()` in the compare function instead of a basic `sort()` to ensure a case-insensitive sort so that `codeberg` appears before `codePen` in the list for example.

docs/src/content/docs/guides/customization.mdx
```diff
@@ -191,7 +191,7 @@ defineConfig({
 
 Starlight has built-in support for adding links to your social media accounts to the site header via the [`social`](/reference/configuration/#social) option in the Starlight integration.
 
-Currently, links to Bitbucket, Codeberg, CodePen, Discord, Email, Facebook, GitHub, GitLab, Gitter, Instagram, LinkedIn, Mastodon, Microsoft Teams, an RSS feed, Stack Overflow, Telegram, Threads, Twitch, Twitter, X, and Youtube are supported.
+You can find a full list of supported social medias in the [Configuration Reference](/reference/configuration/#social).
```
delucis, line 194: I think I might use wording something like this? They’re not all strictly speaking social media I guess (at least e-mail and RSS feel like they stretch that definition), so something more generic probably makes sense:

```suggestion
You can find a full list of supported link icons in the [Configuration Reference](/reference/configuration/#social).
```

docs/src/components/social-links-type.astro
delucis: Looks good. Initially, I wondered if we should sort with the current page’s locale instead of always `en`, but on second thoughts, given this list’s contents, this makes sense 👍

delucis (commented):
Looks good @HiDeoo! I’ll also try to make sure we merge any other pending translations of the two impacted pages ahead of this.

docs/src/components/social-links-type.astro
```diff
@@ -0,0 +1,7 @@
...
+---
+
+<code>{`Partial<Record<${socials.map((social) => `'${social}'`).join(' | ')}, string>>`}</code>
```
kevinzunigacuellar, line 7: This component is very creative!

HiDeoo:
Updated the PR with the changes of #892 & #854.

delucis (approved):
LGTM! Going to merge this while the coast is clear and we don't have too many other translation PRs just yet. Thanks again @HiDeoo for the great idea.

PR #901, opened 2023-10-14
feat: Add support for target and rel attribute for hero action links

timokoessler:

- Changes to Starlight code

#### Description

This pull request changes the CallToAction component so that the target and rel attribute can be set.

### DIFF
--- a/packages/starlight/CHANGELOG.md
+++ b/packages/starlight/CHANGELOG.md
@@ -84,7 +84,7 @@
 
   ```css
   .sl-link-card a {
-  	line-height: 1.6;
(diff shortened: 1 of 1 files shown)


delucis (changes requested):
Thank you for the PR @timokoessler! I don’t think we would want to start adding all potential HTML attributes as properties here.

Instead, I would suggest copying the approach used for sidebar links with an `attrs` prop: https://starlight.astro.build/guides/sidebar/#custom-html-attributes

Would you be interested in updating this PR to use that approach?

timokoessler:
I will create a new clean pull request.

PR #902, opened 2023-10-14
view transitions

rishi-raj-jain:

- Changes to Starlight code
- Something else!

#### Description

- Closes # 

### DIFF
--- a/.changeset/quick-comics-promise.md
+++ b/.changeset/quick-comics-promise.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+added view transitions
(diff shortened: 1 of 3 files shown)


HiDeoo (commented):
Thanks for your contribution.

I am not sure to understand how to test this change. The PR adds a new `viewTransitions` Starlight configuration option (defined in your `astro.config.mjs` file) in the `UserConfigSchema` but reads the `viewTransitions` frontmatter field of a page from `Astro.props.entry.data` so I don't see how to enable the new option.

There may be something to fix here to either always use a frontmatter field (`packages/starlight/schema.ts`) or always use a Starlight configuration option (`packages/starlight/utils/user-config.ts`), or both?

If we were to provide a new convenience option to enable view transitions (it is already possible to use view transitions using component overrides), I think there might also be some other considerations to take into account:

- Should this option be overridable on a per-page basis through frontmatter too?
- Should we include an option to easily set `transition:animate="none"` on the page’s `<html>` element too? Having full control over your page animation on specific elements is a powerful feature but having an option to quickly enable view transitions and having to rely on component overrides to disable them for the rest of the page feels a bit weird (considering that adding an attribute to `html` is not the easiest thing to do with component overrides).

delucis:
This is also a duplicate of #694, so we can probably close and discuss there.

PR #910, opened 2023-10-16
i18n(id): translate `guides/css-and-tailwind.mdx`

oggnimodd:

#### Description

- Translate `guides/css-and-tailwind.mdx` to Indonesian

### DIFF
--- a/docs/src/content/docs/id/guides/css-and-tailwind.mdx
+++ b/docs/src/content/docs/id/guides/css-and-tailwind.mdx
@@ -259,15 +259,15 @@ import ThemeDesigner from '../../../../components/theme-designer.astro';
 			random: 'Random',
 		},
 		editor: {
-			accentColor: 'Accent',
+			accentColor: 'Aksen',
 			grayColor: 'Gray',
 			hue: 'Hue',
 			chroma: 'Chroma',
-			pickColor: 'Pick color',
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/id/guides/css-and-tailwind.mdx
```diff
@@ -0,0 +1,287 @@
...
+<ThemeDesigner
+	labels={{
+		presets: {
+			label: 'Presets',
+			ocean: 'Ocean',
+			forest: 'Forest',
+			oxide: 'Oxide',
+			nebula: 'Nebula',
+			default: 'Default',
+			random: 'Random',
```
delucis, line 259: Noticing that these haven’t been translated. Was that intentional? In other languages we generally have translated these.

delucis (commented):
Thank you for the translation @oggnimodd! Just noticed one small thing, but otherwise looks great 🙌

docs/src/content/docs/id/guides/css-and-tailwind.mdx
oggnimodd: Thanks for catching that, I missed those

delucis (approved):
Great — thank you for making that change. Ready to launch 🚀

PR #930, opened 2023-10-19
Fix translation status builder to get the correct URL for the docs repo

liruifengv:

- Something else!

#### Description

- Fix translation status builder to get the correct URL for the docs repo


delucis (approved):
Ah, good catch — thanks for fixing this @liruifengv! 🚀

PR #933, opened 2023-10-19
i18n(zh-cn): Update i18n.mdx & Update customization.mdx

liruifengv:

#### Description

- Update i18n.mdx
- Update customization.mdx


liruifengv:
Maybe build-in support languages in `i18n.mdx` also can be written as a component, like the social component.

Because we support more and more languages.

HiDeoo:
> Maybe build-in support languages in `i18n.mdx` also can be written as a component, like the social component.

This is an interesting idea.

I guess we could maybe use [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) to get proper language display names based on the current page locale and [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) to get a listing properly formatted for the current locale, e.g. something like:

```js
// Available at build time
const currentPageLang = "fr";
// Generated based on the list of supported languages
const supportedLangs = ["ar", "en", "fr", "zh-CN"];

const langNames = new Intl.DisplayNames([currentPageLang], {
  type: "language",
});

const langList = new Intl.ListFormat(currentPageLang, {
  style: "long",
  type: "conjunction",
}).format(supportedLangs.map((supportedLang) => langNames.of(supportedLang)));
```

This would give for example:

- In English:  `Arabic, English, French, and Chinese (China)`
- In French:  `arabe, anglais, français et chinois (Chine)`
- In Chinese:  `阿拉伯语、英语、法语和中文（中国）`

delucis (approved):
LGTM — thanks @liruifengv!

@HiDeoo Would definitely be interested to see that in a PR! I’d actually considered if it was possible before but dismissed it as not possible because of the need to translate the language names — didn’t know about that part of `Intl`. (Not for the first time you surprise me with `Intl` magic — I should probably brush up on it.)

PR #938, opened 2023-10-20
Document remark/rehype

### LINKED ISSUE #937: Height and width of Embed element are not correct.

0.11.1

v3.3.2

pnpm

Windows

Edge

### Describe the Bug

The height and width are not correct with an embedded YouTube.

In this screen you can see height and width in my code differ from actual height and width.

image

https://stackblitz.com/edit/github-nfsqau?file=src%2Fcontent%2Fdocs%2Fguides%2Fexample.md

### Participation

kevinzunigacuellar:

- Closes #937

#### Description

- Only applies `height: auto` to elements (iframe, img, ...) when height its not set.
- Added `height: auto` to showcase card


docs/src/content/docs/reference/test.mdx
```diff
@@ -0,0 +1,13 @@
...
+title: Testing assets
+description: A test
+---
+
+An image with a height of 100px:
+<img src="https://raw.githubusercontent.com/withastro/docs/main/public/default-og-image.png" height="100px" alt="default og " />
+
+An image without a height:
+<img src="https://raw.githubusercontent.com/withastro/docs/main/public/default-og-image.png" alt="default og" />
+
```
delucis, line 11: Can we also try this with a local Markdown image? Astro’s assets supports will add a `height` automatically I think:

```suggestion

A local image, automatically processed by Astro assets support:
![Local image](../../../assets/showcase/felicity.pages.dev.png)

```

delucis (commented):
Thanks @kevinzunigacuellar!

Not 100% sure if this approach will work or not. For example, not super obvious with the test example of 100px, but shrink the viewport narrow enough and you’ll see this distortion:

image in content that is stretched because its height is 100px even though its width has shrunk with its narrow container

This would be more obvious for larger image sizes.

docs/src/content/docs/reference/test.mdx
kevinzunigacuellar: sure!

delucis: Thanks! As I suspected this is not looking so lovely 😅 

A rather squished image

kevinzunigacuellar:
I am not very happy with this solution either 😓

HiDeoo:
> I am not very happy with this solution either 😓

Could we move this to a rehype plugin?

delucis:
> Could we move this to a rehype plugin?

Yeah I also considered this. I’d honestly love it if Starlight added zero custom rehype plugins to give users a clean slate for their own customisations. Maybe a good opportunity for a third-party plugin that we could recommend?

HiDeoo:
> I’d honestly love it if Starlight added zero custom rehype plugins to give users a clean slate for their own customisations. Maybe a good opportunity for a third-party plugin that we could recommend?

I think that is a valid point and wise idea 👍 

Tried quickly right now to see if something like that already existed, I'm a bit surprised it is not (or I did not find it easily) but would be fairly easy to setup (and yeah, definitely without cheerio).

kevinzunigacuellar:
I think a rehype plugin it's a good idea but what about the support for markdoc files?

HiDeoo:
> I think a rehype plugin it's a good idea but what about the support for markdoc files?

Does Markdoc even allow HTML in markup? I think users would need a custom Markdoc tag for an iframe so the tag could also handle this case at the same time I guess?

docs/src/components/showcase-card.astro
```diff
@@ -56,5 +56,6 @@ const src = (await thumbnail()).default;
 	.image img {
 		aspect-ratio: 16 / 9;
 		object-fit: cover;
+		height: auto;
```
kevinzunigacuellar, line 59:
```suggestion
```

kevinzunigacuellar:
To get the ball rolling I added a small section to the end of the authoring content page about plugins.

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -207,3 +207,7 @@ Long, single-line code blocks should not wrap. They should horizontally scroll i
 ## Other common Markdown features
 
 Starlight supports all other Markdown authoring syntax, such as lists and tables. See the [Markdown Cheat Sheet from The Markdown Guide](https://www.markdownguide.org/cheat-sheet/) for a quick overview of all the Markdown syntax elements.
+
+## Advanced Markdown and MDX Cofiguration
+
+Starlight uses Remark built-in to Astro to parse Markdown and MDX files. You can configure the parser to support additional syntax and plugins by adding `remarkPlugins` or `rehypePlugins` array to your Astro config file. See the [Astro documentation](https://docs.astro.build/en/guides/markdown-content/#configuring-markdown-and-mdx) for more details.
```
delucis, line 213: Great idea!

Small rework here:
- Our headings use sentence case, so lower case “configuration”
- Tweaked some of the explanation to flow better
- Updated the link text to be more descriptive of where the link goes

```suggestion
## Advanced Markdown and MDX configuration

Starlight uses Astro’s Markdown and MDX renderer built on remark and rehype. You can add support for custom syntax and behavior by adding `remarkPlugins` or `rehypePlugins` in your Astro config file. See [“Configuring Markdown and MDX”](https://docs.astro.build/en/guides/markdown-content/#configuring-markdown-and-mdx) in the Astro docs to learn more.
```

delucis (approved):
I think this looks good to me! Do you want to do anything else in this PR @kevinzunigacuellar?

ElianCodes (approved):
Love it! Thanks for adding this @kevinzunigacuellar!

kevinzunigacuellar:
(Pressing the green big button...)

PR #940, opened 2023-10-21
i18n(pt-BR): Update `customization.mdx`, `i18n.mdx`, `index.mdx`, `frontmatter.md` and `configuration.mdx` 

vasfvitor:
- Rename `configuration.md` to `configuration.mdx` and update its contents.
- Updates: `customization.mdx`, `guides/i18n.mdx`, `index.mdx` and `reference/frontmatter.md`

### DIFF
--- a/docs/src/content/docs/pt-br/reference/configuration.mdx
+++ b/docs/src/content/docs/pt-br/reference/configuration.mdx
@@ -157,19 +157,19 @@ Se o seu site é multilíngue, a `label` de cada item é considerada como estand
 
 ```js
 sidebar: [
-  // um exemplo de barra lateral com rótulos traduzidos para o Francês.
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/pt-br/reference/configuration.mdx
HiDeoo, line 1: I think this may be missing the changes from #881 which uses a BCP-47 language tag with a regional subtag for the "Translating labels" section example.

As the example translations are using `pt-BR`, in this case, I think we could use `zh-CN`? Any thoughts?

vasfvitor: agree completely, I believe I saw that but didn't think it through

delucis (approved):
Thanks for the updates @vasfvitor and welcome to the Starlight contributors 🥳

Issue #861, opened 2023-10-11
Component overrides feedback


HiDeoo:
This issue contains various early feedback regarding the new component overrides features after using it for various cases in various projects and integrations from overriding, extending and creating new routes.

# Overriding and extending

This is the main feature/use case (hence the name ^^) and I think it works really well, very easy to use while still being very flexible. I managed to always be able to override what I needed and I guess there may be at some point some edge cases not found yet but this should be fairly easy to handle when they come up.

## Types

I am adding this feedback in this issue but I am not yet sure yet this is directly related to component overrides.

Starting from `v0.11.0`, even tho everything looks fine in an editor like VSC, I can no longer type check my integrations using a basic `tsc --noEmit` command. For some reasons, it now tries to type check Starlight code in `node_modules` and report errors. I guess a potential reason could be the `exports`/`types` shenanigans we had to use to type components but I am not sure yet as I have used this syntax before without issues (but not with non-existing files so maybe that's the reason).

I will `git bisect` and play with [`traceResolution`](https://www.typescriptlang.org/tsconfig#traceResolution) to figure it out, just didn't have the time yet.

## Configuration

In a custom component, especially if shared between projects or coming from a third-party integration, it can be useful to read the Starlight configuration, e.g. to check if the `editLink` is enabled or not, get the default locale, etc.

Of course, we can import `import starlightConfig from 'virtual:starlight/user-config'` but this may be a bit of an unusual import for users, a bit of an implementation detail which is not documenated and we do not provide types for this. This means that for example in most of my integrations, I had to create a `starlight.d.ts` file to provide types for this import:

```ts
declare module "virtual:starlight/user-config" {
  const Config: import("@astrojs/starlight/types").StarlightConfig;

  export default Config;
}
```

This could also be done with the `/// <reference path="../node_modules/@astrojs/starlight/virtual.d.ts" />` syntax but it relies on `node_modules` which is not ideal.

I am wondering if we should not export a function, e.g. `getConfig` or something like that maybe through `@astrojs/starlight/config`, which would be a bit more explicit and easier to type/use for this use case?

## Props types

As `Props` depends on `CollectionEntry<'docs'>`, if you are using it in an integration for example that does not have type for this collection, you get basically no type infos for `Astro.props.entry`. This is a bit annoying as you basically either have to check the source code or see at runtime what you are missing. I am not sure tho there is a lot we can do about this one as the only solution would require some hard-coded types which is not ideal.

# Creating new routes

[...]

Issue #753, opened 2023-09-24
Starlight plugins proposal


HiDeoo:
A plugin system for Starlight is something that has been discussed for a while on Discord. This issue is a proposal to discuss the potential design of such a system, explain why it could be useful, provide a few examples of what it could look like, and discuss / gather feedback on different APIs that could be exposed to plugins.

Feel free to comment on this issue to share your thoughts and ideas.

## Motivation

Starlight is built on top of Astro which already supports [Astro integrations](https://docs.astro.build/en/guides/integrations-guide/) to add new functionalities to a project using Astro (Starlight is actually an Astro integration).

Starlight plugins are not meant to replace Astro integrations but rather build on top of them to provide additional functionalities that are specific to Starlight. A common use case that has already been requested multiple times is to be able to read the Starlight configuration from another integration to customize its behavior based on it. This is something that is not possible at the moment and that could be achieved with a plugin system.

## Usage

Plugins can be added to a Starlight configuration using a new `plugins` option. This option would accept an array of plugins just like Astro integrations do.

```ts
// astro.config.mjs
import { defineConfig } from 'astro/config';
import starlight from '@astrojs/starlight';
import myStarlightPlugin from 'my-starlight-plugin';

export default defineConfig({
  integrations: [
    await starlight({
      title: 'My Docs',
      plugins: [myStarlightPlugin()],
    }),
  ],
});
```

## APIs

A Starlight plugin requires a `name` and a `plugin` function that will be called by Starlight with an object containing various helpers that can be used to interact with Starlight. This function can also optionally return an Astro integration that will be automatically used.

### `config` & `updateConfig()`

A basic plugin that would require reading & updating the Starlight configuration could be similar to [starlight-typedoc](https://github.com/HiDeoo/starlight-typedoc) which generates some Markdown pages on disk, creates some sidebar navigation items for them and updates the Starlight sidebar configuration to display them.

```ts
function starlightTypeDocPlugin(options: StarlightTypeDocOptions) {
  const starlightTypeDoc: StarlightPlugin = {
    name: 'starlight-typedoc',
    async plugin({ config, updateConfig }) {
      // Generate the Markdown pages on disk.
      const pages = await generateTypeDocPages(options);
      // Create a new sidebar group for the generated pages.
      const sidebarGroup = getSidebarGroupFromPages(pages);
      // Add the new sidebar group to the Starlight configuration.
      updateConfig({
        sidebar: [...(config.sidebar ?? []), sidebarGroup],
      });
[...]

PR #942, opened 2023-10-23
Initial Starlight Plugins support

HiDeoo:

- Changes to Starlight code

#### Description

- Implements various features of ~~#753~~#962.

This pull request is the beginning of the implementation of a Starlight Plugin API. A lot of details and discussions are available in ~~#753~~ #962.

The following features are implemented in this first PR:

- `config`
- `updateConfig()`
- `addIntegration()`
- `logger`

I tried to comment various implementation details, specially around Zod, like why the plugin API is using a totally different schema than the configuration one, why the configuration and configuration updates flowing through the plugin API are not validated (but still properly typed for user convenience) and only the final configuration is validated, etc. but if you have any questions/suggestions, please let me know.

#### Plugin Example

This PR adds a temporary `@astrojs/starlight-search-demo` plugin that demonstrates the various features of the plugin API and see how it can be used.

This example is obviously a reference to a potential `starlight-algolia` kind of plugin that has been requested multiple times on Discord and would also be needed for the Astro docs site. The idea was to see the code required to implement such a plugin and having a custom search bar by only adding the following lines to the `astro.config.mjs` file:

```js
import starlightSearchDemo from "@astrojs/starlight-search-demo";

starlight({
  plugins: [starlightSearchDemo({ apiKey: "123456" })],
});
```

Note that this plugin is only meant for demo purposes. It does nothing and even tho we briefly discussed the idea of having official plugins and if we ever decide to make an official Algolia plugin, it feels like this should be done in a separate PR or maybe even a separate repository with all expected features from such a plugin (proper CSS styling, pagefind disabled, tests, etc.).

_Note: I think my personal important takeaway from this example is that the `injectSettings()` API discussed in #753 (and potentially in #861 too for some API changes) could be really useful._

#### Documentation

This PR adds a "Plugins Reference" page inspired by the [Astro Integration API](https://docs.astro.build/en/reference/integrations-reference/) documentation page.
I'm not sure if a guide is needed for the plugin API at this point, as the "use a plugin" section would be pretty much a repeat of the `plugins` section of the Configuration Reference and may change between plugins and a "create a plugin" section, which I think does not exist for the Astro Integration API, would be pretty much merging the various examples from the "Plugins Reference" page.

I also suggested a change to the showcase to add a new section dedicated for plugins so it can be linked from various places. Obviously, we would need to wait for the first plugin to be published before doing that if we decide to go that way.

#### Remaining tasks

These tasks should be completed before merging this PR:

- [x] Remove the Starlight Search Demo plugin in its current state
- [x] Remove the Starlight Search Demo plugin from the docs configuration file
- [x] Remove all `TODO(HiDeoo)` references.
- [x] Add a changeset.


delucis:
Super exciting to see this! Will make sure to get you a thorough review ASAP.

packages/starlight/utils/plugins.ts
```diff
@@ -0,0 +1,143 @@
...
+		throw new Error(
+			'Invalid plugins config passed to starlight integration\n' +
+				parsedPluginsConfig.error.issues.map((i) => i.message).join('\n')
+		);
+	}
+
+	let userConfig = config;
+	const integrations: AstroIntegration[] = [];
+
+	for (const { name, plugin } of parsedPluginsConfig.data.plugins) {
```
delucis, line 28: Should we support processing plugins added by preceding plugins? If so, maybe we need something like this?

```suggestion
	for (let i = 0; i < userConfig.plugins.length; i++) {
		const { name, plugin } = userConfig.plugins[i];
```

packages/starlight/utils/plugins.ts
```diff
@@ -0,0 +1,143 @@
...
+	hooks: z.object({}).passthrough().default({}),
+}) as z.Schema<AstroIntegration>;
+
+/**
+ * A plugin `config` and `updateConfig` argument are purposely not validated using the Starlight
+ * user config schema but properly typed for user convenience because we do not want to run any of
+ * the Zod `transform`s used in the user config schema when running plugins.
+ * The final user config should be validated by the caller after running all the plugins.
+ */
+const starlightPluginSchema = z.object({
```
delucis, line 56: I’m not sure I follow why we can’t include `plugins` in the main user config schema? Would you mind explaining a bit more?

delucis (commented):
Super exciting work @HiDeoo! Left a few initial thoughts and questions.

docs/src/content/docs/reference/plugins.md
```diff
@@ -0,0 +1,99 @@
...
+```ts
+interface StarlightPlugin {
+  name: string;
+  plugin: (options: {
+    config: StarlightUserConfig;
+    updateConfig: (newConfig: StarlightUserConfig) => void;
+    addIntegration: (integration: AstroIntegration) => void;
+    logger: AstroIntegrationLogger;
+  }) => void | Promise<void>;
+}
```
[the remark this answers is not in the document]
HiDeoo: I was never a huge fan of the `plugin()` method name but did not managed at the time to come up with a better name I liked. I personally like `setup()` so I'm all for it.

Regarding the structure, I personally do not have a strong preference, so I think if you have one, we should go with it otherwise I guess mirroring Astro's structure may be the safest bet?

packages/starlight/index.ts
```diff
@@ -7,25 +7,29 @@ import { starlightAsides } from './integrations/asides';
...
-	const parsedConfig = StarlightConfigSchema.safeParse(opts, { errorMap });
+export default function StarlightIntegration(
+	opts: StarlightUserConfigWithPlugins
+): AstroIntegration {
+	return {
+		name: '@astrojs/starlight',
+		hooks: {
+			'astro:config:setup': async ({ config, injectRoute, logger, updateConfig }) => {
+				const { integrations, userConfig } = await runPlugins(opts, logger);
+				const parsedConfig = StarlightConfigSchema.safeParse(userConfig, { errorMap });
```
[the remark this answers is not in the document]
HiDeoo: > Would be nice to add a comment regarding the order here.

Sure, I'll try to describe it better.

> I’m also wondering if we want to parse twice so we can error nicely before plugins run. I’m guessing currently we’d have the risk that a user provides bad config that causes some error in a plugin that Zod could catch but the user would see an error thrown by the plugin?

We definitely could extract the validation part in a function and run it before and after and it could indeed improve the experience when dealing with an error.

packages/starlight/utils/plugins.ts
HiDeoo: I thought about this during the initial prototype and decided to leave it out for now. My reasoning (and maybe wrong) was that it would maybe be more "natural"/explicit to add a `addPlugin` like `addIntegration` considering that we only want to support adding plugins but not remove/re-order them while running them which I think could be less obvious with a helper called `updateConfig`.

If we decide to allow this in `updateConfig`, we would indeed need to tweak the syntax like you suggested and also extract the updated plugins config from the updated Starlight config and also re-run the plugin validation obviously.

packages/starlight/utils/plugins.ts
HiDeoo: Is this question related to the highlighted comment or more of a general question?

If this is directly relateed to the comment, I think my comment is not obvious enough. I try to explain why we are not using something like:

```js
updateConfig: z.function(
  z.tuple([UserConfigSchema]),
  //          ^ this is the problem described in the comment
  z.void()
)
```

If this is not directly to the comment, this would be technically possible to add the plugins to the user config schema. The reason I did not do that are:

- There is __at the moment__ in the code only 1 place (the Starlight integration input) where we need a type that contains a config with the plugins. 
- Past that point, we only need the config without the plugins (when passing it down to plugins, when a plugin updates the config, for passing it to the Vite plugin, etc.)
- I did not want to re-validate the plugins config when re-validating the user config after the plugins have run.
- I thought it would be easier to use a TypeScript union once vs omitting the plugin from the user config schema with the Zod API to get the same result.

This could definitely be revisited if you think it would be better to have the plugins in the user config schema.

docs/src/content/docs/reference/plugins.md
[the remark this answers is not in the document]
delucis: No strong preference from me, so yeah, +1 to matching Astro’s shape 👍

packages/starlight/utils/plugins.ts
delucis: That does make sense to me.

I will share some thoughts from the discussion of when we enabled integrations to add integrations in Astro’s `updateConfig()`. Initially I made an `addIntegration()` helper, but Blu pointed out that ideally `updateConfig()` works for all config options so a user doesn’t have to know what they can and can’t use that hook for (otherwise you risk proliferating `updateSidebar()`, `updateHead()` etc. for any more complex option).

I think the issue in Starlight’s case is that Astro’s [`mergeConfig()`](https://github.com/withastro/astro/blob/262cef2487c7494bd8d23b4ab27bfcdf1870a111/packages/astro/src/core/config/merge.ts) is generally able to not worry about order, so merging `{ option: [something] }` and `updateConfig({ option: [somethingElse] })` as `{ option: [something, somethingElse] }` is safe (and also nicer for the dev to not worry about spreading all the right things). But in Starlight I can totally imagine wanting to do stuff like prepend or filter sidebar items etc., which the `mergeConfig()` approach prevents. So I think I agree with forcing plugin developers to merge config manually even if this does place some trust in them to not break user config.

`plugins` is the only problematic field here right? I wonder if we should error if a plugin tries to mutate it?

packages/starlight/utils/plugins.ts
```diff
@@ -0,0 +1,143 @@
...
+	/** Name of the Starlight plugin. */
+	name: z.string(),
+	/**
+	 * Plugin function called with an object containing various values that can be used by the
+	 * plugin to interact with Starlight.
+	 */
+	plugin: z.function(
```
delucis, line 63: Would it make sense to also pass some [data from the `astro:config:setup`](https://docs.astro.build/en/reference/integrations-reference/#astroconfigsetup) hook here? I think it’s OK to require adding an integration to use any of the functions, but it would probably be helpful for plugins to access [`config`](https://docs.astro.build/en/reference/integrations-reference/#config-option), [`command`](https://docs.astro.build/en/reference/integrations-reference/#command-option), and [`isRestart`](https://docs.astro.build/en/reference/integrations-reference/#isrestart-option). 

Example use cases:
- check if an integration is already in the Astro config before adding it
- only run some initialisation code on dev/build start, but not when the dev server restarts

packages/starlight/utils/plugins.ts
delucis: > Past that point, we only need the config without the plugins (when passing it down to plugins, when a plugin updates the config, for passing it to the Vite plugin, etc.)

Wouldn’t other plugins want to know what plugins they are configured alongside potentially?

Makes sense for the Vite plugin though, where functions would be stripped during serialization.

packages/starlight/utils/plugins.ts
HiDeoo: > … Blu pointed out that ideally `updateConfig()` works for all config options so a user doesn’t have to know what they can and can’t use that hook for …

This is a very good point. 

> I think the issue in Starlight’s case is that Astro’s `mergeConfig()` is generally able to not worry about order

My initial implementation actually had a `mergeConfig()` which was a slimmed down version of Astro’s `mergeConfig()` but as you mentioned, this is not ideal in the case a Starlight.

> `plugins` is the only problematic field here right? 

This is indeed the only one.

> I wonder if we should error if a plugin tries to mutate it?

This would already happen when validating the config after the plugins have run but maybe you mean adding an explicit check in the `updateConfig()` hook?

packages/starlight/utils/plugins.ts
HiDeoo: > Wouldn’t other plugins want to know what plugins they are configured alongside potentially?

I didn't think of this but it's a very good point.

So I guess a potential solution would be:

- Add `plugins` to the user config schema and refactor the various types/calls to follow this change
- Pass down the `plugins` too to the provided `config` to a plugin
- As per your comment above, check if `plugins` is being mutated in the `updateConfig()` hook and error if so

Is this something that looks good to you?

packages/starlight/utils/plugins.ts
delucis: Yeah, because otherwise you could be in the weird situation of using `addIntegration()` to add an integration that reads `config` and _then_ itself adds an integration based on that, which is pretty clunky. (The use case I have in mind is the Algolia plugin, which could add Preact or React depending on what a user is already using.)

packages/starlight/utils/plugins.ts
delucis: Yeah, sounds good to me!

packages/starlight/utils/plugins.ts
delucis: > This would already happen when validating the config after the plugins have run but maybe you mean adding an explicit check in the `updateConfig()` hook?

I think you answered this in https://github.com/withastro/starlight/pull/942#discussion_r1378878069 but validating in `updateConfig()` or after would both be fine. I guess inside `updateConfig()` we might have the advantage of a nice error message that says which plugin was responsible.

HiDeoo:
Thanks a lot for the initial feedback and discussions.

I think all of the points raised are now actionable and I can get back to it and prepare an updated version including all the suggested changes 👍

HiDeoo:
> Add plugins to the user config schema

Turns out this one is a little bit "annoying" due to the recursive types (a user config including plugins which contains a function that pass down a user config that contains plugins…) and our tranforms.

I've setup a small [repro](https://tsplay.dev/WPMPZw) in the TS playground of the issue. I've found so far 1 workaround that I describe in the playgraound but imo this is pretty ugly and I'm not sure if it's even worth it.

I'll continue to explore this as I may be missing something obvious or a better way to do this but so far, no luck.

HiDeoo:
Just pushed a few changes relative to the first review:

- Updated the shape of a plugin to match the one from an Astro integration.
- The configuration passed down to a plugin now also contains the `plugins` list.
- Refactor the entire validation process:
  1. Before running plugins, the user-provided configuration is validated (we can provided an early feedback to the user)
  2. Then, the user-provided plugins are validated (we can provided an early feedback to the user and indicate that a plugin is not valid)
  3. The plugins are executed
     3.1. If a plugin updates the configuration, we check that it does not try to mutate the `plugins` list (we have a dedicated error message indicating which plugin is responsible for the error)
     3.2. If a plugin updates the configuration, we validate the new configuration (if the merged configuration is not valid, we have a dedicated error message indicating which plugin is responsible for the error)
  4. We return the final validated configuration and the list of Astro integrations to add
- I added a few comments to help understand the flow of the code regarding the validation process
- A plugin now receives a few new options (`astroConfig`, `command`, and `isRestart`)
- I updated the tests and documentation to reflect the changes
- I slightly updated the demo
  - Fix the color contrast issue ^^
  - I added a button using Preact to show how a plugin should properly test if an Astro integration is already present before adding it

If I did not miss anything, this should covers all the feedback from the first review except the one about adding `plugins` directly to the user config schema. As commented right above, I still have not a found a proper way to do this that works as expected in all cases but I will keep looking into it.

packages/starlight/utils/plugins.ts
```diff
@@ -0,0 +1,223 @@
...
+
+				// If the updated config is valid, keep track of both the user config and parsed config.
+				userConfig = mergedUserConfig;
+				starlightConfig = mergedConfig;
+			},
+			addIntegration(integration) {
+				// Collect any Astro integrations added by the plugin.
+				integrations.push(integration);
+			},
+			astroConfig: context.config,
```
delucis, line 72: I think as a plugin author I’d expect `astroConfig.integrations` here to include the updated values from previous plugins (or at least to have another way to check this, but here probably make most sense?)

Thoughts on doing something like this? On the one hand it’s slightly lying, but on the other hand, in practice this is what is happening or “will” happen once we update the Astro config after plugins run?

```suggestion
			astroConfig: { ...context.config, integrations: [...context.config.integration, ...integrations] },
```

Effectively mirrors what we do with the `updateConfig()` hook where each plugin is passed `userConfig` as mutated by the previous.

delucis (commented):
Took the liberty of resolving conflicts and updating this branch @HiDeoo! Here are a few comments on the latest version — I’ve looked at everything except docs so far.

What’s the state of the typing issue? Is it kind of resolved in the current code or still needs a solution?

packages/starlight/index.ts
```diff
@@ -37,34 +44,25 @@ export default function StarlightIntegration(opts: StarlightUserConfig): AstroIn
 					pattern: '[...slug]',
 					entryPoint: '@astrojs/starlight/index.astro',
 				});
-				const integrations: AstroIntegration[] = [];
 				if (!config.integrations.find(({ name }) => name === 'astro-expressive-code')) {
```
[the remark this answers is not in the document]
HiDeoo: Really good catch, and I even think this should actually check both 🤔 I updated the PR to do so.

packages/starlight/__tests__/test-config.ts
```diff
@@ -18,14 +21,29 @@ export function defineVitestConfig(
...
+
+class TestAstroIntegrationLogger {
+	options = {} as AstroIntegrationLogger['options'];
+	constructor(public label = 'test-integration-logger') {}
+	fork = (label: string) => new TestAstroIntegrationLogger(label);
+	info = () => undefined;
+	warn = () => undefined;
+	error = () => undefined;
+	debug = () => undefined;
+}
```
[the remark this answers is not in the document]
HiDeoo: This was actually where I started writing this code but unfortunately it seems to not be possible as `test-utils` cannot be imported in `test-config` because it imports `vitest` and would give the following error:

```
Error: Vitest failed to access its internal state.
```

docs/src/content/docs/showcase.mdx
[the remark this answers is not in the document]
HiDeoo: Definitely, I thought I mentioned that in the PR description in the documentation section but I might have forgotten to do so. This change was mostly a suggestion to see what we could do on that side of things and see if it could somehow bring ideas to the table.

This would need to be removed until we get published plugins (it's in the list of remaining tasks at the end of the PR description).

packages/starlight/utils/plugins.ts
HiDeoo: Good catch, and I think this is a good solution you’ve come up with. I implemented this and added a test for it.

packages/starlight/__tests__/test-config.ts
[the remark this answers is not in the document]
delucis: Ahhhh, got it. Wondered if there might be something like that. OK then I guess we can leave it like this for now. Other option would be a separate file like `test-plugin-utils` or something, but probably overkill for so little code.

delucis:
> I was continuing my work locally and did not care about updating it as it was not relevant to my current work.

Ooh, what are you planning? Was wondering if this might be getting towards the point where we could think about releasing it? What would you still like to do before it’s ready for a final review?

HiDeoo:
> Ooh, what are you planning? Was wondering if this might be getting towards the point where we could think about releasing it? What would you still like to do before it’s ready for a final review?

Oh sorry, I was not clear enough, I've been solely working on trying to make that refactor of the user-config schema working with `plugins` being part it. Nothing else.

Feature-wise, I think we are good compared to the initial plan (with potential future improvements as we discussed in the "RFC" issue) and that's also why I suggested in my last point that going with the current implementation without this refactor is a viable option _imo_.

delucis:
OK, great! Then let’s mark this as ready for review and move onto polishing it up. I’ll take a look at the docs in a bit 🙌

delucis (commented):
Here’s the docs review as promised! 🙌

HiDeoo:
- Went ahead and added the `test-plugin-utils` you suggested as I think it makes more sense.
- Added the suggested documentation changes.

A few notes:

- There is still an "easy" reference in `reference/plugins.md`. I think the context is a bit different from the one removed but I prefer to let you know and confirm if it's ok or not in case this was missed.
- I tried to rephrase the `updateConfig` sentence but I'm sure it's still not clear enough so feel free to ditch it.
- I updated the EC highlight for the `addIntegration` example which I think was wrong.

Thanks for the review and especially the comments, I hope to get better at this with time by understanding the reasoning behind the changes. Super helpful! :raised_hands:

docs/src/content/docs/reference/plugins.md
```diff
@@ -0,0 +1,152 @@
...
+export default {
+  name: 'add-twitter-plugin',
+  plugin({ config, updateConfig }) {
+    updateConfig({
+      social: {
+        ...config.social,
+        twitter: 'astrodotbuild',
+      },
+    });
+  },
```
delucis, line 76: Totally missed this on my first read through, but just realising the examples all still use the old shape 😅
````suggestion
```ts {6-11}
// plugin.ts
export default {
  name: 'add-twitter-plugin',
  hooks: {
    setup({ config, updateConfig }) {
      updateConfig({
        social: {
          ...config.social,
          twitter: 'astrodotbuild',
        },
      });
    },
  },
````

File: docs/src/content/docs/reference/plugins.md [hunk 10]
```diff
@@ -0,0 +1,152 @@
...
+  plugin({ addIntegration, astroConfig }) {
+    const isReactLoaded = astroConfig.integrations.find(
+      ({ name }) => name === '@astrojs/react'
+    );
+
+    // Only add the React integration if it's not already loaded.
+    if (!isReactLoaded) {
+      addIntegration(react());
+    }
+  },
```
delucis, line 102:
````suggestion
```ts {14}
// plugin.ts
import react from '@astrojs/react';

export default {
  name: 'plugin-using-react',
  hooks: {
    plugin({ addIntegration, astroConfig }) {
      const isReactLoaded = astroConfig.integrations.find(
        ({ name }) => name === '@astrojs/react'
      );

      // Only add the React integration if it's not already loaded.
      if (!isReactLoaded) {
        addIntegration(react());
      }
    },
  },
````

File: docs/src/content/docs/reference/plugins.md [hunk 11]
```diff
@@ -0,0 +1,152 @@
...
+All logged messages will be prefixed with the plugin name.
+
+```ts {5}
+// plugin.ts
+export default {
+  name: 'long-process-plugin',
+  plugin({ logger }) {
+    logger.info('Starting long process…');
+    // Some long process…
+  },
```
delucis, line 144:
````suggestion
```ts {6}
// plugin.ts
export default {
  name: 'long-process-plugin',
  hooks: {
    plugin({ logger }) {
      logger.info('Starting long process…');
      // Some long process…
    },
  },
````

File: docs/src/content/docs/reference/plugins.md (hunk 9 above)
Comment (HiDeoo) [reply docs/src/content/docs/reference/plugins.md]: Damn, good catch, can't believe I missed that so many times too 😨

**Comment by delucis** [comment]
@HiDeoo do you want to take care of cleaning up the outstanding `TODO`s or would you like a hand getting this ready for release?

**Comment by HiDeoo** [comment]
> @HiDeoo do you want to take care of cleaning up the outstanding `TODO`s or would you like a hand getting this ready for release?

If you think it's ok, I can indeed get started on cleaning the PR when I get the time.

The changeset is probably the one thing where I might need help/review but I'll push one and we can see how it goes.

**Comment by delucis** [comment]
Sounds good, thank you! Yes, we can do a final docs and changeset review once those are in.

**Comment by HiDeoo** [comment]

- Remove demo plugin and its usage in the docs
- Revert the showcase page to a pre-plugin version
- Remove plugin showcase references from `reference/configuration.mdx` & `reference/plugins.md`
- Search and remove all `TODO(HiDeoo)` references
- Add changeset (not sure what level of detail is expected here)

File: docs/src/content/docs/reference/plugins.md [hunk 12]
```diff
@@ -0,0 +1,158 @@
...
+};
+```
+
+#### `addIntegration`
+
+**type:** `(integration: AstroIntegration) => void`
+
+A callback function to add an [Astro integration](https://docs.astro.build/en/reference/integrations-reference/) required by the plugin.
+For example, to add the [React Astro integration](https://docs.astro.build/en/guides/integrations-guide/react/):
```
delucis, line 87: Small suggestion to add the context around first checking before registering a new integration (which I think is a good practice for most integrations):

```suggestion

In the following example, the plugin first checks if [Astro’s React integration](https://docs.astro.build/en/guides/integrations-guide/react/) is configured and, if it isn’t, uses `addIntegration()` to add it:
```

docs/src/content/docs/reference/plugins.md
```diff
@@ -0,0 +1,158 @@
...
+
+```ts {14}
```
delucis, line 89: Maybe makes sense to also highlight the `addIntegration` in the plugin options?

````suggestion
```ts {14} "addIntegration,"
````

File: docs/src/content/docs/reference/plugins.md [hunk 14]
```diff
@@ -0,0 +1,158 @@
...
+  },
+};
+```
+
+#### `astroConfig`
+
+**type:** `AstroConfig`
+
+A read-only copy of the user-supplied [Astro configuration](https://docs.astro.build/en/reference/configuration-reference/).
+This configuration is resolved before any other integrations have run.
```
delucis, line 115: Is this 100% accurate? I’m pretty sure Starlight receives the config as transformed by any preceding integrations. If it _is_ accurate, maybe should be “Astro integrations” for clarity.

delucis (commented):
Really excited we’re getting so close!

I’ve left a last pass of comments on the docs & changeset, but then I think we should be good to go. 🚀

docs/src/content/docs/reference/plugins.md
HiDeoo: Good question, this is totally based on https://docs.astro.build/en/reference/integrations-reference/#config-option :thinking:

HiDeoo: Should we just remove the sentence then?

PS: not sure why everything is red now, looks like it's related to the size check, I'll try to see why 🤔

delucis:
Huh, not sure why CI started failing 🤔

docs/src/content/docs/reference/plugins.md
delucis: Yes, I think removing the sentence is fine.

Yeah, not sure why either — looks like the docs site is failing to build all of a sudden.

delucis:
The CI failure is very odd — just ran the build with no issues locally.

HiDeoo:
> The CI failure is very odd — just ran the build with no issues locally.

Same, working fine locally 🤔

HiDeoo:
Yeah I'm a bit confused as there is no `userConfig` anymore in the `index.ts` file anymore

delucis:
Merged and fixed that to see if it helps!

delucis:
Looks like that worked? I guess somehow there was some weird code in the CI job?

delucis (approved):
STAMP OF APPROVAL! ✅

HiDeoo:
> Looks like that worked? I guess somehow there was some weird code in the CI job?

Yeah, I was afraid to try this, see it works and have no clue on why 🤣 I do not have a better explanation ^^

delucis:
Well, here goes nothing… 🤞

PR #945, opened 2023-10-23
Use GitHub Discussions for feature requests

delucis:

- Something else!

#### Description

- Converts the Feature Request issue template into a template for the Feature Requests category in Discussions
- Updates our new issue config so it links to discussions for feature requests


HiDeoo (approved):
I don't think there is a way to properly check out the form before this is merged, but from the docs, this looks good to me.

PR #988, opened 2023-10-25
Add social icons to mobile menu footer

### LINKED ISSUE #978: Doc template on mobile unable to access social menu

0.11.1

0.11.1

npm

Windows

Firefox/Chrome

### Describe the Bug

In the basic example directly, there is no way to access on mobile the social links on a documentation page (docs template). The menu shows at the end of the navigatio the theme (and the language), but the social icons aren't. This is probably related to https://github.com/withastro/starlight/issues/977 

I haven't mark checkbox to participate with a PR, but I am willing to do so if I have some guidance as I am not familiar with the codebase (nor too much with javascript yet, although I have experience on other programming languages and software development in general).

https://stackblitz.com/github/withastro/starlight/tree/main/examples/basics

### Participation

magicDGS:
_NOTE: taking part in hacktoberfest_

- Changes to Starlight code

#### Description

- Closes #978
- Adds the SocialIcons component with  some styling to the MobileMenuFooter

##### Visual changes

- **Before**:
imagen
- **After**:
imagen


HiDeoo:
Thanks for the pull request :raised_hands: 

On a purely visual level, it looks like in the preview shared in #978, the `<ThemeSelect/>` component is always next to the `<LanguageSelect/>` one whereas now, it seems we get equal spacing between the 3 components. I'm not sure if that's intentional or not, but I thought I'd mention it.

I think the entire `mobile-preferences` could use some gap between the components too:

SCR-20231025-tsdc

I am also not sure what should be the wrapping behavior here, it looks there is no wrapping at all at the moment.

packages/starlight/components/MobileMenuFooter.astro
```diff
@@ -1,14 +1,21 @@
...
+	</div>
 	<ThemeSelect {...Astro.props} />
 	<LanguageSelect {...Astro.props} />
 </div>
 
 <style>
+	.social-icons {
+		gap: 1rem;
+		align-items: center;
+	}
```
delucis, line 18: We should probably also hide this wrapper element when there are no icons so that it only pushes the select menus to the right when there are actually some icons:

```suggestion
	}
	.social-icons:empty {
		display: none;
	}
```

delucis (changes requested):
Thanks @magicDGS! A few suggestions here to make this more closely match our design mock-up.

magicDGS:
Thanks a lot for the review and suggestions, @delucis! I added all of them an now looks much better, even on a viewpoint of 280px with a lot of icons:

imagen

Back to you for review!

packages/starlight/components/MobileMenuFooter.astro
```diff
@@ -1,17 +1,30 @@
...
+	</div>
 	<ThemeSelect {...Astro.props} />
 	<LanguageSelect {...Astro.props} />
 </div>
 
 <style>
+	.social-icons {
+		margin-inline-end: auto;
+		gap: 1rem;
+		align-items: center;
```
delucis, line 18: Looking good!

I noticed one issue when it wraps (thanks for sharing the screenshot):

social icons, language and theme selectors inside Starlight’s mobile menu

This is because when inline with other items, we center the icons but they’re actually not as tall as the select menus so the line collapses in height when they’re the only thing there. Not 100% sure whether this is the correct fix but here’s a quick way to solve it:

```suggestion
		align-items: center;
		padding-block: 1rem;
```

delucis (approved):
Thanks again for taking the initiative with this one @magicDGS! I took care of those last few details so we can get this out in our next release 🌟

magicDGS:
Thanks a lot for merging the PR. I am soglad that I have contributed and I hope to keep doing it in the future, as I am planning to adopt starlight for one of my current personal projects even if not yet on v1.0.0.

PR #991, opened 2023-10-26
i18n(de): fix type import

HiDeoo:

#### Description

This PR fixes a type import in the `/de/reference/overrides.md` file which seems to also [totally break](https://github.com/withastro/starlight/actions/runs/6652578458/job/18076846039) the `format` job.

I tested locally by running the `pnpm format` command before and after the change (not sure why this error leads to error in other files tho 🤔 )


delucis:
Oops, sorry didn’t see this and pushed a quick fix to `main` 😬 

I’ll take this PR as an implicit approval of my reckless commit 😁

PR #995, opened 2023-10-26
feat: extend badge to be available for group links

kevinzunigacuellar:

- Changes to Starlight code

#### Description

- Extend badges to be available for group links.


docs/src/content/docs/reference/configuration.mdx
```diff
@@ -183,6 +183,7 @@ sidebar: [
 type SidebarItem = {
 	label: string;
 	translations?: Record<string, string>;
+	badge?: string | BadgeConfig;
```
HiDeoo, line 186: I cannot suggest a change for this but I think the `badge?: string | BadgeConfig;` right below this one can be removed now.

HiDeoo (commented):
More badges :tada: Amazing.

I left a few questions mostly regarding the documentation and I have some visual feedback as well.

I am not sure if this is expected or part of the initial design, but I think the gap between a label and a badge is different for a link and a group. I am not saying they should be the same, but I personally think the gap is too small for a group at least.

SCR-20231027-kcwl

I think there may be some wrapping behavior that is not handled yet compared to links:

SCR-20231027-kdpx

Last point, maybe we should also update a few tests to ensure in a snapshot we get badges for groups too?

Impatient to see this merged and add more badges to my documentations ^^

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -235,13 +235,14 @@ The `sidebar` frontmatter configuration is only used for autogenerated links and
 
 ## Badges
 
-Links can also include a `badge` property to display a badge next to the link label.
+Links and groups can also include a `badge` property to display a badge next to their label.
```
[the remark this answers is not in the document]
kevinzunigacuellar: Thank, you for the feedback @HiDeoo. Yes, I completely forgot about autogenerated groups. For some reason, I thought they were in the same section as groups but it makes sense for them to have their own section since they have a lot more configurable options.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -235,13 +235,14 @@ The `sidebar` frontmatter configuration is only used for autogenerated links and
...
 
 ```js
 starlight({
 	sidebar: [
 		{
 			label: 'Guides',
+			badge: 'Badge',
```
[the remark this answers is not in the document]
kevinzunigacuellar: Yes, I think this example would be easier to read too. I don't think that the autogenerate option works on the preview componenent but I can manually add some entries to the group.

docs/src/content/docs/reference/configuration.mdx
kevinzunigacuellar: Completely missed that one. Great catch!

kevinzunigacuellar:
One last thing that I would like to get some input. I noticed that the spacing for multi line groups is a bit wide. This gets more noticeable with a badge. Should we decrease the line height for groups?

image

HiDeoo:
> One last thing that I would like to get some input. I noticed that the spacing for multi line groups could get a bit wide. This gets more noticeable with a badge. Should we decrease the line height for groups? or does this looks good?

Indeed, this is something I noticed in the past, and I'm pretty sure we talked about this somewhere else but I cannot find where. I am no designer but for what it's worth, I personally think it could be slightly decreased. Not sure what others think though.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -79,6 +85,8 @@ interface Props {
 		align-items: center;
 		justify-content: space-between;
 		padding-inline: var(--sl-sidebar-item-padding-inline);
+		padding-block-end: 0.3em;
+		line-height: 1.5;
```
delucis, line 89: Nice — thanks @kevinzunigacuellar!

Played around with it a bit and this looks closest to the current styles to me. Preserves an overall spacing of 1.8 (for single-line items), tightens up the line-height for items that wrap, and distributes the extra padding to the top and bottom.

```suggestion
		padding-block: 0.2em;
		line-height: 1.4;
```

| Before either of our changes | After applying this suggestion |
|---|---|
| image | image |

kevinzunigacuellar: Thanks for taking the time to tweak this. Looks perfect on my end

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -235,7 +235,7 @@ The `sidebar` frontmatter configuration is only used for autogenerated links and
 
 ## Badges
 
-Links can also include a `badge` property to display a badge next to the link label.
+Links, groups and autogenerated groups can also include a `badge` property to display a badge next to their label.
```
delucis, line 238: We usually use Oxford commas, so this list needs a comma before the and:

```suggestion
Links, groups, and autogenerated groups can also include a `badge` property to display a badge next to their label.
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -79,6 +85,8 @@ interface Props {
 		align-items: center;
 		justify-content: space-between;
 		padding-inline: var(--sl-sidebar-item-padding-inline);
+		padding-block: 0.2em;
```
delucis, line 88: Nit, but now I see we could collapse these and use the shorthand property:

```suggestion
		padding: 0.2em var(--sl-sidebar-item-padding-inline);
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -120,8 +128,9 @@ interface Props {
 		background-color: var(--sl-color-text-accent);
 	}
 
-	a > *:not(:last-child) {
-		margin-inline-end: 0.25em;
+	a > *:not(:last-child),
+	summary > div.group-label > *:not(:last-child) {
```
delucis, line 132: I think this should work just as well without the extra specificity?

```suggestion
	.group-label > *:not(:last-child) {
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -120,8 +128,9 @@ interface Props {
...
+		margin-inline-end: 0.5em;
```
delucis, line 133: Do we need to increase this spacing? I think the previous value should work OK as long as we reintroduce the whitespace between text and badge.

```suggestion
		margin-inline-end: 0.25em;
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -23,7 +23,6 @@ interface Props {
 						<span>{entry.label}</span>
 						{entry.badge && (
 							<>
-								{' '}
 								<Badge
```
delucis, line 26: Suggestion to reintroduce whitespace here:

```suggestion
								{' '}
								<Badge
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -36,7 +35,14 @@ interface Props {
 						open={flattenSidebar(entry.entries).some((i) => i.isCurrent) || !entry.collapsed}
 					>
 						<summary>
-							<span class="large">{entry.label}</span>
+							<div class="group-label">
+								<span class="large">{entry.label}</span>
+								{entry.badge && (
+									<>
+										<Badge text={entry.badge.text} variant={entry.badge.variant} />
```
delucis, line 42: And then we can also include it here. (Never done this whitespace hack in this kind of context before, but I think it should work — it’s required due to https://github.com/withastro/compiler/issues/852)

```suggestion
										{' '}
										<Badge text={entry.badge.text} variant={entry.badge.variant} />
```

delucis (changes requested):
Yay! I think this is in pretty good shape — thanks @kevinzunigacuellar.

Few more small details, to clean this up.

packages/starlight/components/SidebarSublist.astro
kevinzunigacuellar: This was intentional, I couldn't get the whitespace to work using `{" "}`. it's quite unpredictable. I think it's best to rely more in css for these situations.

kevinzunigacuellar: Oh nevermind, it worked this time 😅.

.changeset/orange-pants-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Extend badges to be available for group links
```
delucis, line 5: What do you think of tweaking this like this?

```suggestion
Adds support for adding sidebar badges to group headings
```

delucis (approved):
Really great work on this one @kevinzunigacuellar. Left one last note to polish up the changeset, but then LGTM! 🌟

kevinzunigacuellar:
Thanks for all the feedback @HiDeoo and @delucis!

delucis:
Oops, I guess we better cut a release then :-D

Usually we wait before merging changes like this because they immediately update docs even before we release them. I totally did not make that clear with my “LGTM” comment though, sorry 😅 

Let me take a look through the open PRs to see if there’s some other stuff we can get out with this.

PR #1003, opened 2023-10-28
Refactor translation system to be reusable in non-Astro code

delucis:

- Changes to Starlight code

#### Description

- Refactors the existing Starlight translations system so it can be instantiated from either `getCollection()` calls (for modules loaded by Astro) or file system calls (for modules running before Astro starts).
- The file-system version makes it possible for work like #742 to re-use our existing logic.
- It will also help us unblock #517.
- This PR instantiates the file-system version in the main Starlight integration. In follow-up PRs we’d be able to pass `useTranslations` to code that needs them (i.e. the Expressive Code integration and the asides plugin)

#### To-do

- [x] Write some tests for the new code

#### Background

This is needed because our current i18n system depends on two bits of code that are only available after Astro starts: `getCollection()` imported from the `astro:content` virtual module is used to load the `i18n` data collection and get any user-configured JSON dictionaries of strings and Starlight user config imported from `virtual:starlight/user-config`.

These dependencies prevented pre-Astro code like remark/rehype plugins or other integration code from accessing our translation system. So far, when trying to support i18n in those cases, we either got stuck or ended up making a copy of our i18n logic that didn’t cover all cases. This refactor helps make sure we share as much logic as possible between the code that runs before and after Astro starts.


packages/starlight/index.ts
```diff
@@ -26,6 +27,7 @@ export default function StarlightIntegration(opts: StarlightUserConfig): AstroIn
 		name: '@astrojs/starlight',
 		hooks: {
 			'astro:config:setup': ({ config, injectRoute, updateConfig }) => {
+				const useTranslations = createTranslationSystemFromFs(userConfig, config);
```
delucis, line 30: This variable is not in use yet, but would be passed into anything that needs it, such as the Starlight integration added in #742.

packages/starlight/utils/createTranslationSystem.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+function localeToLang(
+	locale: string | undefined,
+	locales: StarlightConfig['locales'],
+	defaultLocale: StarlightConfig['defaultLocale']
+): string {
+	const lang = locale ? locales?.[locale]?.lang : locales?.root?.lang;
+	const defaultLang = defaultLocale?.lang || defaultLocale?.locale;
+	return lang || defaultLang || 'en';
+}
```
delucis, line 63: I’ve duplicated this function for now from `slugs.ts`:

https://github.com/withastro/starlight/blob/ba7063556a2fb82b177861d0a612e867643aac0c/packages/starlight/utils/slugs.ts#L31-L39

We could refactor in a similar spirit to make this version the base and then have a module that exports the easier to use methods that depend on `virtual:starlight/user-config`.

packages/starlight/utils/translations.ts
delucis, line 1: Our base `useTranslations()` method is unchanged but is refactored to use the new `createTranslationSystem()` factory, passing in files loaded with `getCollection()`.

packages/starlight/utils/translations-fs.ts
delucis, line 1: This wraps our `createTranslationSystem()` factory with the logic to load user dictionaries using `fs` calls. This is pretty cavalier about error handling currently as users will get content collection validation on their i18n data collection entries.

Might still be a good idea to add _some_ error handling here?

PR #1009, opened 2023-11-01
i18n(pt-BR): Update `css-and-tailwind.mdx`, `authoring-content.md` and `overrides.md`

vasfvitor:
#### Description

- Update `css-and-tailwind.mdx`, `authoring-content.md` and `overrides.md`.
- Improve wording on `overrides.md`.

`overrides.md` isn't showing up in the i18n tracker because it was pushed just minutes before the english version of `overrides.md` being updated, then iirc, ci format kicked in (on the translation) and the tracker considered as update.

Thank you


docs/src/content/docs/pt-br/guides/authoring-content.md
```diff
@@ -207,3 +207,7 @@ Longos blocos de código de linha única não devem quebrar linha. Eles devem ro
 ## Outras funcionalidades comuns do Markdown
 
 Starlight suporta todo o resto da sintaxe de escrita do Markdown, como listas e tabelas. Veja a [Cheat Sheet de Markdown do The Markdown Guide](https://www.markdownguide.org/cheat-sheet/) para uma visão geral rápida de todos os elementos da sintaxe do Markdown.
+
+## Markdown Avançado e Configurando MDX 
+
+O Starlight utiliza o mesmo rendizador Markdown e MDX do Astro, que suporta plugins remark e rehype. Você pode adicionar sintaxe personalizada e _behavior_ adicionando `remarkPlugins` ou `rehypePlugins` no seu arquivo de configuração Astro. Visite [Configurando Markdown e MDX](https://docs.astro.build/pt-br/guides/markdown-content/#configurando-markdown-e-mdx) na documentação do Astro para saber mais.
```
HiDeoo, line 213: - The first sentence seems to slightly differ from the English version which does not mention plugins, is that expected?
- I assume the `_behavior_` is on purpose and not something missed?

HiDeoo (commented):
Thanks for the translations :raised_hands:

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view) and everything looks pretty good to me. I only have a small question regarding the `authoring-content.md` file if you would like to take a look just to double check everything is correct.

docs/src/content/docs/pt-br/guides/authoring-content.md
vasfvitor: 1 - It is not, I'll fix it.

2 - About _behavior_ it is because while it can be translated (_comportamento_), on the [linked Astro docs ](https://docs.astro.build/en/guides/markdown-content/#customizing-a-plugin) page the only instance is in context of an option, so my question is whether it should be translated, this was intentional and I expected it to be discussed:
image
image


this - _[...] adicionar sintaxe personalizada e **behavior**..._
becomes - _[...] adicionar sintaxe e **comportamento** personalizado..._

I'm not sure how the word `behavior` is seen by native speakers in the context of Markdown. 

Correct me if I'm wrong: I imagine that in the current context it is referring to both to the actual behavior but also to the option called `behavior`. If it were to translate it would only refer to the actual behavior, on the other hand, if isn't translated, it would refer only to the option called `behavior`. As I see, either way something is going to be lost.

So my case is that in the same linked docs there's no mention of _comportamento_ and I can imagine someone reading Starlight docs, clicks to Astro docs and CTRL + F "_comportamento_" and doesn't find anything. But if they search for 'behavior' they'd right on the money. What is your opinion?

vasfvitor: note: If it is to not translated behavior, I think it's better to make it `behavior` and not italicized. So I wait what's best to make the necessary change.

HiDeoo: Thanks for the update and the very thorough explanation. I understand your train of thought, really appreciate the effort you put into this :raised_hands:

From my understanding, the Astro docs only reference an option named `behavior` while the Starlight docs reference an actual behavior but considering that English is not my native language either, I think that in this specific case, it might be better like you suggested to wait for Chris or Yan to confirm this before making any changes.

delucis: Yes, I can confirm that in the Starlight context we mean “behavior” in the general sense not the specific option to that remark plugin, so I’d recommend translating it 👍

vasfvitor: done 👍

delucis (approved):
Thanks @vasfvitor and thanks for spotting these details in your review @HiDeoo! 🌟

PR #1014, opened 2023-11-02

jsparkdev:

#### Description

- update `sidebar.mdx`
- The badge's content, `Deprecated` was not translated because I felt it would be more natural to do so.


delucis (approved):
Thanks @jsparkdev — happy to trust your instincts on whether “Deprecated” should be translated in this context or not. 🙌

PR #1023, opened 2023-11-02
fix: handle `build.format` links in sidebar

### LINKED ISSUE #180: Setting Astro `build.format` config breaks sidebar links

kevinzunigacuellar:

- Changes to Starlight code

#### Description

- Closes #180
- Implemented `formatPath` function to adjust path formats according to project configurations.

- Defined conventions for formatting paths:
  - For `build.format: "file"`, the path will conclude with a ".html" extension. For instance, "/reference/" will transform into "/reference.html".
  - For `build.format: "directory"`, the path will include a trailing slash. For instance, "/reference.html" will change to "/reference/".


packages/starlight/integrations/virtual-user-config.ts
```diff
@@ -10,15 +10,15 @@ function resolveVirtualModuleId<T extends string>(id: T): `\0${T}` {
...
+	{ root, srcDir, build }: Pick<AstroConfig, 'root' | 'srcDir' | 'build' >
 ): NonNullable<ViteUserConfig['plugins']>[number] {
 	const resolveId = (id: string) =>
 		JSON.stringify(id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
 
 	/** Map of virtual module names to their code contents as strings. */
 	const modules = {
 		'virtual:starlight/user-config': `export default ${JSON.stringify(opts)}`,
-		'virtual:starlight/project-context': `export default ${JSON.stringify({ root, srcDir })}`,
+		'virtual:starlight/project-context': `export default ${JSON.stringify({ root, srcDir, buildFormat: build.format })}`,
```
delucis, line 21: I’d be tempted to keep this mirroring Astro’s config shape 1:1:

```suggestion
		'virtual:starlight/project-context': `export default ${JSON.stringify({ root, srcDir, build: { format: build.format } })}`,
```

packages/starlight/virtual.d.ts
```diff
@@ -3,7 +3,7 @@ declare module 'virtual:starlight/user-config' {
 	export default Config;
 }
 declare module 'virtual:starlight/project-context' {
-	export default { root: string, srcDir: string };
+	export default { root: string, srcDir: string, buildFormat: import('astro').AstroConfig['build']['format'] };
```
delucis, line 6:
```suggestion
	export default { root: string, srcDir: string, build: Pick<AstroConfig['build'], 'format'> };
```

delucis (commented):
Thanks @kevinzunigacuellar! I see `build.format: 'file'` in the docs to test, but looks like the sidebar in the deploy preview does not use `.html`. Any idea why?

packages/starlight/__tests__/test-config.ts
```diff
@@ -4,11 +4,26 @@ import { getViteConfig } from 'astro/config';
...
+export function defineVitestConfigBuildFormatFile(config: z.input<typeof StarlightConfigSchema>) {
+	const root = new URL('./', import.meta.url);
+	const srcDir = new URL('./src/', root);
+	const build = {
+		format: "file"
+	} as AstroConfig['build']
+	return getViteConfig({
+		plugins: [vitePluginStarlightUserConfig(StarlightConfigSchema.parse(config), { root, srcDir, build })],
 	});
 }
```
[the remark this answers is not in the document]
kevinzunigacuellar: Yes, I wasn't really sure how to test this without copying a directory full of tests. I will try this.

kevinzunigacuellar:
> I see build.format: 'file' in the docs to test, but looks like the sidebar in the deploy preview does not use .html. Any idea why?

Checking the build output it looks that all the href in the sidebar have a .html append to the end. I think it's one of netlify features to strip the .html from the deployed files ([Pretty URL’s](https://docs.netlify.com/configure-builds/file-based-configuration/#pretty-urls))

delucis:
> Checking the build output it looks that all the href in the sidebar have a .html append to the end. I think it's one of netlify features to strip the .html from the deployed files ([Pretty URL’s](https://docs.netlify.com/configure-builds/file-based-configuration/#pretty-urls))

Ahhhhh! Thank you Netlify 🖖

kevinzunigacuellar:
I had to make a few assertions for `AstroConfig['build']` because Partial only makes the top level properties optional and it expects a fully defined build property to be passed to the user config plugin.

packages/starlight/integrations/virtual-user-config.ts
```diff
@@ -10,15 +10,15 @@ function resolveVirtualModuleId<T extends string>(id: T): `\0${T}` {
 /** Vite plugin that exposes Starlight user config and project context via virtual modules. */
 export function vitePluginStarlightUserConfig(
 	opts: StarlightConfig,
-	{ root, srcDir }: Pick<AstroConfig, 'root' | 'srcDir'>
+	{ root, srcDir, build }: Pick<AstroConfig, 'root' | 'srcDir' | 'build' >
```
delucis, line 13: Ah yeah, good point about the assertions. They’re not the end of the world, but maybe we can type more accurately what we actually need here, so we can avoid them:

```suggestion
	{ root, srcDir, build }: Pick<AstroConfig, 'root' | 'srcDir'> & { build: Pick<AstroConfig['build'], 'format'> }
```

packages/starlight/__tests__/test-config.ts
```diff
@@ -4,11 +4,19 @@ import { getViteConfig } from 'astro/config';
...
 import { StarlightConfigSchema } from '../utils/user-config';
+import type { AstroConfig } from 'astro';
+
+export function defineVitestConfig(
+	config: z.input<typeof StarlightConfigSchema>,
+	opts?: Partial<AstroConfig>
+) {
+	const root = opts?.root ?? new URL('./', import.meta.url);
+	const srcDir = opts?.srcDir ?? new URL('./src/', root);
+	const build = opts?.build ?? ({ format: 'directory' } as AstroConfig['build']);
```
delucis, line 15: Should allow us to set this without the assertion I think?

```suggestion
	const build = opts?.build ?? { format: 'directory' };
```

packages/starlight/__tests__/test-config.ts
```diff
@@ -4,11 +4,19 @@ import { getViteConfig } from 'astro/config';
 import type { z } from 'astro/zod';
 import { vitePluginStarlightUserConfig } from '../integrations/virtual-user-config';
 import { StarlightConfigSchema } from '../utils/user-config';
+import type { AstroConfig } from 'astro';
+
+export function defineVitestConfig(
+	config: z.input<typeof StarlightConfigSchema>,
+	opts?: Partial<AstroConfig>
```
delucis, line 11: And then same type here should avoid the assertions in test config:

```suggestion
	opts?: { build: Pick<AstroConfig['build'], 'format'> }
```

packages/starlight/__tests__/sidebar-file-build/vitest.config.ts
```diff
@@ -0,0 +1,52 @@
...
+			{
+				label: 'API v1',
+				autogenerate: { directory: '/api/v1/' },
+			},
+		],
+	},
+	{
+		build: {
+			format: 'file',
+		} as AstroConfig['build'],
```
delucis, line 50:
```suggestion
		},
```

packages/starlight/utils/navigation.ts
```diff
@@ -134,8 +135,19 @@ function makeLink(
...
+		return { type: 'link', label, href, isCurrent: false, badge, attrs: attrs ?? {} };
+
+	href = pathWithBase(href);
+	const isCurrent = href === ensureTrailingSlash(stripExtension(currentPathname));
+
+	if (project.build.format === 'file') {
+		// edge case for index.html
+		if (href === '/') href = '/index/';
+		href = stripTrailingSlash(href) + '.html';
+	}
```
delucis, line 149: Is there a way to extract some of this into a reusable function that formats any relative path according to Astro’s config?

We’ll need to do the same for the `<SiteTitle />` component which currently does this:

https://github.com/withastro/starlight/blob/5499c2524ca958c6a59581eaf6184e14c4fccecf/packages/starlight/components/SiteTitle.astro#L7

Would also make it easier to test because then we can just have a bunch of tests for that function passing different arguments and checking it returns what we expect.

delucis (commented):
Thanks for cleaning up that test config helper. I left a suggestion to avoid the type assertions and then one question.

packages/starlight/utils/navigation.ts
kevinzunigacuellar: I didn't know it was used somewhere else. Yes, I can give it a shot.

kevinzunigacuellar:
I made a function called `formatPathFromConfig` that formats any given path using the astro config options (base and build.format). I am not sure where to put it for now its just on navigation utils. And I wrote a few tests with different edge cases. What do you think?

delucis:
Nice work!

> I am not sure where to put it for now its just on navigation utils.

Might make sense to move it to its own file. We do have `path.ts` and `base.ts`, but those both avoid relying on the virtual imports.

The other option would be to refactor it slightly to take the required config as an argument and put it together in `path.ts`:

```ts
/** Add `.html` extension to a path. */
export function ensureHtmlExtension(path: string) {
	if (path.endsWith('.html')) return ensureLeadingSlash(path);
	path = stripLeadingAndTrailingSlashes(path);
	return path ? ensureLeadingSlash(path) + '.html' : '/index.html';
}

export function formatPath(href: string, format: 'file' | 'directory') {
	// atach base
	href = format === 'file' ? fileWithBase(href) : pathWithBase(href);
	// add html extension
	href = format === 'file' ? ensureHtmlExtension(href) : stripExtension(href);
	return href;
}
```

Then users could import it and pass the `format`:

```js
import project from 'virtual:starlight/project-context';
import { formatPath } from './path';

formatPath(path, project.build.format);
```

kevinzunigacuellar:
I've created a new file for the format function and added `trailingSlash` as an option, given its relevance to the project configuration.

There's some ambiguity regarding the expected behavior for `trailingSlash: "ignore"`. Should we establish a convention similar to the existing one for adding trailing slashes to links? How might this convention apply when the build format is set to "file," or should we disregard it entirely?

In my view, adopting a convention would simplify the comparison of values for "isCurrent" in the sidebar. The convention I propose is as follows:

- `format: "directory", trailingSlash: "ignore"`: Retain the trailing slash (similar to the existing setup).
  - `"/reference"` becomes `"/reference/"`
  - `"/reference.html"` becomes `"/reference/"`

- `format: "file", trailingSlash: "ignore"`: Omit the trailing slash.
  - `"/reference"` becomes `"/reference.html"`
  - `"/reference/"` becomes `"/reference.html"`

Establishing this convention would enhance consistency and aid in comparing values for "isCurrent" in the sidebar, offering a clearer and more predictable structure for the project. What do you think? @delucis @HiDeoo

packages/starlight/utils/format-path.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export function formatPath(href: string, opts?: FormatPathOptions) {
+	const options = {
+		format: opts?.format ?? 'directory',
+		trailingSlash: opts?.trailingSlash ?? 'ignore',
+	};
+
+	const formatStrategy = formatStrategies[options.format];
+	const trailingSlashStrategy = trailingSlashStrategies[options.trailingSlash];
```
delucis, line 35: Nit, but in general for situations like this we use defaults in the signature. My preference is mainly based on avoiding the ambiguity of having both `opts` and `options` in scope. With this way of assigning the defaults you have a single option to consume inside the function body.

```suggestion
export function formatPath(
	href: string,
	{ format = 'directory', trailingSlash = 'ignore' }: FormatPathOptions = {}
) {
	const formatStrategy = formatStrategies[format];
	const trailingSlashStrategy = trailingSlashStrategies[trailingSlash];
```

This also switches from the nullish coalescing `??` logic to logic equivalent to `||`, but I don’t think we want to support stuff like `format: ''` or `trailingSlash: 0`, so should be equivalent?

packages/starlight/utils/navigation.ts
```diff
@@ -341,4 +355,16 @@ function applyPrevNextLinkConfig(
...
+}
+
+/** Add '.html' extension to a path. */
+export function ensureHtmlExtension(path: string) {
+	path = stripTrailingSlash(path);
+	if (path.endsWith('.html')) return ensureLeadingSlash(path);
+
+	path = stripLeadingAndTrailingSlashes(path);
+	return path ? ensureLeadingSlash(path) + '.html' : '/index.html';
+}
```
delucis, line 370: What do you think of moving these to `utils/path.ts`? When they were only used by the navigation logic, keeping them in this file made sense, but now we’re sharing them also in `utils/format-path.ts`, I think that might be the more correct place for them

packages/starlight/utils/navigation.ts
```diff
@@ -134,8 +141,15 @@ function makeLink(
 	badge?: Badge,
 	attrs?: LinkHTMLAttributes
 ): Link {
-	if (!isAbsolute(href)) href = pathWithBase(href);
-	const isCurrent = href === ensureTrailingSlash(currentPathname);
+	href = formatPath(href, {
+		format: project.build.format,
+	});
```
delucis, line 146: Should we not preserve the previous behaviour and avoid formatting absolute URLs? I would guess we shouldn’t enforce our formatting on an external URL. We’re not currently testing any sidebar entries with `{ link: "https://example.com" }`, so may have missed changes in formatting here.

```suggestion
	if (!isAbsolute(href)) {
		href = formatPath(href, { format: project.build.format });
	}
```

packages/starlight/utils/navigation.ts
```diff
@@ -134,8 +141,15 @@ function makeLink(
...
+	currentPathname =
+		project.build.format === 'file'
+			? stripTrailingSlash(currentPathname)
+			: ensureTrailingSlash(currentPathname);
+	const isCurrent = isAbsolute(href) ? false : href === currentPathname;
```
delucis, line 151: Here on the other hand it should be safe to skip the `isAbsolute()` check, as absolute URLs will never match the pathname anyway:

```suggestion
	const isCurrent = href === currentPathname;
```

packages/starlight/utils/navigation.ts
```diff
@@ -134,8 +141,15 @@ function makeLink(
...
 ): Link {
-	if (!isAbsolute(href)) href = pathWithBase(href);
-	const isCurrent = href === ensureTrailingSlash(currentPathname);
+	href = formatPath(href, {
+		format: project.build.format,
+	});
+	currentPathname =
+		project.build.format === 'file'
+			? stripTrailingSlash(currentPathname)
+			: ensureTrailingSlash(currentPathname);
```
delucis, line 150: One question here: `currentPathname` is passed from `Astro.url.pathname`, does Astro provide this already formatted according to `build.format` and `trailingSlash` settings or would we need to call `formatPath(currentPathname)`?

delucis (commented):
Nice refactor @kevinzunigacuellar! Really like the maps of strategies you came up with in `utils/format-path`. Few last comments and questions from me!

packages/starlight/utils/navigation.ts
kevinzunigacuellar: Great question! I believe the reason it doesn't autoformat is to maintain continuity.

When someone requests `/base/reference.html/`, it forwards the URL exactly as provided.

You're correct that we might need to invoke `formatPath`. This ternary operator won't function when a trailing slash is specified. I might have to adjust `formatPath` to avoid adding the base path twice. The current implementation assumes that base is not present.

kevinzunigacuellar:
I may have pinpointed why the test "final parameter can set custom link even if no default link existed" is failing. It seems we're passing the same value to href and currentPath. 

https://github.com/withastro/starlight/blob/0dc4dc793080a96b4f21b72ace264ba6c39d6a17/packages/starlight/utils/navigation.ts#L336

I'm a bit unsure about how to address this—perhaps we could consider getting the current link as a parameter, although I'm not entirely convinced it's necessary. The scenario where this occurs is quite rare—only if someone uses the same link twice (current path name and prev/next link).

Another approach could be to make it optional or pass an empty string. 🤔 Any ideas?

delucis:
Hey @kevinzunigacuellar! Turned out this was fiddlier than I expected, so I spent some time getting everything polished up.

Quick summary of the changes I made:

- Wire up a project’s `trailingSlash` option with the virtual module and apply it when formatting paths.
- Move around some of the logic about when slashes are added/removed to ensure that with `trailingSlash: 'ignore'` (the Astro default) we don’t modify trailing slashes on user-specified links.
- Make sure we never add a trailing slash when `build.format` is set to `file` — I’m pretty sure `/example.html` should never be `/example.html/`.
- Fixed checks for whether a sidebar link was for the current page by always formatting the comparison links with a formatter that sets `trailingSlash: 'always'` so they will be consistent.
- Made the `formatPath()` helper easier to use in Starlight code by exporting it already configured with the project settings. Formatters with different settings can be created with the `createPathFormatter()` factory method.

Would love a quick look if you have time and thanks for the patience waiting for me on this one!

kevinzunigacuellar (commented):
Thanks @delucis for the review and changes. Yes, this PR started very simple but it got very complex very quickly 😅. I left a couple comments, feel free to disregard the nit comments btw.

packages/starlight/__tests__/basics/format-path.test.ts
```diff
@@ -0,0 +1,105 @@
...
+			{ path: '/api/v1/users', expected: '/api/v1/users' },
+			// with file extension
+			{ path: '/guides/components.html', expected: '/guides/components' },
+			// with file extension and trailing slash
+			{ path: '/guides/components.html/', expected: '/guides/components' },
+		],
+	},
+])(
+	'formatPath() with { format: $options.format, trailingSlash: $options.trailingSlash }',
+	({ options, tests }) => {
```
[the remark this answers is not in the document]
delucis: Not sure I understand what this means. Could you explain a little more?

packages/starlight/utils/createPathFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	// Handle extension
+	href = formatStrategy.handleExtension(href);
+
+	// Skip trailing slash handling for `build.format: 'file'`
+	if (format === 'file') return href;
+
+	// Handle trailing slash
+	href ||= '/';
+	href = href === '/' ? href : trailingSlashStrategy(href);
```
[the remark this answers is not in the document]
delucis: It’s hit in my test, although I guess the coverage may not recognise the two branches of the conditional assignment:

test coverage report

Don’t recall exactly, but I’m pretty sure I hit a case where this was needed.

packages/starlight/utils/createPathFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+	const trailingSlashStrategy = trailingSlashStrategies[trailingSlash];
+
+	// Add base
+	href = formatStrategy.addBase(href);
+
+	// Handle extension
+	href = formatStrategy.handleExtension(href);
+
+	// Skip trailing slash handling for `build.format: 'file'`
+	if (format === 'file') return href;
```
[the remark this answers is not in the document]
delucis: I don’t think so — if Astro lets you set that, then that’s on Astro 😄

packages/starlight/utils/createPathFormatter.ts
```diff
@@ -0,0 +1,58 @@
...
+	ensureHtmlExtension,
+	ensureTrailingSlash,
+	stripHtmlExtension,
+	stripTrailingSlash,
+} from './path';
+
+interface FormatPathOptions {
+	format?: AstroConfig['build']['format'];
+	trailingSlash?: AstroConfig['trailingSlash'];
+}
```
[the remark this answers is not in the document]
delucis: Ah, maybe. `base` is it a bit different because you access it with `import.meta.env.BASE_URL` instead of via our project context virtual module. I think we can leave it for now.

packages/starlight/utils/createPathFormatter.ts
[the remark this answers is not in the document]
kevinzunigacuellar: I see. I commented out line 50 when running unit tests and it seem pass all the test cases. Perhaps we should add a few more edge case for when it hits line 50.

packages/starlight/__tests__/basics/format-path.test.ts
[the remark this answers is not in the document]
delucis: Oh right. Yeah, this was cheating a bit just to avoid rewriting the tests too much. I think it’s fine though — it’s OK to do weird stuff in tests sometimes 😅

packages/starlight/__tests__/basics/format-path.test.ts
```diff
@@ -0,0 +1,105 @@
...
+			// index page
+			{ path: '/', expected: '/' },
+			// with trailing slash
+			{ path: '/reference/configuration/', expected: '/reference/configuration/' },
+			// without trailing slash
+			{ path: '/api/v1/users', expected: '/api/v1/users' },
+			// with file extension
+			{ path: '/guides/components.html', expected: '/guides/components' },
+			// with file extension and trailing slash
+			{ path: '/guides/components.html/', expected: '/guides/components' },
```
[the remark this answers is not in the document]
delucis: 🤔 I honestly don’t know… `/guides/components.html/` is already so totally badly formatted that it shouldn’t happen. I even debated if we should handle this case or only strip `.html` from paths without a trailing slash. Because on some hosts this would actually point to `/guides/components.html/index.html` 😅 

I think I’m happy with calling this one of those “bugs” we don’t have to fix until someone actually has a use case where this is an issue.

packages/starlight/utils/createPathFormatter.ts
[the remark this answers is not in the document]
delucis: Thanks for testing! Maybe I just added it out of fear. I’ll commit your suggestion and we can revisit it if anyone runs into issues 🙌

kevinzunigacuellar:
Lost Glasses, Troublesome Moment! ✅

PR #1025, opened 2023-11-02
fix(docs-i18n-tracker): update `translations` import

HiDeoo:

- Something else!

#### Description

This PR fixes an issue with the [Starlight Docs Translation Status](https://i18n.starlight.astro.build/). Following the changes in #1003, it now errors with the following:

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'starlight/packages/starlight/translations' is not supported resolving ES modules imported from starlight/packages/starlight/utils/createTranslationSystem.ts
```

_Note: I got this error locally after noticing that the tracker was not updating, I have no idea where to check the logs for the production one._

This is due to the Node import specifiers spec:

> Directory indexes (e.g. './startup/index.js') must also be fully specified.

And `tsm` [not handling](https://github.com/lukeed/tsm/issues/39#issuecomment-1205714134) this case specifically.

This PR specifies the `index` file in the import statement, which fixes the issue.

Another approach to avoid the change could be:

- Add `"type": "module"` to `docs-i18n-tracker/package.json`
- Replace `tsm` by [`tsx`](https://github.com/esbuild-kit/tsx) which [handles this case](https://github.com/esbuild-kit/tsx/blob/d132c3f37856149aa4b74c33a8b3af8e404c73a8/src/esm/loaders.ts#L234-L236)


HiDeoo:
> I think you should be able to see them, although not easily find them.

Ah thanks, I can indeed see them.

>  I guess hypothetically this could impact Starlight users, so we should probably add a patch changeset to release this fix?

Yeah, thought about that, but was not sure what to say or what the scenario could be (that's also why I included an alternative without the change ^^). I just pushed one with "Internal: fix import issue in translation string loading mechanism" but feel free to change it if you have a better idea.

delucis (approved):
Thanks for adding the changeset — I think this will do!

> Note: I am also not sure if merging this is enough to trigger a rebuild?

Yeah, we very aggressively rebuild that site to keep it up to date by hitting the rebuild webhook for each `pull_request` event on GitHub (only really need to do it when PRs are _opened_ but GitHub’s webhook UI only supports triggering for all PR events) as well as for commits to `main`.

PR #1026, opened 2023-11-02
Overhaul getting started guide

delucis:

#### Description

- This PR overhauls the getting started page in docs.
- Main aims:
  - Mention how to start the dev server after creating your project (feedback from [this video of someone working through the page](https://www.youtube.com/watch?v=2G9sQaJsvB4)).
  - Streamline some of the content and give readers more options to jump into the rest of the docs.
  - More clearly separate the “Getting Started” steps from the other things we’ve shoehorned into this page (“Updating” and “Troubleshooting”).

### DIFF
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -93,7 +93,20 @@ You can add support for Markdoc by installing the experimental [Astro Markdoc in
 Add new pages to your site by creating `.md` or `.mdx` files in `src/content/docs/`.
 Use sub-folders to organize your files and to create multiple path segments.
 
-For example, a file at `src/content/docs/hello-world.md` will be published at `your-site.com/hello-world` and a file at `src/content/docs/guides/faq.md` will be published at `your-site.com/guides/faq`.
+For example, the following file structure will generate pages at `example.com/hello-world` and `example.com/guides/faq`:
+
+import FileTree from '../../components/file-tree.astro';
+
+<FileTree>
+
+- src/
+  - content/
+    - docs/
+      - guides/
+        - faq.md
+      - hello-world.md
+
+</FileTree>
 
 #### Type-safe frontmatter
 
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -7,6 +7,7 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
 
 Starlight is a full-featured documentation theme built on top of the [Astro](https://astro.build) framework.
 This guide will help you get started with a new project.
+See the [manual setup instructions](/manual-setup/) to add the Starlight integration to an existing project.
 
 ## Quick Start
 
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -5,17 +5,19 @@ description: Learn how to start building your next documentation site with Starl
 
 import { Tabs, TabItem } from '@astrojs/starlight/components';
 
-## Creating a new project
-
 Starlight is a full-featured documentation theme built on top of the [Astro](https://astro.build) framework.
+This guide will help you get started with a new project.
 
-You can create a new Astro + Starlight project using the following command:
+## Quick Start
+
+### Create a new project
+
+Create a new Astro + Starlight project by running the following command in your terminal:
 
 <Tabs>
 <TabItem label="npm">
 
 ```sh
-# create a new project with npm
 npm create astro@latest -- --template starlight
 ```
 
@@ -23,7 +25,6 @@ npm create astro@latest -- --template starlight
 <TabItem label="pnpm">
 
 ```sh
-# create a new project with pnpm
 pnpm create astro --template starlight
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/getting-started.mdx
```diff
@@ -45,24 +45,56 @@ Try Starlight in your browser:
...
+
+Starlight supports authoring content in Markdown and MDX with no configuration required.
+You can add support for Markdoc by installing the experimental [Astro Markdoc integration](https://docs.astro.build/en/guides/integrations-guide/markdoc/).
+
+#### Add pages
+
+Add new pages to your site by creating `.md` or `.mdx` files in `src/content/docs/`.
+Use sub-folders to organize your files and to create multiple path segments.
+
+For example, a file at `src/content/docs/hello-world.md` will be published at `your-site.com/hello-world` and a file at `src/content/docs/guides/faq.md` will be published at `your-site.com/guides/faq`.
```
sarah11918, line 95: This is fine! Just thinking: Just because this visually isn't as easy to make the connection between file paths and URLs, is there maybe a way to use the file tree component here to show some sample files in a hierarchy, and then say something like:

For example, this file structure will publish pages at the following URLs:

docs/src/content/docs/getting-started.mdx
```diff
@@ -75,16 +107,18 @@ description: This is a page in my Starlight-powered site
 
 If you forget anything important, Starlight will let you know.
 
-## Deploying your Starlight website
-
-Once you have created and customized your Starlight website, you can deploy it to a web server or hosting platform of your choice including Netlify, Vercel, GitHub Pages and many more.
+### Next steps
```
sarah11918, line 110: I LOVE these next steps!

docs/src/content/docs/getting-started.mdx
```diff
@@ -5,33 +5,33 @@ description: Learn how to start building your next documentation site with Starl
 
 import { Tabs, TabItem } from '@astrojs/starlight/components';
 
-## Creating a new project
-
 Starlight is a full-featured documentation theme built on top of the [Astro](https://astro.build) framework.
+This guide will help you get started with a new project.
```
sarah11918, line 9:
```suggestion
This guide will help you get started with a new project.
See the [manual installation instructions](/manual-setup/) to add the Starlight integration to an existing project.
```
What do you think about something like this?

In Astro docs, manual setup is an analogous process (you still do create a new Astro project, just manually). In this case, "manual" really does mean add to an existing project vs create a new Starlight project, just by hand. You need an existing site already.

I *might* even suggest that that page title should more reflect Add to an Existing project? But, I think it's helpful to explicitly call out a) that you can and b) that that's what you'll find under "manual installation"

sarah11918 (commented):
This looks fabulous, @delucis! Especially love the Next Steps section!

Just a few thoughts going though, but this is wonderful!

docs/src/content/docs/getting-started.mdx
delucis: I did think of this, but worried it might take up more space than required. I could also revert to the previous style, I just wasn’t 100% sure the `path => your-site.com/path` thing was as clear as spelling it out.

Let me try the file tree version though, it would probably help to reinforce the subdirectories especially.

docs/src/content/docs/getting-started.mdx
delucis: Yeah, I like this!

Our manual setup guide does technically work for “`create astro` with the empty template, then add Starlight”, and I think of it as the place to understand better the different components of a Starlight set-up. But yeah, the main user scenario it serves is the “Add to existing site” one.

delucis: (Noting for transparency that I edited your suggestion to change “manual install” to “manual setup”, which is how we refer to the guide generally as it’s more than “installing”.)

HiDeoo (commented):
Wow, this is really great. So many improvements all over the place and I love the dev server/next steps sections.

The only thing that I've always found a bit confusing (at least for me as a non-native reader with my definition of theme and we may have already discussed this) is not really related to the changes, but maybe it's more obvious now, is the fact that we say "Starlight is a full-featured documentation theme" and later "Starlight is an Astro integration". Not sure if anything should be done about it, but I thought I'd mention it.

delucis:
> we say "Starlight is a full-featured documentation theme" and later "Starlight is an Astro integration"

Yeah, I think this is part of the issue for ourselves of answering “What is Starlight?” 😅 

I’d argue that “theme” and “integration” are not mutually exclusive and in both cases we’re drawing attention to the “built on Astro” aspect, which is important because understanding that is really helpful in knowing what else you can do with Starlight. But I still agree the messaging isn’t perfect there. (And there is also the element that you _can_ also use Starlight without really caring about Astro, at least for basic usage.)

I do think we can improve this in the first paragraph though, will make a suggestion.

docs/src/content/docs/getting-started.mdx
```diff
@@ -5,33 +5,34 @@ description: Learn how to start building your next documentation site with Starl
...
+See the [manual setup instructions](/manual-setup/) to add the Starlight integration to an existing project.
```
delucis, line 10: Small suggestion to avoid the “theme” => “integration” whiplash observed by @HiDeoo:

```suggestion
See the [manual setup instructions](/manual-setup/) to add Starlight to an existing Astro project.
```

PR #1031, opened 2023-11-02
Fix docs component details

delucis:

- Something else!

#### Description

- Fixes two small details in our docs components that have been bugging me forever.

- Removes the `not-content` class from the “About Astro” block on the homepage so that the link is the correct colour.

   | Before | After |
   |---|---|
   | image | image |

- Adds just enough margin to file-tree entries so that when two neighbouring entries are highlighted, they don’t touch.

   | Before | After |
   |---|---|
   | image | image |


HiDeoo (approved):
Neat tweaks, looks perfect to me 🌟

PR #1039, opened 2023-11-03
[i18nIgnore] Fix links

delucis:

#### Description

- Fixes several broken links and makes sure links all point to the current locale (e.g. on `/es/` pages, all relative links should be to `/es/` versions of pages).
- Caught the broken links with https://github.com/HiDeoo/starlight-links-validator.
- Caught the wrong-locale links with some RegExp searching in the project (e.g. find `\]\(/(?!ru)` in `docs/src/content/docs/ru/`).

P.S. I considered if we could integrate the link checker into the docs site but https://github.com/HiDeoo/starlight-links-validator/issues/9 is blocking us currently from doing that I think.


HiDeoo (approved):
Took a bit of time to test them all but clearly way less than it took you to fix them all. Impressive work 🤯

delucis:
Thanks for the reviews and testing! Yeah, this is definitely something it would be nicer to stay on top of with each CI check.

@HiDeoo Saw you added the extra context https://github.com/HiDeoo/starlight-links-validator/issues/9 — thank you!

HiDeoo:
> @HiDeoo Saw you added the extra context [HiDeoo/starlight-links-validator#9](https://github.com/HiDeoo/starlight-links-validator/issues/9) — thank you!

Yeah, I also played a little bit with it this morning as I had my head in the code while releasing a new version. The logic should be relatively straightforward altho I think the smart move is to wait for plugins so I can turn the integration into a plugin and easily have access to the Starlight config to check if the site is multilingual or not and check which locales are configured.

delucis:
> the smart move is to wait for plugins so I can turn the integration into a plugin and easily have access to the Starlight config to check if the site is multilingual or not and check which locales are configured.

Absolutely. I’d also love a rule to enforce use of same-language links, which were a bunch of what I’d caught… wait I guess I should make a feature request 😁

PR #1044, opened 2023-11-06
Fix last updated dates for pages displaying fallback content

### LINKED ISSUE #1043: Last Updated fallback not working

0.11.1

3.3.4

pnpm

Windows

Brave

### Describe the Bug

If I create a new page in the default Language, the lastUpdated Field works just fine. 

However, if I change the language and Starlight shows the fallback content, it doesn't use the lastUpdated Date from the fallback content but nothing, which is confusing...

https://starlight.astro.build/tr/guides/i18n/

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1043.

This pull request fixes the last updates dates for pages displaying fallback content by taking into account `isFallback` when generating the file path used with `getFileCommitDate()`.

Preview:

- A docs production fallback page with no last update date: [link](https://starlight.astro.build/tr/guides/overriding-components/#only-override-on-specific-pages)
- The same page after this PR: [link](https://deploy-preview-1044--astro-starlight.netlify.app/tr/guides/overriding-components/#only-override-on-specific-pages)


packages/starlight/utils/route-data.ts
```diff
@@ -66,9 +66,10 @@ function getToC({ entry, locale, headings }: PageProps) {
 	};
 }
 
-function getLastUpdated({ entry, id }: PageProps): Date | undefined {
+function getLastUpdated({ entry, id, isFallback }: PageProps): Date | undefined {
 	if (entry.data.lastUpdated ?? config.lastUpdated) {
-		const currentFilePath = fileURLToPath(new URL('src/content/docs/' + id, project.root));
+		const currentId = isFallback ? localizedId(id, config.defaultLocale.locale) : id;
+		const currentFilePath = fileURLToPath(new URL('src/content/docs/' + currentId, project.root));
```
delucis, line 72: I actually think we can simplify it to this!

```suggestion
		const currentFilePath = fileURLToPath(new URL('src/content/docs/' + entry.id, project.root));
```

`PageProps` includes a root-level `id` localized for the current page, but `entry` also includes the original `id` for the content being used.

packages/starlight/utils/route-data.ts
```diff
@@ -66,9 +66,10 @@ function getToC({ entry, locale, headings }: PageProps) {
 	};
 }
 
-function getLastUpdated({ entry, id }: PageProps): Date | undefined {
+function getLastUpdated({ entry, id, isFallback }: PageProps): Date | undefined {
```
delucis, line 69: And then this is also simpler!

```suggestion
function getLastUpdated({ entry }: PageProps): Date | undefined {
```

delucis (commented):
Thanks for quickly picking this up @HiDeoo — and especially for adding some testing here. Will help us know if my suggestion actually works or not :-D

HiDeoo:
Brilliant idea, did not think of that at all and it works perfectly. Thanks a lot!

I also slightly improved the test as I was taking a route and "faking" it to be a fallback one instead of just using a fallback route ^^ Not sure what I was thinking there.

delucis (approved):
Thanks @HiDeoo — looks great! Should be safe to merge whenever you like 🌟

PR #1045, opened 2023-11-06
Expand testing

delucis:

- Something else!

#### Description

- Adds tests for some areas of the code base that aren’t currently covered:
  - The rehype code used inside the `<Tabs>` component
  - The remark plugin used to render asides (`:::note`, `:::tip` etc.)
  - The wrapper we use to configure `@astrojs/sitemap` for people based on their i18n config
- While working on this I also explored some ways to test our [Zod error map](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/error-map.ts) and found some areas where we could probably improve the error messages, so will move that to a follow-up PR (probably for after #942 given that also is making some changes to config parsing)


packages/starlight/package.json
```diff
@@ -164,6 +164,7 @@
     "astro": "^3.2.0"
   },
   "devDependencies": {
+    "@astrojs/markdown-remark": "^3.2.1",
```
HiDeoo, line 167: Could this mean we could drop the dependency on `unified` and use the `RemarkPlugin` and `RehypePlugin` types exported from `@astrojs/markdown-remark` for our remark and rehype plugins?

Maybe also the [`RemarkPlugins` constructed type](https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/asides.ts#L158) in `packages/starlight/src/integrations/asides.ts` considering `RemarkPlugins` is also exported from `@astrojs/markdown-remark`?

delucis: I added this as a dev dependency for testing only as I’d be a bit concerned about ending up accidentally depending on two different versions (one, the subdependency of a user’s `astro` version, the other a subdependency of Starlight).

I was looking at the `asides` typings the other day in relation to #517. Maybe there’s another way to get those types without `unified`? I might leave figuring that out to until the Astro v4 release as that’s planning to upgrade the whole unified/remark/rehype/MDX ecosystem, and lots of the types shifted in the most recent releases of those packages.

HiDeoo: Looks like we're all looking at these types ^^

I was looking at them too for #983 (which is definitely not as fun as I would have hoped ^^) where I am down to 4 TS errors which are all related to remark/rehype and after that it should get easier (js generation, "assets" copy and .d.ts generation are in place).

delucis: > it's not the easiest eco-system to deal with type-wise imo

Absolutely not 😅 In general I think the types are good, but with the landscape of many, many small packages, it’s really easy to end up with conflicts between versions. Hopefully should be easier with Astro v4 because we should at that point be able to simply upgrade EVERYTHING to `latest`.

HiDeoo: > Hopefully should be easier with Astro v4 because we should at that point be able to simply upgrade EVERYTHING to latest.

That would be so awesome indeed 🤞 

I'm starting to think that #983 will need to wait for that change too. I'm down to 3 errors now all related to rehype/remark that I don't know how to __properly__ fix due to the conflicting versions of `@types/unist` that we have.

PR #1049, opened 2023-11-07
Expose Markdown content styles

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #980 (I don't think this works like issues 🤔 )

This PR exposes the Markdown content styles to the user.

#980 describes a few use cases for this, but a basic example could be an integration/plugin having a page with a totally different UI for the main content where only some portions of the content are Markdown with the expectation for such content to be styled the same way as the rest of Starlight Markdown. For complex UI, wrapping such content in `<MarkdownContent/>` can be a bit cumbersome considering this component expects the [standard component props](https://starlight.astro.build/reference/overrides/#component-props) that would need to be passed down.


delucis (approved):
Thanks @HiDeoo — this looks good to me!

PR #1056, opened 2023-11-08
i18n(de): Update translation of `frontmatter.md` into German

trueberryless:
also fixed an type (Ihrem -> deinem) in line 91


delucis (approved):
Looks great — thank you @trueberryless 🚀

PR #1057, opened 2023-11-08
i18n(de): Update translation of another file into German

trueberryless:
lol


HiDeoo (changes requested):
Thanks for the update.

I think the identical type 4 lines below should be removed (I unfortunately cannot comment on it as it's outside the review limited range).

trueberryless:
Okay, thanks, I didn't notice them!

HiDeoo (approved):
Thanks for the quick fix, looks perfect now.

PR #1069, opened 2023-11-15
Fix sidebar highlighting when path contains non-ASCII characters

### LINKED ISSUE #1067: Sidebar highlighting and navigation buttons not working in pages with non-ascii in path

0.12.0

3.5.3

pnpm

Windows

Chrome

### Describe the Bug

When the current page path contains non-ascii characters, the navigation buttons at the bottom of the page is missing, and the current page is not highlighted in sidebar.

image

Reproduction:

- `pnpm create astro --template starlight`
- Create a new file at `src/content/docs/reference/a中文.md` with any content
- `pnpm dev` and navigate to `http://localhost:4321/reference/a中文/`

https://stackblitz.com/edit/github-gdrppa?file=src%2Fcontent%2Fdocs%2Freference%2Fa%E4%B8%AD%E6%96%87.md

### Participation

Genteure:

- Changes to Starlight code

#### Description

`currentPathname` comes from [`Astro.url` in `index.astro`](https://github.com/withastro/starlight/blob/60edb22374a73bc0d02dbe2e26c2566d6400a3e8/packages/starlight/index.astro#L14), which is

> A full URL object of the request URL. Equivalent to: `new URL(Astro.request.url)`

- Closes #1067

| Before | After |
|---|---|
| image | image |


packages/starlight/utils/navigation.ts
```diff
@@ -135,7 +135,7 @@ function makeLink(
 	attrs?: LinkHTMLAttributes
 ): Link {
 	if (!isAbsolute(href)) href = pathWithBase(href);
-	const isCurrent = href === ensureTrailingSlash(currentPathname);
+	const isCurrent = new URL(href, 'http://example.com').pathname === ensureTrailingSlash(currentPathname);
```
HiDeoo, line 138: I think [`encodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) should work in this case (I haven't tested it though yet) but if it does, it may be easier to understand than using a fake base URL? Any thoughts?

```suggestion
	const isCurrent = encodeURI(href) === ensureTrailingSlash(currentPathname);
```

kevinzunigacuellar:
Perhaps we can re-use a test suite. I was looking around and I think it could be a good fit for `basics/navigation` -> `getSidebar()`. What do you think?

https://github.com/withastro/starlight/blob/60edb22374a73bc0d02dbe2e26c2566d6400a3e8/packages/starlight/__tests__/basics/navigation.test.ts#L16

delucis (approved):
Thanks for the fix @Genteure!

PR #1102, opened 2023-11-20
Add more EC text markers and frames to Starlight docs

HiDeoo:

#### Description

This PR is a follow-up to this [comment](https://github.com/withastro/starlight/pull/742#issuecomment-1739445262) I made in the EC PR and a review I did a while ago regarding all the code blocks in the Starlight documentation to add missing EC text markers and frame informations in a more consistent way.

I edited everything I could find but I think some may be up to discussions, like for example having `src/content/docs/example.md` as frame title for all examples in the frontmatter reference may not be the best, but as I don't personally know the guidelines for this I think this PR could be used as a base to discuss this even if we end up not merging it or only partially (we could also decide to add this kind of changes only when we next edit the relevant portions of the docs).


docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -88,7 +89,8 @@ Add a group using an object with `label` and `items` properties.
 The `label` will be used as the heading for the group.
 Add links or subgroups to the `items` array.
 
-```js
+```js /(label:) 'G/ /(label:) 'St/ "items:"
```
HiDeoo, line 92: I only wanted to highlight the `label` property of the groups and not the ones of the links too.

hippotastic: Haha, nice to see someone finally using the regexp matching!

You could shorten the expression a bit though:
```suggestion
```js /^\s*(label:|items:)/
```

HiDeoo: I was a bit afraid of using `\s` with Prettier which could technically break this one, but seeing how clean your suggestion is, I think it's worth it and no matter which syntax we use, we would still need to make sure it's ok during the review of a new translation so we might as well use the cleanest one.

Thanks for the great suggestion 🤩

hippotastic: Famous last words: I'm pretty confident that Prettier is not going to waltz in and make that expression not match the intended strings anymore. :) As long as there is nothing but optional whitespace (spaces/tabs) in front of the keys, they should match.

And if we really do run into issues with Prettier, I'm looking forward to learning why and how to prevent them. :)

hippotastic: I just tested it and the labels would need to be really long. Even `{ label: 'These are really long labels', link: '/guides/components/' },` still remains on one line. So let's hope for the best! :)

docs/src/content/docs/guides/i18n.mdx
```diff
@@ -152,14 +152,14 @@ You can provide translations for additional languages you support — or overrid
 
 1. Configure the `i18n` data collection in `src/content/config.ts` if it isn’t configured already:
 
-   ```js
+   ```diff lang="js" ins=", i18nSchema"
```
kevinzunigacuellar, line 155: I'm torn about whether to include the comma here. Technically, it's correct, but visually, it seems a bit odd to me 😅. What do you all think?

```suggestion
   ```diff lang="js" ins="i18nSchema"
```

kevinzunigacuellar (commented):
Love the Expressive codification of Starlight 💜

docs/src/content/docs/guides/customization.mdx
```diff
@@ -139,7 +141,7 @@ tableOfContents:
   </TabItem>
   <TabItem label="Global config">
 
-```js
+```js {7}
 // astro.config.mjs
 
 defineConfig({
```
delucis, line 147: Not strictly speaking related to this PR, but noticing again the weird empty title in this example, maybe an opportunity to fix it?

Something like this?

```js
			title: 'Docs with custom table of contents config',
```

docs/src/content/docs/guides/customization.mdx
```diff
@@ -406,7 +409,7 @@ It provides npm modules you can install for the fonts you want to use and includ
 To apply the font you set up to your site, use your chosen font’s name in a [custom CSS file](/guides/css-and-tailwind/#custom-css-styles).
 For example, to override Starlight’s default font everywhere, set the `--sl-font` custom property:
 
-```css
+```css {4}
```
delucis, line 412: This one feels like it might be overkill given it’s almost highlighting the full example and the `:root` selector is relevant too: 

three-line code block with one line highlighted

Might suggest removing the highlight in this case:

````suggestion
```css
````

File: docs/src/content/docs/reference/configuration.mdx
Comment (delucis) [inline docs/src/content/docs/reference/configuration.mdx:1]: Curious what other people think here. A lot of these code snippets are fragments like 

```js
starlight({
  customCss: ['./src/custom-styles.css', '@fontsource/roboto'],
})
```

In these cases, I wonder if it’s misleading to add the `astro.config.mjs` file name because it makes it looks like this is the full file rather than just a part of it?

delucis (commented):
Awesome work @HiDeoo! Left a few comments but most of this is a really nice improvement — cool to see so quickly after we released support for this.

docs/src/content/docs/reference/configuration.mdx
HiDeoo: I think this makes sense and we could definitely remove the frame titles (except the first one).

Altho, if we go with this approach, we have the same pattern in the following files that may need fixing too:

- `docs/src/content/docs/guides/customization.mdx`
- `docs/src/content/docs/guides/sidebar.mdx`

The question would be if we either remove the frame titles for those too or we add the imports? Or maybe this is only relevant in this specific file?

HiDeoo:
Updated the PR with the suggested changes and left a comment regarding the `configuration.mdx` file.

docs/src/content/docs/reference/configuration.mdx
delucis: Scrolling through [`customization.mdx`](https://starlight.astro.build/guides/customization/) on the current production version I think that actually looks good yeah — file names are used for full examples, and not for shorter fragments (usually accompanying a longer one like in the [“Add your logo”](https://starlight.astro.build/guides/customization/#add-your-logo) section).

HiDeoo:

- Remove previously introduced frame titles in `guides/customization.mdx`
- Remove all frame titles in `reference/configuration.mdx` except the first code block
- Remove `astro.config.mjs` frame titles from `guides/sidebar.mdx` as it's explicitely mentioned in the text before any code block (I left the `src/content/docs/example.md` one)

I think this should cover all the feedback we got so far. Let me know if I missed anything or if you have any other suggestions.

docs/src/content/docs/reference/configuration.mdx
kevinzunigacuellar: I personally like the `startlight()` snippets because they are easier to copy and paste. Sometimes I would have a very convoluted astro.config and just having the startlight snippet to copy and paste is very nice.

delucis (approved):
Thank you @HiDeoo! 🚀

PR #1119, opened 2023-11-23
docs: Update various Expressive Code docs links

HiDeoo:

#### Description

This PR updates various links to the Expressive Code documentation in `guides/authoring-content.md` which I think are wrong. I think I'm fairly sure about the first change and for the other two I'm less sure as this leads to two identical links but at least they link to the relevant examples.

### DIFF
--- a/docs/src/content/docs/guides/authoring-content.md
+++ b/docs/src/content/docs/guides/authoring-content.md
@@ -239,7 +239,7 @@ Some of the most common examples are shown below:
   ```
   ````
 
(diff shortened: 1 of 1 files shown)


delucis (approved):
Yes, I think you’re right — looks like better things to link to. We changed up the wording and examples slightly while editing, so could be that we deviated from the original link.

hippotastic:
Thank you @HiDeoo! I should have paid more attention to the wording changes that caused the links to become incorrect. I'm fairly certain they did make sense at some point. Thank you very much for catching and promptly fixing that!

PR #1123, opened 2023-11-24

jsparkdev:
#### Description

update `components.mdx`


HiDeoo (approved):
Thanks for the update, looks perfect to me 👍

PR #1126, opened 2023-11-24
i18n(ko-KR): update `i18n.mdx`

jsparkdev:
#### Description

update `i18n.mdx`


HiDeoo (approved):
Thanks for another great update, looks perfect 👍

PR #1127, opened 2023-11-24

jsparkdev:
#### Description

update `sidebar.mdx`


docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -506,7 +507,7 @@ starlight({
 
 `autogenerate.collapsed` 속성을 정의하여 이 동작을 변경할 수 있습니다.
 
-```js
+```js {5-7} "collapsed: true"
```
HiDeoo, line 510:
```suggestion
```js {5-6} "collapsed: true"
```

The comment only spans 1 line in this version so we can reduce the highlight range.

HiDeoo (changes requested):
Thanks for the update, left a tiny suggestion to reduce an highlight range but except that, looks perfect to me 👍

HiDeoo (approved):
Thanks for the quick fix 🙌

Issue #860, opened 2023-10-11
Cannot access certain imports - Overriding components


TheOtterlord:
### What version of `starlight` are you using?

0.11.0

### What version of `astro` are you using?

latest

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

Certain imports used by overridable components are not exported. `utils` is one I find I often need to access. Either because the builtin component uses it, or I need translation strings, etc.

List of missing exports I've noticed/needed so far.

- `utils`
- `constants`

#### Workaround

You can use a direct path to the `node_modules` folder to work around the missing `exports`.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-bd7al8?file=src%2FTest.astro

### Participation

- [x] I am willing to submit a pull request for this issue.

TheOtterlord:
Adding to this, types are not exposed for the virtual module. The situation is a little different and will require documenting, but you can place this directive in your `env.d.ts` file to load types for the virtual modules. I feel this can be exposed via `@astrojs/starlight/virtual` or something like that, then we can document that you can add this to fix typescript.
```ts
/// <reference path="../node_modules/@astrojs/starlight/virtual.d.ts"/>
```

delucis:
Thanks for the issue @TheOtterlord! This is partly intentional, although we should still find a way to support what you need to do.

Basically one of the aims of my refactors in #709 was to move as much logic as possible up out of components so we can keep `utils/` as internal implementation details, so they’re easy to change without breaking people’s sites.

## Translation strings

The one thing I decided was out of scope there was the translation strings for a couple of reasons:

- Time is limited, so I just decided to put off that work 😁 
- Our current `useTranslations()` system is very basic and not really a proper solution for generic i18n needs. We’ve spoken about replacing it with a more battle-tested solution like `i18next`. Exposing our existing functions only to later change them seemed a shame if avoidable. (More about why a more robust system may be needed in [“Lessons From Linguistics: i18n Best Practices for Front-End Developers”](https://shopify.engineering/internationalization-i18n-best-practices-front-end-developers) on the Shopify blog.)

One interim solution I like the sound of more is just to pre-generate the strings as part of `generateRouteData()` so that component props include a `labels` object which can be accessed directly, something like:

```diff
- import { useTranslations } from '../utils/translations';
- const t = useTranslations(Astro.props.locale);
- const label = t('page.editLink');
[...]

PR #1135, opened 2023-11-24
Expose localized UI strings in route data

delucis:
#### Description

- Closes #1060 and also partially addresses #860.
- Exposes a new `labels` property in Starlight’s route data object for each page.
- This object includes all localized UI strings that up to now were retrieved using `useTranslations()` and the `t()` helper it returned.
- This enables components overriding Starlight’s defaults to easily reuse the built-in UI strings.
- This PR also refactors all our built-in components to use `labels` instead of calling `useTranslations()` themselves.
- Removes the `t.pick()` helper previously returned by `useTranslations()` is it is no longer used.


HiDeoo (approved):
The changes look good to me and I think this properly address the linked issues.

One thing I want to note tho is that with this approach, if we ever move to `i18next` or something similar with a more advanced `t()`, we would not be able to use some of the advanced features like interpolation, formatting, plurals, context, etc. I don't really see us needing those features in the near future, but I just wanted to point that out.

delucis:
> if we ever move to `i18next` or something similar with a more advanced `t()`, we would not be able to use some of the advanced features like interpolation, formatting, plurals, context, etc.

I think if we ever do that, we could still expose those functions on the route data object. Technically this API _could_ have just added `t()` to the route data object, but I decided to go the plain data route because it fits better with the rest of our route data story for now.

HiDeoo:
> I think if we ever do that, we could still expose those functions on the route data object. Technically this API could have just added t() to the route data object, but I decided to go the plain data route because it fits better with the rest of our route data story for now.

Oh definitely, I meant "we would not be able to use some of the advanced features" while keeping the same API without additional changes like exposing `t()`. Should have been more clear about that, my bad.

PR #1137, opened 2023-11-24
i18n(es): update `components`

dreyfus92:
#### Description

Updated `components`


docs/src/content/docs/es/guides/components.mdx
```diff
@@ -185,6 +191,7 @@ import { LinkCard } from '@astrojs/starlight/components';
 </CardGrid>
 
 ### Iconos
+# src/content/docs/example.mdx
```
HiDeoo, line 194: I think this one should be 10 line below (but out of suggestion range).

HiDeoo (changes requested):
Left a comment for one misplaced comment but except that, this looks perfect to me 👍

HiDeoo (approved):
Thanks for the quick fix 🙌

PR #1144, opened 2023-11-24
Add a `pagefind` boolean flag to user config

delucis:
#### Description

- Closes #952
- Adds a `pagefind` boolean to user config that can be set to `false` to disable Pagefind indexing and skip rendering the default search UI.

### DIFF
--- a/.changeset/beige-shoes-whisper.md
+++ b/.changeset/beige-shoes-whisper.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': minor
 ---
 
-Add configuration option to disable Pagefind
+Adds a configuration option to disable site indexing with Pagefind and the default search UI
--- a/.changeset/beige-shoes-whisper.md
+++ b/.changeset/beige-shoes-whisper.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Add configuration option to disable Pagefind
--- a/docs/src/content/docs/reference/configuration.mdx
+++ b/docs/src/content/docs/reference/configuration.mdx
@@ -416,7 +416,7 @@ When using custom themes and setting this to `true`, you must provide at least o
 **type:** `boolean`  
 **default:** `true`
 
-Define if Starlight’s default site search provider [Pagefind](https://pagefind.app/) is enabled.
+Define whether Starlight’s default site search provider [Pagefind](https://pagefind.app/) is enabled.
 
 Set to `false` to disable indexing your site with Pagefind.
 This will also hide the default search UI if in use.
--- a/docs/src/content/docs/reference/configuration.mdx
+++ b/docs/src/content/docs/reference/configuration.mdx
@@ -411,6 +411,16 @@ When `false`, the colors provided by the active syntax highlighting theme are us
 When using custom themes and setting this to `true`, you must provide at least one dark and one light theme to ensure proper color contrast.
 :::
 
+### `pagefind`
+
+**type:** `boolean`  
+**default:** `true`
+
+Define if Starlight’s default site search provider [Pagefind](https://pagefind.app/) is enabled.
+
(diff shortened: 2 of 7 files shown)


.changeset/beige-shoes-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Add configuration option to disable Pagefind
```
sarah11918, line 5:
```suggestion
Adds a configuration option to disable site indexing with Pagefind
```
less harsh wording?

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -411,6 +411,16 @@ When `false`, the colors provided by the active syntax highlighting theme are us
 When using custom themes and setting this to `true`, you must provide at least one dark and one light theme to ensure proper color contrast.
 :::
 
+### `pagefind`
+
+**type:** `boolean`  
+**default:** `true`
+
+Define if Starlight’s default site search provider [Pagefind](https://pagefind.app/) is enabled.
```
sarah11918, line 419:
```suggestion
Define whether Starlight’s default site search provider [Pagefind](https://pagefind.app/) is enabled.
```

packages/starlight/utils/user-config.ts
```diff
@@ -190,6 +190,13 @@ const UserConfigSchema = z.object({
 	 */
 	expressiveCode: ExpressiveCodeSchema(),
 
+	/**
+	 * Define if Starlight’s default site search provider Pagefind is enabled.
```
sarah11918, line 194:
```suggestion
	 * Define whether Starlight’s default site search provider Pagefind is enabled.
```

sarah11918 (commented):
Docs review! Some comments suggested.

packages/starlight/components/Header.astro
```diff
@@ -8,14 +9,20 @@ import {
 	SocialIcons,
 	ThemeSelect,
 } from 'virtual:starlight/components';
+
+/**
+ * Render the `Search` component if Pagefind is enabled or the default search component has been overridden.
```
HiDeoo, line 14: Should this behavior referenced somewhere like in the search component override reference?

delucis: Ah, maybe. I initially thought about disabling rendering inside `<Search>` itself, but that would result in some CSS & JS still being loaded even if the UI isn’t, so moved it here. I guess I assumed if you override `Search` you don’t also want your override to be hidden if `pagefind` is disabled.

.changeset/beige-shoes-whisper.md
delucis: Expanded a little more in that case:
```suggestion
Adds a configuration option to disable site indexing with Pagefind and the default search UI
```

packages/starlight/components/Header.astro
delucis: Added a note in accf0a2!

HiDeoo (approved):
Looks perfect and ready to me 👍

dreyfus92 (approved):
LGTM ✅

kevinzunigacuellar:
I'm considering an API structure like this:

```js
search: 'pagefind' // or false
```
I find this setup more explicit in its configuration. Additionally, this API structure could potentially pave the way for other official plugins like 'algolia', although that's beyond the scope of this comment.

What are your thoughts on this approach?

delucis:
> this API structure could potentially pave the way for other official plugins

Interesting idea! But I'm pretty confident that we won't have other search providers built-in, so this option would end up having only one possible value, which suggests the Boolean toggle probably works well I think.

Alternative search providers could still be added via the new plugins API but they would then be configured in the plug-in options rather than in the global Starlight config.

kevinzunigacuellar (approved):
Lively Groove, Tune Mastered!

PR #1114, opened 2023-11-23
i18n(fr): Update multiple files with EC markers

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content

#### Description

This PR updates the french version of the following files with the changes of #1102:

- `guides/components.mdx`
- `guides/css-and-tailwind.mdx`
- `guides/customization.mdx`
- `guides/sidebar.mdx`
- `reference/frontmatter.md`
- `reference/overrides.md`

PR #1150, opened 2023-11-25
i18n(fr): translate some frame titles

HiDeoo:

#### Description

This pull request is a follow-up to [this comment](https://github.com/withastro/starlight/pull/1114#pullrequestreview-1746836085) and translates various frame titles to French, e.g. `example.md` -> `exemple.md`.

There are still some frame titles that are not translated:

- `font-face.css`
- `custom.css`
- `EmailLink.astro`
- `Title.astro`
- `ConditionalFooter.astro`
- `Custom.astro`

If this makes sense to translate some of them like `example` or `my-test-file` in a documentation context, I explicitly left those out for multiple reasons:

- It's usually not recommended to have file names in french in the industry and even forbidden sometimes
- Some are very specific to a concept which could be hard to translate, e.g. `font-face.css`
- Some I just didn't know how to translate, e.g. `custom.css` -> `personnalisé.css`? This makes no sense especially with the `é` which is usually not recommended in file names


delucis (approved):
LGTM! And happy for you to make the call about when it really makes sense to translate these and when keeping the original English is more helpful.

PR #1156, opened 2023-11-27
Upgrade `@astrojs/sitemap` to latest

votemike:
#### Description

Upgrade sitemap package to fix Zod security issue.

### DIFF
--- a/.changeset/quiet-trains-know.md
+++ b/.changeset/quiet-trains-know.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Updated sitemap dependency to fix Zod security issue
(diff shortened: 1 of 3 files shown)


delucis:
Thanks for the PR @votemike. I assume this relates to https://github.com/advisories/GHSA-m95q-7qp3-xv42?

To be clear, this vulnerability does not impact Starlight — firstly because we do not parse any e-mails, and secondly because this is a DDoS vulnerability that would only apply if you were using Zod to parse unsanitized user input, for example in a server application, which is not the case with Starlight.

Still happy to update versions so people’s security analysis tools don’t complain, but wanted to mention this here for the record.

votemike:
Yep. I guessed as much. But I thought other people's dependabots would be complaining.

Also, other parts of @astrojs/starlight seem to have already updated to 3.22.4, this seemed to be the only thing that hadn't.

PR #1168, opened 2023-11-29
Add Algolia DocSearch plugin

delucis:
#### Description

- Adds our first official plugin — for Algolia DocSearch.
- Adds docs for site search in Starlight, including how to add this new plugin.
- ~~WIP because depends on several of our inflight PRs. Currently this is built on @HiDeoo’s #942 branch so will show many unrelated changes. I will rebase once that is merged.~~


packages/docsearch/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "variables.css",
+    "virtual.d.ts"
+  ],
+  "exports": {
+    ".": "./index.ts",
+    "./DocSearch.astro": "./DocSearch.astro",
+    "./schema": "./schema.ts"
+  },
+  "peerDependencies": {
+    "@astrojs/starlight": ">=0.9.0"
```
HiDeoo, line 28: Shouldn't this be way higher?

delucis: Oh yeah 😅 

Copy/pasted from the Tailwind package, but will need v0.14. I’ll update once that’s released.

delucis: Fixed in 77164f6

docs/src/content/docs/guides/site-search.mdx
```diff
@@ -0,0 +1,167 @@
+---
+title: Site Search
+description: Learn about Starlight’s built-in site search features and how to customize them.
```
HiDeoo, line 3: Random thought: should we move the "New" badge from `guides/overriding-components.md` to this page?

docs/src/content/docs/guides/site-search.mdx
```diff
@@ -0,0 +1,167 @@
...
+   import starlightDocSearch from '@astrojs/starlight-docsearch';
+
+   export default defineConfig({
+   	integrations: [
+   		starlight({
+   			title: 'Site with DocSearch',
+   			plugins: [
+   				starlightDocSearch({
+   					appId: 'YOUR_APP_ID',
+   					apiKey: 'YOUR_INDEX_NAME',
```
HiDeoo, line 96:
```suggestion
   					apiKey: 'YOUR_SEARCH_API_KEY',
```

docs/src/content/docs/guides/site-search.mdx
```diff
@@ -0,0 +1,167 @@
...
+   					indexName: 'YOUR_SEARCH_API_KEY',
```
HiDeoo, line 97:
```suggestion
   					indexName: 'YOUR_INDEX_NAME',
```

HiDeoo (commented):
I do not have API keys to actually test this but the code looks fine to me so far and I think the comment you left will be a good template/starting point for plugin authors 👍 

Left a few comments regarding the documentation.

delucis:
> I do not have API keys to actually test this

If you wanted to test, you can use the test credentials provided on the DocSearch site: https://docsearch.algolia.com/docs/DocSearch-v3#testing

About to do that myself now that this has all our recent changes merged in.

docs/src/content/docs/guides/site-search.mdx
HiDeoo: I think this page which contains valuable infos for **both** users of Pagefind and the first actual plugin may be way more interesting to people but it's your call ^^

HiDeoo:
> If you wanted to test, you can use the test credentials provided on the DocSearch site: [docsearch.algolia.com/docs/DocSearch-v3#testing](https://docsearch.algolia.com/docs/DocSearch-v3#testing)

Oh I did not know about that, thanks 👍

HiDeoo:
Well, after playing with for a while, I can say that it's working amazingly well, great job.

The main thought I had was not related to this PR but mostly about `extend` which I wonder if accepting an array would not have been easier for users with multiple plugins extending the same schema.

Back on topic, I am not a big user of DocSearch but do you see people asking for other options? Like [`getMissingResultsUrl`](https://docsearch.algolia.com/docs/api#getmissingresultsurl), [`initialQuery`](https://docsearch.algolia.com/docs/api#initialquery), [`searchParameters `](https://docsearch.algolia.com/docs/api#searchparameters) and such? Is this common? I see the docs are [using them](https://github.com/withastro/docs/blob/main/src/components/Header/DocSearch.tsx#L48-L92) quite heavily?

delucis:
> The main thought I had was not related to this PR but mostly about `extend` which I wonder if accepting an array would not have been easier for users with multiple plugins extending the same schema.

I did consider this but typing it was a nightmare. Especially as we support things like unions and intersections in the docs schema, which are even harder to generically combine.

For the i18n case specifically, it’s actually pretty OK because we only support `z.object()`:

```js
i18nSchema({
  extend: customSchema()
  	.merge(otherSchema())
  	.merge(z.object({ 'one.more': z.string() }),
})
```

For the `docsSchema` it will depend how the custom schemas are written how easy it is. With objects the same kind of approach can apply. With unions etc. it may be more complex.

> Back on topic, I am not a big user of DocSearch but do you see people asking for other options? Like [`getMissingResultsUrl`](https://docsearch.algolia.com/docs/api#getmissingresultsurl), [`initialQuery`](https://docsearch.algolia.com/docs/api#initialquery), [`searchParameters `](https://docsearch.algolia.com/docs/api#searchparameters) and such? Is this common? I see the Astro docs are [using them](https://github.com/withastro/docs/blob/main/src/components/Header/DocSearch.tsx?rgh-link-date=2023-11-29T22%3A05%3A42Z#L48-L92) quite heavily?

**VERY** good question. I kind of skipped over these during implementation as not being able to pas across the config boundary (and some are not really designed well for the plain JavaScript version of the component we’re using). I should probably take a look tomorrow at what we could do though. One option would be to let users provide a DocSearch options module or something, where they would export a function like:

```js
export function getDocSearchOptions(routeData) {
  return {
    transformItems(items) {
      return items.map((item) => ({
        ...item,
        content: item.content.toUpperCase(),
      }));
    }
  }
}
```

delucis:
> do you want me to setup a PR as a follow-up to this one to add back the plugin showcase and its references I removed from various pages that was in my plugin PR?

I think the references removed in https://github.com/withastro/starlight/pull/942/commits/0682f56c8ebf5d623314e7ae1c63ca68b32392b5 definitely already make sense! Don’t know if we need a separate plugins page (like a “Community Plugins” page) yet or not, although the nice thing would be giving some more visibility to them as a concept, so I’m not against either if you want to do that.

delucis:
@HiDeoo What to do you think of treating support for those DocSearch options as a feature request to be added after an initial release? I don’t _think_ there’s another way to build this plugin that would make supporting them easier, so releasing early shouldn’t be too risky (and we’ll release as a 0.1, so can always break later if needed).

HiDeoo:
> I did consider this but typing it was a nightmare.

I see, this makes sense (and I'm not gonna be the one contradicting you on this one after my lost battle with it in the plugins PR :rofl:).

> One option would be to let users provide a DocSearch options module or something

Yeah, this is one we talked about on Discord for something else a long time ago and that I have used a few times to cross over. Not the most elegant but in the current state of things, I personally do not have other "easy" solutions and I think this _may_ be ok for more "advanced" configuration scenarios.

> I think the references removed in [0682f56](https://github.com/withastro/starlight/commit/0682f56c8ebf5d623314e7ae1c63ca68b32392b5) definitely already make sense!

Perfect, I'll setup a PR for this as a follow-up to this one.

> although the nice thing would be giving some more visibility to them as a concept

I agree, altho it may be worth waiting for at least 2 or 3 plugins to get a dedicated page I think but that's something we can easily tweak later on.

> What to do you think of treating support for those DocSearch options as a feature request to be added after an initial release?

I would think it's fine but as I'm not a huge user, I may be totally wrong on this one. But like you said, I don't think what you built would need a total rewrite to support this, and even more for a `v0.1` so I would go for it imo and built on top of it later on based on what is needed for Astro docs and user feedback.

docs/src/content/docs/guides/site-search.mdx
delucis:
```suggestion
description: Learn about Starlight’s built-in site search features and how to customize them.
badge: New
```

HiDeoo (approved):
Plugin #1 is cleared for takeoff in my book 🚀

docs/src/content/docs/guides/site-search.mdx
delucis: Accepted your suggestion and added the badge here for now! Maybe can move it to a plugin showcase in a bit once we have that.

PR #1170, opened 2023-11-29
Fix timezone-reliance in LastUpdated

### LINKED ISSUE #1166: lastUpdated displays incorrect date

0.13.1

3.6.3

npm

macOS

Firefox

### Describe the Bug

I set `lastUpdated` in frontmatter to 2023-03-24:

CleanShot 2023-11-29 at 13 40 48@2x

And get March 23, 2024 as the output on the page:

CleanShot 2023-11-29 at 13 40 23@2x

But the `<time>` element includes the correct date.

The underlying component uses `toLocaleDateString`:

https://github.com/withastro/starlight/blob/64d25c125c80375ff7b2cbf749f3c35d49b8bc98/packages/starlight/components/LastUpdated.astro#L14

Which [uses the user-agent's timezone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString). So I'm at -5 offset, so it'll always show the wrong day on my computer.

http://stackblitz.com/sorry-i-can-submit-a-pr-this-is-pretty-well-described-right

### Participation

tmcw:
#### Description

- Closes #1166 
- This fixes an issue in which LastUpdated was server-rendered using the server's timezone, yielding incorrect dates.


kevinzunigacuellar:
LGTM!

tmcw:
👍 Added a changeset.

delucis (commented):
Thanks for the PR @tmcw! And sorry it got buried as we were prepping for Astro v4 and migrating Astro’s Docs to run on Starlight.

Have we tested this works as expected with git-based last updated times as well?

The updated date can come from git or from frontmatter, so ideally it would be great to handle here:

https://github.com/withastro/starlight/blob/7526059df69750f6fe4b6f14f90659906aeedf02/packages/starlight/utils/route-data.ts#L72-L84

This would also mean anyone overriding this component would get the correct behaviour without having to replicate this logic. May not be possible though given we’re providing a `Date` object in route data.

.changeset/new-starfishes-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
```
delucis, line 2:
```suggestion
'@astrojs/starlight': patch
```

delucis (approved):
Yeah, you’re probably right @kevinzunigacuellar! Let’s give it a try and if anyone notices broken behaviour with this change, hopefully they’ll let us know — thanks again @tmcw 🙌

PR #1175, opened 2023-11-30
"Virtual" pages prototype

HiDeoo:

- Changes to Starlight code

#### Description

This pull request is an attempt to:

- Closes #982
- Fix integrations and plugins injecting routes using [`injectRoute()`](https://docs.astro.build/en/reference/integrations-reference/#injectroute-option) being broken starting with Starlight `v0.14.0`

**Note: This PR is a draft as it is not ready to be merged and is still a work in progress regarding most aspects from code, to tests, to documentation which have been mostly put in place to get feedback on the approach which is what I'm mostly interested in before spending more time on this.**

This PR introduces a new component `<VirtualPage />` (name up for discussion if the approach is accepted, I didn't find anything better yet) that can be used by integrations and plugins injecting routes to render dynamically generated content using the Starlight layout.

The preview includes a set of pages generated by a plugin using this approach:

- [basic page with default sidebar](https://starlight-git-fork-hideoo-hd-virtual-pages-astrodotbuild.vercel.app/virtual-pages-demo/a/)
- [page with no sidebar](https://starlight-git-fork-hideoo-hd-virtual-pages-astrodotbuild.vercel.app/virtual-pages-demo/b/)
- [page with custom sidebar and a toc](https://starlight-git-fork-hideoo-hd-virtual-pages-astrodotbuild.vercel.app/virtual-pages-demo/c/)

<details>
<summary>Technical details</summary>

A lot of details are already included in the [feature request](https://github.com/withastro/starlight/discussions/982) but I'll provide some additional details here.

Before this PR, an integration or plugin injecting a new route would need to render the `@astrojs/starlight/components/Page.astro` and provide all the expected route data to it. This was mostly unsafe as there was no way for a plugin to provide all expected data and was a bit cumbersome, e.g. this example from my blog integration:

```ts
export function getPageProps(title: string, slug: string): Props {
  const entryMeta = getEntryMeta();

  return {
    ...entryMeta,
    editUrl: undefined,
    entry: {
      data: {
        head: [],
        pagefind: false,
        title,
      },
      slug,
    },
    entryMeta,
    hasSidebar: true,
    headings: [],
    id: slug,
    lastUpdated: undefined,
    pagination: {
      next: undefined,
      prev: undefined,
    },
    sidebar: [],
    slug,
    toc: {
      items: [],
      maxHeadingLevel: 0,
      minHeadingLevel: 0,
    },
  };
}
```

Another (and maybe more) important point is that this was impossible for a plugin to use the generated sidebar based on the Starlight configuration meaning all pages generated by a plugin would not have a sidebar, e.g. this breaking change in my OpenAPI integration:

> Due to some API changes in Starlight, all OpenAPI documentation pages will no longer display a sidebar. If you wish to preserve a sidebar on these specific pages until this is fixed, please stick to the 0.2.2 release and Starlight <0.11.0 for the time being.

Starting with `v0.14.0`, the [change](https://github.com/withastro/starlight/pull/1135) that exposes localized UI strings in route data and having all internal components using the route data to access localized UI strings makes is now impossible for a plugin to use such an approach as it would require the plugin to provide localized UI strings which is not possible.

To address this, the proposed approach is to introduce a new component `<VirtualPage />` that can (and should) be used by integrations and plugins to render dynamically generated content:

- This component is a wrapper around the `<Page/>` component
- It accepts a different (but slightly similar) set of props
- Some optional props will be automatically provided by the component, e.g. if a page does not provide a custom sidebar and does not explictly hide it, the default sidebar will be used
- It makes generating props a bit easier by removing the complexity of `entry` as they technically do not exist for virtual pages

Rendering a virtual page would look like this:

```astro
---
// plugin/src/Example.astro
import VirtualPage, {
  type VirtualPageProps,
} from '@astrojs/starlight/components/VirtualPage.astro';
import CustomComponent from './CustomComponent.astro';

export function getStaticPaths() {
	// Omitted for brevity...
}

const props = {
  title: 'My custom page',
  slug: 'custom-page/example',
  template: 'doc',
  hasSidebar: true,
  headings: [{ depth: 2, slug: 'description', text: 'Description' }],
  dir: 'ltr',
  lang: 'en',
  pagefind: true,
  head: [],
} satisfies VirtualPageProps;
---

<VirtualPage {...props}>
  <h2 id="description">Description</h2>
  <CustomComponent />
</VirtualPage>
```

If this an approach that is found acceptable, there is still a lot of work to be done as I mentioned earlier but so far I have not been able to find a better approach to address these issues.

</details>


Genteure:
Thank you for working on this feature!

I just made another prototype with this prototype: https://github.com/withastro/docs/pull/5551
I think the approach works pretty well and injected pages are easy to write and add.

From DX point of view, making all props but `title` optional or providing a helper that returns a set of defaults would be nice, since only `title` is required in page frontmatter and it's a little confusing where should I get those values.

A lot of things are not working correctly in https://github.com/withastro/docs/pull/5551 right now, most of I assume are just because this PR is WIP and not implemented yet so I'm not too worried.
One that might need extra attention is, `<Content/>` from content collection entries are rendered as `<p set:html="text here"></p>`.
Preview link: https://docs-n9dghxdqp-astrodotbuild.vercel.app/fr/reference/glossary/

HiDeoo:
Thanks for taking the time to try it out and the super valuable feedback, greatly appreciated! :raised_hands:

After seeing your PR, I understand what you meant the other day on Discord and I'm not sure what I read, but I totally did not imagine that ^^ This makes way more sense to me now :thumbsup:

> From DX point of view, making all props but `title` optional or providing a helper that returns a set of defaults would be nice, since only `title` is required in page frontmatter and it's a little confusing where should I get those values.

My original thinking was that if a plugin emits a page with some content, the plugin must knows the content so it also knows the language, the text direction, if it should be searchable or not, etc. Another point is that I tried to minimize the changes for the initial draft (we basically get what is the minimal expected props to get it working).

That's definitely something to think about more just like the `<Content/>` component you mentioned, this is not something I even tried personally yet.

ematipico:
> My original thinking was that if a plugin emits a page with some content, the plugin must knows the content so it also knows the language, the text direction, if it should be searchable or not, etc.

Although some websites don't have that information, or don't need to provide all of them because they are just adding _a subset_ of the props.

The proposal looks sound to me and it's highly needed because `v0.14.0` breaks the current usage of `Page`.

delucis (commented):
Thanks for the amazing work on this @HiDeoo! And apologies for breaking all your stuff with the i18n labels change — I didn’t think through how that could impact upstream stuff using `<Page>` like this.

On the whole I think this looks like a good direction. I think I tend to agree with people arguing for more default values so a minimal set of props are required — maybe even just `title` and `slug`? I think we have the power to choose sensible defaults for everything else. (For frontmatter, we could even parse it using Starlight’s frontmatter schema to ensure consistency and equivalent defaults.)

One thing I’d push for us to think about is how this fits into our i18n system. Are we able to have consistent fallback behaviour for virtual roots? (Right now other languages 404.) And will virtual routes be fully compatible with the language switcher? Admittedly this could be quite tricky because we’re not actually in charge of the routing — like https://github.com/withastro/roadmap/discussions/688 would allow us to be — so we don’t actually have access to the content. I wonder if we could provide helpers for `getStaticPaths()` that would handle that somehow?

But overall — big thumbs up! If I’m able to champion my roadmap proposal for programmatic content collection entries, we might want to eventually revisit this, but this looks ideal for unblocking this sooner rather than later!

HiDeoo:
Thanks for the review :raised_hands: (and no worries, I think none of us thought about that potential change but imo, this was due to happen at some point anyway, as we have no dedicated story for the `<Page/>` component yet).

Considering the approach seems reasonable, I'll start thinking more about some of the raised points and will try to come up with an update to this PR. Regarding defaults, I'll review all the props and see what we can do, I would guess indeed `title` and `slug` could be the only required ones but I'll double check.

Regarding i18n, this is something I left as an open question for this first draft but indeed, definitely something to consider, although I'm not sure yet what the best approach would be.

packages/starlight/utils/route-data.ts
```diff
@@ -54,6 +61,94 @@ export function generateRouteData({
...
+					z.literal('jpg'),
+					z.literal('jpeg'),
+					z.literal('tiff'),
+					z.literal('webp'),
+					z.literal('gif'),
+					z.literal('svg'),
+					z.literal('avif'),
+				]),
+			}),
+	}).parse(props);
```
HiDeoo, line 149: Is there a better way to do this?

HiDeoo:
Just updated the PR with various changes based on previous feedback.

The `<VirtualPage />` component now only requires a `slug` and a `title` props and everything else is optional and will be inferred when not provided. This allows for a much simpler API and flexible usage.

> For frontmatter, we could even parse it using Starlight’s frontmatter schema to ensure consistency and equivalent defaults.

I refactored the code to use this approach which makes a lot of sense (altho I left a GH comment regarding the `ImageFunction` as I'm not sure if there's a better way to do this).

> One thing I’d push for us to think about is how this fits into our i18n system. [...] I wonder if we could provide helpers for `getStaticPaths()` that would handle that somehow?

This new update includes a `makeVirtualStaticPaths()` and `getLocalizedSlugs()` helper functions to generate dynamic localized routes for multilingual sites, supporting multiple languages and marking some routes as fallback routes. I'm not sure if this is what you had in mind but the approach allows to generate all localized routes from a plugin for monolingual and multilingual sites and marking some routes as fallback routes.

I don't think we can provide a magic helper that would handle all of this automatically while still allowing for flexibility as like you said, we’re not actually in charge of the routing and the content.

I updated the preview examples to use the new helpers (switching language in the preview examples won't 404 now, and will provide content in English and French as well as fallback English content for the other languages) and also added some documentation for it. If this is the approach we end up going with, I think the documentation should be improved but I'm not sure how to do that considering it's fairly "low-level" and also involves a lot of Astro specific concepts where I struggled to find the right balance between being too verbose and repeating some of the Astro documentation vs being too terse, not explaining enough and linking to the Astro documentation.

levino:
As a remark: I think this approach is too complex because of upstream design choices. The problem is that currently the hirarchy looks like this:
```
astro
  -> starlight
    -> blog
```
However I think would be much better if it looked like this
```
astro
  -> starlight
    -> docs
    -> blog
```
So basically starlight itself would be more of a basic website boilerplate which would only support pages with a nice header and footer which I potentially could customize with additional links, etc. Then "docs feature" would enable a route `/docs` with the docs and the "blog feature" would enable a route `/blog` with a blog functionality. I think right now a lot of complexity and fuctionality for the docs feature is bleeding into the code of the blog feature. I think docusaurus has solved these issues in a very nice way and one really should adapt their API for configuration of [docs](https://docusaurus.io/docs/docs-introduction) and [blog](https://docusaurus.io/docs/blog). I cannot stress how much better I like their API over the API of astro at the moment. And I really think it would remove a lot of complexity removing the tight coupling of "theme / website boilerplate" (header, footer, navigation, styles) and "docs" (autogenerating sidebar navigation + some docs and i18n magic).

delucis (commented):
Thanks again for your patience on this one @HiDeoo! Some big picture feedback

- As [discussed earlier today](https://discord.com/channels/830184174198718474/1197146528289136772), let’s remove the `makeVirtualStaticPaths()` routing helper for now to focus on the component. Once that’s shipped we can think about ways to help people with routing and i18n in particular.

- What do you think about centralising all the virtual page code in one place for now rather than mixing it into several modules? Basically wondering if instead of having some of our existing things extend the virtual versions (e.g. `schema.ts` and `utils/routing.ts`) we can flip that and omit the things unavailable to the virtual version? That way a single file could include the schema, types, and route data helpers and avoid leaking virtual page stuff into the default logic. Might not work as a refactor, but would be nice if it does, so all the feature code is together!

- Re: docs. Given we’re removing the routing part I wonder if we can also simplify documenting this as “This is a component. Its props are the same as frontmatter, except for…” to avoid repeating the same things across frontmatter reference and this component? So the virtual page docs would only document the _differences_.

- Because of this I do wonder whether keeping the frontmatter/other props separation would be good idea. Making everything top level is admittedly nice for the simple `{ title, slug }` case, but I wonder if `{ slug, headings, hasSidebar, frontmatter: { title, description } }` helps both with documentation and also helps clarify what the “extra” properties like `sidebar` etc. are. (It could also be called `data` instead of `frontmatter` for consistency with the content collection type.) What do you think? `sidebar` in particular could catch people out as we do have the `sidebar` frontmatter property, but in this context the `sidebar` prop means something else.

HiDeoo:
Thanks for the review, really appreciate it and very great comments as usual 🙌 

> - As [discussed earlier today](https://discord.com/channels/830184174198718474/1197146528289136772), let’s remove the `makeVirtualStaticPaths()` routing helper for now to focus on the component. Once that’s shipped we can think about ways to help people with routing and i18n in particular.

This is now gone.

> - What do you think about centralising all the virtual page code in one place for now rather than mixing it into several modules? Basically wondering if instead of having some of our existing things extend the virtual versions (e.g. `schema.ts` and `utils/routing.ts`) we can flip that and omit the things unavailable to the virtual version? That way a single file could include the schema, types, and route data helpers and avoid leaking virtual page stuff into the default logic. Might not work as a refactor, but would be nice if it does, so all the feature code is together!

I'm fine with that, this keeps the virtual page code in one place and makes the diff way less scary as I mostly add 1 or 2 `export` to the mentioned files.

Type-wise, I think the approach I used before and the omit approach now used are the more practial ones, as picking instead of omitting would require to constantly making sure we pick new fields in the future. Now, `VirtualPageProps` is as most as possible based on `PageProps`.

Altho, the omit approach is not the easiest to read either and required a bit of "trickery" as we needed to remove an index signature and the final type was pretty much unreadable when hovering it so I added a small helper to "prettify" it so people hovering it from a plugin can have a better idea of what's inside.

| Before | After |
| :----: | :---: |
|  SCR-20240123-nutq   |  SCR-20240123-nund  |

> Because of this I do wonder whether keeping the frontmatter/other props separation would be good idea.

The frontmatter props are now in a `frontmatter` object, way cleaner and indeed, the documenation is now way more readable. I tried to emphasize the fact that `title` is required too with the new approach (not sure if the way I listed the differences is clear enough tho).

> Couldn’t a virtual page theoretically provide a URL where it can be edited? e.g. if you knew a page was generated from a specific code file, you could link to that?

Excellent catch, this is now supported.

---

Question unrelated to the review: as we're moving towards with this solution, what do we do regarding the name? I picked "Virtual pages" as I failed to find a better name but I'm not sure at all it's the best one ^^

packages/starlight/utils/virtual-page.ts
```diff
@@ -0,0 +1,169 @@
...
+					z.literal('jpg'),
+					z.literal('jpeg'),
+					z.literal('tiff'),
+					z.literal('webp'),
+					z.literal('gif'),
+					z.literal('svg'),
+					z.literal('avif'),
+				]),
+			}),
+	}).parse(frontmatter);
```
delucis, line 152: One thing I’m thinking about here is what happens to a user’s schema extensions in `src/content/config.ts`. We don’t really have access to them and this `parse()` call will only use the default schema (although the type will be inferred from their extensions if I followed everything correctly, because it’s based on `StarlightDocsEntry` and therefore `CollectionEntry`).

There are some quick solutions that get us _some_ of the way there, like:

```suggestion
	})
		.passthrough()
		.parse(frontmatter);
```

That would at least preserve additional values passed to the frontmatter prop rather than stripping them. But it wouldn’t solve cases where the frontmatter schema extension includes default values or transforms.

Not sure how big of a concern this is — we could add it as a future to-do I guess. In theory, importing `collections` from a user’s `src/content/config.ts` is do-able and `collections.docs.schema` could give us the full schema instead of using `StarlightFrontmatterSchema` directly.

packages/starlight/utils/virtual-page.ts
```diff
@@ -0,0 +1,169 @@
...
+		? never
+		: symbol extends Tkey
+		? never
+		: Tkey]: TType[Tkey];
+};
+
+// https://www.totaltypescript.com/concepts/the-prettify-helper
+type Prettify<TType> = {
+	[TKey in keyof TType]: TType[TKey];
+} & {};
```
delucis, line 169: This is totally a stylistic nit, but we don’t generally prefix type names with `T` in other code. I think this could even be as simple as:

```suggestion
// https://stackoverflow.com/a/66252656/1945960
type RemoveIndexSignature<T> = {
	[K in keyof T as string extends K
		? never
		: number extends K
		? never
		: symbol extends K
		? never
		: K]: T[K];
};

// https://www.totaltypescript.com/concepts/the-prettify-helper
type Prettify<T> = {
	[K in keyof T]: T[K];
} & {};
```

delucis: Or if we prefer being a bit less terse, we could use the full words:

```suggestion
// https://stackoverflow.com/a/66252656/1945960
type RemoveIndexSignature<Type> = {
	[Key in keyof Type as string extends Key
		? never
		: number extends Key
		? never
		: symbol extends Key
		? never
		: Key]: Type[Key];
};

// https://www.totaltypescript.com/concepts/the-prettify-helper
type Prettify<Type> = {
	[Key in keyof Type]: Type[Key];
} & {};
```

delucis (commented):
Thanks @HiDeoo — a couple of quick comments! I also spun up [a thread on Discord](https://discord.com/channels/830184174198718474/1200406557951795300) to chat about the component naming. @sarah11918 and I so far landed on maybe `<StarlightPage>` being better than “Virtual” (now that the concept is not really including virtual data etc.), but still up for discussing a bit more.

Will think a bit more about how to document this afternoon as well, given the component use case is not strictly speaking only for plugins, maybe it will make sense in some different context.

packages/starlight/utils/virtual-page.ts
HiDeoo: Damn, total oversight on my end, thanks for pointing that out.

The `passthrough()` is indeed a solution and you are correct, the user, e.g. in an override would get the value properly typed (just tested it). Altho, now this makes me think of the same issue from a plugin author perspective:

- A plugin want to use virtual routes.
- This plugin also provide a schema that the user should extend from which add a mew optional value.
- When using a virtual route, the plugin author will fill the props based on `VirtualPageProps` and for the `frontmatter` property, it's based on `StarlightVirtualFrontmatter` which is inferred from the `StarlightVirtualFrontmatterSchema` which extends the `StarlightFrontmatterSchema` which has no knowledge about any potential extension.

If the plugin author want to provide this new additional value, they would get an error when doing something like this:

```ts
const props = {
  slug: 'custom-page/example',
  frontmatter: {
    title: 'My custom page',
    customPluginValue: 'foo',
    // ^^^^^^^^^^^^^^^^^^^^^^
  },
} satisfies VirtualPageProps;
```

I'm wondering if `VirtualPageProps` should be a generic type that would allow the plugin author to provide their own extra props for the frontmatter. Something like:

```ts
type PluginExtend = z.infer<typeof PluginExtendSchema>;

const props = {
  slug: 'custom-page/example',
  frontmatter: {
    title: 'My custom page',
    customPluginValue: 'foo',
  },
} satisfies VirtualPageProps<PluginExtend>;
```

And we would just need to tweak the `VirtualPageProps` type to allow for this generic type, e.g. `VirtualPageProps<T = {}>` and the frontmatter key would become `frontmatter: StarlightVirtualFrontmatter & T;`.

Any thoughts?

delucis: Oh right, just seeing that my assumption about types was wrong above, because `StarlightVirtualFrontmatter` is inferred from the schema, not `StarlightDocsEntry` as I thought (which also makes sense because `StarlightDocsEntry` is an output type not the input type).

I think a generic here might be kind of annoying to use. I think we can probably solve this somehow, perhaps with a virtual module that re-exports a user’s content collections config (or a default config if that is not found):

```ts
// virtual:starlight/collection-config
import { defineCollection } from 'astro:content';
import { docsSchema, i18nSchema } from '@astrojs/starlight/schema';

let userCollections;
try {
	userCollections = await import('resolved/path/to/src/content/config.ts').collections;
} catch {}
if (!userCollections) {
	userCollections = {
		docs: defineCollection({ schema: docsSchema() }),
		i18n: defineCollection({ type: 'data', schema: i18nSchema() }),
	};
}
export const collections = userCollections;
```

I’ll also ask if exporting the [`ContentConfig` type](https://github.com/withastro/astro/blob/a3df9d83ca92abb5f08f576631019c1604204bd9/packages/astro/content-types.template.d.ts#L186) from `astro:content` is possible — that would make it easier to get that type in our code. Currently it generates a line like this in `.astro/types.d.ts`:

```ts
type ContentConfig = typeof import("../src/content/config");
```

If that was exported, we could easily type our virtual module as something like:

```ts
declare module 'virtual:starlight/collection-config' {
	export const collections: import('astro:content').ContentConfig['collections'];
}
```

HiDeoo: Thanks for the input, always appreciated. Before jumping into the code and to make sure I get it right:

- We would import the schema from the virtual module and use `collections.docs.schema` to actually define the `StarlightVirtualFrontmatterSchema` and omit/extend it like we actually do with the `StarlightFrontmatterSchema`.
- We would then have the user schema extension when using `StarlightVirtualFrontmatterSchema` to do the parsing.
- A user using the `VirtualPageProps` type in its own project would have proper typing for the virtual frontmatter including the schema extension.

Altho, I don't think this addresses the issue I mentioned above of a plugin author developing a plugin in isolation  and getting error when trying to add custom property to the frontmatter, correct? We would need something else for that (and I have no idea yet what that could be if not a generic)

delucis: Yes, that all sounds right to me!

> I don't think this addresses the issue I mentioned above of a plugin author developing a plugin in isolation and getting error when trying to add custom property to the frontmatter, correct?

Yes, I think this a similar issue to Starlight’s where types based on `astro:content` are not available (except for the fact we kind of hackily have them around from the docs site thanks to some specifics of our monorepo set-up). That said, I _think_ because the type in question here would fail to resolve, it would be `any` for plugin authors, so shouldn’t cause errors?

HiDeoo: Just played with it a bit locally and here are my first impressions:

- Getting hold of the schema is relatively easy indeed.

- We cannot use it yet to validate the data as when extended, we are using an [intersection](https://github.com/withastro/starlight/blob/main/packages/starlight/schema.ts#L156) which prevents us to refine the schema as stated in the Zod docs:

  > `A.and(B)` returns a less useful `ZodIntersection` instance that lacks common object methods like `pick` and `omit`.

  Not sure yet if we can work around this, I'll need to dig a bit more.

- > That said, I think because the type in question here would fail to resolve, it would be `any` for plugin authors, so shouldn’t cause errors?

  I think this is correct but I wonder if it's better or worse. As a plugin author, I would at least expect to get auto completion/errors for at least the known properties. :thinking: Need to think about this a bit more.

HiDeoo: Small update after more investigations, it looks like the following could work:

```ts
const StarlightVirtualFrontmatterSchema = (context: SchemaContext) => {
  const docsSchema = collections.docs.schema!;
  const schema =
    typeof docsSchema === "function" ? docsSchema(context) : docsSchema;

  return schema
    .and(
      z.object({
        editUrl: z.union([z.string().url(), z.literal(false)]).default(false),
      })
    )
    .transform((frontmatter) => {
      const { sidebar, ...others } = frontmatter;
      return others;
    });
};
```

After applying your change from the Astro PR and adding the types for `virtual:starlight/collection-config`, this works as expected **in the Starlight code base**. I get proper typing when the Docs schema is not extended and the same when extended, I can see the extended properties and everything. Same goes for the validation, it works as expected from my initial tests.

So now, even if I simplify the schema by returning it unmodified with no `.and()` and `.transform()` and directly exporting its type:

```ts
export type StarlightVirtualFrontmatter = z.input<
  ReturnType<typeof StarlightVirtualFrontmatterSchema>
>;
```

And re-exporting it from `packages/starlight/props.ts` (or in the virtual page component like it's currently done in the PR), I cannot get the type to resolve anywhere even in the Starlight docs when importing it. It's always `any`.

I'll keep investigating and see if there are something that could help me figure out why this is not resolving properly, maybe I'm missing something obvious, not sure.

If you have any idea in the meantime, I'm all ears and I'll keep you posted if I find anything.

delucis: I’m not sure I follow all the ins and outs of code at this point, but if you want to push some WIP code for me to look at, I’d be happy to try playing with it locally!

HiDeoo: Sure, I just pushed a [WIP commit](https://github.com/withastro/starlight/pull/1175/commits/2f6905ed53472f8741cca82321d3b794a36c2856) (tests are breaking, did not look into it yet, just trying to get the basics working).

- The main issue is in `docs/src/components/about-astro.astro` where the `StarlightVirtualFrontmatter` imported type is `any`.
- In `packages/starlight/utils/virtual-page.ts` it is not tho.

Note that it needs the change from https://github.com/withastro/astro/pull/9840

delucis: > It looks like this starts happening when the file being imported by the virtual module in build mode also import astro:content.

You mean the `import { defineCollection } from 'astro:content';` in the `virtual:starlight/collection-config` module? We could probably get rid of that if so — `defineCollection()` is an identity function just used for type checking, I think, so `defineCollection({ schema: foo })` is actually identical to `{ schema: foo }`.

(Maybe I misunderstood though — I’m too sleepy on this train and also agree you deserve a break 😁)

HiDeoo: I think it's more the `import { defineCollection } from 'astro:content';` in the `src/content/config.ts` file which is imported by the virtual module. I can successfully import another file from the virtual module, e.g. `src/content/test.ts` which contains random code, but as soon as this file import `astro:content` like the config is doing, the build will not build any page (but not fail).

I'll investigate more tomorrow. Have a good ride home!

HiDeoo: Ok, finally fixed and we should have virtual pages in the preview. For one part, I definitely needed a break, for the other part, I'm not quite sure yet but I had to move outside of the virtual module the fallback to our default schema if we could not import the user config.

I pushed the changes and also addressed your other feedback regarding type naming, I don't care at all, just an old habit ^^

delucis (commented):
Just had a play around with the APIs and they worked great! Lovely stuff @HiDeoo ✨ 

Some notes:

- I did run into the `frontmatter` prop types being missing without `astro:content` types available. Definitely a small annoyance, but I think kind of inevitable (and not something users would experience directly mostly as they would have these types in their projects if they were using the component).

- I think we can improve error messages for malformated frontmatter using https://github.com/withastro/starlight/blob/main/packages/starlight/utils/error-map.ts — but this can happen in a follow-up PR rather than adding extra work to this one, it’s just a nice quality of life add-on.

- We can go ahead and work on the renaming from `VirtualPage` to `StarlightPage` as previously discussed, and cleaning up this PR to remove the demos I think.

- I’ll take a look at the docs in a bit and think about where is best for us to present the new possibilities.


Thanks again for all the dedicated work and patience on this PR @HiDeoo — really not an easy one and I think we’ve arrived somewhere super helpful!

HiDeoo:
> I did run into the `frontmatter` prop types being missing without `astro:content` types available. Definitely a small annoyance, but I think kind of inevitable (and not something users would experience directly mostly as they would have these types in their projects if they were using the component).

Yeah, we talked a bit about this one but like you said, I think we don't have a way to provide some sort of proper workaround for this. And indeed, this should only hit plugin developers and even in this case, if you use a monorepo approach similar to what we are doing, this could be mitigated.

> * I think we can improve error messages for malformated frontmatter using [`main`/packages/starlight/utils/error-map.ts](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/error-map.ts?rgh-link-date=2024-02-09T12%3A19%3A24Z) — but this can happen in a follow-up PR rather than adding extra work to this one, it’s just a nice quality of life add-on.

Excellent catch. Considering we still have some clean-up, renaming, doc'ing, etc. I think I would prefer to do that now as this should not be too much work and it would be a nice improvement.

---

I'll start working on the renaming and cleanup (and update this branch, did not even notice it has conflicts since I last worked on it) and let you know when I'm done.

HiDeoo:
Just updated the PR with the following changes:

- Updated the development and starter templates version of Astro to `4.3.5`
  - This would have been annoying to see red squiggly lines everywhere while renaming things without https://github.com/withastro/astro/pull/10022
  - Maybe this could be done in another PR too
- `VirtualPage` is now `StarlightPage`
  - The only remaining references to "virtual" in the code and docs should be virtual modules.
- I added better validation errors as suggested
  - I moved the `throwValidationError()` function we had for plugins to the file containing the `errorMap` as they are related and we often would want to use them together
StarlightPage frontmatter validation error preview

- I removed the `virtual-pages-demo` package
  - This is done last in a separate commit so it's easy to check it out and play with it with the renamed `StarlightPage` component

On the front of remaining tasks, I so far noted the following:

- Docs obviously
  - Small note: while updating the existing docs to use `StarlightPage` instead of `VirtualPage`, I noticed that not having a specific keyword (not saying `virtual` was good, I'm the first one saying it's not, but something different from `Starlight`) may be confusing at some point. For example, if we ever say "pages generated by Starlight", a user might assume it's talking or including `StarlightPage`.
- For sanity, I think I may want to test `virtual-pages-demo` (by checking out the commit removing it) when the language server is updated and contains the latest Volar fixes as seeing the red squiggly lines for the import was not reassuring even tho it should be fine.
- Add a changeset
  - I thought the changeset may include a link to the new docs for the `StarlightPage` component so I figured it would be better to wait for the docs to be done and stabilized before adding the changeset

delucis:
>  * Updated the development and starter templates version of Astro to `4.3.5`
>     * Maybe this could be done in another PR too

I think it’s fine as part of this PR — it’s internal and we likely won’t need it urgently for other work I don’t think.

> I added better validation errors as suggested

Looks amazing! 🤩

> For sanity, I think I may want to test `virtual-pages-demo` (by checking out the commit removing it) when the language server is updated and contains the latest Volar fixes as seeing the red squiggly lines for the import was not reassuring even tho it should be fine.

This should be fixed and released in https://github.com/withastro/language-tools/releases/tag/astro-vscode%402.7.3 🎉 

> I thought the changeset may include a link to the new docs for the `StarlightPage` component so I figured it would be better to wait for the docs to be done and stabilized before adding the changeset

Sounds good 👍 

>  * Docs obviously
>       
>     * Small note: while updating the existing docs to use `StarlightPage` instead of `VirtualPage`, I noticed that not having a specific keyword (not saying `virtual` was good, I'm the first one saying it's not, but something different from `Starlight`) may be confusing at some point. For example, if we ever say "pages generated by Starlight", a user might assume it's talking or including `StarlightPage`.

Will look at docs now. Yes, I can see that concern — also perhaps in search results. Will keep it in mind as I review.

delucis:
Took a pass at new docs for this feature:

- Added a new “Pages” guide
- Moved a lot of the “Add content” copy from the “Getting Started” guide to the new page as part of docs for “content pages” (not sure if that’s the best name)
- Wrote a “custom pages” section which for `src/pages/` routes, and moved the plugin reference docs about `<StarlightPage>` there
- Did a pass at the prop docs standardising verb tense and tweaking a few things.

HiDeoo:
Pushed a new update following the conversation on Discord:

- The `slug` is now inferred from `Astro.url`
- Updated the associated doc (not sure if we should go into more details or if this what you had in mind)
- Is it correct that we cannot properly stub `import.meta.env.BASE_URL` in our tests?

> > For sanity, I think I may want to test `virtual-pages-demo` (by checking out the commit removing it) when the language server is updated and contains the latest Volar fixes as seeing the red squiggly lines for the import was not reassuring even tho it should be fine.
> 
> This should be fixed and released in [withastro/language-tools@`astro-vscode%402.7.3` (release)](https://github.com/withastro/language-tools/releases/tag/astro-vscode%402.7.3) 🎉

Took the opportunity to test this, works like a charm!

packages/starlight/__tests__/basics/slugs.test.ts
```diff
@@ -76,3 +77,32 @@ describe('localizedSlug', () => {
...
+		expect(urlToSlug(new URL('https://example.com/dir/sub-dir/page/'))).toBe('dir/sub-dir/page');
+	});
+
+	test('returns slugs with `build.output: "file"`', () => {
+		expect(urlToSlug(new URL('https://example.com/slug.html'))).toBe('slug');
+		expect(urlToSlug(new URL('https://example.com/dir/page/index.html'))).toBe('dir/page');
+		expect(urlToSlug(new URL('https://example.com/dir/sub-dir/page.html'))).toBe(
+			'dir/sub-dir/page'
+		);
+	});
```
delucis, line 94: Might also be worth testing `https://example.com/` and `https://example.com/index.html`?

HiDeoo:
> Unfortunately, yes, this is correct. (Or at least was when I last tested.)

Thanks for the confirmation, this is pretty much what I assumed altho turns out this may have changed and my understanding was based on an pretty weird behavior:

- To test my `vi.stubEnv('BASE_URL', '/base/');`, I used a VSC snippet to insert the following code in `urlToSlug()`
  ```javascript
  console.log('🚨 [slugs.ts:108] import.meta.env.BASE_URL:', import.meta.env.BASE_URL);
  ```
- The log while running this specific test was showing me that the base was `/` so I assumed that the `vi.stubEnv` was not working.
- Turns out that if you replace the log with the following code:
  ```javascript
  console.log('🚨 [slugs.ts:108] SOMETHING ELSE:', import.meta.env.BASE_URL);
  ```
  The log will show that `import.meta.env.BASE_URL` is `/base/`.

Super sketchy behavior so I think it's safer right now to keep this test as a `test.todo` and introduce `base` tests at a later time? Altho, the good news is that the `test.todo` test for the `base` seems to be working as expected.

delucis:
Wow, now that is odd behaviour 😅 Thanks for testing though — I’ll mention this to the team as well in case there’s a simple way to unblock `stubEnv` support. Blu might know.

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+title: Hello, World!
+description: This is a page in my Starlight-powered site
+---
+```
+
+If you forget anything important, Starlight will let you know.
+
+## Custom pages
+
+For advanced use cases, you can create custom pages in Astro’s `src/pages/` directory.
```
sarah11918, line 54:
```suggestion
For advanced use cases, you can create custom pages by adding Astro’s reserved `src/pages/` directory.
```
reserved? special? But in any case, the point I think we really need to get across is from a default Starlight site, you'll have to MAKE this folder yourself!

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+This is helpful if you need to build pages with a completely custom layout or generate a page from an alternative data source.
+The `src/pages/` directory uses file-based routing and includes support for `.astro` files.
```
sarah11918, line 56: Just checking: ONLY `.astro` pages? Not HTML, MDX etc. Because that's what this implies.

sarah11918, line 56:
```suggestion
The `src/pages/` directory uses [Astro's file-based routing](https://docs.astro.build/en/basics/astro-pages/#file-based-routing) and includes support for `.astro` files.
```

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+**default:** the language of the current locale
+
+Set the BCP-47 language tag for this page’s content, e.g. `en`, `zh-CN`, or `pt-BR`.
+
+##### `isFallback`
+
+**type:** `boolean`  
+**default:** `false`
+
+Indicate if this page is untranslated in the current language and using [fallback content](/guides/i18n/#fallback-content).
```
sarah11918, line 137:
```suggestion
Indicate if this page is using [fallback content](/guides/i18n/#fallback-content) because there is no translation for the current language.
```
I feel like the important part here is that it's using fallback content (and the untranslated bit is secondary). So I might do something like this.

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+- The [`slug`](/reference/frontmatter/#slug) property is not supported and is automatically set based on the custom page’s URL.
+- The [`editUrl`](/reference/frontmatter/#editurl) option requires a URL to display an edit link.
+- The [`sidebar`](/reference/frontmatter/#sidebar) property is not supported. In Markdown frontmatter, this option allows customization of [autogenerated link groups](/reference/configuration/#sidebar), which is not applicable to pages using the `<StarlightPage />` component.
+
+##### `sidebar`
+
+**type:** `SidebarEntry[] | undefined`  
+**default:** the sidebar generated based on the [global `sidebar` config](/reference/configuration/#sidebar)
+
+Provide a custom site navigation sidebar for this page.
```
sarah11918, line 100: Just pointing out that the structure of all of these definitions is an action the user might want to take, vs e.g.

"A property indicating whether Starlight's default global sidebar is used" or "The sidebar entry to use"

Also, I have no idea how I would set this. I think an example here would be helpful to show what that looks like.

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+If not set, the page will use the default global sidebar.
+
+##### `hasSidebar`
+
+**type:** `boolean`  
+**default:** `false` if [`frontmatter.template`](/reference/frontmatter/#template) is `'splash'`, otherwise `true`
+
+Set if the sidebar should be displayed on this page.
```
sarah11918, line 108:
```suggestion
Whether or not the sidebar should be displayed on this page.
```

Similarly, as a more concrete example, this would be another way to frame all these options. Not a deal breaker, just pointing out that these rely on the type and default to do all the defining here, vs explainy text. Simply observing that it stood out to me!

sarah11918 (commented):
Gotta board a plane, so here are thoughts! Love this new page!

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
+---
+title: Pages
+description: Learn how to create and manage your documentation site’s pages with Starlight.
+sidebar:
+  order: 1
+---
+
+Starlight generates your site’s HTML pages based on your content, with flexible options provided via Markdown frontmatter.
+In addition, Starlight projects have full access to Astro’s powerful page generation tools.
```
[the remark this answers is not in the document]
kevinzunigacuellar: I was thinking we could have a local content and remote content section in the new "pages" page. My assumption is that Virtual pages main use case is tied with fetching content and displaying it in an Starlight layout.

docs/src/content/docs/guides/pages.mdx
```diff
@@ -0,0 +1,137 @@
...
+This guide shows how page generation works in Starlight.
+
+## Content pages
```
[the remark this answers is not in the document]
delucis: Yeah the naming was hard here. My thinking was that these are in `src/content/`, but yeah it’s not perfect. They’d probably _both_ be “docs pages”, but maybe something like “autogenerated” could work… 🤔

docs/src/content/docs/guides/pages.mdx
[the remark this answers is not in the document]
delucis: Yeah the use cases are:
- fetching content from an external source
- generating content from non-Markdown local data
- plugins injecting a route that do the above
- probably something else I haven’t though of yet

docs/src/content/docs/guides/pages.mdx
delucis: No, all formats Astro supports are supported, this is what the “includes” was intended to imply. I didn’t want to duplicate the full list from Astro docs when people should use that as the source of truth.

docs/src/content/docs/guides/pages.mdx
delucis: > the structure of all of these definitions is an action the user might want to take

That was intentional. Is it a good or bad thing? 😅 

+1 on an example being helpful.

docs/src/content/docs/guides/pages.mdx
delucis:
```suggestion
Control whether or not the sidebar should be displayed on this page.
```

docs/src/content/docs/guides/pages.mdx
delucis: Ah, right. That’s true that it might not exist. The suggestion makes it sound like just adding a directory creates pages — will have a thing about wording.

docs/src/content/docs/guides/pages.mdx
delucis: Ah, so the example prompt was very helpful! Turns out using this is not very fun, which I hadn’t tested 😅 

Basically the types are quite strict and there’s none of the helpful massaging that you get if you’re setting a sidebar in the Starlight config.

An example:

```astro
<StarlightPage
	frontmatter={{ title: 'Demo Page' }}
	sidebar={[
		{
			type: 'link',
			href: '/',
			label: 'Home',
			isCurrent: false,
			badge: undefined,
			attrs: {},
		},
		{
			type: 'group',
			label: 'Pages',
			entries: [
				{
					type: 'link',
					href: '/demo/',
					label: 'Demo',
					isCurrent: true,
					badge: undefined,
					attrs: {},
				},
			],
			badge: undefined,
			collapsed: false,
		},
	]}
/>
```

There’s a lot of boilerplate here, it could probably be more like this with some nice defaults:

```astro
<StarlightPage
	frontmatter={{ title: 'Demo Page' }}
	sidebar={[
		{ href: '/', label: 'Home' },
		{
			label: 'Pages',
			entries: [
				{ href: '/demo/', label: 'Demo', isCurrent: true },
			],
		},
	]}
/>
```

Wondering if we could just hold off documenting `sidebar` and ship a follow-up release that adds a nicer API and we document it then

delucis: Pulled the trigger on that — will follow up in a future PR.

HiDeoo: Random thoughts: If we don't like the API, wouldn't it be better to do the change now rather than later and force all plugins and `<StarlightPage/>` users to manually do the change? This will definitely be used if released as-is or people will ask for it considering all the broken plugins or previous `<Page>` users like Biome were specifying custom sidebars and it's been broken since October. I'm not sure if it's a good idea to release it as-is after all this time and then change it to an entirely different API in a few PRs.

delucis (approved):
OK! Let’s go! Can’t believe we’re finally here!

@HiDeoo, you have poured your heart and soul into this one and I want to express my gratitude for your patience, thoughtfulness, and care as we inched this to this point. Amazing dedication 🌟

PR #1162, opened 2023-11-28
Add `extend` options to docs and i18n schemas

delucis:
#### Description

- Adds an `extend` option to both the docs and i18n content collection schemas, which allows users to add additional fields to the default schemas.
- Adds docs for for the docs schema to the frontmatter reference. Added docs:
	- [New frontmatter section](https://starlight-git-dx-755-schema-extensions-astrodotbuild.vercel.app/guides/authoring-content/#frontmatter) in Authoring Markdown content guide
	- [New extending docs schema reference](https://starlight-git-dx-755-schema-extensions-astrodotbuild.vercel.app/reference/frontmatter/#customize-frontmatter-schema)
	- [New extending i18n schema docs](https://starlight-git-dx-755-schema-extensions-astrodotbuild.vercel.app/guides/i18n/#extend-translation-schema) (annoyingly Vercel seems not to be rebuilding? So no preview for this one just yet)

PR #1187, opened 2023-12-01
i18n(fr): update `guides/i18n.mdx`

HiDeoo:

#### Description

This pull request update the french version of the `guides/i18n.mdx` page with the changes of #1162.


dreyfus92 (approved):
ÇMSB

delucis (approved):
Looks good to me!

Interesting that "donné" was used in the old translation given that the i18n is technically a "data collection" rather than a content collection, but I guess we always call the feature "content collections", so the consistency makes sense 👍

PR #1190, opened 2023-12-01
i18n(zh-CN): translate `i18n.mdx`

Genteure:
#### Description

Update translation for `guides/i18n.mdx`.
Corresponding source file changes:

- c6a4bcb7982c54c513f20c96a9b2aaf9ac09094b #742
- 4b3143b34076a0306fe88eb9d27977d38390295c #1102 
- 00d101b159bfa4bb307a66ccae53dd417d9564e0 #1162

I also made some formatting changes and now the translated file is matching the English version line by line.

### DIFF
--- a/docs/src/content/docs/zh-cn/guides/i18n.mdx
+++ b/docs/src/content/docs/zh-cn/guides/i18n.mdx
@@ -45,7 +45,8 @@ Starlight 提供了内置的多语言支持，包括路由、回退内容和完�
 
    你的 `defaultLocale` 将用于回退内容和 UI 标签，所以选择你最有可能开始编写内容的语言，或者已经有内容的语言。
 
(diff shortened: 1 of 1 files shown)


HiDeoo (changes requested):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). Thanks for clarifying for `schema`, I think you're the best to know if it's relevant/better to keep it in English or not and if at some point someone wants to translate it, this could be easily discussed in a new PR.

One tiny details, this is missing the highlight of the very first code block added in https://github.com/withastro/starlight/pull/1102/files#diff-4c65fc770580b991c60e2b800f0c7ed48935c7bf31959943e2118b7e243ed7c1R14 and unfortunately I cannot suggest a change as it's oustide of the reviewable area.

Except that, everything looks good to me 👍

HiDeoo (approved):
Thanks for the quick fix.

PR #1194, opened 2023-12-01
i18n(fr): add `guides/site-search.mdx`

HiDeoo:

#### Description

This pull request adds the french version of the `guides/site-search.mdx` page.

### DIFF
--- a/docs/src/content/docs/fr/guides/site-search.mdx
+++ b/docs/src/content/docs/fr/guides/site-search.mdx
@@ -1,0 +1,166 @@
+---
+title: Recherche
+description: Découvrez les fonctionnalités de recherche intégrées à Starlight et comment les personnaliser.
+sidebar:
+  badge: Nouveau
+---
+
+import { Tabs, TabItem } from '@astrojs/starlight/components';
+
+Par défaut, les sites utilisant Starlight incluent une recherche plein texte utilisant [Pagefind](https://pagefind.app/), un outil de recherche rapide et à faible bande passante pour sites statiques.
+
+Aucune configuration n'est requise pour activer la recherche. Créez et déployez votre site, puis utilisez la barre de recherche dans l'en-tête du site pour trouver du contenu.
+
+## Cacher du contenu dans les résultats de recherche
+
+### Exclure une page
+
+Pour exclure une page de votre index de recherche, ajoutez [`pagefind: false`](/fr/reference/frontmatter/#pagefind) au frontmatter de la page :
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/site-search.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+Par défaut, les sites utilisant Starlight incluent une recherche plein texte utilisant [Pagefind](https://pagefind.app/), un outil de recherche rapide et à faible bande passante pour sites statiques.
+
+Aucune configuration n'est requise pour activer la recherche. Créez et déployez votre site, puis utilisez la barre de recherche dans l'en-tête du site pour trouver du contenu.
+
+## Cacher du contenu dans les résultats de recherche
+
+### Exclure une page
+
+Pour exclure une page de votre index de recherche, ajoutez [`pagefind: false`](/fr/reference/frontmatter/#pagefind) au frontmatter de la page :
```
delucis, line 18: Is it at all common to use `&nbsp;` instead of a regular space before `:` in French texts? Or would something like that usually be taken care of by a plug-in? Noticing on mobile how two of these wrap onto a new line, which looks kind of funny.

delucis (approved):
LGTM! Left a comment but it can be tackled in a follow-up if you like to not block getting this page translated. Thanks for the translation!

docs/src/content/docs/fr/guides/site-search.mdx
HiDeoo: So I would say this depends on the context. 

Technically, we do use a non-breaking space before a colon in French. Altho, most of the time, nobody does, care about it or even know how to do it (at least in my experience) except for people who work in publishing/editing/studies/etc. This is also something a bit annoying to check during reviews on GH if this is enforced.

Now back to CommonMark and Remark, I could technically use ` ` ([`0xA0`](https://www.compart.com/en/unicode/U+00A0)) as nobody would want to explicitly write `&nbsp;` in their markdown. This would work in a Markdown file but not in a `.mdx` file so I personally never do that for consistency.

delucis: Good to know! I guess I was thinking of a remark plugin which looks for text strings that end in ` :` and automatically replaces this with `&nbsp;:` more than manually editing this (and I guess could also be expanded to ` !` and ` ?` for example, although already mentioning that makes me realise there are even more cases to cover like ` : ` in the middle of a paragraph, so maybe too much to think about 😅

PR #1195, opened 2023-12-01
i18n(fr): add `reference/plugins.md`

HiDeoo:

#### Description

This pull request adds the french version of the `reference/plugins.md` page.


docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -0,0 +1,162 @@
...
+
+Les modules d'extension Starlight peuvent personnaliser la configuration, l'interface utilisateur et le comportement de Starlight, tout en étant faciles à partager et à réutiliser.
+Cette page de référence documente l'API à laquelle ces modules d'extension ont accès.
+
+Consultez la [référence de configuration](/fr/reference/configuration/#plugins) pour en savoir plus sur l'utilisation d'un module d'extension Starlight.
+
+## Référence rapide de l'API
+
+Un module d'extension Starlight a la forme suivante.
+Voir ci-dessous pour plus de détails sur les différentes propriétés et paramètres des hooks.
```
HiDeoo, line 16: [Reference](https://fr.wikipedia.org/wiki/Hook_(informatique)) for keeping "hook" as other would be literal translations without a lot of meaning in this context.

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -0,0 +1,162 @@
...
+**Type :** `StarlightUserConfig`
+
+Une copie en lecture seule de la [configuration de Starlight](/fr/reference/configuration) fournie par l'utilisateur.
+Cette configuration peut avoir été mise à jour par d'autres modules d'extension configurés avant celui en cours.
+
+#### `updateConfig`
+
+**Type :** `(newConfig: StarlightUserConfig) => void`
+
+Une fonction de rappel pour mettre à jour la [configuration de Starlight](/fr/reference/configuration) fournie par l'utilisateur.
```
HiDeoo, line 63: [Reference](https://developer.mozilla.org/fr/docs/Glossary/Callback_function) for "fonction de rappel"

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Type :** `boolean`
+
+`false` lorsque le serveur de développement démarre, `true` lorsqu'un rechargement est déclenché.
+
+#### `logger`
+
+**Type :** `AstroIntegrationLogger`
+
+Une instance du [journaliseur (logger) d'intégration Astro](https://docs.astro.build/fr/reference/integrations-reference/#astrointegrationlogger) que vous pouvez utiliser pour écrire des messages de journalisation.
```
HiDeoo, line 142: [Reference](https://fr.wiktionary.org/wiki/journaliseur) for "journaliser" (but I left the English version as it's not a common word)

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -0,0 +1,162 @@
...
+      logger.info("Démarrage d'un long processus…");
+      // Un long processus…
+    },
+  },
+};
+```
+
+L'exemple ci-dessus affichera un message qui inclut le message d'information fourni :
+
+```plaintext frame="terminal"
```
HiDeoo, line 160: This change is to avoid weird highlighting due to the `'` character:

image

delucis (approved):
Looks great! And nice catch with that terminal frame and the `'` highlighting.

Only thing I want 100% sure of was whether `plugin-longue-processus` would be more natural than `longue-processus-plugin`, but I'm probably wrong and it's totally unimportant in any case 😁

HiDeoo:
> Only thing I want 100% sure of was whether `plugin-longue-processus` would be more natural than `longue-processus-plugin`, but I'm probably wrong and it's totally unimportant in any case

I went with the more programmatic approach but re-reading your suggestion, I think you're totally right and it makes more sense for a reader in the context of the docs. I'll make a tiny PR to change it.

PR #1217, opened 2023-12-03
i18n(ja): Add plugins.md

morinokami:
#### Description

- i18n(ja): Add ja/reference/plugins.md.


morinokami:
@HiDeoo Thanks for pointing that out, updated the link!

PR #1218, opened 2023-12-03
i18n(ja): Add overrides.md

morinokami:
#### Description

- Add ja/reference/overrides.md.

### DIFF
--- a/docs/src/content/docs/ja/reference/overrides.md
+++ b/docs/src/content/docs/ja/reference/overrides.md
@@ -1,0 +1,356 @@
+---
+title: オーバーライド
+description: Starlightのオーバーライドでサポートされているコンポーネントとコンポーネントプロパティの概要。
+tableOfContents:
+  maxHeadingLevel: 4
+---
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/ja/reference/overrides.md
```diff
@@ -0,0 +1,356 @@
...
+#### `entry`
+
+現在のページのAstroコンテンツコレクションのエントリー。`entry.data`には、現在のページのフロントマターの値が含まれます。
+
+```ts
+entry: {
+  data: {
+    title: string;
+    description: string | undefined;
+    // etc.
```
HiDeoo, line 85: Just to double check, is this fine untranslated or something missed?

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view).

I left a tiny question regarding one term where I'm not sure if it should be translated or not. Other than that, everything looks good to me 👍

docs/src/content/docs/ja/reference/overrides.md
morinokami: @HiDeoo I intentionally left the expression 'etc.' as it is, thinking that most Japanese people would understand it. However, I now believe it might be more natural to translate it, so I will update it!

HiDeoo (approved):
Thanks for the update 🙌

PR #1226, opened 2023-12-05
Dutch aside text

tlandmangh:
Updated the aside text for note, caution and danger.

#### Description

- Translated the aside text into Dutch.


ElianCodes (commented):
Hey there! Thanks for tackling this!

just left you a suggestion

ElianCodes:
This PR still needs a changeset, which I'll write in the coming hours. Busy now releasing Astro 4.0, but after that I should have time for that!

PR #1232, opened 2023-12-05
i18n(ko-KR): add `plugins.md`

jsparkdev:
#### Description

- add `plugins.md`
- udpate link

### DIFF
--- a/docs/src/content/docs/ko/reference/plugins.md
+++ b/docs/src/content/docs/ko/reference/plugins.md
@@ -65,7 +65,7 @@ Starlight가 초기화될 때 호출되는 플러그인 설정 함수입니다 (
 중첩된 구성 값을 업데이트하려면 전체 중첩 객체를 제공해야 합니다.
 
 기존 구성 옵션을 재정의하지 않고 확장하려면 기존 값을 새 값에 전개하여 확장하세요.
-다음 예시에서는 `config.social`에 전개 연산자를 사용하여 새로운 `social` 객체를 확장합니다. 이를 통해, 새로운 [`소셜`](/ko/reference/configuration/#social) 미디어 계정을 기존 구성에 추가합니다.
+다음 예시에서는 `config.social`에 전개 연산자를 사용하여 새로운 `social` 객체를 확장합니다. 이를 통해, 새로운 [`social`](/ko/reference/configuration/#social) 미디어 계정을 기존 구성에 추가합니다.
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/ko/reference/plugins.md
```diff
@@ -0,0 +1,163 @@
...
+#### `updateConfig`
+
+**타입:** `(newConfig: StarlightUserConfig) => void`
+
+사용자가 제공한 [Starlight 구성](/ko/reference/configuration)을 업데이트하는 콜백 함수입니다. 
+재정의하려는 루트 수준 구성 키를 제공합니다. 
+중첩된 구성 값을 업데이트하려면 전체 중첩 객체를 제공해야 합니다.
+
+기존 구성 옵션을 재정의하지 않고 확장하려면 기존 값을 새 값에 전개하여 확장하세요.
+다음 예시에서는 `config.social`에 전개 연산자를 사용하여 새로운 `social` 객체를 확장합니다. 이를 통해, 새로운 [`소셜`](/ko/reference/configuration/#social) 미디어 계정을 기존 구성에 추가합니다.
```
HiDeoo, line 68:
```suggestion
다음 예시에서는 `config.social`에 전개 연산자를 사용하여 새로운 `social` 객체를 확장합니다. 이를 통해, 새로운 [`social`](/ko/reference/configuration/#social) 미디어 계정을 기존 구성에 추가합니다.
```

I think this should still be `social` as this references the `social` configuration option. Any thoughts? If yes, not sure if my suggestion is valid tho.

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view).

I left a small question regarding a configuration key which I think should not be translated, but except for that, everything looks perfect to me :thumbsup:

docs/src/content/docs/ko/reference/plugins.md
jsparkdev: I was wondering whether to translate everything into Korean for linguistic uniformity of the expression `"social media account"` or whether to use English for this part as it is because I use the `social` option.

I chose the former, but after hearing what you said, I think the latter is the right choice :)

HiDeoo (approved):
Thanks for the quick fix 👍

Issue #1236, opened 2023-12-06
Support for Astro v4


dhruvkb:
### What version of `starlight` are you using?

0.14.0

### What version of `astro` are you using?

4.0.2

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### What browser are you using?

Chrome

### Describe the Bug

The lastest release of Starlight doesn't support Astro v4. I was a bit shocked because the v4 launch post mentioned Starlight being used for the Astro docs. Turns out the docs are still on Astro v3 themselves.

I tried updating Astro to v4 and the app still builds and runs, except with a couple of deprecation warnings.

<img width="1519" alt="Screenshot 2023-12-06 at 7 07 55 AM" src="https://github.com/withastro/starlight/assets/16580576/f3621048-b412-44c3-92c8-05ae9c15dc98">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-fctspu?file=package.json

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #1237, opened 2023-12-06
Specify `entrypoint` in addition to `entryPoint`

dhruvkb:
#### Description

Fixes the warnings logged with Astro v4 as shown in #1236.

Astro v4 deprecated `entryPoint` in favour of `entrypoint`. The migration docs suggest [specifying both](https://docs.astro.build/en/guides/upgrade-to/v4/#what-should-i-do-2) to support older versions of Astro in addition to v4.

### DIFF
--- a/packages/starlight/index.ts
+++ b/packages/starlight/index.ts
@@ -41,11 +41,13 @@ export default function StarlightIntegration({
 
 				injectRoute({
 					pattern: '404',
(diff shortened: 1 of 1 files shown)


delucis:
Thanks for the PR @dhruvkb! We’ll actually need to do a bit more work and drop v3 support, so will close this in favour of #1238, but thanks for getting the ball rolling 🙌

PR #1238, opened 2023-12-06
Add support for Astro v4, drop support for Astro v3

delucis:
#### Description

- Closes #615  
- Updates Starlight’s dependencies to Astro v4 and updates integrations to match.
- Drops support for Astro v3 — unfortunately necessary because of how we bundle MDX in Starlight. Before v1, it might be worth considering whether `@astrojs/mdx` should be a peer dependency we include in starter templates, but not directly as a dependency of `@astrojs/starlight`, so that we can avoid these situations.
- WIP: next step is updating all remark/rehype-related dependencies.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@astrojs/starlight": "workspace:*",
     "@types/culori": "^2.0.0",
-    "astro": "^3.2.3",
+    "astro": "^4.0.1",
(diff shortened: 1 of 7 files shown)


jeanlucaslima:
This is breaking the install command through the template. It will be fixed (I think) if there's a release with up to date library.

delucis:
Yup! Just getting a release out any minute now — thanks for reporting 🙌

PR #1249, opened 2023-12-07
Add an empty i18n folder to templates to avoid warning

Genteure:
#### Description

Currently a default starlight site created with `pnpm create astro --template starlight` logs the following message:

```
[WARN] [content] The i18n collection is defined but no content/i18n folder exists in the content directory. Create a new folder for the collection, or check your content configuration file for typos.
```

The warning message is a common source of confusion.

This PR adds an empty `i18n` folder to both templates by creating a `.gitkeep` file.

I have tested by creating a new starlight project with `pnpm create astro --template starlight` then add the `.gitkeep` file at the same location manually, and no warning messages are printed anymore.


HiDeoo:
While this would avoid the warning (which has been indeed reported multiple times so it's clearly something that should be addressed somehow), my initial thinking is:

- I didn't research this or give it much thought yet, but I wonder if this is the right place for the fix. I mean, if something like Starlight is something that Astro wants to support, e.g. we could imagine a similar concept entirely dedicated to blogging instead of documentation, I wonder if `defineCollection` could not have a flag for marking some collections as entirely optional to avoid the warning or something like that.
- If we go with the suggested solution, I wonder if some teaching is needed. As a new user, I could not know what a `.gitkeep` is, why is it empty? And `git` prefixed file names usually have a special meaning, should I commit it? Should I not? I think we should at least add a comment explaining what this file is for.

lukasborawski:
This one will generate another issue that will require handling.

`Unsupported file type i18n/.gitkeep found.` 

I think this should be configurable from the configuration itself. However, if you're not creating any UI extensions and you don't need translations for it, this feature should be disabled.

delucis:
Raised this with the team today. I think I agree with @HiDeoo that this may not be the ideal fix — requires extra explanation what this file is doing and as @lukasborawski mentioned, causes new logging 😀 

My proposal is actually to not warn in this case in Astro itself. I think the rationale for warning here was to help user catch when their content config accidentally doesn’t match their directory structure (e.g. they configured an `i18n` collection, but made a typo and the directory is named `i81n` or something).

PR #1255, opened 2023-12-08
Add support for SSR

Fryuni:
### Summary

Add configuration for not prerendering Starlight pages.

This enables using components with SSR logic on Starlight pages, whether in the content or as component overrides.

#### Double index

To do this without triggering multiple warnings, there are now two `.astro` pages for the index entrypoint:

- The pre-existing `index.astro` is always prerendered and uses `getStaticPaths` and `Astro.props` to receive information about the route.
- A new `indexSSR.astro` that is _never_ prerendered and retrieves the information about the route from the dynamic route parameter.

#### Route parameter indexing

To make the SSR rendering efficient, a lookup map for all the content was created using the parameter that will be received from the dynamic route.

#### Pagefind

Pagefind indexing requires access to the HTML file generated by the routes, but those are not available when building for SSR, so Pagefind is not supported when using SSR.

~~The solution that I implemented for this is that when both SSR and Pagefind are enabled for Starlight, the content of the docs is pre-rendered for indexing but not included in the final build. The final content served will be from SSR, but the index will be built from that static rendering.~~

#### Default behavior

~~Starlight will default to the same as any other Astro page, prerendering when the output mode is not defined, when it is `static`, and when it is `hybrid`. When the output mode is `server`, Starlight will default to SSR.~~

~~Starlight previously didn't build at all when the site had output mode set to `server`, so the different default is not, AFAICT, a breaking change.~~

Starlight will default to prerender pages, which differs from the default for Astro pages. This is to maintain compatibility and because docs pages are most commonly not dynamic. Users must set `prerender: false` in their Starlight configuration to enable on-demand docs page rendering. At the moment, this applies to _all_ Starlight pages. Future work could be done to support SSR on only some of the pages.

#### TODO

- [x] Disable Pagefind by default and make it incompatible
- [x] Move all `child_process` logic to build plugins (git information)
- [x] Add tests for SSR builds


Fryuni:
For anyone that want to try this out on their website to check for problems you can install it with:

```terminal
npm install https://gitpkg.now.sh/Fryuni/starlight/packages/starlight?ssr-support

yarn add https://gitpkg.now.sh/Fryuni/starlight/packages/starlight?ssr-support

pnpm add https://gitpkg.now.sh/Fryuni/starlight/packages/starlight?ssr-support
```

Fryuni:
Probably blocked by https://github.com/withastro/astro/issues/9392, due to that bug any project using Starlight with SSR cannot add any other page to the project as everything will be overridden by Starlight's `[...slug]` route.

Fryuni:
This depended on withastro/astro#9439, released on 4.2. It's on my pipe to come back to this and make any necessary changes. I'll probably ping people to review it by next Tuesday.

ronildo:
waiting for this to be released 🙏

ryanrussell:
Looking good. Like the decision to exclude ssr pages from pagefind.

Fryuni:
How the new errors look like for reference:
image

Fryuni:
How the new errors look like:
Pagefind conflict

static output

HiDeoo:
Not a review per se as I did not jump into the code yet, but I started testing the PR with different scenarios and I noticed some weird behavior when running the tests in watch mode.
The output seems a bit weird but more importantly, the tests are endlessly running in a loop and making my CPU go crazy ^^


https://github.com/withastro/starlight/assets/494699/940d9dc8-406a-4437-9126-4fd01878cd00

HiDeoo:
Mirroring here some Discord messages here in case someone encounter this:

- @Fryuni did not manage to repro the issue on macOS and Linux.
- This is still happening for me, even tried on a fresh clone. I'll try to investigate a bit more when I get the time.

HananoshikaYomaru:
I want to know does this PR solve pageFind for SSR or just disable it in SSR?

HiDeoo (commented):
Found a bit of time this week-end to play a bit more with the PR and left a few high-level or docs related comments.

An additional question: one of your messages on Discord states that this PR should enable the use of the Cloudflare adapter to deploy a Starlight website to Cloudflare with SSR. I tried to set this up (not a huge Cloudflare user so maybe I did something wrong, altho I mostly followed the adapter guide) but this is breaking for me at build time due to unsupported Node dependencies (which makes sense to me with the current setup but considering your message, is this expected to work?).

packages/starlight/package.json
```diff
@@ -167,17 +170,22 @@
       "default": "./expressive-code.mjs"
     },
     "./index.astro": "./index.astro",
+    "./indexSSR.astro": "./indexSSR.astro",
     "./404.astro": "./404.astro",
+    "./404SSR.astro": "./404SSR.astro",
+    "./pagefind.ts": "./pagefind.ts",
```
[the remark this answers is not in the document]
Fryuni: From a previous solution that allowed Pagefind with SSR, forgot to remove it here

Fryuni:
> An additional question: one of your messages on Discord states that this PR should enable the use of the Cloudflare adapter to deploy a Starlight website to Cloudflare with SSR. I tried to set this up (not a huge Cloudflare user so maybe I did something wrong, altho I mostly followed the adapter guide) but this is breaking for me at build time due to unsupported Node dependencies (which makes sense to me with the current setup but considering your message, is this expected to work?).

It should work. I moved the Node APIs that were needed for git to be used only on build time.

I'll double check to see what else is missing, but the goal of this PR is for it to work on all SSR adapters and have only the same requirements as Astro itself

mmv-dev:
When can we use this feature? Very much needed for my latest project. Just need to be able to deploy the site on cloudflare. Please help.

mmv-dev:
@Fryuni. Thanks for your support. I have a tested this new changes and tried to deploy to Cloudflare. I was able to deploy it. But there is a catch, I have to get a Workers Paid subscription. Because in cloudflare there is limit for function file size (1 MiB), but the basic starlight function size is around 8-9 MiB. In Workers paid plan, the limit is 10 MiB. So i was able to deploy it. But if your planning to implement this in a much bigger project which uses more server side capabilities, this might be something to look out for.

But once again, Thank you very much @Fryuni

Fryuni:
> but the basic starlight function size is around 8-9 MiB (output = "server"). In Workers paid plan, the limit is 10 MiB. So i was able to deploy it. But for someone planning to implement this in a much bigger project which uses more server side capabilities, this might be something to look out for.

Yeah, I don't think there is much for us to do on that front. In server output mode the entire content and all the tooling to render the content has to be shipped into a single function to be processed on each request, so that is a lot of data.

packages/starlight/__tests_e2e__/.gitignore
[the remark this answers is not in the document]
Fryuni: No it isn't

marcelovicentegc:
Hey folks! Thanks for this effort @Fryuni. I just wanted to let you know that I shipped a website with your patched version from `https://gitpkg.now.sh/Fryuni/starlight/packages/starlight?ssr-support` to make Starlight work with https://authjs.dev/ and it worked seamlessly! I'm using Vercel's adapter.

Fryuni:
Hey @marcelovicentegc, thanks for the ping! There is not much blocking this PR; it is just a big change, and the team hasn't had the time to review it yet.

@delucis any chance this gets closer to the top of your to-do list? 🥺

HiDeoo:
To share a small update, I continued over the past few days the list of SSR scenarios I had, which was blocked at Cloudflare last time, and all of them worked without any problems with all the adapters I tried, no issue at all :tada: Amazing work :clap:

I'll continue with a more code-oriented review when I get a bit more free time, but so far, personally, this is looking super great!

packages/starlight/utils/route-data.ts
```diff
@@ -78,11 +77,10 @@ function getLastUpdated({ entry }: PageProps): Date | undefined {
 	const { lastUpdated: configLastUpdated } = config;
 
 	if (frontmatterLastUpdated ?? configLastUpdated) {
-		const currentFilePath = fileURLToPath(new URL('src/content/docs/' + entry.id, project.root));
```
HiDeoo, line 81: Removing this changes the argument being passed down to `getNewestCommitDate()` to no longer being an absolute path.

This means that `getNewestCommitDate()` in dev would use `.` for the `cwd` option of `spawnSync` for a file like the getting started guide of the docs resulting in the following error:

> fatal: ambiguous argument 'getting-started.mdx': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'

And the last update date not being visible in dev mode.

A small reproduction is to just add `lastUpdated: true` to the `docs/astro.config.mjs` file.

HiDeoo (changes requested):
First, I'd like to apologize for the delay in reviewing this PR. It's obviously a very large PR, we have other tasks on the side, we all have our own lives, but I still feel bad for not having enough time to review properly this PR earlier. Thanks a lot for your patience.

Second, and this could have been my first point, I'd like to say "wow". I obviously already played a lot with the PR in the past, and being familiar with your work, I was already expecting a high quality PR, but I have to say that while I was reviewing the code, I was still amazed by the quality of the work you did. :clap:

To give a global overview of my review, I think there is only 1 major issue that would need to be fixed, a few spots that could be refactored to avoid some code duplication, some nitpicks regarding the new example but overall, I think the PR is in a very good shape.

To add a few extra details that I didn't mention in the review:

- I did not review yet the documentation changes (altho they should be pretty straightforward and relatively small)
- Nit: we have the `__tests__/` folder for our unit tests, `__e2e__/` for our Playwright tests so far. `__tests_functional__` feels a bit weird, maybe we could just go with `__functional__/` or `__integration__/` or something like that which would not repeat the `tests` word?
- We should document the new type of tests in the "Testing" section of the `CONTRIBUTING.md` file and just like we did with e2e tests, add some description to help users identify when they really need functional tests and always prefer unit tests when possible.

packages/starlight/__tests__/i18n-root-locale/routing.test.ts
```diff
@@ -82,8 +82,8 @@ test('fallback routes use fallback entry last updated dates', () => {
 
 	expect(getNewestCommitDate).toHaveBeenCalledOnce();
 	expect(getNewestCommitDate.mock.lastCall?.[0]).toMatch(
-		/src[/\\]content[/\\]docs[/\\]guides[/\\]authoring-content.md$/
```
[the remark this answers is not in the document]
Fryuni: Oh right, it's missing the `^` at the beginning.

packages/starlight/index.ts
```diff
@@ -39,20 +39,23 @@ export default function StarlightIntegration({
 
 				const useTranslations = createTranslationSystemFromFs(starlightConfig, config);
 
+				injectRoute({
```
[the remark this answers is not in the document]
Fryuni: Not anymore! This was a workaround for before the routing change I did in core

Fryuni:
> First, I'd like to apologize for the delay in reviewing this PR. It's obviously a very large PR, we have other tasks on the side, we all have our own lives, but I still feel bad for not having enough time to review properly this PR earlier. Thanks a lot for your patience.

It's not a problem at all. It was a fun journey, especially teasing @delucis on Discord (sorry if I overdid it 😅).

Now I'm closer to becoming a first-time contributor to Starlight 🤣🤣🤣

> Second, and this could have been my first point, I'd like to say "wow". I obviously already played a lot with the PR in the past, and being familiar with your work, I was already expecting a high quality PR, but I have to say that while I was reviewing the code, I was still amazed by the quality of the work you did. 👏

Oohh! Thank you 😊
This was my first time using a lot of the tools involved on Starlight (like Vite and Vitest), so looking at it now there is a lot of things to improve. But I did the best I could at the time.

> To give a global overview of my review, I think there is only 1 major issue that would need to be fixed, a few spots that could be refactored to avoid some code duplication, some nitpicks regarding the new example but overall, I think the PR is in a very good shape.

Oh god 🙈 The example was mostly for debugging as I developed, I haven't done anything to make it useful for others yet. I hope you haven't spent too much time on it.

There are many conflicts to resolve, so it's a great opportunity for refactors. I'll work on it this week(end).

> To add a few extra details that I didn't mention in the review:
> 
> * I did not review yet the documentation changes (altho they should be pretty straightforward and relatively small)
> * Nit: we have the `__tests__/` folder for our unit tests, `__e2e__/` for our Playwright tests so far. `__tests_functional__` feels a bit weird, maybe we could just go with `__functional__/` or `__integration__/` or something like that which would not repeat the `tests` word?
> * We should document the new type of tests in the "Testing" section of the `CONTRIBUTING.md` file and just like we did with e2e tests, add some description to help users identify when they really need functional tests and always prefer unit tests when possible.

The use case for functional tests seems quite limited for Starlight. It is for when we want to test the final output of the whole build and serving process (so they can't be unit tests) but don't care about what is rendered in the browser (so it doesn't _need_ to be e2e). Although not strictly necessary, it might be better to turn the functional tests into e2e tests so there is not yet another test bed in the repo. (there was no e2e tests back then)

HiDeoo:
> The example was mostly for debugging as I developed

Oh yeah, and it helped me a lot too. I also think it'll bring a lot of value for people who want to get started with SSR or even just to debug/setup a repro quickly at a fairly low maintenance cost once it's done.

> (there was no e2e tests back then)

I totally forgot about that and that's a very good point. I don't see either a huge need of "functional" test for Starlight on the long run so I really like your idea of refactoring the few you added to be e2e ones as I don't think they need anything Vitest specific. `@playwright/test` would do the job just fine for that very limited usage and nothing prevent us from not interacting with a browser in such tests, we could even [tag](https://playwright.dev/docs/test-annotations#tag-tests) these tests differently if we want.

packages/starlight/utils/route-data.ts
Fryuni: Receiving the absolute path would require inlining it in the SSR bundle, which exposes the project's path on the user machine.

I think it is better to resolve the relative path to the absolute path in the `getNewestCommitDate` when running the dev server.

Fryuni: And actually this breaks for projects using a custom `srcDir`. So I'll have to fix that one together 😅

aaronmondal:
FYI I've rebased this PR to a recent version of starlight (https://github.com/aaronmondal/starlight/pull/1) to test whether this works with Qwik. Works pretty much out-of-the-box :+1:

It might be worth noting that in the experiments I ran with `@astrojs/cloudflare`, the SSR variant with expressive code enabled caused a large server-side bundle size (I think it was ~8MB for shiki or something, and compression/treeshaking didn't seem to have an effect) which ultimately prevented deploying the SSR variant to Clouflare Pages/workers/functions (https://github.com/TraceMachina/nativelink/blob/f5e72760e722a34023e9196073d23fc38443e5ef/docs/astro.config.mjs#L28).

examples/ssr/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+[![Built with Starlight](https://astro.badg.es/v2/built-with-starlight/tiny.svg)](https://starlight.astro.build)
+
+```
+npm create astro@latest -- --template starlight
+```
+
+[![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/github/withastro/starlight/tree/main/examples/basics)
+[![Open with CodeSandbox](https://assets.codesandbox.io/github/button-edit-lime.svg)](https://codesandbox.io/p/sandbox/github/withastro/starlight/tree/main/examples/basics)
+[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/clone?repository-url=https%3A%2F%2Fgithub.com%2Fwithastro%2Fstarlight%2Ftree%2Fmain%2Fexamples%2Fbasics&project-name=my-starlight-docs&repository-name=my-starlight-docs)
```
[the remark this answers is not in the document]
Fryuni: I think that would be best. We could have the example check for `process.env.VERCEL` and switch to the Vercel adapter, but that would be unnecessarily complex for an example for others to start with.

Fryuni:
@HiDeoo, any idea how I could optimize the e2e git test?

It runs on 4s locally on a considerably old Intel Mac Air, but for some reason, it is reaching the 30s timeout on CI. I'm out of ideas 😞

packages/starlight/utils/plugins.ts
```diff
@@ -16,8 +16,24 @@ export async function runPlugins(
...
+							.literal(true, {
+								errorMap: () => ({
+									message:
+										'astro:Disabling prerendering is not supported when using the `static` output mode.',
+								}),
+							})
+							.optional(),
+					})
+			  )
+			: StarlightConfigSchema;
```
HiDeoo, line 33: I assume this change is due to me commenting in my previous review that the 2 previous checks should be moved to the schema. I think the change is great for the other check altho I now realize that I missed in my previous review the fact that this other check also use the `context`.

I'm wondering if for this one specific check, we could only validate once, when all the plugins have ran with a simple `if` like you had before (basically just moving the previous code near the end of the function)? This would avoid this schema "dynamic" change and also the changes in `packages/starlight/utils/error-map.ts` with the `astro:` prefix which I'm not a huge fan of.

Do you have any thoughts?

.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v3
+      - run: corepack enable
```
HiDeoo, line 65: Could you explain why this was necessary as this would be different from all other jobs? Maybe Windows? Should we change this in some other places?

packages/starlight/__e2e__/fixtures/ssr/src/content/docs/demo.mdx
```diff
@@ -0,0 +1,59 @@
+---
+title: Tabs
```
HiDeoo, line 2: We should probably change the title of this page and also remove the content of this page involving `Tabs` as it's a duplicate of another set of tests which should not be needed here.

packages/starlight/__e2e__/fixtures/git/src/content/docs/tabs-variable-height.mdx
HiDeoo, line 1: I don't think this specific page is used in a `git` test? We can probably remove it from the fixture (or if it used, we should change it's title and content).

packages/starlight/__e2e__/fixtures/git/src/content/docs/tabs.mdx
HiDeoo, line 1: Same as the previous comment, I don't think this specific page is used in a `git` test? We can probably remove it from the fixture (or if it used, we should change it's title and content).

packages/starlight/__e2e__/test-utils.ts
```diff
@@ -1,34 +1,74 @@
...
 
-	test.beforeAll(async () => {
-		const root = fileURLToPath(new URL(fixturePath, import.meta.url));
-		await build({ logLevel: 'error', root });
-		previewServer = await preview({ logLevel: 'error', root });
+				const newServer = await makeServer(params);
+				customServers.push(newServer);
+
+				return new StarlightPage(newServer, page);
+			}),
```
HiDeoo, line 60: What do you think of exposing 2 helpers, e.g. `getServer` and `makeServer`. Most tests would just use `getServer` to get the default server and the few tests needing a second one would also use `makeServer`?

packages/starlight/__e2e__/fixtures/git/.gitignore
HiDeoo, line 1: If I am correct, I don't think this file is needed anymore?

packages/starlight/__e2e__/fixtures/git/astro.config.mjs
```diff
@@ -0,0 +1,11 @@
+import starlight from '@astrojs/starlight';
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+	integrations: [
+		starlight({
+			title: 'Basics',
```
HiDeoo, line 7: We usually name our fixtures based on what they test/setup, we could just call it Git here or something related to last updated to make it different from the Basics fixture.

packages/starlight/__e2e__/fixtures/ssr/astro.config.mjs
```diff
@@ -0,0 +1,16 @@
...
+import starlight from '@astrojs/starlight';
+import node from '@astrojs/node';
+
+export default defineConfig({
+	output: 'server',
+	adapter: node({ mode: 'standalone' }),
+	compressHTML: false, // for easier debugging
+	integrations: [
+		starlight({
+			title: 'Docs',
```
HiDeoo, line 11: Same as the previous comment, this one could just be titled SSR.

packages/starlight/__e2e__/fixtures/ssr/src/content/docs/404.mdx
```diff
@@ -0,0 +1,21 @@
...
+  title: Make your docs shine with Starlight
+  tagline: Everything you need to build a stellar documentation website. Fast, accessible, and easy-to-use.
+  actions:
+    - text: Get started
+      icon: right-arrow
+      variant: primary
+      link: /getting-started/
+    - text: View on GitHub
+      icon: external
+      link: https://github.com/withastro/starlight
```
HiDeoo, line 15: I think we are only testing the `#server-check` element on a 404 page and not anything related to the hero so maybe we could remove this entirely to simplify the fixture page?

packages/starlight/__e2e__/fixtures/ssr/src/content/docs/content.mdx
```diff
@@ -0,0 +1,6 @@
+---
+title: Tabs
```
HiDeoo, line 2: Let's update this fixture page title to avoid mentioning unrelated content from another test.

packages/starlight/vitest.workspace.ts
HiDeoo, line 1: I think we should be able to remove entirely these changes, right?

packages/starlight/vitest.config.ts
HiDeoo, line 1: I think when updating the branch and maybe resolving conflicts, the change may have gone in the wrong direction. We want something closer to what we now have on `main` rather than going back to the manual list of excludes copied from the Vitest repo.

packages/starlight/utils/route-data.ts
```diff
@@ -82,11 +81,10 @@ function getLastUpdated({ entry }: PageProps): Date | undefined {
...
```
HiDeoo, line 85: As you mentioned, considering you also fixed this for users with a custom `srcDir`, I think we should add a second changeset file for this change.

packages/starlight/__e2e__/test-utils.ts
Fryuni: What would be the difference between `getServer` and `makeServer` without any parameters?

HiDeoo: My idea was that `getServer` would never take arguments, and would always return the default prod server created. On the opposite, `makeServer` would require a mode to create a new one (I don't think we ever pass an inline config to `makeServer` so maybe this part could even be removed).

packages/starlight/utils/plugins.ts
Fryuni: I think validating only once after the plugins would be great. The tradeoff is that we lose the report of which plugin caused the problem in case one of them disables prerender. I think that is fine tho.

.github/workflows/ci.yml
Fryuni: I changed it on some of the iterations to fix things on Windows. I'll try removing it to see if it actually is needed for Windows to work of if I just forgot to undo it before.

packages/starlight/__e2e__/fixtures/git/.gitignore
Fryuni: It is. Without this file, the git test tries to index the entire node_modules and Astro-generated files.

packages/starlight/__e2e__/test-utils.ts
```diff
@@ -1,34 +1,79 @@
...
+			// Let the cleanup happen on the background
+			timers.setTimeout(30000),
+			(async () => {
+				// Stop all started servers.
+				await prodServer?.stop();
+				for (const server of servers.values()) {
+					await server.stop();
+				}
+			})(),
+		]);
```
HiDeoo, line 76: I think the Playwright behavior described in the comment makes sense but I think something may be wrong here.

At this point, the last test likely succeeded and we only have to cleanup resources. `context.setTimeout(context.timeout + 40000);` would add 40 seconds to the timeout just to cleanup resources. And after that, the `Promise.race` is basically saying, you have 30 seconds to cleanup resources otherwise I'll continue.

I think the code I would expect here would be:

```ts
await prodServer?.stop();
for (const server of servers.values()) {
  await server.stop();
}
```

and nothing else.

If I clone this branch, change the `test.afterAll` code to the one above, edit `packages/starlight/__e2e__/git.test.ts` to transform the last `test` (`include git information while developing`) to a `test.only` and run the E2E tests, this single test would fail due a timeout issue.
If I add a bit of logging, I can see that the last (and single) expectation of this test is met in 2s on my machine so this means that the test is taking more than 38s to cleanup resources.
If I add back at the beggining of the `test.afterAll` the line `context.setTimeout(context.timeout + 40000);`, and even use a ridiculous high value, the test still times out.
If I edit the `packages/starlight/playwright.config.ts` file and switch `headless` to `false`, the test passes in 2s altho an `✘ [ERROR] The build was canceled` message is logged.

I think something doesn't make sense here. I'm not even sure what this "build" message is about as we should not be building anything by running only this specific test. But it definitely looks like closing a dev server is hanging the test indefinitely.

I'm assuming this problem and the current approach (with the `Promise.race`) may also explain why the time to run the tests before and after these changes almost doubled).

I'll try to investigate this further but I don't think I've seen anything obvious so far that could be causing this issue.

HiDeoo (changes requested):
Amazing job on the last changes and all the cleanup, this is shaping up very nicely! 🌟 

I think personally I'm down to one last comment/issue regarding the tests and then we should be able to move on to the documentation (which is a fairly small part considering the size of the PR and should be quick to review) :tada:

packages/starlight/__e2e__/test-utils.ts
Fryuni: I see this problem only on CI. Localy it passes in about 5s even when stopping one server at a time and waiting.

HiDeoo: Wow, that's even more weird. For me, this single test definitely hangs at `await server.stop();`. I definitely wonder what could be the cause of such weird/inconsistent behavior.

HiDeoo: Adding a link to a Discord message containing the investigation I did on this topic so I don't lose it: https://discord.com/channels/830184174198718474/1070481941863878697/1271515653840568341

packages/starlight/utils/git.ts
```diff
@@ -39,3 +40,76 @@ export function listGitTrackedFiles(directory: string): string[] {
 	const output = result.stdout.trim();
 	return output.split('\n').map((file) => join(directory, file));
 }
+
+function getRepoRoot(directory: string): string {
```
HiDeoo, line 44: If I am correct, with this new refactor, `listGitTrackedFiles()` (just above but outside of review range) is no longer used? If correct, we should remove it entirely.

packages/starlight/utils/git.ts
```diff
@@ -39,3 +40,76 @@ export function listGitTrackedFiles(directory: string): string[] {
...
+	}
+
+	try {
+		return realpathSync(result.stdout.trim());
+	} catch {
+		return directory;
+	}
+}
+
+export function getAllNewestCommitDate(directory: string): [string, number][] {
```
HiDeoo, line 61: We should add a third changeset I think to mention this new performance improvement as this would benefit to all type of projects?

packages/starlight/utils/git.ts
```diff
@@ -39,3 +40,76 @@ export function listGitTrackedFiles(directory: string): string[] {
...
+
+		const currentLatest = latestDates.get(fileName) || 0;
+		latestDates.set(fileName, Math.max(currentLatest, runningDate));
+	}
+
+	return Array.from(latestDates.entries()).map(([file, date]) => {
+		const fileFullPath = resolve(repoRoot, file);
+		const fileInDirectory = relative(directory, fileFullPath);
+
+		return [fileInDirectory, date];
```
HiDeoo, line 113: To keep consistency with `getNewestCommitDate()` and the function name, would it make more sense to actually create the dates here instead of in the virtual module?

packages/starlight/__e2e__/test-utils.ts
```diff
@@ -9,27 +9,63 @@ process.env.ASTRO_DISABLE_UPDATE_CHECK = 'true';
...
+		options: {
+			mode?: 'build' | 'dev';
+		} = {}
+	): Promise<Server> {
+		const { mode } = options;
+		if (mode === 'dev') {
+			return await dev({
+				logLevel: 'error',
+				root,
+				vite: { optimizeDeps: { noDiscovery: true } },
```
HiDeoo, line 32: Let's add a small comment that this was necessary to avoid some timeouts.

Now that it's confirmed that this was the issue, when I get the chance to make a small repro and open an issue on the Vite repo, I'll also make a new PR to add the link here too.

HiDeoo (commented):
Thanks for putting the last finishing touches on the PR, making sure even the tests run smoothly but also bringing a new optomization too :tada: Amazing work again!

I left 2 or 3 comments but very minor. I'll take a look at the documentation changes later but I personally think we are really really (really) close to the finish line now. I also recently noticed on Discord another new user mentioning that they are using the PR already too so that's a good sign. 🌟

packages/starlight/utils/git.ts
Fryuni: It will have to be turned into a number to be serialized in the source and turned back into a Date inside the virtual module, so I don't think there is much advantage in that.

The `getNewestCommitDate` is used directly when using the dev server, so it returns the final value.

.changeset/ninety-singers-film.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/starlight': patch
+---
+
+Improve performance of computing the `lastUpdated` date from Git history.
```
HiDeoo, line 5:
```suggestion
Improves performance of computing the last updated times from Git history.
```

To keep consistency with the next changeset.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -455,6 +455,18 @@ Define whether Starlight’s default site search provider [Pagefind](https://pag
 Set to `false` to disable indexing your site with Pagefind.
 This will also hide the default search UI if in use.
 
+Pagefind cannot be enabled when the [`prerender`](#prerender) option is set to `false`.
+
+### `prerender`
+
+**type:** `boolean`
```
HiDeoo, line 462:
```suggestion
**type:** `boolean`  
```

To get the `default` value on a new line.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -455,6 +455,18 @@ Define whether Starlight’s default site search provider [Pagefind](https://pag
...
+**default:**: `true`
```
HiDeoo, line 463:
```suggestion
**default:** `true`
```

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -455,6 +455,18 @@ Define whether Starlight’s default site search provider [Pagefind](https://pag
...
+
+Define whether Starlight pages should be prerendered or handled using SSR.
+This option behave just like the [`export const prerender`](https://docs.astro.build/en/guides/server-side-rendering/#enable-on-demand-server-rendering) from Astro pages and allow overriding the default prerendering behavior of the selected output mode.
+
+Starlight pages are prerendeered by default regardless of the output mode of the Astro project. This differs from the default behavior of Astro pages, which are prerendered by default on `output: 'hybrid'` but not `output: 'server'`.
```
HiDeoo, line 468:
```suggestion
Starlight pages are pre-rendered by default regardless of the output mode of the Astro project. This differs from the default behavior of Astro pages, which are pre-rendered by default on `output: 'hybrid'` but not `output: 'server'`.
```

To match the wording from [Astro Docs](https://docs.astro.build/en/guides/server-side-rendering/#enable-on-demand-server-rendering)

docs/src/content/docs/manual-setup.mdx
```diff
@@ -125,6 +125,6 @@ In the future, we plan to support this use case better to avoid the need for the
 
 ### Use Starlight with SSR
 
-You can use Starlight alongside custom on-demand rendered pages in your project by following the [“On-demand Rendering Adapters”](https://docs.astro.build/en/guides/server-side-rendering/) guide in Astro’s docs.
+Starlight supports [SSR](https://docs.astro.build/en/guides/server-side-rendering/) out of the box using Astro's server adapters.
 
-Currently, documentation pages generated by Starlight are always prerendered regardless of your project's output mode. We hope to be able to support on-demand rendering for Starlight pages soon.
+Contrary to your own pages, documentation pages generated by Starlight are always prerendered by default regardless of your project's output mode. To opt-out of prerendering your Starlight pages, set the [`prerender` config option](/reference/configuration/#prerender) to `false`.
```
HiDeoo, line 130:
```suggestion
Contrary to your own pages, documentation pages generated by Starlight are always pre-rendered by default regardless of your project's output mode. To opt-out of pre-rendering your Starlight pages, set the [`prerender` config option](/reference/configuration/#prerender) to `false`.
```

To match the wording from [Astro Docs](https://docs.astro.build/en/guides/server-side-rendering/#enable-on-demand-server-rendering)

HiDeoo (dismissed):
Thanks for the updates. As promised, this should personally be my last review on the PR before approving this **_amazing_** PR: a few documentation/changeset tweaks and a suggestion that should fix the last CI error. :clap: 🎉

packages/starlight/integrations/virtual-user-config.ts
```diff
@@ -82,3 +88,37 @@ export function vitePluginStarlightUserConfig(
...
+
+	return `
+const trackedDocsFiles = new Map(${JSON.stringify(trackedDocsFiles)});
+
+export function getNewestCommitDate(file) {
+	const timestamp = trackedDocsFiles.get(file);
+	if (!timestamp) throw new Error(\`Failed to retrieve the git history for file "\${file}"\`);
+	return new Date(timestamp);
+}`;
+}
```
delucis, line 124: In general all the other modules in this file aim to keep the generative stuff as minimal and clean as possible. Would it be possible to refactor this to move as much as possible to imported code? That way we get the benefit of typechecking, editor tooling etc. and avoid too much JS written in template strings.

i.e. a quick naïve take, make a file a bit like the following (or these could also be inlined into `utils/git` itself):

```ts
import { getNewestCommitDate, getAllNewestCommitDate } from '../utils/git';

export function getDevGitUtil(docsPath: string) {
	return (file: string) => getNewestCommitDate(resolve(docsPath, file));
}
export function getBuildGitUtil(docsPath: string) {
	const trackedDocsFiles = new Map(getAllNewestCommitDate(docsPath));
	return (file: string) => {
		const timestamp = trackedDocsFiles.get(file);
		if (!timestamp) throw new Error(`Failed to retrieve the git history for file "${file}"`);
		return new Date(timestamp);
	};
}
```

and then the virtual module does the bare minimum:

```ts
const resolvedDocsPath = resolveId(srcDir, 'content/docs');
const gitUtilName = command !== 'build' ? 'getDevGitUtil' : 'getBuildGitUtil';
const modules = {
	// ...
	'virtual:starlight/git-info': `import { ${gitUtilName} as getGitUtil } from "...";
export const getNewestCommitDate = getGitUtil(${resolvedDocsPath});`
	// ...
};
```

packages/starlight/package.json
```diff
@@ -168,7 +168,9 @@
       "default": "./integrations/expressive-code/hast.mjs"
     },
     "./index.astro": "./index.astro",
+    "./indexSSR.astro": "./indexSSR.astro",
     "./404.astro": "./404.astro",
+    "./404SSR.astro": "./404SSR.astro",
```
delucis, line 173: @HiDeoo mentioned a question around naming of the different entrypoints. I don’t have a strong opinion, but one idea would also be to move all four into a dedicated `routes/` directory to help us keep organised. Up to you @Fryuni if you want to do that or not — it’s not essential and can also happen later in some work tidying up the repo.

packages/starlight/__e2e__/ssr.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	const b$ = cheerio.load(b);
+
+	a$('script[src]').attr('src', '');
+	a$('link[href]').attr('href', '');
+
+	b$('script[src]').attr('src', '');
+	b$('link[href]').attr('href', '');
+
+	expect(a$.html()).toEqual(b$.html());
+}
```
delucis, line 68: Wonder if we could use something lighter than Cheerio for this. I quite like `linkedom` and in a quick test it worked just as well and is about half the size ([2.3 MB](https://pkg-size.dev/linkedom) vs [5.1 MB](https://pkg-size.dev/cheerio)).

It’s also nice because its API is the standard DOM API, instead of using cryptic `.attr()` calls:
```diff
- import * as cheerio from 'cheerio';
+ import { parseHTML } from 'linkedom';
```
```suggestion
function expectEquivalentHTML(a: string, b: string) {
	expect(getNormalizedHTML(a)).toEqual(getNormalizedHTML(b));
}

function getNormalizedHTML(html: string) {
	const window = parseHTML(html);
	window.document.querySelectorAll('script[src]').forEach((el) => el.setAttribute('src', ''));
	window.document.querySelectorAll('link[href]').forEach((el) => el.setAttribute('href', ''));
	return window.toString();
}
```
What do you think?

delucis (commented):
OK, didn’t quite get to everything yet, but need to break for a bit so sharing some initial review comments!

Thanks as always @Fryuni for your dedication and patience on this PR. Amazing work from you as well as from @HiDeoo working through the details to this point 🙌

packages/starlight/indexSSR.astro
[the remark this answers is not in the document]
Fryuni: The wiring in the index is quite a bit different between them so I don't think it would work to just import `index.astro`. The shared part can be extracted tho, I'll refactor the pages a bit

packages/starlight/__e2e__/fixtures/git/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Astro generates this during tests, but we want to ignore it.
+env.d.ts
+.astro
+/node_modules/
+/dist/
```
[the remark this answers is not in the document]
Fryuni: This fixture is used for the e2e test using git as a separate repo root. This file is for that test repo to ignore it and not index all of node_modules.

packages/starlight/integrations/virtual-user-config.ts
Fryuni: The point is exactly to the git module during build and inline the result, so there is no dependency on git or Node APIs during runtime.

Calling `getAllNewestCommitDate` from the virtual module (even if indirectly) inverts the whole thing, it would compute the git information every time the server initializes.

packages/starlight/__e2e__/ssr.test.ts
Fryuni: I've never heard of linkedom, but if works that's fine by me.

packages/starlight/__e2e__/fixtures/git/.gitignore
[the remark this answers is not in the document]
delucis: Ah, gotcha, thanks!

packages/starlight/integrations/virtual-user-config.ts
delucis: OK! Then can we run with the general principle of simplifying? i.e. still compute those for the build version, but then the above example looks something like:
```ts
const argument = command !== 'build' ? resolveId(srcDir, 'content/docs') : getAllNewestCommitDate(...);
const gitUtilName = command !== 'build' ? 'getDevGitUtil' : 'getBuildGitUtil';
const modules = {
	// ...
	'virtual:starlight/git-info': `import { ${gitUtilName} as getGitUtil } from "...";
export const getNewestCommitDate = getGitUtil(${argument});`
	// ...
};
```

As I mentioned, my code was a “naïve take,” but the principle is to move as much as possible out of a template literal and into regular code.

packages/starlight/__e2e__/ssr.test.ts
delucis: It’s a really nice and fast alternative to JSDOM built with linked lists — https://github.com/WebReflection/linkedom

packages/starlight/utils/plugins.ts
```diff
@@ -72,6 +73,13 @@ export async function runPlugins(
 		});
 	}
 
+	if (context.config.output === 'static' && !starlightConfig.prerender) {
+		throw new AstroError(
+			'Starlight prerendering cannot be disabled when the project is set to "static" output mode.',
+			'Either set output mode to "server"/"hybrid" or enable Starlight prerendering.'
```
delucis, line 79: Suggesting a small rework here. My thinking here is that usually people don’t actively set `output: 'static'` — it’s the implicit default — so they may not understand how/why their project is set that way. Also seems a good opportunity to add a docs link.

```suggestion
			'Starlight’s `prerender: false` option requires `output: "hybrid"` or `"server"` in your Astro config.',
			'Either set `output` in your Astro config or set `prerender: true` in the Starlight options.\n\n' +
				'Learn more about rendering modes in the Astro docs: https://docs.astro.build/en/basics/rendering-modes/'
```

(Instructing the user to set `prerender: true` may feel a bit odd, given it’s the default, but I think in this scenario it gives them a really clear course of action — “I already have `prerender: false`, I just need to updated that” — vs trying to explain removing the option entirely, and an error message’s aim is getting things working again as simply as possible.)

packages/starlight/utils/routing.ts
```diff
@@ -100,6 +100,22 @@ function getRoutes(): Route[] {
...
+	}
+
+	return map;
+}
+const INDEX_SLUG_PARAM = Symbol('index');
+const routesBySlugParam = getParamRouteMapping();
+
+export function getRouteBySlugParam(slugParam: string | undefined): Route | undefined {
+	return routesBySlugParam.get(slugParam?.replace(/\/$/, '') ?? INDEX_SLUG_PARAM);
+}
```
delucis, line 117: Couple of questions here:
- is there a particular advantage of using a `Symbol` vs just using `undefined` as the index key?
- should the `??` be `||`? I’d guess a param of `""` should be handled the same as a param of `undefined`

```suggestion
function getParamRouteMapping(): ReadonlyMap<string | undefined, Route> {
	const map = new Map<string | undefined, Route>();
	for (const route of routes) {
		map.set(slugToParam(route.slug), route);
	}
	return map;
}
const routesBySlugParam = getParamRouteMapping();

export function getRouteBySlugParam(slugParam: string | undefined): Route | undefined {
	return routesBySlugParam.get(slugParam?.replace(/\/$/, '') || undefined);
}
```

packages/starlight/utils/user-config.ts
```diff
@@ -209,15 +209,31 @@ const UserConfigSchema = z.object({
 	/** Disable Starlight's default 404 page. */
 	disable404Route: z.boolean().default(false).describe("Disable Starlight's default 404 page."),
 
+	/**
+	 * Define whether Starlight pages should be prerendered or not.
+	 * Defaults to always prerender Starlight pages, even when the project is
+	 * set to "server" output mode.
+	 * Does nothing when Astro's output mode is undefined or set to "static".
```
delucis, line 216: We throw an error in if you set it to false for static output right? I might just remove this line:
```suggestion
```

delucis (commented):
Thanks for taking my suggestions on board @Fryuni! This is in great shape. Left a last couple of comments on the files I didn’t get to yesterday.

packages/starlight/utils/routing.ts
Fryuni: > is there a particular advantage of using a `Symbol` vs just using `undefined` as the index key?

Matter of preference, I dislike using `undefined` as a marker for things, but it works the same.

> should the `??` be `||`? I’d guess a param of `""` should be handled the same as a param of `undefined`

Both `slugToParam` and the Astro params from the URL transform the root index to `undefined` so there is no case for the empty string. So it could be `||` with no real difference.

packages/starlight/utils/plugins.ts
```diff
@@ -72,6 +73,13 @@ export async function runPlugins(
 		});
 	}
 
+	if (context.config.output === 'static' && !starlightConfig.prerender) {
```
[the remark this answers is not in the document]
Fryuni: This PR is almost the entirety of v4. It started two days after 4.0.0 and now we have to consider how v5 just around the corner will break it 😅

delucis (approved):
Thanks for the last updates @Fryuni! Looks good to me! 🎊 

I think one thing we can think about for the future is if this needs any more documentation (the current SSR section is a bit hidden in the manual setup guide), but I’m happy to run with this as it’s a rarer option for people.

Will take a look to see if there are other things we’d like to get released with this and decide when to cut the next minor.

delucis:
Updated the end-to-end fixtures to the latest Astro release to match the updates in #2287 — tests still all seem good!

sarah11918:
Just wanted to say I'm excited! :partying_face:  Congrats, @Fryuni !

marcelovicentegc:
Hey, folks! I just wanted to let you know that I've been following this PR for the last couple of months and wanted to congrats everyone involved on this effort. You did an amazing job here @Fryuni @HiDeoo and @delucis! I'm saving this PR as an example of open source contribution. 

Really cool to see these in-depth discussions happening async, and moving forward! 

Again, congrats and thank you @Fryuni !

rodolphoasb:
Trabalho incrível, @Fryuni! Congrats team!

OliverSpeir:
🎉

PR #1250, opened 2023-12-07
docs(i18n): Add Indonesian translation for site search documentation

herbras:
**Description**
What does this PR change? Give us a brief description.

Translate site-search.md docs into Indonesian.

PR #1256, opened 2023-12-08
docs: fix formatting related build issue

HiDeoo:

#### Description

This PR fixes a build issue with the docs following the merge of #1250. If I get this right, I think this is what happened:

- #1250 was building fine in the PR
- #1250 was rendering fine in the PR
- After merging #1250, the live docs are rendering fine
- After CI applied formatting, it somehow formatted the file in way that broke the build

My assumption was that if some Markdown was rendering fine, applying Prettier would never break it but I guess this is a wrong assumption?


HiDeoo:
> I've noticed that if there isn't a new line before and after codeblocks the formatter sometimes will break the syntax

Oh, thanks for sharing, I totally did not know about that kind of edge cases 🤔

HiDeoo:
Fixed by @delucis with https://github.com/withastro/starlight/commit/5df5798405ed0a29f0a2b2ca9252cda6e28e72a3

delucis:
Oh boy — sorry for not noticing this PR! I saw the failing build on `main` and just jumped to getting it passing, sorry. Should have checked for anyone else working on it first!

PR #1266, opened 2023-12-11
perf: Remove redundant subprocess calls

### LINKED ISSUE #1265: `lastUpdated` option causes performance degradation

0.15.0

4.0.3

pnpm

Windows

Chrome

### Describe the Bug

When using `lastUpdated: true` in the astro (Starlight) configuration performance of route generation seems to be severely degraded.
When setting the option to `false` I can see generation times of ~100ms, whereas enabling the option it rises to ~800ms.

https://stackblitz.com/edit/github-8ld5g5

### Participation

alex-way:
#### Description

Closes #1265 

A performance improvement when checking the lastUpdated time.

Render times for docs using `lastUpdated = true` are reduced from ~800ms to ~200ms.


HiDeoo:

I have not yet had time to test the changes or give much thoughts to them but I wonder what would be the result in the same setup if instead of removing the checks, we:

- Check only once if `git` is installed and cache the result. I think it's a nice error to have and gives a good hint to the user.
- Instead of using `execa` and the `test` command to check if a path exists and is a file, we use `fs.accessSync`, `fs.existsSync` or something similar.

alex-way:
Hey @HiDeoo I had the same thinking in my original issue I raised. However, I opted for just removing the code which checks whether git is installed and also whether the file path exists as in the original implementation it was raising an exception which was immediately caught and not re-raised. As such the code (I believe) wasn't _really_ doing anything.

If we'd rather have explicit error messages when git isn't installed, or the filepath can't be found then I can look into doing so. However, the current implementation doesn't do so due to the catch-all exceptions?

I agree that using the `fs` module will be a better fit to checking if a file exists, rather than relying on a sub-command so happy to replace if we move forward with explicit checks.

Along the same lines of thinking as yourself, I think one other way we could improve the performance is to gather the last updated dates for all files under the `content` directory and cache them so we aren't performing costly sub-process calls for each document. This changeset will likely be much larger though so it may take a lot longer for me to investigate and implement.

I can have a look into the above if you think it's useful? The current changes I've made were a `quick win` but happy to look further into optimisation where I can  :smile:

HiDeoo:
> as in the original implementation it was raising an exception which was immediately caught and not re-raised

Oh, you are right, I totally missed that. I quickly read that on my phone and based my assumption on the JSDoc but didn't realize that `GitNotFoundError` was actually never bubbled up. Good catch!

Considering this worked like that for a long time, I would then assume this is a safe change to make indeed.

Altho, now it makes me wonder, if `GitNotFoundError` was explicitly created, I guess the original intent was to not catch it and let it bubble up to the user and if we should either remove it as suggested or actually tweak the code to make it work as intended (and maybe cached) :thinking:

I wonder what Chris thinks about this.

alex-way:
It looks like the implementation was copied directly from docusaurus as per the comment, including the `GitNotFoundError`: https://github.com/withastro/starlight/pull/1266/files#diff-dbaa2909a7c752ba9604ed8f39fa172a91057d4dd370a18fed5531f86394ab84L3

There is also no references to `GitNotFoundError` so it should be safe to remove?

Again happy either way (To keep as is, or to refactor to remove the catch-all and correctly raise the Exception) - Just let me know what you'd rather and I'll get to work!

delucis (approved):
As @alex-way spotted, this code was adapted from the Docusaurus implementation a while back, so has a few things — like those custom error types — which we were never actually using. The “git not installed” and “file does not exist” errors would potentially be helpful for debugging, but as noted above, if we weren’t every doing anything with them, then… they had limited utility 😁

As such I’m happy with these changes.

PR #1270, opened 2023-12-12
fix: remove link to incompatible plugin

levino:
#### Description

As [mentioned](https://github.com/HiDeoo/starlight-blog/issues/19) in the starlight-blog repo, starlight-blog does not work with starlight 0.14 and above. There has been some [work](https://github.com/withastro/starlight/pull/1175) by @HiDeoo to reenable the plugin and I hope that this topic gets more interest and push from the maintainers soon. It seems to be non-trivial though and it looks as it requires decision making from the maintainers of starlight. This is also why I open this PR.

I think it makes sense not to advertise for the plugin for the time being since it creates some really difficult to debug crashes. I tried to move from docusaurus to astro + starlight + starlight-blog and this topic cost me several hours pulling my hair.

The changes should be easy to revert once the plugin has been fixed.

### DIFF
--- a/docs/src/content/docs/de/showcase.mdx
+++ b/docs/src/content/docs/de/showcase.mdx
@@ -31,11 +31,6 @@ Diese Community-Tools, Plugins und Integrationen arbeiten mit Starlight zusammen
 		description="Füge deinen Dokumentseiten ein Benutzer-Feedback-System hinzu."
 	/>
 	<LinkCard
-		href="https://github.com/HiDeoo/starlight-blog"
-		title="starlight-blog"
-		description="Erweitere deine Dokumentationsseite mit einen Blog."
-	/>
-	<LinkCard
 		href="https://github.com/HiDeoo/starlight-links-validator"
 		title="starlight-links-validator"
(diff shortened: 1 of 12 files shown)


delucis:
Thanks for the PR @levino! From what I understand, @HiDeoo is actively working to bring `starlight-blog` fully up to date, so I think I’m happy keeping it listed for now rather than removing and re-adding it in a week or two. Happy to revisit this later if it turns out the plugin can’t be fully updated. Short periods of incompatibility are unfortunately unavoidable sometimes as software updates and especially with ecosystem projects for something like Starlight which is still working towards v1. Once we hit that milestone next year, we can aim for much more stability and this should happen very rarely!

levino:
Thank you for your consideration. Good luck. I will revisit in 2 weeks.

narration-sd:
Just a note; I got caught.  Possibly a note on the marketing that it's temporarily unavailable?

MIght be nice for the person trying to catch up here, shouldn't we think?

levino:
My reminder about this also rang yesterday. I am sad to say it, but as I expected, no fix yet. I think if disabling a broken feature takes significantly less time than the fix, it should be done.

PR #1275, opened 2023-12-13
docs: Add plugins showcase

HiDeoo:

#### Description

Following the release of the version [`0.5.0`](https://github.com/HiDeoo/starlight-links-validator/releases/tag/v0.5.0) of the `starlight-links-validator` which is now a Starlight plugin, this PR brings back a plugins showcase that was initially part of https://github.com/withastro/starlight/pull/942 and removed as we only had one plugin at the time.

~~I tried to follow the structure from one of your [suggestion](https://github.com/withastro/starlight/pull/942/commits/d3ef3114e88b6b59d91d1c5e94435ec43b7213cc) but this might need some adjustments as you might have a better idea of how you want to showcase plugins.~~

~~I deliberately left the `Add plugins to your site` section out as a showcase did not seem to be the right place for it but I might be wrong.~~

~~Personal note: I feel like it would be weird to use a `<LinkCard/>` for individual official plugins as they have dedicated sections/headings altho I feel like official plugins are not emphasized enough compared to community plugins with their link cards.~~

Update: I updated the PR with this [new suggestion](https://github.com/withastro/starlight/pull/1275#discussion_r1425843128) which I personally think looks a bit better.


docs/src/content/docs/showcase.mdx
```diff
@@ -18,11 +18,34 @@ Starlight is already being used in production. These are some of the sites aroun
 
 See all the [public project repos using Starlight on GitHub](https://github.com/withastro/starlight/network/dependents).
 
-## Community plugins
+## Plugins
+
+Starlight [plugins](/reference/plugins/) can customize Starlight configuration, UI, and behavior, while also being easy to share and reuse.
+This section lists the official plugins supported by the Starlight team and also community plugins that are maintained by Starlight users.
+
+### Algolia DocSearch
```
kevinzunigacuellar, line 26: I don't know if we are planning to support more official plugins in the future but I was thinking that maybe we could have an official plugins section and list the algolia plugin under that section.

I think this will help level the importance of official and community plugins which will encourage more people to share their creations.

HiDeoo: Yeah, I think you're totally right. I updated the PR to add an "Official plugins" section and if we want to go back to a structure closer to the [suggested one](https://github.com/withastro/starlight/pull/942/commits/d3ef3114e88b6b59d91d1c5e94435ec43b7213cc), we can easily revert or tweak it.

docs/src/content/docs/showcase.mdx
```diff
@@ -18,11 +18,36 @@ Starlight is already being used in production. These are some of the sites aroun
 
 See all the [public project repos using Starlight on GitHub](https://github.com/withastro/starlight/network/dependents).
 
-## Community plugins
+## Plugins
+
+Starlight [plugins](/reference/plugins/) can customize Starlight configuration, UI, and behavior, while also being easy to share and reuse.
```
delucis, line 23: Tiny nit suggestion to avoid the repetition of “Starlight”:

```suggestion
[Plugins](/reference/plugins/) can customize Starlight configuration, UI, and behavior, while also being easy to share and reuse.
```

docs/src/content/docs/showcase.mdx
```diff
@@ -18,11 +18,36 @@ Starlight is already being used in production. These are some of the sites aroun
...
+This section lists the official plugins supported by the Starlight team and also community plugins that are maintained by Starlight users.
```
delucis, line 24: Bit subjective this one, but I wanted to avoid “this section lists…” because it seemed a bit pedestrian and obvious, but I like how the sentence explains the maintenance difference between official and community plugins. What do you think of this?

```suggestion
Extend your site with official plugins supported by the Starlight team and community plugins maintained by Starlight users.
```

delucis (approved):
Thanks for circling back to this @HiDeoo! Left a couple of small notes but overall this looks great and a nice improvement 🙌

docs/src/content/docs/showcase.mdx
HiDeoo: Great suggestion, this is indeed way better while keeping the maintenance difference :thumbsup:

PR #1276, opened 2023-12-14
Bump EC dependency to ^0.30.1

hippotastic:
#### Description

- Bumps the `astro-expressive-code` dependency to `^0.30.1`.
- The new version now uses Shikiji as syntax highlighter and allows loading custom language grammars.


kevinzunigacuellar:
LGTM I haven't try adding new syntax but I went around docs and there seems to no change 💜

HiDeoo:
Small question for my personal curiosity as this was asked a few times already. In the context of Starlight, to add custom languages, I would now use the following code:

```js
starlight({
  expressiveCode: {
    shiki: {  langs: [...]  },
  },
});
```

Is that a correct assumption?

hippotastic:
> Small question for my personal curiosity as this was asked a few times already. In the context of Starlight, to add custom languages, I would now use the following code:
> 
> ```js
> starlight({
>   expressiveCode: {
>     shiki: {  langs: [...]  },
>   },
> });
> ```
> 
> Is that a correct assumption?

Yes, that's correct! This is the Expressive Code-specific option that works in all the integrations and that provides types which match best what the underlying `vscode-textmate` actually requires. In contrast to the types exported by Shikiji and used in Astro, this will not complain about missing `$self` or `$base` properties in the `repository` property, or a missing `repository` property altogether, which all aren't really needed and grammars work fine without them. You can `import` most JSON-style grammars I found online and pass them into this list without getting TypeScript errors.

If you don't specify this option, `astro-expressive-code` will also automatically look for Astro's `shikiConfig: { langs: [...] }` config setting and add any languages defined there.

HiDeoo:
I see, thanks for the detailed explanation, super appreciated :raised_hands:

This will also be a great reference to link to if ever needed :thumbsup:

delucis (approved):
LGTM! If I understood the [expressive-code release notes](https://github.com/expressive-code/expressive-code/releases/tag/expressive-code%400.30.0), the `langs` config API is entirely new and not a replacement for another API? If that’s the case, it is indeed safe as a patch. If it’s replacing another API someone might be using, it would need to be a minor.

As mentioned on Discord, merging this should be safe at any time as it doesn’t introduce docs changes and is a patch. We’ll then release it together with other recent PRs sometime next week.

Thanks for keeping this updated @hippotastic!

hippotastic:
> LGTM! If I understood the [expressive-code release notes](https://github.com/expressive-code/expressive-code/releases/tag/expressive-code%400.30.0), the `langs` config API is entirely new and not a replacement for another API? If that’s the case, it is indeed safe as a patch.

This is correct. The update should have no effect whatsoever on existing sites. The change to Shikiji is purely under the hood and does not change the HTML output as I'm not using Shikiji's rendering at all. It just made it possible to introduce the new `langs` option which didn't exist before.

> As mentioned on Discord, merging this should be safe at any time as it doesn’t introduce docs changes and is a patch. We’ll then release it together with other recent PRs sometime next week.

Ok! I'll merge it then! :)

PR #1278, opened 2023-12-15
Export `StarlightUserConfig` type

HiDeoo:

- Changes to Starlight code

#### Description

I've been finding myself often writing the following type when working on plugins and having to pass down the received Starlight user config to a function that needs multiple of its properties:

```ts
type StarlightUserConfig = Parameters<StarlightPlugin['hooks']['setup']>['0']['config']
```

This PR exports this type which is already referenced multiple times in the ["Plugins reference" page](https://starlight.astro.build/reference/plugins/) of the docs.

### DIFF
--- a/.changeset/tame-squids-yawn.md
+++ b/.changeset/tame-squids-yawn.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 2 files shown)


delucis (approved):
LGTM! Thanks @HiDeoo 🌟

PR #1288, opened 2023-12-18
[i18nIgnore] docs: revert some translated default UI strings

HiDeoo:

#### Description

While looking for something, I noticed that some of the default UI strings were translated in the docs site (I also checked all the other languages and they were fine).

This PR reverts those changes (note that the Spanish change is slightly bigger as it also fixes an indentation issue).


delucis (approved):
Well spotted! Looks good to me and the `i18nIgnore` keyword is also good 👍

PR #1298, opened 2023-12-19
fix: autogenerated sidebar alphabetical sort

### LINKED ISSUE #1296: alphabetical sorting not functioning properly in autogenerated sidebars

0.15.1

4.0.1

pnpm

Mac

Chrome

### Describe the Bug

Sidebar alphabetical sort is not doesn't sort correctly in the presence of nested directories and different page titles.

**Example**
image

https://stackblitz.com/edit/github-efyknp

### Participation

kevinzunigacuellar:
#### Description

- Closes #1296
- This PRs replaces the use of the `slug` property with the `filename` for sorting purposes when sorting file entries.


HiDeoo:
Left some comment in https://github.com/withastro/starlight/pull/1151#issuecomment-1862371840 as I think this may break the expected behavior based on the current documentation.

kevinzunigacuellar:
Thanks for the review @HiDeoo 💜 . In that case, we can use only the filename name, without its slug path to do the sorting. I will update this PR accordingly.

packages/starlight/__tests__/basics/navigation-labels.test.ts
```diff
@@ -59,6 +51,14 @@ describe('getSidebar', () => {
...
 			    "type": "group",
 			  },
+			  {
+			    "attrs": {},
+			    "badge": undefined,
+			    "href": "/",
+			    "isCurrent": true,
+			    "label": "Home Page",
+			    "type": "link",
+			  },
```
delucis, line 61: Noting that this change moves `index` files down to the bottom here. We haven’t really decided/documented what sorting is like for `index`, but I’d kind of assume they should come first. 🤔

packages/starlight/utils/navigation.ts
```diff
@@ -224,7 +224,7 @@ function sortDirEntries(
 		// Pages are sorted by order in ascending order.
 		if (aOrder !== bOrder) return aOrder < bOrder ? -1 : 1;
 		// If two pages have the same order value they will be sorted by their slug.
-		return collator.compare(isDir(a) ? keyA : a.slug, isDir(b) ? keyB : b.slug);
+		return collator.compare(isDir(a) ? keyA : stripExtension(basename(a.entry.id)), isDir(b) ? keyB : stripExtension(basename(b.entry.id)));
```
delucis, line 227: Couple of notes:

- We can use `a.id` and `b.id` here. These are localized for the current route, where the `entry.id` are the original ID (in the case of fallback content). Don’t think it matters too much in this case, but might as well be correct plus they’re shorter.

- I think we need to consider `index` handling here. Maybe something like:

   ```js
   function getComparisonId(id: string) {
     const basename = stripExtension(basename(id));
     return basename === 'index' ? '' : return basename;
   }
   ```
   
   Not 100% sure what the right return value is here to ensure index gets sorted first. In English, `''` works from my testing, but not impossible empty string sorting varies by locale.

delucis (commented):
Thanks @kevinzunigacuellar! Love you digging into these tricky corners 💜 Left some thoughts.

packages/starlight/__tests__/basics/navigation-labels.test.ts
kevinzunigacuellar: Indeed, `index` stands out as a particular case when sorting by `filename`. 

Considering this, using the frontmatter title could offer more control over the sorting sequence and aesthetically aligns with the sidebar's order, given that the sidebar label relies on the frontmatter title. 

However, it's essential to note the potential tradeoffs, such as the scenario where the title significantly differs 😆 from the slug mentioned by @HiDeoo. Another concern is that each i18n variation could present a distinct order based on its language spelling, which might change the intended order set by the default lang.

packages/starlight/utils/navigation.ts
kevinzunigacuellar: great idea! I can further test what happens in other languages.

kevinzunigacuellar: On second thought this fix might work wonderfully. I found a bug withtin a bug, so since the `filename` is consistent across locales. We don't really need to pass a locale based on the route. We can just use the default locale which will be, most likely, the language that author will write the filename in.

Fryuni (approved):
The current behavior will sort the sidebar page differently depending on the locale of the page it is being shown. With this PR it will have a consistent ordering. But what is the ordering is not properly documented.

I noticed this recently when translating the docs that it is mentioned there that the sidebar is sorted alphabetically, but the implementation uses collation, which is not the same thing. Not all languages sort alphabetically and even the definition of "alphbetically" may change between them.

More information and examples of how things can differ [here](https://unicode-org.github.io/icu/userguide/collation/).

I don't think this should be changed in this same PR, but just leaving it here as a note that I think this should probably be mentioned in the docs. Both on the `defaultLocale` and on the sidebar guide.

.changeset/many-bags-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
```
delucis, line 2: I think this is probably safe to release as a patch. It fixes incorrect sorting in some edge cases, so probably needn’t be tagged as a minor.
```suggestion
'@astrojs/starlight': patch
```

.changeset/many-bags-press.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix issue causing incorrect sorting behavior for autogenerated sidebars
```
delucis, line 5: And while I’m here I guess we can also update this to Sarah’s preferred verb tense:

```suggestion
Fixes incorrect sorting behavior for some autogenerated sidebars
```

delucis (approved):
Let’s patch and GO! ✨

PR #1303, opened 2023-12-20
chore: fix type errors

lilnasy:
#### Description
- Adds type info or pragmas in a bunch of places that fail with `tsc`.

### DIFF
--- a/packages/starlight/__tests__/remark-rehype/asides.test.ts
+++ b/packages/starlight/__tests__/remark-rehype/asides.test.ts
@@ -118,9 +118,7 @@ describe('translated labels in French', () => {
 :::${type}
 Some text
 :::
-`,
-			{ fileURL: new URL('./_src/content/docs/fr/index.md', import.meta.url) }
-		);
+`);
 		expect(res.code).includes(`aria-label="${label}"`);
 		expect(res.code).includes(`</svg>${label}</p>`);
 	});
--- a/packages/starlight/__tests__/remark-rehype/asides.test.ts
+++ b/packages/starlight/__tests__/remark-rehype/asides.test.ts
@@ -118,7 +118,10 @@ describe('translated labels in French', () => {
 :::${type}
 Some text
 :::
-`);
+`,
+// @ts-expect-error fileURL is part of MarkdownProcessor's options
+{ fileURL: new URL('./_src/content/docs/fr/index.md', import.meta.url) }
+);
 		expect(res.code).includes(`aria-label="${label}"`);
 		expect(res.code).includes(`</svg>${label}</p>`);
 	});
--- a/packages/starlight/components.ts
+++ b/packages/starlight/components.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 export { default as Card } from './user-components/Card.astro';
 export { default as CardGrid } from './user-components/CardGrid.astro';
 export { default as Icon } from './user-components/Icon.astro';
--- a/packages/starlight/components.ts
+++ b/packages/starlight/components.ts
@@ -1,3 +1,4 @@
+// @ts-nocheck
 export { default as Card } from './user-components/Card.astro';
 export { default as CardGrid } from './user-components/CardGrid.astro';
 export { default as Icon } from './user-components/Icon.astro';
--- a/packages/starlight/utils/createTranslationSystem.ts
+++ b/packages/starlight/utils/createTranslationSystem.ts
@@ -72,12 +72,12 @@ function buildDictionary(
 	base: (typeof builtinTranslations)[string],
 	...dictionaries: (i18nSchemaOutput | undefined)[]
 ) {
-	const dictionary = { ...base };
+	const dictionary: typeof base = { ...base };
 	// Iterate over alternate dictionaries to avoid overwriting preceding values with `undefined`.
 	for (const dict of dictionaries) {
 		for (const key in dict) {
-			const value = dict[key as keyof typeof dict];
-			if (value) dictionary[key as keyof typeof dict] = value;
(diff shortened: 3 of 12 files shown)


delucis (commented):
Thanks @lilnasy! As we don’t have any type-checking CI set up, would you mind describing what you did to test these changes? Commands you ran, etc.

Also noted one change I think may be unintentional and I think is the cause of the test failure.

lilnasy:
Here are the commands I ran to get to these file changes:
```
pnpm add -wD typescript
pnpm exec tsc
```
Run at the project root.

packages/starlight/__tests__/remark-rehype/asides.test.ts
```diff
@@ -118,9 +118,7 @@ describe('translated labels in French', () => {
 :::${type}
 Some text
 :::
-`,
-			{ fileURL: new URL('./_src/content/docs/fr/index.md', import.meta.url) }
-		);
+`);
```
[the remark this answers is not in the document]
lilnasy: Thanks for the pointer! Looking into this now.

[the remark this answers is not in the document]
lilnasy: Now that I try to bring it back I remember why I removed it. It's not a real option apparently.
image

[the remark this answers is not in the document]
lilnasy: What do you know, that was it!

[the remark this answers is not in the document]
delucis: Looks like the types in `@astrojs/markdown-remark` intentionally mark `fileURL` as internal: https://github.com/withastro/astro/blob/da307e4a080483f8763f1919a05fa2194bb14e22/packages/markdown/remark/src/types.ts#L66-L71

The [`render()` method](https://github.com/withastro/astro/blob/da307e4a080483f8763f1919a05fa2194bb14e22/packages/markdown/remark/src/index.ts#L122-L123) clearly uses `fileURL`, but yeah guess it’s not really intended for external use.

```js
		async render(content, renderOpts) {
			const vfile = new VFile({ value: content, path: renderOpts?.fileURL });
```

packages/starlight/utils/translations.ts
```diff
@@ -8,6 +8,7 @@ let userTranslations: Record<string, i18nSchemaOutput> = {};
 try {
 	// Load the user’s i18n collection and ignore the error if it doesn’t exist.
 	userTranslations = Object.fromEntries(
+		// @ts-expect-error
```
delucis, line 11: I think it might be safer to ignore this one — this is only an error if a user doesn’t have an i18n collection:
```suggestion
		// @ts-ignore — may be an error in projects without an i18n collection

```

packages/starlight/utils/createTranslationSystem.ts
```diff
@@ -72,12 +72,12 @@ function buildDictionary(
 	base: (typeof builtinTranslations)[string],
 	...dictionaries: (i18nSchemaOutput | undefined)[]
 ) {
-	const dictionary = { ...base };
+	const dictionary: typeof base = { ...base };
```
delucis, line 75: This is implicit and in my testing it looks like we can do without:
```suggestion
	const dictionary = { ...base };
```

packages/starlight/components.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
delucis, line 1: [Chatted](https://discord.com/channels/830184174198718474/1070481941863878697/1192244027223973978) with @Princesseuh about this and sounds like we should actually fix this properly by setting `"jsx": "preserve"` in `tsconfig.json`.

```suggestion
```

delucis (commented):
Thanks @lilnasy! Spotted a couple more things now I looked closely.

It would also be interesting to know if these changes resolve issues for Starlight consumers (I know some people encountered type errors when _using_ Starlight).

lilnasy:
Yes, that was the original motivation. I'd think that's all covered unless something else popped up in the past few days.

packages/starlight/utils/translations.ts
lilnasy: Worth checking whether that leaks into starlight. It might need to become a js file instead. I'll take a look.

packages/starlight/components.ts
lilnasy: Worth checking whether that leaks into starlight. It might need to become a js file instead. I'll take a look.

lilnasy: All good. This file did not error on the user-side in the first place.

delucis (approved):
Looks good to me! Thanks @lilnasy. If you're interested in a follow-up PR, it might be smart to add a CI job to prevent us reintroducing type errors in the future.

lilnasy:
Hmm that might be a little tricky because `tsc` also included errors related to `astro:content`.

delucis:
> Hmm that might be a little tricky because `tsc` also included errors related to `astro:content`.

Ah, ok. At one point I had those kind of working but it was a pretty janky setup relying on types across monorepo packages (basically making `packages/starlight` see the types from `docs`, so wouldn't be surprised if it broke or only works in the editor but not tsc or something.

HiDeoo:
> It would also be interesting to know if these changes resolve issues for Starlight consumers (I know some people encountered type errors when _using_ Starlight).

Just tested it on various projects or integrations/plugins, and yes, it does indeed resolves some issues as expected. Fixing all of them properly for all possible TS configs of any consumers will require additional changes from https://github.com/withastro/starlight/discussions/983 but this PR lays the groundwork for that perfectly.

PR #1304, opened 2023-12-20
i18n(ko-KR): update `showcase.mdx`

jsparkdev:
#### Description
update `showcase.mdx`


HiDeoo (approved):
Thanks for the quick update, looks good to me :star2:

PR #1308, opened 2023-12-22
Add `<FileTree>` component

HiDeoo:

- Changes to Starlight code

#### Description

This PR adds a `<FileTree>` component to display the structure of a directory.

_**This pull request is a draft as some decisions still need to be made regarding various points described below and maybe discuss some implementation details.**_

##### Rehype processor and component

Based on the version in the documentation, there is not a lot of changes, only minor ones:

- Documented a bit more the code as I discovered some features while working on this.
- Validation is no longer done with a regex but on the HAST tree itself and more errors are reported.
- Reported errors contain a hint with a link to the component documentation.
- Adds a bunch of tests.

##### Icons

- Cleaned the icon definitions a bit to only include information that is actually used.
- All icons have been updated as some of them were outdated, just the first letter of the extension thrown in an SVG, etc. When possible, icons from the Unicons icon set have been used.
- All icons have been ran through SVGOMG.
- All icons have been made usable with the `<Icon>` component.
- The following file icons have been removed for either not having Visual Studio Code icons or sometimes not even finding one: `argdown`, `asm`, `bsl`, `cake`, `cake_php`, `code-search`, `cu`, `d`, `go` (Slidesgo, not the language), `grails`, `hacklang`, `jade`, `makefile`, `mdo`, `odata`, `pddl`, `pddl-happenings`, `pddl-plan`, `prolog`, `rescript`, `slim`, `smarty`, `spring`, `sublime`, `twig`, `wat`, `wgt`.
- The following icons have been added (either made sense to me or are official integrations): `alpine`, `biome`, `cloudflare`, `lit`, `mdx`, `netlify`, `node`, `pnpm`, `preact`, `solid`, `tailwind`, `vercel`.

Regarding my train of thought for the icons, I imagined receiving feature requests for some icons being used in the file tree but not being available in Starlight so I made them all available. And if we decide to not do that, it's easier to remove icons vs. adding them all (I can attest to that ^^).

We may also decide to not include all the icons from my initial push, I just added everything possible based on the existing code and figured it would be easier to remove some of them if they are unwanted.

##### Docs

I added some documentation for the component and updated the english version of the docs site to use the new component altho I'm not sure what is the best way to handle the removal of the old one to not trigger too much the i18n tracker considering it is a drop-in replacement but this PR should still be tracked as it contains new documentation.

My current idea I have would be:

- Remove the changes to the english version of the docs site to use the new component.
- Prepare a follow-up PR that should be merged right after this one that would:
  - Use `[i18nIgnore]`
  - Update all documentation pages in all languages to use the new component
  - Remove the old component
  - Remove the following dependencies: `hast-util-from-html`, `hast-util-to-string`, `hastscript`, `hast`, `rehype`, and `unist-util-visit`.

Another approach could be to let translators handle the replacement of the component in their language and wait for them to do so before removing the old component and its dependencies.

#### If merged, the following tasks should be done before merging:

- [ ] Delete the `docs/src/content/docs/guides/file-tree-demo.mdx` demo file.


delucis (commented):
Amazing work on this @HiDeoo as usual! Thanks for your patience waiting for a review. As you know, I’m very much in favour of adding this to our built-in components.

Before diving into the detail wanted to give some big picture feedback about some of the changes.

> Adds a bunch of tests.

I love it! This was my big intended chunk of work when moving this into Starlight, so very happy to see it in place&nbsp;🙌

> All icons have been updated as some of them were outdated, just the first letter of the extension thrown in an SVG, etc. When possible, icons from the Unicons icon set have been used.

The existing `<FileTree>` in docs uses the exact same [Seti icon set](https://github.com/elviswolcott/seti-icons) as VS Code uses by default. I’m not 100% sure if we should move away from that.

Not so much because the alternative icons added in this PR are not nice, just because having an opinion has a cost, and it could be nice to say “We use the default VS Code icon set” and avoid discussion about whether icon X for file type Y is better than icon Z. (It also could make it easier to update when VS Code updates, although I didn’t make the process particularly automated when building out the initial icon set.)

I think filling obvious gaps makes sense, but not sure if we should deviate much more than that. (Although noting there’s still plenty of scope for regular maintenance when dealing with these. For example, since this PR was opened, Biome has already changed icon&nbsp;😅)

What do you think? There may be more things to consider here, so I’m happy to discuss.

> All icons have been made usable with the `<Icon>` component.

At first this made me nervous, but I think you’re right that this is sensible. (We might want to revisit the icon picker UI in the component docs in the future, but not required for this PR.)

I think the only thing to double check is if any of these icons could clash with names in the Unicons set? I might try adding all the Unicons in a separate PR to see if that helps us spot these things. If it is an issue, maybe we would come up with a prefix for file-type icons?

> I'm not sure what is the best way to handle the removal of the old [docs `<FileTree>` component] to not trigger too much the i18n tracker

We do have special commit message syntax that allows us to opt certain files out, although Lunaria doesn’t document that yet. Nevertheless, I think your proposed strategy is good as it will help translators clearly see the important changes in this PR and easily ignore the follow-up one.

HiDeoo:
> it could be nice to say “We use the default VS Code icon set” and avoid discussion about whether icon X for file type Y is better than icon Z.

I think I am mostly fine with that. Altho, considering we plan to make these icons available in the `<Icon>` component, and having them all visible in the documentation, I'll have to check but I think we would need to have a list of icons to exclude as some of them do not even render anything.

> It also could make it easier to update when VS Code updates

I think considering my previous point and this one, I should make a script to quickly fetch the Seti icons, have an exclusion list, optimize them and then generate what we need in a temporary file or something.

> I think filling obvious gaps makes sense

Definitely, I think having at least official integrations, all package managers (so that #1453 is possible), etc. would be a small but nice improvement.

> We might want to revisit the icon picker UI in the component docs in the future, but not required for this PR.

Totally agree, maybe even an Icons dedicated page listing at first the various places/components where you can use icons and then the icons themselves with a basic search input could be enough. I'll play around with that idea and see if I can come up with something nice.

> I think the only thing to double check is if any of these icons could clash with names in the Unicons set?

I think I could bake that into the script I mentioned earlier so it's easier to maintain on the long run.

> I think your proposed strategy is good as it will help translators clearly see the important changes in this PR and easily ignore the follow-up one.

Perfect, I'll start preparing that PR when we are almost done with this one.

---

I'll start thinking and working on that "icon refactor" but if you have any questions or suggestions, feel free to ask.

HiDeoo:
Finally updated this PR with a script to automatically generate the file icons and definitions used by the `<FileTree>` component but this tooks me on a small detour ^^

- There is now a script to generate everything and it's using the same repo Visual Studio Code is using.
- The SVG paths are not extracted from the SVG files but from the glyphs included in the Seti UI icon font (this is actually the approach used by Visual Studio Code).
- We scale up the glyphs until their generated paths dimensions fit perfectly in our `<Icon>` component viewBox.
- There are a few configuration options to tweak the entire process (e.g. renaming some icons, aliasing them to other icons, etc.). I tried to document every options and why I think they are useful but if you have any question or suggestion, feel free to ask.
- Only 1 icon is excluded from the Visual Studio Code set, I commented why (it's also an issue in Visual Studio Code).
- The scripts check and warn if some generated icons conflict with the built-in icons.

Compared to the previous iteration, the new icons manually added to the built-in icons are the following:

- `astro`
- `pnpm`
- `biome` (the new one)
- `bun`
- `folder`

As suggested in my initial post, I also removed the docs changes to use the new `<FileTree>` component but I did not yet get the time to play with a potential new documentation page for all icons, with maybe a search input to filter them, etc. but this is not worth delaying a review of the new approach.

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -0,0 +1,245 @@
...
+
+	if (rootElements.length === 0) {
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unique unordered list but found no elements.`
+		);
+	}
+
+	if (rootElements.length !== 1) {
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unique unordered list but found the following elements: ${rootElements
```
delucis, line 212: Language nit: “unique” in English suggests the list is different from any other list rather than the French “unique” meaning singular. Maybe also helps to repeat that the issue is “multiple” children?

```suggestion
			`The <FileTree> component expects its content to be a single unordered list but found multiple child elements: ${rootElements
```

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/** Validate that the user provided HTML for a file tree is valid. */
+function validateFileTree(tree: Element) {
+	const rootElements = selectAll('body > *', tree);
+	const [rootElement] = rootElements;
+
+	if (rootElements.length === 0) {
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unique unordered list but found no elements.`
```
delucis, line 206: Same here:

```suggestion
			`The <FileTree> component expects its content to be a single unordered list but found no child elements.`
```

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -0,0 +1,245 @@
...
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unique unordered list but found the following elements: ${rootElements
+				.map((element) => `<${element.tagName}>`)
+				.join(' - ')}.`
+		);
+	}
+
+	if (!rootElement || rootElement.tagName !== 'ul') {
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unordered list but found the following element: <${rootElement?.tagName}>.`
```
delucis, line 220: Typo:

```suggestion
			`The <FileTree> component expects its content to be an unordered list but found the following element: <${rootElement?.tagName}>.`
```

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -0,0 +1,245 @@
...
+		);
+	}
+
+	const listItemElement = select('li', rootElement);
+
+	if (!listItemElement) {
+		throwFileTreeValidationError(
+			`The <FileTree> component expects its content to be a unordered list with at least one list item.`
```
delucis, line 228:
```suggestion
			`The <FileTree> component expects its content to be an unordered list with at least one list item.`
```

packages/file-icons-generator/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Generates Starlight file icons based on the Seti UI theme",
+  "author": "Chris Swithinbank <swithinbank@gmail.com>",
+  "scripts": {
+    "build": "tsx ."
+  },
+  "license": "MIT",
+  "dependencies": {
+    "opentype.js": "^1.3.4",
+    "tsx": "^4.7.1"
+  },
```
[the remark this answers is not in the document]
HiDeoo: Yeah, I guess it depends on how you look at it, e.g. these are needed to run the generator.

delucis:
> Do we want to prefix the file icons?

Let’s go for it. I also thought through various prefixes like `file`, `ext` etc. but always felt weird in some combos, whereas a theme-name prefix feels quite good I think. A colon separator would also mirror how projects like Iconify do it to namespace different icon sets.

> Do the current UI/UX in the docs is good enough to ship the file icons?

I think it is, yes (just about 😁). I wonder if we should move the icon component docs to be _last_ on the page? I intentionally made them last when we first added it, but we since added new components to the bottom. Doesn’t solve all the UX problems, but does at least mean that other components don’t get too buried below the long icon list.

> something like a weekly CI would be more than enough (with a manual trigger too)

Sounds good to me! It’s almost more as a check that the script still _works_ more than ensuring we pull in regular updates.

HiDeoo:
Thanks for the review and feedback. The PR should be updated with all changes discussed so far (minus the CI part, which will be done in a separate PR).

When we are ok with the changes, I'll start preparing the follow-up `i18nIgnore` docs PR.

docs/src/content/docs/guides/components.mdx
```diff
@@ -296,3 +269,79 @@ The code above generates the following on the page:
 import importedCode from '/src/env.d.ts?raw';
 
 <Code code={importedCode} lang="ts" title="src/env.d.ts" />
+
+### FileTree
+
+Use the `<FileTree>` component to display a file tree in your documentation representing the structure of a directory.
+
+A `<FileTree>` requires an [unordered Markdown list](https://www.markdownguide.org/basic-syntax/#unordered-lists) of files and directories.
+Directories can either end with a `/` or contain a nested list of files and directories.
```
delucis, line 278: Suggesting a small rewrite here to flip the order to start with “user needs” rather than the more passive “component needs”.

```suggestion
Specify the structure of your files and directories with an [unordered Markdown list](https://www.markdownguide.org/basic-syntax/#unordered-lists) inside `<FileTree>`.
Create a sub-directory using a nested list or add a `/` to the end of a list item to render it as a directory without specific content.
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -296,3 +269,79 @@ The code above generates the following on the page:
...
+
+The following syntax can be used to customize the appearance of the file tree:
+
+- If a file or directory name is bolded, e.g. `**README.md**`, it will be highlighted in the file tree.
+- Comments can be appended after a file or directory name to provide additional information.
+- Placeholder files and directories can be added by using either `...` or `…` as the name.
```
delucis, line 284: Same principle here, start with the goal and follow with the how:

```suggestion
- Highlight a file or directory by making its name bold, e.g. `**README.md**`.
- Add a comment to a file or directory by adding more text after the name.
- Add placeholder files and directories by using either `...` or `…` as the name.
```

delucis (commented):
Thanks for all the great updates @HiDeoo! It’s looking really amazing. Left a few comments and here are a couple of other observations that didn’t make sense on specific code:


- Are all the mappings working as expected in your file tree demo?

	I noticed MDX uses a default file icon, for example:
	
	File tree entries for MD and MDX files. MD has a downward pointing arrow icon, MDX uses a default file icon
	
	In the current `<FileTree>`, it uses the same as Markdown. Or in VS Code (maybe only because I have the MDX extension?) it displays the MDX logo, which is a bit tiny:

	
	Current file tree showing Markdown and MDX files with the same icon VS Code sidebar showing MDX file entry with the MDX logo as its icon

	Plus there are a few other formats like `.less`, `.ejs`, `.coffee`, and `.erb`, which show icons in VS Code but not in the preview. (Some of these are also in the generator script config so maybe something’s up there?)

- Here’s a fun one:

	Part of the file tree with a guideline off-center relative to the icon The same file tree with the guideline now centred correctly
	
	The position of the left line varies depending on what screen I use 🙃 
	
	This is already the case with the current `<FileTree>`, so no need to fix it in this PR necessarily, but super fun. I’m guessing it’s because our horizontal alignment values somehow add up to fractional pixels which render differently on high resolution screens (right-hand screenshot) and regular screens (left-hand screenshot).

docs/src/content/docs/guides/file-tree-demo.mdx
```diff
@@ -0,0 +1,176 @@
...
+```
+// TODO(HiDeoo) Remove this page
+```
+
+<FileTree>
+
+- folder/
+- file
+- another_folder/
+  - astro.config.mjs this an **important** file
```
santiemanuel, line 20: Could be changed to: this is an **important** file

packages/file-icons-generator/config.ts
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * A list of Seti UI icons to ignore.
+	 * Each entry should be commented with the reason why the icon is ignored.
+	 */
+	ignores: [
+		// The ReasonML SVG icon contains a path issue that makes it render a plain square instead of a
+		// square with the "RE" letters when converted to a font.
+		// This is also an issue in Visual Studio Code but considering that in Starlight, all the icons
+		// are also available with the `<Icon>` component, it's better to ignore it for now instead of
+		// providing an icon with no meaning.
```
santiemanuel, line 56: Previous code comments were in blocks instead of line comments

delucis:
Awesome — I think this should be pretty much good to go I guess. Will do a final pass through things now.

> * My guess would be that you manually "aliased" the `.mdx` files to use the `.md` icon but I have no proof of that

I’m fairly sure I did that too, although I don’t quite remember 😁 

Of the options I guess we should either do nothing or use the official icon, even if we don’t like it. I’m fine with whatever you prefer.

delucis (commented):
As I thought — looks ready to go for the next minor! Just left one tiny suggestion.

HiDeoo:
Update the PR:

- I added the MDX icon
- Added Markdown backticks to the errors (added a few more compared to the suggestion for elements, hope it's fine)
- Fixed a link issue in the file tree error hint.
- Removed the file tree demo page

I think for this PR we should be good to go (except if we find any other issues ofc).

The documentation follow up PR is available in https://github.com/withastro/starlight/pull/1563

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -0,0 +1,245 @@
...
+ * @returns The processed HTML for the file tree.
+ */
+export function processFileTree(html: string, directoryLabel: string) {
+	const file = fileTreeProcessor.processSync({ data: { directoryLabel }, value: html });
+
+	return file.toString();
+}
+
+/** Rehype processor to extract file tree data and turn each entry into its associated markup. */
+const fileTreeProcessor = rehype().use(function fileTree() {
```
delucis, line 35: Ah, just noticed one thing while referencing this PR during work on `<Steps>`! Should `rehype` be run in “fragment” mode here?

```suggestion
const fileTreeProcessor = rehype()
	.data('settings', { fragment: true })
	.use(function fileTree() {
```

This will change the shape of the parsed HTML (no `<body>` etc.), but is what our other component rehype pipelines do (e.g. tabs)

HiDeoo: Great catch, did not even notice that, just [pushed](https://github.com/withastro/starlight/pull/1308/commits/d9dfc9f09afeef757a4464e27077b08769802e18) a new version fixing that, and it even avoids relying on `selectAll` :tada:

The change seems huge but it's mostly due to indentation changes, the actual change is just a few lines.

delucis: Nice! LGTM 🙌

delucis (approved):
Giving this an official stamp of approval for Friday’s release. Lovely work @HiDeoo! 🌟

PR #1310, opened 2023-12-24
Migrate localization tooling to Lunaria

Yan-Thomas:
### Description

This PR migrates the localization tooling to use Lunaria. (PR removing all the old code as follow-up)

Note: this PR will need some Vercel deploy magic. locally, the dashboard will be built to `docs/dist/lunaria/index.html`.

A few screenshots for y'all

image
image

That's all! And don't forget:

MERRY CHRISTMAS, BABY!


docs/lunaria/components.ts
```diff
@@ -0,0 +1,15 @@
...
+export const TitleParagraph = () => html`
+	<p>
+		If you're interested in helping us translate
+		<a href="https://starlight.astro.build/">starlight.astro.build</a> into one of the languages
+		listed below, you've come to the right place! This auto-updating page always lists all the
+		content that could use your help right now.
+	</p>
+	<p>
+		Before starting a new translation, please read our
+		<a href="https://contribute.docs.astro.build/guides/i18n/">i18n Guide</a>
```
HiDeoo, line 12: Wow, such a great and exciting PR 🤩

Small, non-technical question: I wonder if instead of only linking to the Astro Docs Docs, we should also link to https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#translating-starlights-docs too? The AD² page contains a lot of Astro specific informations that could be confusing for people who want to contribute to the Starlight docs, e.g. the "Adding a new language" section. Any thoughts?

Yan-Thomas: Yeah, that's a good point, we'll eventually be able to merge most of the AD² guidance since it will be using Starlight instead of our custom logic for i18n. I am not exactly sure how exactly to phrase/organize the paragraph here, so I'll let @delucis make the call.

delucis: I think we can just straight up link to the Starlight contributing docs in this case and incorporate an AD² link in our `CONTRIBUTING.md` eventually:

```suggestion
		<a href="https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#translating-starlights-docs">translation guide</a>
```

docs/lunaria.config.json
```diff
@@ -0,0 +1,94 @@
...
+    },
+    {
+      "label": "Русский",
+      "lang": "ru"
+    },
+    {
+      "label": "Türkçe",
+      "lang": "tr"
+    },
+    {
```
delucis, line 60: We recently merged in Ukrainian too!

```suggestion
    {
      "label": "Українська",
      "lang": "uk"
    },
    {
```

delucis (commented):
This looks great @Yan-Thomas, thank you!

Could you also add a step to our [guidance on adding a new language](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#adding-a-new-language-to-starlights-docs) to say that the Lunaria config also needs updating?

delucis (approved):
Thanks again Yan! Added that note to `CONTRIBUTING.md` so will merge this and start setting up the new deployment!

PR #1314, opened 2023-12-26
showcase: Add docs.ghostfam.com

cuongtqtran:
Adding a new showcase, this is the documentation of TanaFlows theme docs from ghostfam.com, a ghost theme docs.
Available at https://docs.ghostfam.com/


HiDeoo (approved):
Thanks for the great addition to the showcase :raised_hands:

I took the liberty of slightly updating the PR to move the new entry at the bottom of the list as we append new entries and I also updated the thumbnail to be less zoomed out.

PR #1322, opened 2023-12-31
[i18nIgnore] Add src alias

at-the-vr:
#### Description

- [x] brings alias for `docs/src` path
- [ ] allow users to code navigate using ctrl/cmd + left click - edit: this might be a "~Path  Intellisense~ MDX" extension issue 🤔
- [x] update all necessary markdown files with alias
- [ ] update all necessary .astro components with alias - edit: the components in `docs/src` path **example:** [**sidebar-preview.astro**](https://github.com/withastro/starlight/blob/main/docs/src/components/sidebar-preview.astro) use dependencies from `packages/starlight/components`. There will another path alias for `packages/starlight/components` which I am doubtful since its a bit out of scope for this PR


delucis (approved):
Thanks for the work on this @at-the-vr! I just pushed a small commit to also add these changes to a few translations that came in after your initial commits and updated the title to include `[i18nIgnore]` so that this PR won’t impact translation status on https://i18n.starlight.astro.build/

Re: your two unchecked checkboxes in your PR description:

1. We [discussed the code navigation issue on Discord](https://discord.com/channels/830184174198718474/1190945940668952596) and concluded it’s an issue with the MDX language server. I think it’s OK to continue with this PR and lose that benefit for now.

2. I think all content has been updated! The relative imports from `packages/starlight` I think probably should not be updated and also are imported from internal component code rather than docs content, so I don’t think are as much of an issue.

Given that, I think this is ready to go. Thanks again 🚀

at-the-vr:
thanks for the assist 🔥🔥

PR #1324, opened 2024-01-02
[i18nIgnore] docs: `pnpm install` → `pnpm add`

HiDeoo:

#### Description

As discussed in the last Talking and Doc'ing session, it seems that even tho `pnpm install <pkg>` works and do the expected job, it might be better to align with the `pnpm` docs and use `pnpm add <pkg>` instead:

- `pnpm install` is used to install all dependencies for a project ([reference](https://pnpm.io/cli/install))
- `pnpm add <pkg>` installs a package and any packages that it depends on ([reference](https://pnpm.io/cli/add))

This PR updates the docs to use `pnpm add` instead of `pnpm install` when installing a specific package.

If there are any reasons to keep `pnpm install` for reasons that none of the session attendees thought of, we can close this PR altho we might want to notify the docs team so they are aware of it before continuing the same process (https://github.com/withastro/docs/pull/5978/files already includes this changes).


at-the-vr (approved):
LGTM, thanks for the PR. I do recall this was discussed on last week Talking and Doc'ing

ElianCodes (approved):
Good catch! Thanks Hideoo!

kevinzunigacuellar (approved):
pnpm add this PR!

delucis (commented):
This looks good to me!

@HiDeoo any thoughts on doing this across languages and adding an `[i18nIgnore]`? Feels like the kind of change that would be safe to do that way and avoid marking more stuff out of date than necessary.

dreyfus92 (approved):
LGTM ✅

HiDeoo:
> @HiDeoo any thoughts on doing this across languages and adding an `[i18nIgnore]`? Feels like the kind of change that would be safe to do that way and avoid marking more stuff out of date than necessary.

Sorry for the late reply. This is a great idea, I don't quite remember if I did not think about it or if I wanted to get an approval before doing it but no matter what, I'll update the branch, update all locales and add the `[i18nIgnore]` tag in the next few days.

HiDeoo:
Just pushed a new commit applying the changes to all languages and added the `[i18nIgnore]` tag to the PR title, we should be good for a final review.

There is now only 1 reference to `pnpm install` in the `docs/` folder which is in the `docs/README.md` file:

```sh
$ rg 'pnpm install'
README.md
16:| `pnpm install`          | Installs dependencies                            |
```

delucis (approved):
Thank you @HiDeoo! This looks good to me. Time to `pnpm add` this PR to the docs 🚀

PR #1338, opened 2024-01-04
i18n(pt-BR): Update `reference/configuration`

Fryuni:
#### Description

Update the Portuguese translation of the configuration reference.

PR #1274, opened 2023-12-13
[i18nIgnore] docs: fix more documentation links

HiDeoo:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)

#### Description

Following the release of the version [`0.5.0`](https://github.com/HiDeoo/starlight-links-validator/releases/tag/v0.5.0) of the `starlight-links-validator` plugin, I ran a new validation pass on the documentation. This PR fixes the broken links that were found and are fixable.

The plugin configuration used was:

```js
plugins: [
  starlightLinksValidator({
    errorOnFallbackPages: false,
    errorOnInconsistentLocale: true,
    errorOnRelativeLinks: false,
  }),
],
```

PR #1339, opened 2024-01-04
[i18nIgnore] docs: Add `starlight-links-validator` to docs

Fryuni:
#### Description

Adds [`starlight-links-validator`](https://starlight-links-validator.vercel.app/) plugin to the docs, ensuring that all links are valid.

The plugin is configured such that pages should link to the same locale, even when there is no translation for that page in the locale (which will show the fallback content in English but the rest of the UI will be in the previous locale).

Adding the plugin uncovered an issues on the current Indonesian (Bahasa Indonesia) translation of `guides/authoring-content`. That page links to the `plugins` option in `reference/configuration`, which also _is_ translated, but doesn't have the section about the `plugins` yet. 

The same also happens for the Portuguese translation of the `reference/plugins` page, pointing to the same section, but I have an open PR updating the Portuguese translation of `reference/configuration` open already (#1338). 

What we can do about those cases are:
- Replace them temporarily with external links to the English page, so the plugin doesn't validate them
- Not configure the plugin to require links to the same locale, and once this one instance gets fixed we can enable that option
- Link to the page without the anchor
- Block this until there are translations for the `plugins` option on the two offending languages


HiDeoo:
Neat, just quickly dumping a few notes from my phone:

- https://github.com/HiDeoo/starlight-links-validator/issues/25#issuecomment-1862578037 contains a few comments regarding the remaining broken links, and considering there is only one left after your other PR, we may be able to ping / find someone to fix the remaining one before merging?
- If we do not want to change the relative links, there is a `errorOnRelativeLinks` option that can be used as shown in https://github.com/withastro/starlight/pull/1274 (I personally prefer the explicitness added by the changes in this PR)
- I don't think this should run three times:
  - I don't think blocking the preview is a good experience for the author and even more for reviewers as it forces any reviewers to check a PR locally in dev mode if there is a broken link or wait for them to be fixed before potentially pointing out other issues.
  - The a11y check is already slow enough
  - The Windows run is pretty slow too
  - I wonder if we should just not add another run on Linux which will be the only one to check for broken links? Any thoughts?

docs/astro.config.mjs
```diff
@@ -199,6 +200,10 @@ export default defineConfig({
 					autogenerate: { directory: 'reference' },
 				},
 			],
+			plugins: [starlightLinksValidator({
+			  errorOnFallbackPages: false,
+			  errorOnInconsistentLocale: true,
+			})],
```
delucis, line 206: Can we move this behind an environment variable flag and add a dedicated CI job that runs it? Right now a broken link causes many tests and the preview deployment to fail. It would be nice to separate it out so we have a clear indication that a broken link is the cause (can can still see the preview while waiting for the fix).

Something like:

```suggestion
			plugins: process.env.CHECK_LINKS ? [starlightLinksValidator({
			  errorOnFallbackPages: false,
			  errorOnInconsistentLocale: true,
			})] : [],
```

And then a new job in `.github/workflows/ci.yml` like:

```yml
  links:
    name: Check for broken links
    runs-on: ubuntu-20.04
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Setup PNPM
        uses: pnpm/action-setup@v2

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version: ${{ env.NODE_VERSION }}
          cache: 'pnpm'

      - name: Install Dependencies
        run: pnpm i

      - name: Build docs site and check for links
        working-directory: ./docs
        run: pnpm build
        env:
          CHECK_LINKS: true
```

delucis (commented):
Lovely stuff @Fryuni — thanks for proposing and tackling this! Left one suggestion to improve CI DX.

docs/astro.config.mjs
delucis: (Heh, I’m too slow for @HiDeoo as usual, but always on the same page 😁)

delucis (commented):
Looking great @Fryuni! I think once we fix that one broken link this should be good to go 🥳

delucis (approved):
Lovely work! Let’s get link checking 🥳

PR #1342, opened 2024-01-04
i18n(pt-BR): Update `showcase`

Fryuni:
#### Description

- Update the Portuguese translation of the showcase page
- Update the link on the Portuguese translation of `reference/configuration` that was pointing to the incorrect section while the correct section was had not been translated.  
  Rewrite happened on #1339


delucis (approved):
Thanks for tying up the loose ends with that link update @Fryuni too! 🚀

PR #1343, opened 2024-01-04
i18n(pt-BR): Update `reference/plugins`

Fryuni:
#### Description

- Update the Portuguese translation of the plugins reference.
- The new link is broken as it points to a section added on #1342, once that gets merges the link here will be valid


delucis (approved):
Thanks! And so fun to get to update branch and see the link checker go from red to green 🥳

PR #1355, opened 2024-01-10
docs: Clarify auto-generated group default ordering

Fryuni:
#### Description

Auto-generated groups on the sidebar are sorted by the collation of the site's default locale. This PR makes that distinction explicit in the docs that previously stated that it was alphabetical. In many locales, including English, those two are the same but for some locales, this distinction may be unexpected, especially so when reading a documentation in one locale and creating a site in another.


docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -138,7 +138,8 @@ The configuration above generates the following sidebar:
 
 Starlight can automatically generate a group in your sidebar based on a directory of your docs.
 This is helpful when you do not want to manually enter each sidebar item in a group.
-Pages will be sorted alphabetically by filename by default.
+
+By default, pages will be sorted according to the filename in your [`defaultLocale`](/reference/configuration/#defaultlocale). This order generally follows the native speakers' agreed-upon “alphabetical” arrangement. However, variations exist between locales and even within a locale, where the definition of “alphabetical” may differ for formal versus informal text or based on case sensitivity. The exact ordering that Astro uses is the one defined by the [ICU Collation](https://unicode-org.github.io/icu/userguide/collation/) of your default locale without any [customization](https://unicode-org.github.io/icu/userguide/collation/customization/).
```
sarah11918, line 142:
```suggestion
By default, pages will be sorted in alphabetical order according to the filenames in your [`defaultLocale`](/reference/configuration/#defaultlocale). 

:::tip
The definition of “alphabetical” may differ between locales or even within a locale where conventions may differ based on context or case sensitivity. Astro follows the order defined by the [ICU Collation](https://unicode-org.github.io/icu/userguide/collation/) of your default locale without any [customization](https://unicode-org.github.io/icu/userguide/collation/customization/).
:::
```
Was thinking something like this might work! (Haven't viewed a preview, just editing in place.) What do we think?

Fryuni: There are some edge cases if the locale includes a variation, like Traditional European Spanish (es-ES-u-co-trad) which allow letter groups interfere with the ordering, so it is not alphabetical even by their own definition. For example, in Traditional European Spanish "cosas" is sorted before "chica" even though they sort "h" before "o".

That is why I replaced the sentence stating that it _is_ alphabetical.

But that may be too purist of me. And as Tim Peters said, practicality beats purity.

sarah11918: Acknowledged, but I also think we've maybe given people *enough* of a heads up that "alphabetical" might not be what they expect, so we probably don't have to describe the specifics to that full extent here?

What do we think visually of the addition? Tip not too big/distracting?

Fryuni: If we go with a tip I think we can leave the previous sentence on the same paragraph as before.

Fryuni: I think it looks fine. It is at the end of the explanation and before the examples so it doesn't break the flow IMO.

In the same paragraph as before:
image

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -138,7 +138,11 @@ The configuration above generates the following sidebar:
 
 Starlight can automatically generate a group in your sidebar based on a directory of your docs.
 This is helpful when you do not want to manually enter each sidebar item in a group.
-Pages will be sorted alphabetically by filename by default.
+By default, pages will be sorted in alphabetical order according to the filenames in your [`defaultLocale`](/reference/configuration/#defaultlocale).
```
sarah11918, line 141:
```suggestion

By default, pages will be sorted in alphabetical order according to the filenames in your [`defaultLocale`](/reference/configuration/#defaultlocale). 
```
I will let you decide, or some 3rd-party tie breaker, but I *do* prefer this as a separate line. I don't like burying key information like "what the default is!" at the end of an already long-ish paragraph. And, I think this sets up the tip nicely. There's no way someone is not going to take away "alphabetical" after seeing all this, AND FURTHERMORE that "alphabetical" might not even be what they expect.

So, non-blocking, and I'll allow merging either way, but this is def my personal preference.

sarah11918 (approved):
Let's do this! I'm approving, will let Chris do a final check and the honors.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -138,7 +138,12 @@ The configuration above generates the following sidebar:
 
 Starlight can automatically generate a group in your sidebar based on a directory of your docs.
 This is helpful when you do not want to manually enter each sidebar item in a group.
-Pages will be sorted alphabetically by filename by default.
+
+By default, pages will be sorted in alphabetical order according to the filenames in your [`defaultLocale`](/reference/configuration/#defaultlocale).
```
kevinzunigacuellar, line 142: If possible to hold the docs update. The sorting may change, again (by bad 😅 ).

For context #1151, we might use the slug instead of the filename.
```suggestion
By default, pages will be sorted in alphabetical order according to the [`slug`](reference/overrides/#slug) in your [`defaultLocale`](/reference/configuration/#defaultlocale). 
```

kevinzunigacuellar: #1151 merged 💜

Fryuni: Wonderful

Fryuni:
@delucis Ball is with you on this one

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -138,7 +138,12 @@ The configuration above generates the following sidebar:
 
 Starlight can automatically generate a group in your sidebar based on a directory of your docs.
 This is helpful when you do not want to manually enter each sidebar item in a group.
-Pages will be sorted alphabetically by filename by default.
+
+By default, pages will be sorted in alphabetical order according to the file [`slug`](reference/overrides/#slug) in your [`defaultLocale`](/reference/configuration/#defaultlocale). 
+
+:::tip
+The definition of “alphabetical” may differ between locales or even within a locale where conventions may differ based on context or case sensitivity. Astro follows the order defined by the [ICU Collation](https://unicode-org.github.io/icu/userguide/collation/) of your default locale without any [customization](https://unicode-org.github.io/icu/userguide/collation/customization/).
+:::
```
delucis, line 146: I wonder if the nitty gritty of the collation logic is really important enough to warrant a big callout? I think we can probably get away with the simple initial statement. (I’ll also be curious to see if this ever causes people issues, i.e. whether we should always sort in `'en'` or something if in some languages people get unexpected results, but I think we can hedge our bets and document more detail if we see this is something that trips people up.)

```suggestion
By default, pages are sorted in alphabetical order according to the file [`slug`](reference/overrides/#slug).
```

delucis (commented):
Thanks for working on this! Hope the big red delete lines in this review don’t make me look TOO enthusiastic with the scissors 😅

Making this suggestion to avoid overloading the reader with information if there is a high likelihood they don’t need to know it.

docs/src/content/docs/guides/sidebar.mdx
Fryuni: Maybe a footnote would be better and less intrusive then an aside?

delucis: We don’t really have footnotes as an established pattern and they’re a bit tricky to make accessible, but really I’d argue that information we’re thinking about hiding in a footnote is perhaps not information that is essential to include.

It can be tempting to include everything we know in docs, but docs aren’t 1:1 descriptions of implementation — they are guides that help people get where they need to go and sometimes it’s as important to decide when NOT to tell someone something, so as not to overload or confused them, as it is to decide WHAT to tell them.

Fryuni: The thing is, and that is the reasons I opened this PR, saying it is sorted alphabetically is _technically_ incorrect. It is correct in English.

Event the docs for [the higher level API `String.localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare), which doesn't require the use of the lower-level Collator object, only mentions "alphabetically" once, in one English example, because it doesn't guarantee alphabetical order in other cases.

In fact, for some languages there is a definition for alphabetical ordering and that is _not_ what the collator will yield (for a variety of reasons).

And this doesn't apply only to non-English text, English words are sorted differently in other languages. Even if we assume that the slugs would all be in English, if you take [this word list](https://github.com/powerlanguage/word-lists/blob/master/common-7-letter-words.txt), which is all English, no special characters and no diacritics, and sort it using a Collator from multiple locales you'll find that just those words differ in order in at least 68 locales (the locale list is not necessarily complete).

<details>
<summary>Code for reproduction</summary>

```js
const fs = require('node:fs');
const {all: locales} = require('locale-codes');

const defaultCollator = new Intl.Collator('en-US');
const displayName = new Intl.DisplayNames(['en'], { type: 'language' });

const words = fs.readFileSync('words.txt', 'utf8')
  .trim()
  .split('\n')
  .filter(line => !!line)
  .sort((a, b) => defaultCollator.compare(a, b));

const differenceSamples = [];

const localeTags = new Set(locales.map(locale => locale.tag));

console.log('Testing locales:', localeTags.size);

for (const localeTag of localeTags.values()) {
  try {
    const collator = new Intl.Collator(localeTag);

    const localeWords = Array.from(words)
      .sort((a, b) => collator.compare(a, b));

    for (let i = 0; i < words.length; i++) {
      if (localeWords[i] !== words[i]) {
        const sample = {
          locale: displayName.of(localeTag),
          tag: localeTag,
          sampleEnglish: words.slice(i - 2, i + 2),
          sampleLocale: localeWords.slice(i - 2, i + 2),
        };

        differenceSamples.push(sample);
        break;
      }
    }
  } catch {
    // Nothing
  }
}

fs.writeFileSync('differences.json', JSON.stringify(differenceSamples), 'utf8');
```
</details>

And to make it worse, most people don't know that this difference exist. The local alphabetical order is so deeply rooted in everyone's mind that they don't even entertain the notion that there could be a different order in other places/scenarios. So it will just seem wrong for people making sites in any of those locales.

So I think this is something important to add.


_P.S.: I'm aware of this because the official alphabetical order in my own language changed while I was in school because 3 letters (K, W and Y) that were used in a few words but not considered part of the alphabet and were sorted after everything else (I don't know what is more absurd, that or changing the order)._

sarah11918: I'm going to agree with cutting the tip for the same reason that we "don't document how JavaScript works" -- we don't document how alphabetical order works.

But also, I clicked the link provided, and it may be a source of truth, but it's not very user friendly. I spent minutes trying to find *helpful* information there, as if, I spoke language X and wanted to check my language. And I failed.

So, I think the tip is doubly unhelpful: it provides a technicality that most people will not encounter, is potentially overwhelming and distracting to include (e.g. making people think they need to go check here), one that is not our place to guide them on, one that they probably have better local resources for and are aware of if they regularly use this language, because they'd be aware of its alphabetization (standard or not, or current issues)... and then lastly the resource we *do* provide also seems to be overwhelming to a novice in the area.

I am totally OK with someone noticing, "Hey, my files aren't sorted the way I thought they were" and then doing their own investigation, potentially with better context since they presumably would use a language where this is an occurrence even outside of software.

Fryuni: Ok, that sounds fair. No more tip then :)

delucis (approved):
Thanks for the discussion everyone! Really appreciate the care and attention that went into discussing this 💖

PR #1359, opened 2024-01-11
docs(showcase): add OpenSaaS.sh

vincanger:
adds OpenSaaS.sh docs to the showcase


delucis (approved):
Yay! Thanks for taking the time to come back to this @vincanger 🙌

PR #1360, opened 2024-01-11
i18n(es): Update `index`

dreyfus92:
#### Description

Update `index` based on this commit 238e3615f4d334bf948778a3f71cea0237ae3068

### DIFF
--- a/docs/src/content/docs/es/index.mdx
+++ b/docs/src/content/docs/es/index.mdx
@@ -47,7 +47,7 @@ import Testimonial from '~/components/testimonial.astro';
 	</Card>
 </CardGrid>
 
-<TestimonialGrid title="Qué es lo que dice la gente">
+<TestimonialGrid title="Lo que nuestros usuarios opinan">
   <Testimonial
     name="Rachel"
     handle="rachelnabors"
--- a/docs/src/content/docs/es/index.mdx
+++ b/docs/src/content/docs/es/index.mdx
@@ -24,6 +24,8 @@ hero:
 
 import { CardGrid, Card } from '@astrojs/starlight/components';
 import AboutAstro from '~/components/about-astro.astro';
+import TestimonialGrid from '~/components/testimonial-grid.astro';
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/es/index.mdx
```diff
@@ -45,6 +47,113 @@ import AboutAstro from '~/components/about-astro.astro';
 	</Card>
 </CardGrid>
 
+<TestimonialGrid title="Qué es lo que dice la gente">
```
kevinzunigacuellar, line 50: "gente" is a touch informal for my taste but I couldn't decide on a good alternative. Here are some options:
- Lo que dicen las personas
- Lo que expresan las personas

I am also open to other alternatives

dreyfus92: What do you think about "Qué es lo que opinan lo demás"?

delucis: Feel free to interpret it liberally! Excuse my bad Spanish, but just to show how free we can be, it could also be flipped around to something like “Opiniones que hemos oído” for example if that worked better in translation. No need to be literal on a heading like this that is kind of punchy and more “marketing” in context.

kevinzunigacuellar: Marketing is hard 😅

dreyfus92: Yeah, it can be 😁 which suggestion would make more sense:
- Qué es lo que opinan lo demás
- Opiniones que hemos oído
- Lo que estamos oyendo

kevinzunigacuellar: I like the style of "Lo que estamos oyendo", I tweak it a bit to follow the translation. "Lo que nuestros usuarios opinan" What do you think? @dreyfus92

kevinzunigacuellar (approved):
LGTM! Thanks @dreyfus92 💜

PR #1361, opened 2024-01-12
i18n(zh-cn): Update index.mdx

liruifengv:
#### Description

- Update index.mdx
- Follow  https://github.com/withastro/starlight/pull/1104

PR #1363, opened 2024-01-12
i18n(ko-KR): update `index.mdx`

jsparkdev:
#### Description

- update `index.mdx`
- #1104

### DIFF
--- a/docs/src/content/docs/ko/index.mdx
+++ b/docs/src/content/docs/ko/index.mdx
@@ -24,6 +24,8 @@ hero:
 
 import { CardGrid, Card } from '@astrojs/starlight/components';
 import AboutAstro from '~/components/about-astro.astro';
+import TestimonialGrid from '~/components/testimonial-grid.astro';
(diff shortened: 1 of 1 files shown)


delucis (commented):
Hey @jsparkdev! Thanks for the PR. Do you think we should also translate the text of the posts? This is what I had assumed we’d do and #1360 and #1361 did translate for Chinese and Spanish, so maybe this should be consistent?

jsparkdev:
I was debating whether I should translate it or not, but after looking at the work of other translators, I thought I should translate it, so I completed it! thank you for telling me :)

delucis (approved):
Amazing — thank you! 🌟 

Yes, I could understand also the idea that these are direct quotes and so to keep them in the original language, but given not every reader can speak English, I think the translations are helpful and people can still click through to the original posts to read the English if they want.

jsparkdev:
Thank you for providing a new perspective. This was helpful!

PR #1365, opened 2024-01-12
extend config options to pagefind

### LINKED ISSUE #1364: config options are not followed in pagefind search

0.15.3

4.0.1

pnpm

Mac

Chrome

### Describe the Bug

Pagefind UI links do not follow build options from astro config file

Issue discussed in #1358

https://stackblitz.com/edit/github-f5e5ew?file=astro.config.mjs

### Participation

kevinzunigacuellar:
#### Description

- Closes #1364
- This PR exposes astro config options to Pagefind UI to correctly format search result


packages/starlight/components/Search.astro
```diff
@@ -53,6 +53,8 @@ const pagefindTranslations = {
 </site-search>
 
 <script>
+	import { formatPathNoBase } from '../utils/format-path';
```
delucis, line 56: I know it might require some code duplication but I’m hesitant about importing this utility into client-side code. Can we move this into a minimal version of the code just for this use case inside the `<script>` tag?

delucis (changes requested):
Thanks for the speedy work @kevinzunigacuellar! Left a comment.

packages/starlight/components/Search.astro
kevinzunigacuellar: Sure!

delucis: Yes. Basically to isolate the client-side script from modules that are used elsewhere and might import all kinds of things unintentionally at some point.

delucis (commented):
@kevinzunigacuellar Thanks for your patience on this one! Was hoping to get a review before today’s release, but because this can go out in a patch, prioritised minor changes instead.

But now I had some time to do some research and tests to see if all this extra client-side JS is needed and I have some fun news: I don’t think so!

I spun up a test project and built with various settings to test out the current behaviour and reached the conclusion that the only change needed is to support `trailingSlash: 'never'`.

- With `build.format: 'file'`, Pagefind correctly recognises the format and generates URLs like `/guides/example.html`
- With `base: 'foo'`, because we already pass that to Pagefind, it’s handled correctly as `/foo/guides/example/`

From what I can tell Pagefind always assumes a trailing slash, so we just need to account for the case where it needs stripping. Should be nice to simplify this!

I also think we’ll need a slightly different function than the current `stripTrailingSlash()` because these paths can be stuff like `/guides/example/#heading` where the `/` before the fragment needs stripping.

You still up for reworking this?

packages/starlight/components/Search.astro
```diff
@@ -108,10 +136,12 @@ const pagefindTranslations = {
 			});
 
 			let translations = {};
+			let config = {};
 			try {
 				translations = JSON.parse(this.dataset.translations || '{}');
+				config = JSON.parse(this.dataset.config || '{}');
 			} catch {}
-
+			const formatURL = createURLFormatter(config);
```
delucis, line 144: And skip lots of the indirection:

```suggestion
			const formatURL = (href: string) => shouldStrip ? stripTrailingSlash(href) : href;
```

packages/starlight/components/Search.astro
```diff
@@ -61,6 +73,22 @@ const pagefindTranslations = {
...
+				{ trailingSlash = 'ignore' }: ConfigOptions = {}
+			) => {
+				if (trailingSlash !== 'never') return path;
+				return stripTrailingSlash(path);
+			};
+
+			const createURLFormatter = (opts: ConfigOptions) => {
+				return (href: string) => pagefindFormatURL(href, opts);
+			};
+
```
delucis, line 91: Allows to delete all of this:

```suggestion
```

packages/starlight/components/Search.astro
```diff
@@ -53,6 +61,10 @@ const pagefindTranslations = {
 </site-search>
 
 <script>
+	interface ConfigOptions {
+		trailingSlash?: 'ignore' | 'always' | 'never';
+	}
+
```
delucis, line 67: And this:
```suggestion
```

packages/starlight/components/Search.astro
```diff
@@ -108,10 +136,12 @@ const pagefindTranslations = {
 			});
 
 			let translations = {};
+			let config = {};
 			try {
 				translations = JSON.parse(this.dataset.translations || '{}');
+				config = JSON.parse(this.dataset.config || '{}');
```
delucis, line 142: And this:
```suggestion
```

packages/starlight/components/Search.astro
```diff
@@ -13,9 +14,16 @@ const pagefindTranslations = {
 			.map(([key, value]) => [key.replace('pagefind.', ''), value])
 	),
 };
+
+const projectConfig = {
+	trailingSlash: project.trailingSlash,
+};
```
delucis, line 20: And this also goes away:

```suggestion
```

delucis (commented):
Nice, thanks @kevinzunigacuellar! I think we can go even simpler still.

kevinzunigacuellar:
Should we add tests for this new `striptrailingslash`? I tested locally a few searchs. The only instance where it didn't work was when referencing an id `#_top` in the landing page. it strips the slash but chrome (I think) automatically adds it back because we are in the `/` page.

delucis (approved):
> Should we add tests for this new `striptrailingslash`? I tested locally a few searchs. The only instance where it didn't work was when referencing an id `#_top` in the landing page. it strips the slash but chrome (I think) automatically adds it back because we are in the `/` page.

I’m not sure testing would add all that much in this case given how much we don‘t control in this context and how relatively simple the function is. I updated the regular expression to handle the case you mentioned and avoid stripping `/` in the case of things like `/` or `/#something`.

Thanks again for tackling this @kevinzunigacuellar! ✨

PR #1368, opened 2024-01-14
fix: add open new tab to social links

jsparkdev:
#### Description

Most websites' social link buttons are expected to open in a new window. (including [Astro Docs](https://astro.build))
However, the social link buttons on the [Starlight website](https://starlight.astro.build/) do not.
Fix this to improve UX.

### DIFF
--- a/.changeset/warm-hotels-tie.md
+++ b/.changeset/warm-hotels-tie.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+fix socialicons component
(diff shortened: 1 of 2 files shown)


delucis:
Thank you for the PR @jsparkdev! However, we intentionally avoid these opening in a new tab. Here is some UX research that recommends against this pattern: https://www.nngroup.com/articles/new-browser-windows-and-tabs/

In general forcing a user to open something in a new tab is not a good idea because it can be confusing and interferes with their standard way of deciding where to open a link (e.g. click = same window, cmd/ctrl + click = new window/tab).

Individual Starlight sites can use an override and the `<Icon>` component if they really want this, but I’d recommend against this as the default.

(And I might need to talk to the people working on astro.build… 😅)

jsparkdev:
Thank you for a detailed description!

PR #1372, opened 2024-01-15
Update ToC highlighting styles to prevent UI shifts

### LINKED ISSUE #1120: ToC highlighting UI shifts

0.13.0

3.6.0

pnpm

Mac

Chrome

### Describe the Bug

While browsing the new Astro Docs in the Starlight migration branch, I noticed that the table of contents highlighting styles can lead to some visually distracting UI shifts that I never noticed before but this is happening a lot in the Astro v3 migration guide for example.

Here is a video of the issue where I am just scrolling the page:

https://github.com/withastro/starlight/assets/494699/39039404-bd68-49b0-84a8-79ab1b9213aa

_Note: I pasted the content of this page in our repo and this repro too so this is not related to some changes that may have been made in the migration branch._

https://docs-git-fork-theotterlord-starlight-migration-astrodotbuild.vercel.app/en/guides/upgrade-to/v3/

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1120

This PR applies the [suggested changes](https://github.com/withastro/starlight/issues/1120#issuecomment-1883903946) from @doodlemarks to prevent UI shifts in the table of contents when scrolling through a page.

| Old | New |
|---|---|
| SCR-20240115-pkop | SCR-20240115-pkrj |

> Currently those styles are also applied on :hover and :focus, but with this update, I think we can remove that and let the default hover/focus styles kick in.

I applied this suggestion too which means that hovering the currently highlighted ToC item will change its color to the default link hover color whereas before hovering the currently highlighted ToC item would have no effect. Just wanted to mention this to make sure this is the intended behavior.


.changeset/young-roses-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
```
delucis, line 2: I think we should probably make this a minor given it will cause a change in the visual appearance of people’s sites.

```suggestion
'@astrojs/starlight': minor
```

.changeset/young-roses-tickle.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Updates the table of contents highlighting styles to prevent UI shifts when scrolling through a page
```
delucis, line 5: I can’t decide if this is a good idea or not 😅 

We sometimes provide migration advice in case someone is keen to preserve previous behaviour, so I guess in theory we could here too? Don’t know what you think — I guess we don’t really _want_ people to do this, but could still be handy if someone does.

````suggestion
Updates the table of contents highlighting styles to prevent UI shifts when scrolling through a page.

If you want to preserve the previous, buggy styling, you can add the following custom CSS to your site:

```css
starlight-toc a[aria-current='true'],
starlight-toc a[aria-current='true']:hover,
starlight-toc a[aria-current='true']:focus {
	font-weight: 600;
	color: var(--sl-color-text-invert);
	background-color: var(--sl-color-text-accent);
}
```
````

delucis (commented):
Oh nice! This was on my to-do list this week — fun for someone else to do it instead 😁 

Left a couple of notes, but the style change looks good to me 🙌

delucis (approved):
Thanks! This looks great now.

Let’s hold before merging — maybe we can get a few more changes in a new minor. I’d love to review #1175 today and have a few other styling things that could pair well with this.

delucis:
Although I won’t release the minor just yet, merging this early because I can see a sidebar style issue in #1383 and want to see if it is also present with this change and then fix it once, instead of fixing it for the old styling on that branch and then maybe having to fix again for these changes. Hoping to release v0.16 tomorrow anyway, so hopefully shouldn’t block any other patches between then and now.

PR #1375, opened 2024-01-16
docs: ignore testimonials from search results

HiDeoo:

#### Description

This PR is a follow-up to this [Discord message](https://discord.com/channels/830184174198718474/1070481941863878697/1196378884950720512) that got some positive reactions so I decided to go ahead and make a PR for it (happy to close it if it's not wanted for whatever reason).

The idea is to ignore testimonials from search results as searching for "subdomain", "component", or "monorepo" for example can return results redirecting to the top of the homepage which I don't think is very useful from a user perspective.

### DIFF
--- a/docs/src/components/testimonial-grid.astro
+++ b/docs/src/components/testimonial-grid.astro
@@ -6,7 +6,7 @@ interface Props {
 const { title } = Astro.props;
 ---
 
-<div class="testimonial-wrapper not-content">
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks @HiDeoo! I’d missed that comment, but yes, agree that these are pretty noisy for search results. Appreciate the attention to detail 💖

PR #1376, opened 2024-01-16
Tweak vertical spacing in Markdown content styles

delucis:
#### Description

This is a subtle change to Starlight’s default content styling that should improve most sites:

- Default vertical spacing between content items is reduced from `1.5rem` to `1rem`.
- Spacing before headings is now relative to font size, meaning higher-level headings have slightly more spacing and lower-level headings slightly less.

The overall impact is to tighten up content that belongs together and improve the visual hierarchy of headings to break up sections.

Helpful pages to check:
- Authoring content guide: [before](https://starlight.astro.build/guides/authoring-content/) / [after](https://starlight-git-dx-921-vertical-rhythm-astrodotbuild.vercel.app/guides/authoring-content/) — has a big range of content types to see the overall impact
- Overrides props: [before](https://starlight.astro.build/reference/overrides/#props) / [after](https://starlight-git-dx-921-vertical-rhythm-astrodotbuild.vercel.app/reference/overrides/#props) — example of tightening up API docs-style sections


HiDeoo (commented):
I may not the best to review UI changes, but personally, I like the changes a lot. At the same time subtle like you said but still quite impactful 🌟 

Small question regarding the changeset:

> Spacing before headings is now relative to font size

Wasn't this already the case?

delucis:
> > Spacing before headings is now relative to font size
> 
> Wasn't this already the case?

Previously all headings had a top margin of `2.5rem`, i.e. relative to base font-size of `16px` a static value of `40px`. This is now `1.5em`, so relative to the _heading’s_ size. In practice this means `<h2>` has a bigger margin than before, `<h3>` is more or less the same, and `<h4>` and down have less margin than before.

We don’t have heavy usage of `<h4>` in the Starlight docs, so this isn’t super obvious but the [“Code blocks” section](https://starlight-git-dx-921-vertical-rhythm-astrodotbuild.vercel.app/guides/authoring-content/#code-blocks) of our Markdown docs helps show this nicely I think.

HiDeoo:
Thanks for the explanation that made me understand my confusion: I'm blind, I read in the changes that the old value was `2.5em;` but it is in fact `2.5rem;` :sweat_smile:

dreyfus92 (approved):
Very subtle and easier to read, LGTM 😁

ElianCodes (approved):
Makes it much better! Nice work Chris!

PR #1383, opened 2024-01-18
Refactor virtual module system for layout components

### LINKED ISSUE #1374: Hot Updates Reporting Errors Causing Project Crashes
### 📝 Issue Description

I am translating the docs. 

Update, save, hot update, open the page again and find it crashes. 

### error:

```
Cannot read properties of null (reading 'EditLink')

Astro detected an unhandled rejection. Here's the stack trace:
TypeError: Cannot read properties of null (reading 'EditLink')
```

### Some screenshots:

image

image
image

image

### 📋 On which page(s) it occurs

/zh-cn/reference/api-reference/

maybe all page. Not just one page.

### 🤔 Expected Behavior

Development servers don't crash

### 👀 Current Behavior

Development servers crashed.

### 🖥️ Browser

Chrome

### 📄 Additional Information

windows 11. Powershell.

delucis:
#### Description

- Closes #1374 

- Refactors Starlight’s internal virtual module system for components to avoid circular references
- Previously a single `virtual:starlight/components` exported all the components used in Starlight’s layouts. These are now split into an individual virtual entrypoint per component, e.g. `virtual:starlight/components/EditLink`.
- The previous design introduced a circular reference between `virtual:starlight/components` and some of the components it re-exported. This solves that and should fix an occasional bug where modules were evaluated in the wrong order during HMR, causing dev server errors.
- While this is technically an internal change, I’m aware some people have used the old entrypoint, so marking this as a `minor` and including a bit of migration advice for people who are.

### DIFF
--- a/packages/starlight/integrations/virtual-user-config.ts
+++ b/packages/starlight/integrations/virtual-user-config.ts
@@ -25,7 +25,7 @@ export function vitePluginStarlightUserConfig(
 	const virtualComponentModules = Object.fromEntries(
 		Object.entries(opts.components).map(([name, path]) => [
 			`virtual:starlight/components/${name}`,
-			`export { default as default } from ${resolveId(path)};`,
+			`export { default } from ${resolveId(path)};`,
 		])
 	);
 
--- a/packages/starlight/integrations/virtual-user-config.ts
+++ b/packages/starlight/integrations/virtual-user-config.ts
@@ -22,6 +22,13 @@ export function vitePluginStarlightUserConfig(
 	const resolveId = (id: string) =>
 		JSON.stringify(id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
 
+	const virtualComponentModules = Object.fromEntries(
+		Object.entries(opts.components).map(([name, path]) => [
+			`virtual:starlight/components/${name}`,
+			`export { default as default } from ${resolveId(path)};`,
+		])
+	);
+
 	/** Map of virtual module names to their code contents as strings. */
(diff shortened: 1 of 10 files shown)


HiDeoo (approved):
Well, I tried a bunch of scenarios locally, in dev containers, in GitPod and did not manage to break anything or repro the related issue.
Altho, to be honest, this is not an issue I was hitting myself or could repro easily except during some Talking & Doc'ing sessions with others so my tests may have been for nothing or not that relevant ^^

packages/starlight/integrations/virtual-user-config.ts
```diff
@@ -22,6 +22,13 @@ export function vitePluginStarlightUserConfig(
 	const resolveId = (id: string) =>
 		JSON.stringify(id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
 
+	const virtualComponentModules = Object.fromEntries(
+		Object.entries(opts.components).map(([name, path]) => [
+			`virtual:starlight/components/${name}`,
+			`export { default as default } from ${resolveId(path)};`,
```
dreyfus92, line 28:
```suggestion
			`export { default } from ${resolveId(path)};`,
```

dreyfus92 (commented):
Amazing work @delucis, just wondering if this could work.

packages/starlight/integrations/virtual-user-config.ts
delucis: I believe so, yes!

PR #1390, opened 2024-01-22
i18n(fr): add `resources/plugins`

HiDeoo:

#### Description

This PR adds the french translation of the `resources/plugins` page (which is mostly moved content from the previous showcase) and also updates the link to it in `reference/configuration`.


docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: Modules d'extension et intégrations
```
delucis, line 2: Non blocking, but noting that this title is quite long, causing it to wrap when it’s the current page:

sidebar with page title

It’s an accurate translation though, so I’m happy if you are. I guess one other option could be “Modules d'extension et outils”?

delucis (approved):
LGTM! Left one comment but up to you whether it needs any action or not.

docs/src/content/docs/fr/resources/plugins.mdx
HiDeoo: Great catch, and based on the content, your suggestion is totally valid so I went with it 👍

HiDeoo:
> LGTM! Left one comment but up to you whether it needs any action or not.

As I went with your suggestion, I am using the previous approval to merge the PR without a new review 🫣

Issue #1272, opened 2023-12-12
Built-in components are not tree shakable


yasnbouz:
### What version of `starlight` are you using?

0.15.0

### What version of `astro` are you using?

4.0.1

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Edge

### Describe the Bug

Even though my homepage only uses the `Card` component, the bundle still includes `CardGrid` and `Tabs`.
<img width="1064" alt="shapes at 23-12-12 13 08 29" src="https://github.com/withastro/starlight/assets/45861586/aba1f22c-5996-415b-bfc2-9a4ad846ea39">

The cause of this problem is the use of barrel files in [components.ts](https://github.com/withastro/starlight/blob/5df5798405ed0a29f0a2b2ca9252cda6e28e72a3/packages/starlight/components.ts), which re-export modules from one file. According to the [Vite.js documentation](https://vitejs.dev/guide/performance#avoid-barrel-files), we should avoid barrel files. Therefore, the solution is to enable individual imports of built-in components:

```
import { Card } from '@astrojs/starlight/components;
```
to
```
import Card from "@astrojs/starlight/components/Card.astro";
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-mufzhv?file=src%2Fcontent%2Fdocs%2Findex.mdx

### Participation

- [X] I am willing to submit a pull request for this issue.

delucis:
This is a known issue in Astro’s tree shaking. There have been some attempts to improve this — available with the [`experimental.optimizeHoistedScript`](https://docs.astro.build/en/reference/configuration-reference/#experimentaloptimizehoistedscript) option — but so far they are buggy for Starlight’s case (breaking JS inclusion even when used in some cases).

We did also consider non-barrel imports, but barrel entrypoints do have some nice advantages when thinking of imports that need to be used by content authors:

- Single import location to think of: `@astrojs/starlight/components`
- Import everything you need with one line — especially helpful for combos like `Tabs` and `TabItem` or `Card` and `CardGrid`

I agree that loading extra JS is not ideal though. Ideally, the fix comes from Astro (I can check with the team next week whether further work on this is planned), but otherwise I guess we could probably also export each component independently as well, so people can choose that path if they prefer.

Fryuni:
> but otherwise I guess we could probably also export each component independently as well, so people can choose that path if they prefer.

The components are already available independently just by how ESM resolution works isn't it?

[...]

PR #1395, opened 2024-01-23
Add Expressive Code `<Code>` component to Starlight

hippotastic:
#### Description

- Adds the new `<Code>` component by Expressive Code to Starlight. For the original component, see here: https://expressive-code.com/key-features/code-component/
- Re-exports the component from Starlight to match the import syntax of all other user components:\
  `import { Code } from '@astrojs/starlight/components'`
- Adds an `@astrojs/starlight/internal` export that provides access to the `useTranslations` function to components
- Changes the default behavior of the `build:examples` script so that the projects are no longer built concurrently. The previous configuration could lead to "Port already in use" errors, and building the examples sequentially fixes this.
- Adds documentation for the new component to the Components page


packages/starlight/components.ts
```diff
@@ -4,3 +4,4 @@ export { default as Icon } from './user-components/Icon.astro';
 export { default as Tabs } from './user-components/Tabs.astro';
 export { default as TabItem } from './user-components/TabItem.astro';
 export { default as LinkCard } from './user-components/LinkCard.astro';
+export { Code } from 'astro-expressive-code/components';
```
delucis, line 7: Just to double check, I think you said this only adds CSS/JS when actually used right (same way the standard code block feature does)? Only because as noted in #1272, this module doesn’t currently tree shake. I remember you saying you had it set up that way, but wanted to double check.

packages/starlight/integrations/shared/pathToLocale.ts
```diff
@@ -22,11 +22,11 @@ export function pathToLocale(
...
 	const docsDir = new URL('content/docs/', srcDir);
-	const slug = path
-		// Format path to unix style path.
-		?.replace(/\\/g, '/')
-		// Strip docs path leaving only content collection file ID.
-		// Example: /Users/houston/repo/src/content/docs/en/guide.md => en/guide.md
-		.replace(docsDir.pathname, '');
+	// Format path to unix style path.
+	path = path?.replace(/\\/g, '/')
+	if (path && !path.startsWith('/') && docsDir.pathname.startsWith('/')) path = '/' + path;
```
delucis, line 27: Can we add a comment here about why this step is necessary. You explained it to me, but I’ve already forgotten, so seems like a good sign a comment is a good idea 😁

packages/starlight/integrations/expressive-code/index.ts
```diff
@@ -60,56 +61,55 @@ export type StarlightExpressiveCodeOptions = Omit<AstroExpressiveCodeOptions, 't
...
 };
 
-export const starlightExpressiveCode = ({
-	astroConfig,
+type StarlightEcIntegrationOptions = {
+	starlightConfig: StarlightConfig;
+	useTranslations?: ReturnType<typeof createTranslationSystemFromFs> | undefined;
+};
+
+export function getStarlightEcConfigPreprocessor({
```
delucis, line 69: Given we technically export this publicly (even if only for internal use really), it might be helpful to have a small JSDoc comment explaining that (feel free to tweak this suggestion to something you feel represents this better):

```suggestion
/**
 * Create an Expressive Code configuration preprocessor based on Starlight config.
 * Used internally to set up Expressive Code and by the `<Code>` component.
 */
export function getStarlightEcConfigPreprocessor({
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -195,3 +195,31 @@ The code above generates the following on the page:
 A list of all available icons is shown below with their associated names. Click an icon to copy the component code for it.
 
 <IconsList />
+
+### Code
+
+Starlight uses [Expressive Code](https://expressive-code.com/) to render code blocks with extended formatting possibilities. In addition to fenced code blocks in markdown & MDX documents, Expressive Code can display syntax-highlighted code snippets using its [`<Code>` component](https://expressive-code.com/key-features/code-component/).
+
+The `<Code>` component provides props like `code`, `lang` or `meta` that allow you to dynamically define a code block's contents. Using this component, you can render code blocks from variables or data coming from external sources like files, databases or APIs.
```
delucis, line 203: Suggesting a couple changes here to link to the MD/MDX docs and switch up some of the order:

```suggestion
Starlight uses Expressive Code to render [Markdown and MDX code blocks](/guides/authoring-content/#code-blocks). Use the `<Code>` component to render syntax highlighted code when using a Markdown code block is not possible, for example, to render data coming from external sources like files, databases, or APIs.

See the [Expressive Code “Code Component” docs](https://expressive-code.com/key-features/code-component/) for full details of the props `<Code>` supports.
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -195,3 +195,31 @@ The code above generates the following on the page:
...
+
+The code above generates the following on the page:
+
+import { Code } from '@astrojs/starlight/components';
+
+export const exampleCode = `console.log('This could come from a file or CMS!');`;
+export const fileName = 'example.js';
+export const highlights = ["file", "CMS"];
+
+<Code code={exampleCode} lang="js" title={fileName} mark={highlights} />
```
delucis, line 225: Since it came up several times recently, what do you think of adding something like this to explicitly document the import version of usage?

````suggestion
<Code code={exampleCode} lang="js" title={fileName} mark={highlights} />

#### Code transclusion

Use [Vite’s `?raw` import suffix](https://vitejs.dev/guide/assets#importing-asset-as-string) to import any code file as a string.
You can then pass this imported string to the `<Code>` component to include it on your page.

```mdx title="src/content/docs/example.mdx" "?raw"
import { Code } from '@astrojs/starlight/components';
import importedCode from '/src/env.d.ts?raw';

<Code code={importedCode} lang="ts" title="src/env.d.ts" />
```

import importedCode from '/src/env.d.ts?raw';

<Code code={importedCode} lang="ts" title="src/env.d.ts" />
````

**Review by delucis — COMMENTED** [review]
Super exciting to see you crack this challenging feature — I know so many of our users will be delighted to be able to use this. Thank you @hippotastic! Left a few comments, but I think it’s all small things and docs for the most part. The main feature seems to be working great! Amazing work 🙌

File: package.json [hunk 6]
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "build:examples": "pnpm --filter '@example/*' build",
+    "build:examples": "pnpm --no-bail --workspace-concurrency 1 --filter '@example/*' build",
```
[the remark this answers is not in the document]
hippotastic: Oh, sorry about not mentioning this! The change was intentional - I noticed that the default behavior of concurrently running these build processes led to "Port already in use" errors on my machine, so I changed this to build the examples sequentially, which fixed this error. The `--no-bail` flag ensures that you still see all errors and it doesn't skip building other projects after the first error.

packages/starlight/integrations/shared/pathToLocale.ts
hippotastic: Sure, will do! It's a Windows-specific issue where the input path does not have a leading slash, and therefore your "remove the matching beginning" logic doesn't work otherwise.

docs/src/content/docs/guides/components.mdx
hippotastic: Love it! Great idea to add this example. Btw, this is the first time I've read the word "transclusion", I'm just trusting you that this is the correct term. :)

package.json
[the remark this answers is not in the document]
hippotastic: Updated the PR description to include this.

packages/starlight/integrations/shared/pathToLocale.ts
hippotastic: Done! :)

package.json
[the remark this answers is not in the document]
delucis: Oh, super weird! I wonder why the build process needs to access a port at all 🤔 Thanks for the explanation though.

docs/src/content/docs/guides/components.mdx
delucis: Haha, I honestly only learned it after working on Starlight and seeing it used by technical writers and some other tools. It’s a bit specialised (so I might double check with Sarah if she thinks there’s a better way to say it), but basically a technical term meaning including other content by reference in a page ([Wikipedia](https://en.wikipedia.org/wiki/Transclusion)).

delucis (approved):
OK, added a changeset and will request one last docs check from @sarah11918, but otherwise looks good to go out tomorrow! Thanks @hippotastic 🚀

docs/src/content/docs/guides/components.mdx
```diff
@@ -195,3 +195,49 @@ The code above generates the following on the page:
 A list of all available icons is shown below with their associated names. Click an icon to copy the component code for it.
 
 <IconsList />
+
+### Code
+
+Starlight uses Expressive Code to render [Markdown and MDX code blocks](/guides/authoring-content/#code-blocks). Use the `<Code>` component to render syntax highlighted code when using a Markdown code block is not possible, for example, to render data coming from external sources like files, databases, or APIs.
```
sarah11918, line 201:
```suggestion
Use the `<Code>` component to render syntax highlighted code when using a [Markdown code block](/guides/authoring-content/#code-blocks) is not possible, for example, to render data coming from external sources like files, databases, or APIs.
```
I'm having a hard time putting my finger on it, but it seemed weird to state Expressive Code here without linking to it (but then there was immediately an important link), and the fact it was at the very beginning of the sentence/introduction seemed like it was "the main point."

I think this might get to the point, and still retain all the useful information?

docs/src/content/docs/guides/components.mdx
```diff
@@ -195,3 +195,49 @@ The code above generates the following on the page:
...
+
+import { Code } from '@astrojs/starlight/components';
+
+export const exampleCode = `console.log('This could come from a file or CMS!');`;
+export const fileName = 'example.js';
+export const highlights = ["file", "CMS"];
+
+<Code code={exampleCode} lang="js" title={fileName} mark={highlights} />
+
+#### Code transclusion
```
sarah11918, line 227:
```suggestion
#### Imported Code
```
I don't hate "transclusion"! (though, I also didn't know what it was!) ... but wonder whether something more funtional here like "Imported Code" might help, since that's in fact how it's talked about in the explanation.

If you're going to say, "you wouldn't *always* do this with code you import" then I would reply that the wording below can be shifted to talk about the process of importing code, and then the part below becomes an "and if you want it imported as a string..." description.

delucis: Yup I’m happy to use a simpler and more common word here too. I’d seen a few people ask us about this and use “transclusion”, but VitePress’s docs also say [“Import Code Snippets”](https://vitepress.dev/guide/markdown#import-code-snippets) for their equivalent feature if we need more convincing.

PR #1408, opened 2024-01-24
[docs] community content: article description copy edit

olets:
#### Description

You included my Starlight vs VuePress comparison in the community content page. Thanks!

There was a typo in the description.

The first commit fixes the typo.

The second two are a naive pass at correcting the existing localizations. I do not speak French or Korean, so used Google Translate. Google Translate translates the "before" English into the "before" French and Korean, making me think it's getting it right — or perhaps that the "before" translation was also done with Google Translate.

(There is also a Chinese localization. Running the "after" English through Google Translate results in no change in the translation.)

//

- See also #1396, which adds a Spanish translation based on the mistaken English copy


delucis (approved):
Thanks for the update (and the article!) @olets.

I’ve made this a “typo fix” change only to the English version as we usually avoid updating multiple languages at once so we defer to our translators for that.

PR #1414, opened 2024-01-27
i18n(fr): Update components, configuration, sidebar and customization files

thomasbnt:
#### Description

Update some files in French language.

PR #1418, opened 2024-01-27
docs: fix `disable404Route` default type

HiDeoo:

#### Description

This PR fixes the default type of the new `disable404Route` in the docs to be a boolean (`false`) instead of a string (`'false'`) which I think was a typo that I noticed while reviewing a translation.

This PR also apply the changes to the `ko/reference/configuration` page which has already been translated. When merged, I'll apply a suggestion to https://github.com/withastro/starlight/pull/1414 too (and any other translation PRs that may appear until then).


delucis (approved):
Thanks @HiDeoo! I totally missed this in my review 🙌

PR #1420, opened 2024-01-27
fix : font not rendered properly on firefox #1419

### LINKED ISSUE #1419: font not rendered correctly on firefox

0.17.0

v20.10.0

pnpm

Windows

Firefox

### Describe the Bug

The [https://starlight.astro.build](https://starlight.astro.build/) and [https://docs.astro.build/](https://docs.astro.build/en/getting-started/) websites have a problem showing the correct font on firefox
astro starlight
astro docs

https://stackblitz.com/edit/github-brdc4n

### Participation

abdelhalimjean:
#### Description

- Closes #1419 
this PR fixes an error using the css variable for font-family which caused it not to render properly on firefox
Before:
astro starlight
After:
image

### DIFF
--- a/packages/starlight/style/props.css
+++ b/packages/starlight/style/props.css
@@ -87,8 +87,8 @@
 		'Segoe UI Symbol', 'Noto Color Emoji';
 	--sl-font-system-mono: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono',
 		'Courier New', monospace;
-	--__sl-font: var(--sl-font, ''), var(--sl-font-system);
-	--__sl-font-mono: var(--sl-font-mono, ''), var(--sl-font-system-mono);
+	--__sl-font: var(--sl-font-system);
+	--__sl-font-mono: var(--sl-font-system-mono);
 
 	/** Key layout values */
 	--sl-nav-height: 3.5rem;


packages/starlight/style/props.css
```diff
@@ -87,8 +87,8 @@
 		'Segoe UI Symbol', 'Noto Color Emoji';
 	--sl-font-system-mono: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono',
 		'Courier New', monospace;
-	--__sl-font: var(--sl-font, ''), var(--sl-font-system);
-	--__sl-font-mono: var(--sl-font-mono, ''), var(--sl-font-system-mono);
+	--__sl-font: var(--sl-font-system);
+	--__sl-font-mono: var(--sl-font-system-mono);
```
delucis, line 91: This change would break the intended behaviour here: that a user can set `--sl-font` and/or `--sl-font-mono` to apply a custom font, while still preserving the fallback system font stack Starlight defines.

In the existing code, this worked (on most devices 😅), because when those two variables weren’t set, you’d end up with font stacks something like this (if I expand the variables):
```css
--__sl-font: '', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
--__sl-font-mono: '', ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
```

The `''` value is ignored usually, then as soon as `--sl-font` and `--sl-font-mono` are set — let’s say to imaginary `Astro Sans` and `Astro Mono` fonts — this switches to:
```css
--__sl-font: 'Astro Sans', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
--__sl-font-mono: 'Astro Mono', ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
```

I was thinking how to keep the current behaviour while avoiding the risk of hitting an unintended font, which is apparently possible with the `''` value in your case, and I think we can do this instead:

```suggestion
	--__sl-font: var(--sl-font, var(--sl-font-system)), var(--sl-font-system);
	--__sl-font-mono: var(--sl-font-mono, var(--sl-font-system-mono)), var(--sl-font-system-mono);
```

This might look a bit odd, but in the fallback case, where no user-defined fonts are used, we basically double up the system font stack. The second value is effectively pointless until a user sets a custom font, and then the second value makes sure the system fallbacks are still in place.

delucis (changes requested):
Thanks for the bug report and PR @abdelhalimjean!

It’s a weird bug — I _think_ what is happening is that you happen to have a font installed on your computer that somehow has a font name which is an empty string and therefore matches the `""` variable fallback value. I honestly didn’t know that was even possible, hence why I used the empty string here basically as a null value in the case `--sl-font` wasn’t set.

Still, rare as it might be, I think we can still fix it! Suggesting an alternative approach as your proposed change would actually break some intended behaviour.

packages/starlight/style/props.css
abdelhalimjean: @delucis thanks for taking the time to look at this PR, It's definitely and odd behavior because it works with chrome and edge without an issue, someone else had the same issue here : [https://github.com/withastro/astro.build/issues/943](Weird font rendering on Firefox ) 

at first i didn't understand the use of `--sl-font` because on my branch that variable didn't exist and that's why i ignored it, but now with your explanation i understand why, it's user specific font preference, so we need to keep it as the first option and fall back on the `--sl-font-system` (same with mono)

i put in place your suggested fix in a new commit

delucis (approved):
Thanks @abdelhalimjean! I hadn’t realised this is probably the same bug as https://github.com/withastro/astro.build/issues/943

Yes, I think this change seems a good one to make. I’ll add a changeset and get this released in our next patch. Appreciate you taking the time to make a PR!

PR #1429, opened 2024-01-29
i18n(zh-cn): Update components.mdx

liruifengv:
#### Description
- Update components.mdx
- https://github.com/withastro/starlight/pull/1395

### DIFF
--- a/docs/src/content/docs/zh-cn/guides/components.mdx
+++ b/docs/src/content/docs/zh-cn/guides/components.mdx
@@ -195,7 +195,7 @@ import { Icon } from '@astrojs/starlight/components';
 
 <IconsList />
 
-### Code
+### 代码块
 
 当使用 [Markdown 代码块](/zh-cn/guides/authoring-content/#代码块) 行不通时，可以使用 `<Code>` 组件来渲染语法高亮的代码。例如，渲染来自文件、数据库或 API 等外部来源的数据。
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/zh-cn/guides/components.mdx
```diff
@@ -194,3 +194,49 @@ import { Icon } from '@astrojs/starlight/components';
 下面显示了所有可用图标的列表及其关联的名称。点击图标以复制其组件代码。
 
 <IconsList />
+
+### Code
```
HiDeoo, line 198: It seems that all the previous component headings have been translated, I guess it would make sense to translate this one too, for consistency. Any thoughts?

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). I left one small question regarding a heading but ignoring that, this looks good to me :thumbsup:

docs/src/content/docs/zh-cn/guides/components.mdx
liruifengv: en.. I'll think of a suitable translation.

HiDeoo (approved):
Thanks for the quick update 👍

PR #1437, opened 2024-01-29
Add Starlight types to `defineEcConfig`, export `StarlightExpressiveCodeOptions`

hippotastic:
#### Description

- Adds Starlight-specific types to the `defineEcConfig` function and exports `StarlightExpressiveCodeOptions`. Both can be imported from `@astrojs/starlight/expressive-code`.

This provides Starlight types and IntelliSense support for your Expressive Code configuration options inside an `ec.config.mjs` file. See the [Expressive Code documentation](https://expressive-code.com/key-features/code-component/#using-an-ecconfigmjs-file) for more information.

### DIFF
--- a/.changeset/calm-roses-joke.md
+++ b/.changeset/calm-roses-joke.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/starlight': patch
+---
+
+Adds Starlight-specific types to `defineEcConfig` function and exports `StarlightExpressiveCodeOptions`.
+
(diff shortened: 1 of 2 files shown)


delucis (commented):
This looks good to me @hippotastic!

One question though: shouldn’t this be documented somewhere? Or is the intention people start with the `astro.config` version and then only use this based on terminal logging?

hippotastic:
It is documented on the EC website, including examples that show usage of the function. A majority of users will never need this as most parts of the config are serializable. Only advanced users who know plugins or advanced configurations through the EC website will run into the scenario where the Code component tells them that they need to move their EC config to the file.

The EC website contains examples which use the function, and I included the full Code component error message to ensure it pops up when using the search function: https://expressive-code.com/key-features/code-component/#using-an-ecconfigmjs-file

delucis (approved):
Ok sounds good! Thanks @hippotastic 🙌

PR #1438, opened 2024-01-30
Update Vitest to v1

delucis:
#### Description

- Updates Starlight’s test dependencies to use the latest version of Vitest, updating from a pre-release to v1.

- A couple of test snapshots changed to a nicer string escaping algorithm — I inspected them visually and I don’t think anything important changed, mostly just stuff like:

   ```diff
   - "<el attr=\\"val\\">"
   + `<el attr="val">`
   ```
- Updated Vitest coverage thresholds config to match new format as changed in https://github.com/vitest-dev/vitest/pull/4442

### DIFF
--- a/packages/starlight/__tests__/remark-rehype/code-rtl-support.test.ts
+++ b/packages/starlight/__tests__/remark-rehype/code-rtl-support.test.ts
@@ -10,7 +10,7 @@ test('applies `dir="auto"` to inline code', async () => {
 	expect(output).not.toEqual(input);
 	expect(output).includes('dir="auto"');
 	expect(output).toMatchInlineSnapshot(
-		'"<p>Some text with <code dir=\\"auto\\">inline code</code>.</p>"'
+		`"<p>Some text with <code dir="auto">inline code</code>.</p>"`
 	);
 });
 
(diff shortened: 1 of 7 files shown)


delucis:
> Locally (and CI too actually), I get a few `[ERROR] [vite] WebSocket server error: Port is already in use` errors but I don't think they're relevant to us.

Yes, I noticed these too. I think it’s something to do with things happening in parallel? @hippotastic reported the same errors in the parallel builds of the example projects and added a `--workspace-concurrency 1` option to that package script as part of #1395. I can try that for the tests too, to see if it helps.

HiDeoo:
> Yes, I noticed these too. I think it’s something to do with things happening in parallel? @hippotastic reported the same errors in the parallel builds of the example projects and added a `--workspace-concurrency 1` option to that package script as part of #1395. I can try that for the tests too, to see if it helps.

Interesting, to add more details, I got them when running only the `starlight` package tests from its directory so not sure this would help (as I think it's a pnpm workspace flag).

delucis:
Ah, good point — I think I saw that too. We do have multiple Vitest workspaces — perhaps would need _those_ to be run sequentially.

HiDeoo:
I wonder if it could be https://github.com/vitejs/vite/issues/14328 and we should just ignore in this specific context of testing.

delucis:
> I agree, and it looks like Vitest is supposed to [have this workaround](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/vite.ts?rgh-link-date=2024-01-30T16%3A03%3A19Z#L9) but it's not working for some reasons.

Oh yeah… Well, overriding a global like that is always a bit flaky if some other code somewhere also does something to it, so I guess not totally shocking it’s not working. Still, think I’m happy to merge this for now.

PR #1440, opened 2024-01-30
Add JS support to `@astrojs/starlight/expressive-code` export

hippotastic:
#### Description

- Converts Starlight's `@astrojs/starlight/expressive-code` export from a TS-only export to an `.mjs`+`.d.ts` pair, using the exact configuration and file structure also used by Astro's `astro/config` export.
  - Note: To keep this now `.mjs`-based export working in projects using Astro's `strictest` TypeScript configuration that contains `allowJs: false`, it was required to move the exporting files into the package root folder using names that are identical to the export's name.
- This change allows using the recently introduced `defineEcConfig` helper function and all other re-exports from `astro-expressive-code` (which are important for advanced customization scenarios) to also work in `ec.config.mjs` outside of monorepos, where Astro's Vite instance does not process integration imports and therefore no TS support is available.
- Moves the exported `getStarlightEcConfigPreprocessor` function to `@astrojs/starlight/internal`, because it uses TS and therefore cannot remain in the previous `@astrojs/starlight/expressive-code` export which now no longer depends on TS.


delucis (approved):
LGTM! Thanks for wrangling the complexity of these different environments 🫡

PR #1454, opened 2024-02-03
Make Starlight compatible with server output mode

Fryuni:
#### Description

Make projects using Starlight and `output: 'server'` build successfully.

All Starlight pages will still be pre-rendered as though they were emitted for a static build.
The built-in search, powered by Pagefind, will index any custom route that is also pre-rendered, SSR routes won't be indexed.

This PR is the simplified version of #1255, making Starlight compatible with SSR projects, but not allowing any SSR on Starlight pages themselves.


Fryuni:
It took me a while to open this because I was testing locally with Starlight docs itself. It doesn't work there but not because of this change, it is due to https://github.com/withastro/astro/issues/9674, which breaks the page when using the Node adapter and `trailingSlash: 'always'`.

If anyone wants to try this locally, make sure to comment out the `trailingSlash` option until that bug is fixed :)

packages/starlight/index.ts
```diff
@@ -43,11 +43,15 @@ export default function StarlightIntegration({
 					injectRoute({
 						pattern: '404',
 						entrypoint: '@astrojs/starlight/404.astro',
+						// Ensure page is pre-rendered even when project is on server output mode
+						prerender: true,
```
delucis, line 47: Mainly out of curiosity, why do we need both this AND the `export const prerender`? Seems like an API smell in Astro to me, so if these are both required I’d love to understand why.

packages/starlight/index.ts
```diff
@@ -78,6 +82,13 @@ export default function StarlightIntegration({
 					scopedStyleStrategy: 'where',
 					// If not already configured, default to prefetching all links on hover.
 					prefetch: config.prefetch ?? { prefetchAll: true },
+					experimental: {
+						// For projects in server mode enable global route priority to allow custom
+						// routes to customize any path.
+						...(config.output === 'server' && {
+							globalRoutePriority: true,
+						}),
+					}
```
delucis, line 91: Could you say a bit more about what does and doesn’t work with and without this?

.changeset/cool-days-look.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/starlight': patch
```
delucis, line 2: I think this counts as a minor! It’s essentially adding a feature, and if we’re changing some of our set-up code, I guess theoretically could even break the odd person in a rare case.

```suggestion
'@astrojs/starlight': minor
```

.changeset/cool-days-look.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Make Starlight compaticle with server output mode.
+
+Starlight pages are always prerendered, even when `output: 'server'`.
```
delucis, line 7: Suggesting a couple of small tweaks to match our changelog style (third-person “makes” over imperative “make”) and the Astro docs terminology of “on-demand server rendering”:

```suggestion
Makes Starlight compatible with [on-demand server rendering](https://docs.astro.build/en/guides/server-side-rendering/) (sometimes referred to as server-side rendering or SSR).

Starlight pages are always prerendered, even when using `output: 'server'`.
```

docs/src/content/docs/manual-setup.mdx
```diff
@@ -125,4 +125,7 @@ In the future, we plan to support this use case better to avoid the need for the
 
 ### Use Starlight with SSR
 
-Currently, Starlight does not support [SSR deployment](https://docs.astro.build/en/guides/server-side-rendering/) using Astro’s server adapters. We hope to be able to support this soon.
+Currently, Starlight does not support [Server-Side rendering](https://docs.astro.build/en/guides/server-side-rendering/) for its pages, but you can use Starlight alongside SSR pages in your project. Starlight pages will always be prerendered regardless of your project's output mode.
+
+We hope to be able to support SSR for Starlight pages soon.
```
delucis, line 130: In the knowledge that people often stop reading quickly, let’s flip this to not say “does not support” right away 😁

```suggestion
You can use Starlight alongside custom on-demand rendered pages in your project by following the [“On-demand Rendering Adapters”](https://docs.astro.build/en/guides/server-side-rendering/) guide in Astro’s docs.

Currently, documentation pages generated by Starlight are always prerendered regardless of your project's output mode. We hope to be able to support on-demand rendering for Starlight pages soon.
```

delucis (commented):
Looks great @Fryuni! Thanks also for the thorough testing, notes, and backlinks to related issues. I think that issue should obviously be fixed, but doesn’t have to block us in the meantime.

I’ve left a couple of queries — mostly requests for enlightenment — and a suggested a couple docs tweaks. Thanks again for tackling this!

packages/starlight/index.ts
Fryuni: Since everything will be pre-rendered basically everything works as intended for most use cases.

The only difference is in the case of a pre-rendered custom route generating the same path as a docs page.

Ideally, that should be an error or at least a warning (tracked on https://github.com/withastro/astro/issues/9832), but it currently is silently ignored I believe it is better to have the custom page overriding the docs page instead of the opposite.

packages/starlight/index.ts
Fryuni: It is something that I noticed while working on the routing logic. Different parts of Astro core have different behavior for prerendered and on-demand routes, but those different parts don't all have access to both those methods of defining whether something is prerendered.

For example, the conflict warning about on-demand routes with the same dynamic segments runs before the file is read, so it only knows about this prerender parameter here. Another part, that generates another warning about not matching a static path in the dev server, only knows about the prerender constant from the file, not the value passed here.

So to have a consistent behavior everywhere this needs to be passed in both places.

I haven't found anything but warning that check only one of those, but I haven't checked too deeply either.

Fryuni: This _could_ also relate to endpoints not being prerendered if just this flag is passed in (which I mentioned here: https://github.com/withastro/astro/issues/7721#issuecomment-1847763893). But that is mostly speculation on my part, I haven't actually tried to debug that at all.

packages/starlight/index.ts
delucis: Thanks for the clarification! I’m thinking about it. I agree with the basic premise that “it is better to have the [user’s] custom page overriding the docs page instead of the opposite”.

The only thing I’m wondering about is if setting this experimental mode invisibly could mess with a user’s wider project in other unexpected ways, which is something we need to consider given this applies globally to all their routes. Have you seen this option cause any issues for folks so far?

Maybe a good compromise is to set this _only_ if the user themselves didn’t already set it:

```suggestion
					experimental: {
						// For projects in server mode enable global route priority to allow custom
						// routes to customize any path.
						...(config.output === 'server' && {
							globalRoutePriority: config.experimental.globalRoutePriority ?? true,
						}),
					}
```

That way we can default to the new experimental behaviour, but still let people opt out if they need.

delucis: OK, so important question from above is:

> Have you seen this option cause any issues for folks so far?

(or are there obvious potential issues you can imagine?)

That can inform whether we set this for folks or let them opt in to it themselves if they need it.

Fryuni: We can leave this out for now... Docs page will win in case of conflicts.

It is a requirement for the other PR that makes Starlight pages SSR since then it would override any custom route.

I hope cases of conflicts will have at least a warning in the future anyway, so I don't really mind which one wins here that much.

Fryuni: Let users opt-in if they need/want it.

Fryuni: Just so I don't leave the question open

> Have you seen this option cause any issues for folks so far?

No, I haven't. In fact I haven't seen anyone even mentioning that they are using it, so the only use case I know of is mine and it is working fine for me.

delucis: Ok, here’s a radical proposal, how about we straight up enable this for all projects?

```suggestion
					experimental: {
						globalRoutePriority: true,
					},
```

This will give us some more data about if it causes issues for people and prepare for when it becomes the default priority algorithm in a future Astro major.

Couple of notes:
- We’ll also need to bump the minimum supported Astro version to 4.2 in this case in the peer dependencies of `packages/starlight`
- We should probably add a couple of extra changesets specifically for these two things for visibility (new minimum version support and enabling the `globalRoutePriority` flag)

Fryuni: > Also basically makes the recently added disable404Route option obsolete I think? 😁

I don't think it does, both would be static routes. Neither would be guaranteed to win, and I'm not sure which _would_ win. It would also generate a warning about the conflict.

---

But I agree about enabling it for everyone.

delucis: Oh right, we still have two `404.astro` routes in that case. We still would need a way for Starlight to mark it's default one as lower priority.

Fryuni: Changed to enable it always and added a changeset for it

delucis (approved):
Thanks for taking care of all that @Fryuni! Should be ready to go. Suggesting one small tweak to the changeset.

Hoping to release this tomorrow — weekend SSR fans get ready 🙌

PR #1458, opened 2024-02-05
Silence i18n content collection warnings

delucis:
Closes https://github.com/withastro/starlight/discussions/1174

Currently, there are two warnings logged related to the i18n content collection when running our basic starter kit (i.e. with no i18n files set up):

1. A warning that an i18n collection is configured but doesn’t exist on disk (on dev/build start up):

   > The **i18n** collection is defined but no **content/i18n** folder exists in the content directory. Create a new folder for the collection, or check your content configuration file for typos.

2. A runtime error when Starlight actually calls `getCollection('i18n')` to again say that the collection directory doesn’t exist:

   > The collection **i18n** does not exist or is empty. Ensure a collection directory with this name exists.

Many of us Starlight users have trained ourselves to ignore these warnings, but they remain confusing especially for newer users who assume they need to fix something.

This PR addresses both of these points:

1. Removes the `i18n` collection set-up from our two starter templates, which stops Astro emitting that first warning. Means people have to add that manually if they need the custom i18n features, but we do already document that in our [“Translate Starlight’s UI”](https://starlight.astro.build/guides/i18n/#translate-starlights-ui) docs.

2. Hackily overrides `console.warn()` briefly while we make our `getCollection('i18n')` call, so that we swallow the runtime error. Admittedly, this is a bit gross, but I considered the options and I feel like the warning does have quite a bit of value for your average Astro users, so removing it upstream seemed kind of unfair.

   @Princesseuh also suggested adding a “noWarn” option to the query APIs to support this officially. I think that could be a nice idea, but I’m not 100% sure the use case is widespread enough to justify that yet.


packages/starlight/utils/translations.ts
```diff
@@ -3,16 +3,23 @@ import config from 'virtual:starlight/user-config';
...
-
-		(await getCollection('i18n')).map(({ id, data }) => [id, data] as const)
-	);
-} catch {}
+/** Get all translation data from the i18n collection, keyed by `id`, which matches locale. */
+async function loadTranslations() {
+	let userTranslations: Record<string, i18nSchemaOutput> = {};
+	// Briefly override `console.warn()` to silence logging when a project has no i18n collection.
+	const warn = console.warn;
+	console.warn = () => {};
```
natemoo-re, line 11: 😞 This obviously works, but I'm wondering if we can make the warnings for `getCollection` lazy in core? Ideally you'd only get the warning if you try to _access_ a collection that does not exist.

delucis: That’s exactly what’s happening here! We call `getCollection('i18n')` _in case_ it exists (there’s no `hasCollection('i18n')` of `getCollections()` API we could use to check for existence first).

delucis:
Thanks for the review Nate 💖

PR #1459, opened 2024-02-05
Fix table of contents highlighting glitch

### LINKED ISSUE #1448: Table of Contents Glitch
### 📝 Issue Description

When selecting the "Content Collect Errors" section header, the page slides down to the section. However the highlight on the Table of Contents is not updated. Whichever section was previously highlighted, remains highlighted. This behaviour occurs when the browser window is maximized or in full screen mode and page is at normal zoom level. 

image

### 📋 On which page(s) it occurs

https://docs.astro.build/en/reference/error-reference/#content-collection-errors

### 🤔 Expected Behavior

When the browser window is maximized or in full screen mode and page is at normal zoom level, when the "Content Collect Errors" section header is selected, the page scrolls down to the section and the section header is highlighted in the Table of Contents.

### 👀 Current Behavior

Currently, while the page slides down to the section, the section header on the table of contents is not highlighted. This is probably because there isn't enough room at the bottom of the page for the heading to come into focus.

### 🖥️ Browser

Arc Version 1.26.0, Chrome Version 120.0.6099.234

### 📄 Additional Information

This problem is relevant for large window sizes and screen resolutions.

delucis:
#### Description

- Fixes #1448
- See https://github.com/withastro/starlight/issues/1448#issuecomment-1927969598 for full details. TL;DR — larger heading top margins increased the likelihood of the table of contents intersection observer falling _between_ elements.
- By increasing the height of the zone we watch for intersections, we can avoid this (unless someone adds content with more extreme top margins, at some point this could still break).

### DIFF
--- a/.changeset/sharp-cows-buy.md
+++ b/.changeset/sharp-cows-buy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a bug where table of contents highlighting could break given very specific combinations of content and viewport size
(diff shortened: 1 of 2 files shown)


TheOtterlord (approved):
Thanks, this looks great! Tested on a few resolutions and pages, and everything worked as expected!

SatanshuMishra:
This is great, Nice job @delucis!

PR #1475, opened 2024-02-08
Add CodingCat.dev stream video to community content page

delucis:
#### Description

- Updates our community content section with yesterday’s live stream

PR #1479, opened 2024-02-09
i18n(fr): update `resources/community-content`

HiDeoo:

#### Description

This PR updates the French version of the `resources/community-content` page with the changes of #1475.

### DIFF
--- a/docs/src/content/docs/fr/resources/community-content.mdx
+++ b/docs/src/content/docs/fr/resources/community-content.mdx
@@ -106,6 +106,12 @@ import YouTubeGrid from '~/components/youtube-grid.astro';
 <YouTubeGrid
 	videos={[
 		{
+			href: 'https://www.youtube.com/watch?v=sF6UcV3moZg',
+			title: 'Construire une documentation avec Starlight et Astro',
+			description:
+				'Regardez Chris et Alex plonger au cœur de Starlight dans Coder avec CodingCat.dev',
+		},
+		{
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/resources/community-content.mdx
```diff
@@ -105,6 +105,12 @@ import YouTubeGrid from '~/components/youtube-grid.astro';
 
 <YouTubeGrid
 	videos={[
+		{
+			href: 'https://www.youtube.com/watch?v=sF6UcV3moZg',
+			title: 'Construire une documentation avec Starlight et Astro',
+			description:
+				'Regardez Chris et Alex plonger au cœur de Starlight dans Coder avec CodingCat.dev',
```
HiDeoo, line 112: This is not technically a 1:1 translation of the English version which would be something like "Regardez Chris et Alex plonger dans Starlight dans Coder avec CodingCat.dev" but I did not like the "dans" repetition which does not exist in the English version with "into"/"on" so I slightly tweaked it to avoid the repetition while keeping the same meaning.

Happy to change it if we prefer the 1:1 translation.

delucis: Definitely counts as an accurate translation to me! Almost always better to choose an idiomatic translation like this rather than aiming for word-for-word equivalence  🙌

delucis (approved):

PR #1485, opened 2024-02-10
feat: Add html attributes to hero action links

timokoessler:
#### Description

This pull request changes the CallToAction component to add support for setting custom html attributes of hero action links.

Relates to @delucis request on #901


docs/src/content/docs/reference/frontmatter.md
```diff
@@ -166,6 +169,7 @@ interface HeroConfig {
     link: string;
     variant: 'primary' | 'secondary' | 'minimal';
     icon: string;
+    attrs?: HTMLAttributes<"a">;
```
delucis, line 172: Let’s use an explicit type here as `HTMLAttributes` is an Astro type which we don’t document anywhere:
```suggestion
    attrs?: Record<string, string | number | boolean>;
```

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -119,6 +119,9 @@ hero:
     - text: View on GitHub
       link: https://github.com/astronaut/my-project
       icon: external
+      attrs:
+        target: _blank
+        rel: noopener
```
delucis, line 124: I’m a little bit torn about showing this as in general opening in a new window for this link in this example is not really considered a best practice (see https://www.nngroup.com/articles/new-browser-windows-and-tabs/ for example for some UX research), although I know plenty of people will still want to do exactly this.

Something like this could make sense:

```suggestion
    - text: View on GitHub
      link: https://github.com/astronaut/my-project
      icon: external
      attrs:
        rel: me
```

The other link attribute that could make sense sometimes would be `download`, but we’re missing a good icon for that one.

packages/starlight/components/CallToAction.astro
```diff
@@ -1,17 +1,19 @@
 ---
+import type { HTMLAttributes } from 'astro/types';
 import Icon from '../user-components/Icon.astro';
 import type { Icons } from './Icons';
 
 interface Props {
 	variant: 'primary' | 'secondary' | 'minimal';
 	link: string;
 	icon?: undefined | { type: 'icon'; name: keyof typeof Icons } | { type: 'raw'; html: string };
+	attrs?: HTMLAttributes<"a">;
```
delucis, line 10: This doesn’t really impact frontmatter validation, but to be a bit more accurate, we should omit `href` from the type here as it is already set elsewhere:

```suggestion
	attrs?: Omit<HTMLAttributes<'a'>, 'href'>;
```

(I was also unsure about `class` given we‘re using `class:list` on the `<a>` element, but just tested and Astro does merge `class` and `class:list`, so should be fine.)

delucis (commented):
Thanks for the quick response @timokoessler — I Left a couple of small notes, but basically looks great! 🚀 

One more thing, usually we only update the English docs and let our translators take care of updating their languages. In this case, it does seem to be safe to do, except for Indonesian and German (`id` and `de`). Those two pages are currently outdated and updating them in this PR would cause our [translation tracker](https://i18n.starlight.astro.build/) to treat them as “updated” even though this PR doesn’t actually do that.

packages/starlight/components/CallToAction.astro
```diff
@@ -1,17 +1,19 @@
...
 }
 
-const { link, variant, icon } = Astro.props;
+const { link, variant, icon, attrs } = Astro.props;
```
delucis, line 13: Ah, follow-up sorry. I just tested again and actually the `class` and `class:list` merging only seems to work when both are declared directly (i.e. `<a class:list={[]} class="">`), not when `class` is in a spread object.

So we’ll need to special case that one:
```suggestion
const { link, variant, icon } = Astro.props;
const { class: customClass, ...attrs } = Astro.props.attrs;
```

packages/starlight/components/CallToAction.astro
```diff
@@ -1,17 +1,19 @@
...
 ---
 
-<a class:list={['sl-flex action', variant]} href={link}>
+<a class:list={['sl-flex action', variant]} href={link} {...attrs}>
```
delucis, line 16:
```suggestion
<a class:list={['sl-flex action', variant, customClass]} href={link} {...attrs}>
```

docs/src/content/docs/reference/frontmatter.md
timokoessler: If I undo the changes to the docs it will not help with the translation tracker problem because there are commits modifying this files, or am I wrong?

delucis (approved):
Looks good to me! Thanks for taking care of those updates @timokoessler.

I can take care of removing the German and Indonesian docs pages from this PR.

Will aim to bundle this in the next minor with #1175, so will merge when that is ready.

delucis:
Thanks again for contributing this! Will be released shortly in v0.19 🙌

PR #1489, opened 2024-02-12
Fix rendering issue for text with unhandled generic Markdown directives

HiDeoo:

- Changes to Starlight code

#### Description

This PR is a follow-up to this Discord [discussion](https://discord.com/channels/830184174198718474/1206539738824314912).

The `remark-directive` plugin used to handle [asides](https://starlight.astro.build/guides/authoring-content/#asides) based on [generic directives](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) is responsible of parsing 3 different types of directives:

- text directives
- leaf directives
- container directives

Starlight only uses the last type but all types are parsed by the plugin. This means that content that may accidentally contain a directive of another type will be parsed as such and may lead to unexpected results. For example:

> AWS re:Invent is a learning conference hosted by AWS.

The above sentence will be rendered as:

SCR-20240212-ocsm

This PR fixes the issue by transforming the unhandled directive so that they render verbatim.

SCR-20240212-ocvn

Note: this relies on `mdast-util-to-markdown` so that we can pass down to it the official `directiveToMarkdown` extension and automatically get the correct rendering for the directive including any attributes.

### DIFF
--- a/packages/starlight/__tests__/remark-rehype/asides.test.ts
+++ b/packages/starlight/__tests__/remark-rehype/asides.test.ts
@@ -143,3 +143,22 @@ test('runs without locales config', async () => {
 	const res = await processor.render(':::note\nTest\n::');
 	expect(res.code.includes('aria-label=Note"'));
 });
+
+test('tranforms back unhandled text directives', async () => {
+	const res = await processor.render(
+		`This is a:test of a sentence with a text:name[content]{key=val} directive.`
+	);
+	expect(res.code).toMatchInlineSnapshot(`
+		"<p>This is a:test
+		 of a sentence with a text:name[content]{key="val"}
+		 directive.</p>"
+	`);
+});
+
+test('tranforms back unhandled leaf directives', async () => {
+	const res = await processor.render(`::video[Title]{v=xxxxxxxxxxx}`);
+	expect(res.code).toMatchInlineSnapshot(`
+		"<p>::video[Title]{v="xxxxxxxxxxx"}
+		</p>"
+	`);
+});
--- a/.changeset/many-knives-type.md
+++ b/.changeset/many-knives-type.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a potential text rendering issue with text containing colons.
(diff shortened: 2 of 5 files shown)


packages/starlight/__tests__/remark-rehype/asides.test.ts
```diff
@@ -143,3 +143,22 @@ test('runs without locales config', async () => {
...
+
+test('tranforms back unhandled text directives', async () => {
+	const res = await processor.render(
+		`This is a:test of a sentence with a text:name[content]{key=val} directive.`
+	);
+	expect(res.code).toMatchInlineSnapshot(`
+		"<p>This is a:test
+		 of a sentence with a text:name[content]{key="val"}
+		 directive.</p>"
+	`);
```
delucis, line 155: Noting a subtle difference here in the transformation back to text where the attribute value becomes wrapped in `"`. It should be extremely rare that this is an issue though, as it only applies in a relatively complex string that is a correctly formed directive, so I guess that’s fine?

```suggestion
		`This is a:test of a sentence with a text:name[content]{key=val} directive.`
	);
	expect(res.code).toMatchInlineSnapshot(`
		"<p>This is a:test
		 of a sentence with a text:name[content]{key="val"}
		 directive.</p>"
	`);
```

delucis (approved):
Thanks for the quick diagnosis and fix @HiDeoo!

Left a note, but this looks good to me 🚀

packages/starlight/__tests__/remark-rehype/asides.test.ts
HiDeoo: Excellent catch, I totally missed that.

After a bit of digging, I found that `mdast-util-directive` respects [`options.quote`](https://github.com/syntax-tree/mdast-util-to-markdown#optionsquote) altho in our case, this would not be enough to fix the issue.

I think the current approach is a good enough solution for now but if we ever hit the rare case you mentioned where we cannot restore the node to its exact original content including the quotes, we would need to revisit this. I think the 2 options would be to either:

- Only restore the node if it's a directive with no attributes and document/emit a warning when encountering a case where we cannot restore (this seems to be the Docusaurus approach)
- Find another way to properly restore the node (which I think would mean to have a copy of the AST before the directive was transformed)

Are you still okay with this approach?

- If no, I can work a refactor based on the solution we would choose
- If yes, I can merge this PR as it contains no doc changes in its current state

delucis: Yes, 100% approve — we can merge and release this as is.

If we run into this rare issue, we could also look into an alternative to `remark-directive` that _only_ provides the container directive and not also text/lead directives.

PR #1499, opened 2024-02-16
Add `<Aside>` component

delucis:
#### Description

- Adds an `Aside` Astro component to our `@astrojs/starlight/components` module.
- Produces the same markup as the existing `:::note`-style asides available in Markdown and MDX.

#### Notes

- I had initially thought it would be ideal to share the HTML-building logic between this component, but when I looked at it, it felt kind of silly: the component is so simple that pulling the full hastscript pipeline feels wrong. It does mean some duplication — changes would require updating in two places — but given how minimal this is, I think this is OK.

- This PR copy/pastes the `urlToSlug()` helper added in #1175 so it can use it — will resolve the conflicts once #1175 is merged.


docs/src/content/docs/guides/components.mdx
```diff
@@ -169,6 +169,59 @@ import { LinkCard } from '@astrojs/starlight/components';
...
+	A warning aside *with* a custom title.
+</Aside>
+
+<Aside type="tip">
+    Other content is also supported in asides.
+
+    ```js
+    // A code snippet, for example.
+    ```
+
```
HiDeoo, line 195: I am guessing the extra line is auto added by Prettier?

HiDeoo (approved):
> I had initially thought it would be ideal to share the HTML-building logic between this component, but when I looked at it, it felt kind of silly: the component is so simple that pulling the full hastscript pipeline feels wrong. It does mean some duplication — changes would require updating in two places — but given how minimal this is, I think this is OK.

I guess one could say that we could extract the variants and `isAsideVariant` from the remark plugin but considering we are already duplicating the SVGs and that we are not likely to have to update the variants often, I think it's fine to keep it as is and that it was a good call to not pull everything :thumbsup:

Really simple and nice component with a great error message for invalid types :star2:

docs/src/content/docs/guides/components.mdx
delucis: Yes, unfortunately :-(

Not sure if there are better Prettier rules we could configure? The MDX ones are honestly kind of odd.

delucis: Played around in the Prettier playground a bit, but seems just to be how it is? Most of the time when things go wrong it even collapses the whole component children onto a single line 😬

delucis: Another valid formatting option could be:

````mdx
<Aside type="tip">

Other content is also supported in asides.

```js
// A code snippet, for example.
```

</Aside>
````

This is at least balanced even if it takes some more space.

kevinzunigacuellar (approved):
Looks good to me! 💜

PR #1500, opened 2024-02-16
fix: "Adding a new Language" link points to new i18n guides 

rgilsimoes:
#### Fixed broken "Adding a new Language" link.

- The "Adding a new Language" link on the Contributing.md doc was pointing to a broken link because guides where moved to new location.


delucis (approved):
Thanks for spotting and fixing @rgilsimoes! 🚀

PR #1509, opened 2024-02-17
i18n(ja): Add Japanese translation for "Resources"

morinokami:
#### Description

- Add Japanese translation for "Resources" in the sidebar.


HiDeoo (approved):
Thanks for the update :star2:

PR #1529, opened 2024-02-18
Add Astro-GhostCMS to showcase

Adammatthiesen:
#### Description

Add Astro-GhostCMS.xyz to starlight showcase

This website was recently converted from a Astro/Starlight site to a full Starlight site!


delucis (approved):
Nice! Thanks @Adammatthiesen 🙌 

Looked like the little ghost dude was a bit stretched in the original screenshot, so took the liberty of making a new one — they’re looking nice and round now 👻

PR #1530, opened 2024-02-18
Add `<Badge>` component

kevinzunigacuellar:
#### Description

- Adds an `<Badge>` component to @astrojs/starlight/components module.

### Notes

- Move outline style to be applied with css under `current-aria="page"`

### DIFF
--- a/packages/starlight/components/SidebarSublist.astro
+++ b/packages/starlight/components/SidebarSublist.astro
@@ -1,7 +1,7 @@
 ---
 import { flattenSidebar, type SidebarEntry } from '../utils/navigation';
 import Icon from '../user-components/Icon.astro';
-import Badge from './Badge.astro';
+import Badge from '../user-components/Badge.astro';
 
 interface Props {
 	sublist: SidebarEntry[];
@@ -28,7 +28,7 @@ const { sublist, nested } = Astro.props;
 								{' '}
 								<Badge
 									text={entry.badge.text}
-									variant={entry.isCurrent ? 'outline' : entry.badge.variant}
+									variant={entry.badge.variant}
 								/>
 							</>
 						)}
--- a/packages/starlight/schemas/badge.ts
+++ b/packages/starlight/schemas/badge.ts
@@ -1,8 +1,10 @@
 import { z } from 'astro/zod';
 
+export const badgeVariants = ['note', 'danger', 'success', 'caution', 'tip', 'default'] as const;
+
 const badgeSchema = () =>
 	z.object({
-		variant: z.enum(['note', 'danger', 'success', 'caution', 'tip', 'default']).default('default'),
+		variant: z.enum(badgeVariants).default('default'),
 		text: z.string(),
 	});
 
--- a/packages/starlight/user-components/Badge.astro
+++ b/packages/starlight/user-components/Badge.astro
@@ -1,11 +1,20 @@
 ---
-import type { Badge } from '../schemas/badge';
+import { AstroError } from 'astro/errors';
+import { type Badge, badgeVariants } from '../schemas/badge';
 
 interface Props {
-	variant?: Badge['variant'] | 'outline';
+	variant?: Badge['variant']
 	text?: string;
 }
+
 const { variant = 'default', text } = Astro.props;
+
+if (!badgeVariants.includes(variant)) {
+	throw new AstroError('Invalid `variant` prop passed to the `<Badge>` component.\n',
+		`Received: ${JSON.stringify(variant)}\n` +
+		`Expected: ${badgeVariants.slice(0,-1).map((i) => JSON.stringify(i)).join(', ')} or "${badgeVariants.at(-1)}".`
+	);
+}
 ---
 
 <span class:list={['sl-badge', variant]} set:html={text} />
@@ -16,21 +25,19 @@ const { variant = 'default', text } = Astro.props;
 		border: 1px solid var(--sl-color-border-badge);
 		border-radius: 0.25rem;
 		font-family: var(--sl-font-system-mono);
-		font-size: var(--sl-text-xs);
-		font-weight: 400;
 		padding: 0.125rem 0.375rem;
-		line-height: 1;
+		line-height: normal;
 		color: #fff;
 		background-color: var(--sl-color-bg-badge);
 		overflow-wrap: anywhere;
 	}
 
-	.outline {
-		--sl-color-bg-badge: transparent;
-		--sl-color-border-badge: currentColor;
-		color: inherit;
+	:global(#starlight__sidebar) .sl-badge {
+		line-height: 1;
+		font-size: var(--sl-text-xs);
 	}
 
+	/* Color variants */
 	.default {
 		--sl-color-bg-badge: var(--sl-color-accent-low);
 		--sl-color-border-badge: var(--sl-color-accent);
@@ -83,5 +90,11 @@ const { variant = 'default', text } = Astro.props;
 
 	:global([data-theme='light']) .tip {
 		--sl-color-bg-badge: var(--sl-color-purple-high);
+	}
+
+	:global(a[aria-current="page"]) > .sl-badge {
+		--sl-color-bg-badge: transparent;
+		--sl-color-border-badge: currentColor;
+		color: inherit;
 	}
 </style>
--- a/docs/src/content/docs/reference/test.mdx
+++ b/docs/src/content/docs/reference/test.mdx
@@ -1,0 +1,69 @@
+---
+title: Testing badges
+description: Learn how to start building your next documentation site with Starlight by Astro.
+sidebar:
+  badge: Test
+---
+
+import { Tabs, TabItem, Badge } from '@astrojs/starlight/components';
+
+## APIs
+
+Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.
+
+### <Badge text="POST" variant="success" /> `/api/users`
+
+Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.
+
+#### <Badge text="PUT" variant="tip" /> `/api/users`
+
+Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.
+
+##### <Badge text="DELETE" variant="danger" /> `/api/users`
+
+Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.
+
+### Lists
+
+- *Italic <Badge text="caution" variant="caution" />* lorem ipsum
+- **Bold <Badge text="tip" variant="tip" />** lorem ipsum
+- ***Bold Italic <Badge text="note" variant="note" />*** lorem ipsum
+
+### Asides
+
+:::danger
+Using this is <Badge text="deprecated" variant="caution" /> and should be avoided.
+:::
+
+:::note
+This is a <Badge text="danger" variant="danger" /> message.
+:::
+
+### Tabs
+
+<Tabs>
+  <TabItem label="Tab 1">
+    This API is <Badge text="deprecated" variant="caution" /> and should be avoided.
+
+    ```js
+    const foo = 'bar';
+    ```
+
+  </TabItem>
+  <TabItem label="Tab 2">
+    This API is <Badge text="deprecated" variant="danger" /> and should be avoided.
+
+    ```js
+    const foo = 'bar';
+    ```
+  </TabItem>
+</Tabs>
+
+### Tables
+
+| Method | Endpoint | Description |
+| ------ | -------- | ----------- |
+| <Badge text="GET" variant="success" /> | `/api/users` | Get all users |
+| <Badge text="POST" variant="tip" /> | `/api/users` | Create a new user |
+| <Badge text="PUT" variant="note" /> | `/api/users` | Update a user |
+| <Badge text="DELETE" variant="danger" /> | `/api/users` | Delete a user |
--- a/packages/starlight/components.ts
+++ b/packages/starlight/components.ts
@@ -1,4 +1,5 @@
 export { default as Aside } from './user-components/Aside.astro';
+export { default as Badge } from './user-components/Badge.astro';
 export { default as Card } from './user-components/Card.astro';
 export { default as CardGrid } from './user-components/CardGrid.astro';
 export { default as Icon } from './user-components/Icon.astro';


HiDeoo:
Kevin and badges, as a huge badge consumer, I'm always happy to see your PRs and this one is definitely gonna remove a lot of duplicated code for me :raised_hands:

I did not yet closely look at the code, but some general (and personal) thoughts:

- Overall, I find the badges/font a bit too big outside of the sidebar. I usually find the Vitepress version [too small](https://vitepress.dev/reference/default-theme-badge) but I think the current version is the opposite and wonder if we could find a middle ground. Note that I'm usually not the best person to ask for design feedback, so take this with a grain of salt.
- Would it make sense to also provide the `outline` variant to users?
- I wonder if allowing a `class` prop would be nice to allow users to customize the badge even further. To explain, some of the use cases I encountered where a custom class would help:
  - Defining more custom colors, for example in the requests/OpenAPI land, there are more request types than variants we have, so being able to define a custom color would be nice.
  - Having badges themed to a brand color for a company or project.

> Is an update in the package.json required?

If I understand correctly, the `packages/starlight/components.ts` update should be enough.

kevinzunigacuellar:
Thank you for the review @HiDeoo 💜

> Overall, I find the badges/font a bit too big outside of the sidebar. I usually find the Vitepress version [too small](https://vitepress.dev/reference/default-theme-badge) but I think the current version is the opposite and wonder if we could find a middle ground. Note that I'm usually not the best person to ask for design feedback, so take this with a grain of salt.

You are right. They do look a bit bigger, I will follow the same style as the inline code block which uses a smaller font size.

> Would it make sense to also provide the outline variant to users?

I have mixed feelings about exposing `outline`. it is not an official variant since it's only applied when the sidebar is active. But then I don't see why it couldn't be an official one.

I am not sure what the best way to go about this. I could use some guidance here.

> I wonder if allowing a class prop would be nice to allow users to customize the badge even further. To explain, some of the use cases I encountered where a custom class would help:
Defining more custom colors, for example in the requests/OpenAPI land, there are more request types than variants we have, so being able to define a custom color would be nice.
Having badges themed to a brand color for a company or project.

It would be nice to have more theme versions. A class would be a very nice addition to include more customizations. 

In terms of customizing current variants. I think the nicest DX would be to override a global variable similar to what [VitePress does](https://vitepress.dev/reference/default-theme-badge#customize-type-color). 

Then `class` could override variables locally resulting in a new badge theme.

After adding `class` prop it should the user should be able to create a new theme like this

```css
.custom-badge {
  --sl-color-bg-badge: cyan;
  --sl-color-border-badge: purple;
  --sl-color-text-badge: red;
}
```

And use it like

```jsx
<Badge text="custom" class="custom-badge" />
```

HiDeoo (commented):
Sorry it took me so long to get back to this PR after my first review and also for the new updates you have been working on since then. Super appreciate your patience and hard work on this. 🙌 

I left a few new comments and suggestions on the code, nothing major. I'd also like to recap everything else remaining to be done so we can push this PR to the finish line.

- ~~Confirm that the `class` prop is a good idea (I'm personally in favor as I suggested it and have use-cases for it, but I'd like to hear more opinions)~~ I talked a bit with Chris who suggested that we could actually support all valid HTML attributes for `<span>` which I think is a great idea.
- ~~Check if the new light mode styles are something we want to include (may need to check Starlight design documents)~~ Regarding the new light mode styles, after talking with Chris, a follow-up PR would be better to discuss and implement them.
- Update the branch so that we get the new updated `docs/src/content/docs/guides/components.mdx` file
- Move the new section in `docs/src/content/docs/guides/components.mdx` just above the `Icon` section
- Remove the `docs/src/content/docs/reference/test.mdx` file

Does this sound correct to you? Do you see anything missing?

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,143 @@
...
+if (!badgeVariants.includes(variant)) {
+	throw new AstroError(
+		'Invalid `variant` prop passed to the `<Badge>` component.\n',
+		`Received: ${JSON.stringify(variant)}\n` +
+			`Expected one of ${badgeVariants.map((i) => JSON.stringify(i)).join(', ')}`
+	);
+}
+---
+
+<span class:list={['sl-badge', variant, className]}><slot /></span>
```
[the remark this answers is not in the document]
kevinzunigacuellar: I agree, after taking a look after some time I think it's better to have a text prop. My previous idea was to pass in a slot in case users would like to pass svgs or other elements but I would like to keep it simple and keep it just with text.

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,143 @@
...
+
+const { variant = 'default', class: className } = Astro.props;
+
+if (!badgeVariants.includes(variant)) {
+	throw new AstroError(
+		'Invalid `variant` prop passed to the `<Badge>` component.\n',
+		`Received: ${JSON.stringify(variant)}\n` +
+			`Expected one of ${badgeVariants.map((i) => JSON.stringify(i)).join(', ')}`
+	);
+}
```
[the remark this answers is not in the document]
kevinzunigacuellar: Yeah! go `props` 🥳

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	.sl-badge {
+		display: inline-block;
+		border: 1px solid var(--sl-color-border-badge);
+		border-radius: 0.25rem;
+		font-family: var(--sl-font-system-mono);
+		padding: 0.125rem 0.375rem;
+		line-height: normal;
+		color: var(--sl-color-text-badge, #fff);
```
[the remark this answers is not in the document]
kevinzunigacuellar: I do not remember 😅 but happy to remove it

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,143 @@
...
+		background-color: var(--sl-color-bg-badge);
+		overflow-wrap: anywhere;
+	}
+
+	/* Sidebar overrides */
+	:global(#starlight__sidebar) .sl-badge {
```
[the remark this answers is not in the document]
kevinzunigacuellar: Yeah, it seems that `#starlight__sidebar` is mostly use for media queries happy to change it.

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	/* Markdown content overrides */
+	:global(.sl-markdown-content :not(h1, h2, h3, h4, h5, h6)) .sl-badge,
+	:global(.sl-markdown-content) > .sl-badge {
+		font-size: var(--sl-text-sm);
+	}
+
+	/* outline variant */
+	:global(a[aria-current='page']) > .sl-badge {
```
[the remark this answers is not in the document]
kevinzunigacuellar: yeah, very good observation!

kevinzunigacuellar:
Thanks for the review @HiDeoo. I added the changes you suggested, I don't think it's missing anything else besides from doing some more testing for CSS edge cases. I will do some local testing to see what pops up.

HiDeoo:
Additional note that this PR will close #1764 by removing the `outline` CSS class which collides with Tailwind CSS [`outline`](https://tailwindcss.com/docs/outline-style) utility adding `outline-style: solid;` and leading to a thicker outline for sidebar badges for the current page:

SCR-20240418-kfwo

kevinzunigacuellar:
I think it would be a bit complicated to handle the vertical alignment for user because it depends in two things: 
1. the size of the badge
2. the size of the parent font. 

I think baseline is a good default, users can always pass in a custom class to tweak their designs if they need to.

Thanks again for the review @HiDeoo 💜

HiDeoo:
Are you referring to my first point, the second point, or both?

kevinzunigacuellar:
Second point, I don't think we should support position.

HiDeoo:
> Second point, I don't think we should support position.

Ah perfect, thanks for the clarification. Then yeah, I think we totally agree and should only try to improve the first one :thumbsup:

HiDeoo:
Had a bit of time to play with the heading issue people mentioned during Talking & Doc'ing. It looks like it's mostly something very visible with larger headings.

I wonder if just adding a rule close to something like this could help a bit:

```css
/* Badge in headings */
:global(.sl-markdown-content :is(h1, h2, h3, h4, h5, h6)) .sl-badge {
  vertical-align: middle;
}
```

Here is a preview before and after the change. Do you have any thoughts on this? I think this may be a bit better and less something that catches the eye.

headings

kevinzunigacuellar:
It looks good 💜, middle in only heading might work because of the bigger font sizes.

packages/starlight/schemas/badge.ts
```diff
@@ -4,8 +4,13 @@ const badgeSchema = () =>
 	z.object({
 		variant: z.enum(['note', 'danger', 'success', 'caution', 'tip', 'default']).default('default'),
 		text: z.string(),
+		class: z.string().optional(),
```
HiDeoo, line 7: If we keep this prop here instead of `BadgeComponentSchema`, we would need to document it [here](https://starlight.astro.build/reference/configuration/#badgeconfig).

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,139 @@
+---
+import { type Badge, BadgeComponentSchema } from '../schemas/badge';
+import { parseWithFriendlyErrors } from '../utils/error-map';
+
+type Props = Badge;
```
HiDeoo, line 5: At the moment, this is equivalent to `z.output<ReturnType<typeof badgeSchema>>` but I guess we would want the props to be `z.output<typeof BadgeComponentSchema>` instead so we may need to create a type for this new component schema.

HiDeoo (commented):
Amazing. I just pushed this heading changes so we can have a preview including it.

I left a few comments regarding your latest changes, very minor, but this is looking really good. I would guess the documentation may still need some work but feature-wise and UI-wise I am personally happy with it.

packages/starlight/schemas/badge.ts
kevinzunigacuellar: Yes, it will update the corresponding docs.

HiDeoo:
Tried to work a bit on the documentation as I was personally not super happy with the "custom badges" section as we usually have examples which includes a demo but coming up with a concise example where the demo would be looking good in both light and dark mode is a bit tricky imo.

To do that, I rewrote some parts and sharing it here at the moment as it's only a rough draft and want to get opinions first.

````mdx
### Badges

import { Badge } from '@astrojs/starlight/components';

Use the `<Badge>` component to display small pieces of information, such as status or labels.

Pass the content you want to display to the `text` attribute of the `<Badge>` component.
Optionally, you can specify a `variant` attribute with values like `tip`, `note`, `caution`, `danger`, `success` or `default` (the default using the theme accent color) to adjust the badge's appearance.

To adjust the badge's size, you can specify a `size` attribute with values like `small` (the default), `medium`, or `large`.
You can also include other `<span>` attributes such as `class`.

```mdx title="src/content/docs/example.mdx"
import { Badge } from '@astrojs/starlight/components';

<Badge text="New" variant="tip" size="small" />
<Badge text="Deprecated" variant="caution" size="medium" />
<Badge text="Starlight" variant="note" size="large" />
```

<Badge text="New" variant="tip" size="small" />
<Badge text="Deprecated" variant="caution" size="medium" />
<Badge text="Starlight" variant="note" size="large" />
````

Altho, this made me realize some points:

- We forgot [Chris suggestion](https://github.com/withastro/starlight/pull/1530#pullrequestreview-1988899462) to support all valid HTML attributes for `<span>`.
  - For consistency, we should probably remove the `class` attribute from the schema and use the same approach as [sidebar link custom HTML attributes](http://localhost:4321/guides/sidebar/#custom-html-attributes) and use an `attrs` property.
- We can probably stop using `set:html`.

What do you think?

**Comment by HiDeoo** [comment]
Amazing, love to hear that.

> I completely forgot about that set:html, definitely changing that for a set:text.

Or just `<span class:list={['sl-badge', variant, size, className]}>{text}</span>`. Considering this is equivalent and way more common, it could potentially be more readable for others.

> I wasn't very happy with the custom badges section, too. By including attributes I think it would be easier to explain that the user may override this component however they see fit.

Yeah, that's what I tried to do in my rewritten version with the "You can also include other `<span>` attributes such as `class`." sentence. It's a pattern we have used for [Link Cards](https://starlight.astro.build/guides/components/#link-cards). Of course, it could be improved as I'm not the best at documentation but I think we are on the right track.

**Comment by HiDeoo** [comment]
Thanks for the new update :tada:

I think it's looking good, and how you were able to reduce the documentation shows that imo. One thing I noticed:

- I mentioned `attrs` as a parallel with sidebar link custom HTML attributes, so that would only apply to the sidebar (not the component) but as we decided to not support that, I think we could drop entirely the new `attrs`. This would mean that in the sidebar a user can use `class` (this is currently the case) and on the component and a user could directly use `<Badge text="New" style="…" />`.
- As we would not support all attributes for sidebar badges (only `class`), I think it would be fine to just not include attributes in the Zod schema at all, pass w/e extra attributes to the `<span>` without validating them when the user is using the component directly and have the component `Props` typed with something close to `z.input<typeof badgeSideBarSchema> & HTMLAttributes<'span'>`.
  - We basically would only validate `variant`, `text` and `class` and that's it. This would allow for quite a few simplifications in the schemas I assume.

Any thoughts?

**Comment by kevinzunigacuellar** [comment]
I see what you meant. I have applied the requested changes. Thanks for the through review @HiDeoo

File: packages/starlight/schemas/badge.ts [hunk 8]
```diff
@@ -1,10 +1,14 @@
...
-	});
+const badgeSchema = () => z.object({
+	variant: z.enum(['note', 'danger', 'success', 'caution', 'tip', 'default']).default('default'),
+	text: z.string(),
+	class: z.string().optional(),
+});
+
+export const BadgeComponentSchema = badgeSchema().extend({
+	size: z.enum(['small', 'medium', 'large']).default('small'),
+});
```
HiDeoo, line 11:
```suggestion
const badgeSchema = () =>
	z.object({
		variant: z.enum(['note', 'danger', 'success', 'caution', 'tip', 'default']).default('default'),
		text: z.string(),
		class: z.string().optional(),
	});

export const BadgeComponentSchema = badgeSchema()
	.extend({
		size: z.enum(['small', 'medium', 'large']).default('small'),
	})
	.passthrough();

export type BadgeComponentProps = z.input<typeof BadgeComponentSchema>;
```

We can use [`.passthrough`](https://github.com/colinhacks/zod#passthrough) so that unknown keys are not stripped.

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,153 @@
+---
+import { BadgeComponentSchema } from '../schemas/badge';
```
HiDeoo, line 2:
```suggestion
import { BadgeComponentSchema, type BadgeComponentProps } from '../schemas/badge';
```

packages/starlight/user-components/Badge.astro
```diff
@@ -0,0 +1,153 @@
...
+const { text, variant, size, class: customClass } = parseWithFriendlyErrors(
+	BadgeComponentSchema,
+	{
+		class: classProp,
+		variant: variantProp,
+		text: textProp,
+		size: sizeProp,
+	},
+	'Invalid prop passed to the `<Badge/>` component.'
+);
```
HiDeoo, line 30:
```suggestion
type Props = BadgeComponentProps & HTMLAttributes<'span'>;

const {
	text,
	variant,
	size,
	class: customClass,
	...attrs
} = parseWithFriendlyErrors(
	BadgeComponentSchema,
	Astro.props,
	'Invalid prop passed to the `<Badge/>` component.'
);
```

The previous change allows for a bit of simplification in the component.

HiDeoo (commented):
Amazing. I left 2 small suggestions to simplify a little bit the code and avoid some duplication, but personally I'm very happy with the new component :tada:

You did an amazing work 👏

packages/starlight/user-components/Badge.astro
kevinzunigacuellar: Thanks Hideo 💜

docs/src/content/docs/guides/components.mdx
```diff
@@ -368,6 +368,35 @@ import { Steps } from '@astrojs/starlight/components';
...
+import { Badge } from '@astrojs/starlight/components';
+
+Use the `<Badge>` component to display small pieces of information, such as status or labels.
+
+Pass the content you want to display to the `text` attribute of the `<Badge>` component.
+Optionally, you can specify a `variant` attribute with values like `tip`, `note`, `caution`, `danger`, `success` or `default` (the default uses the theme accent color) to adjust the badge's appearance.
+
+To adjust the badge's size, you can specify a `size` attribute with values like `small` (the default), `medium`, or `large`.
+
+You can also include other `<span>` attributes such as `class` or `style` to further customize the badge.
```
sarah11918, line 382:
```suggestion
Pass the content you want to display to the `text` attribute of the `<Badge>` component.

By default, the badge will use the theme accent color of your site. To use a built-in badge color, set the `variant` attribute to one of the following values: `note` (blue), `tip` (purple), `danger` (red), `caution` (orange), or `success` (green).

The `size` attribute (default: `small`) controls the size of the badge text. `medium` and `large` are also available options for displaying a larger badge.

For further customization, use other `<span>` attributes such as `class` or `style` with custom CSS.
```
I'm thinking something like this? Adjust to taste, or ignore!

sarah11918 (commented):
Docs look great, @kevinzunigacuellar ! Just left some polishing thoughts below!

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -294,12 +294,14 @@ The configuration above generates the following sidebar:
...
 
-### Badge variants
+### Badge variants and custom styling
 
-Customize the badge styling using an object with `text` and `variant` properties.
+Customize the badge styling using an object with `text`, `variant` and `class` properties.
 
 The `text` represents the content to display (e.g. "New").
-Override the `default` styling, which uses the accent color of your site, by setting the `variant` property to one of the following values: `note`, `tip`, `danger`, `caution` or `success`.
+Change the appearance, by setting the `variant` property to one of the following values: `note`, `tip`, `danger`, `caution` or `success`.
```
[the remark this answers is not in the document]
kevinzunigacuellar: Variant is optional and defaults to `default`. Also sizing is not available for sidebar items and groups.

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -340,7 +340,7 @@ sidebar:
 
 Add a badge to the page in the sidebar when displayed in an autogenerated group of links.
 When using a string, the badge will be displayed with a default accent color.
-Optionally, pass a [`BadgeConfig` object](/reference/configuration/#badgeconfig) with `text` and `variant` fields to customize the badge.
+Optionally, pass a [`BadgeConfig` object](/reference/configuration/#badgeconfig) with `text`, `variant` and `class` fields to customize the badge.
```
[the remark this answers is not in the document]
kevinzunigacuellar: Can't change the size in the sidebar.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -203,6 +203,7 @@ type SidebarItem = {
 interface BadgeConfig {
 	text: string;
 	variant: 'note' | 'tip' | 'caution' | 'danger' | 'success' | 'default';
```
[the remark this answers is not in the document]
kevinzunigacuellar: Yes, variant is optional and size is not available.

kevinzunigacuellar, line 205:
```suggestion
	variant?: 'note' | 'tip' | 'caution' | 'danger' | 'success' | 'default';
```

.changeset/wicked-melons-study.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/starlight': minor
+---
+
+Adds the `<Badge>` component
```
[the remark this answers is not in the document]
kevinzunigacuellar: yes, it is the same component but now is generally available.

docs/src/content/docs/reference/frontmatter.md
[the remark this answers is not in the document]
delucis: Cool — guess we still should add the Oxford comma (if we’re being picky):

```suggestion
Optionally, pass a [`BadgeConfig` object](/reference/configuration/#badgeconfig) with `text`, `variant`, and `class` fields to customize the badge.
```

.changeset/wicked-melons-study.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/starlight': minor
+---
+
+Expands the `<Badge>` component to be generally available
```
delucis, line 5: I know we think of this as “expanding”, but I actually think we can describe this as “adding” a component — the previous component doesn’t really exist for users currently, they only enable it via config options.

```suggestion
Adds a new `<Badge>` component
```

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -294,12 +294,14 @@ The configuration above generates the following sidebar:
 	]}
 />
 
-### Badge variants
+### Badge variants and custom styling
 
-Customize the badge styling using an object with `text` and `variant` properties.
+Customize the badge styling using an object with `text`, `variant` and `class` properties.
```
delucis, line 299: Ooh, spotted another Oxford comma opportunity 👀 
```suggestion
Customize the badge styling using an object with `text`, `variant`, and `class` properties.
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -24,13 +24,11 @@ const { sublist, nested } = Astro.props;
...
-								<Badge
-									text={entry.badge.text}
-									variant={entry.isCurrent ? 'outline' : entry.badge.variant}
-								/>
-							</>
+							<Badge
+								variant={entry.badge.variant}
+								class={entry.badge.class}
+								text={entry.badge.text}
+							/>
```
delucis, line 31: It’s worth double checking that removing the fragment and explicit `{' '}` whitespace here doesn’t cause issues — it was added as a workaround for issues caused by https://github.com/withastro/compiler/issues/852

packages/starlight/components/SidebarSublist.astro
```diff
@@ -41,10 +39,11 @@ const { sublist, nested } = Astro.props;
 							<div class="group-label">
 								<span class="large">{entry.label}</span>
 								{entry.badge && (
-									<>
-										{' '}
```
delucis, line 45: Same here

delucis (commented):
Thanks again for all the work advocating for this component @kevinzunigacuellar! Can’t wait to release it. Left a couple of final small notes.

packages/starlight/components/SidebarSublist.astro
delucis: I don’t see the whitespace stripped in the preview, but maybe I’m missing something?

kevinzunigacuellar: Tested a local build in chrome and safari, it seems to be working fine.

delucis (approved):
I hereby approve this PR! ✅ 

Will release it later today — thanks again for your work @kevinzunigacuellar and to @HiDeoo for shepherding it through 💖

PR #1534, opened 2024-02-19
Improve DX for `sidebar` prop in `<StarlightPage>` and document it

delucis:
#### Description

This PR adds a parsing step to process the value of `sidebar` passed to `<StarlightPage>`. This allows us to improve DX by not requiring lots of boilerplate setting default values, for example:

##### Before
```astro
<StarlightPage
	sidebar={[
		{
			type: 'link',
			label: 'Custom link 1',
			href: '/test/1',
			isCurrent: false,
			badge: undefined,
			attrs: {},
		}
	]}
/>
```
##### After
```astro
<StarlightPage
	sidebar={[
		{ 
			label: 'Custom link 1',
			href: '/test/1',
		}
	]}
/>
```

Setting these extra props is still valid, but not required. `type` has been marked as `@deprecated` as it can be inferred so there is no real need to set it manually although you still can.

While working on this and the documentation, I realised we leaked a difference in naming between the internal `SidebarEntry` shape and the public sidebar config used in `astro.config.mjs`. In user config, link groups have an `items` array, whereas internally we named this `entries`. It’s probably a good idea to align these so that we have some consistency.

In this PR I’ve added support for setting `items` in link groups so we can document that. Setting `entries` instead is still also supported although it is marked as `@deprecated` to steer people away from it. In a future version we can start warning about use of `entries`, and later on remove support entirely. For now, I was aiming for full backwards compatibility so this can go out as a patch.

**Aside:** there’s a fair amount of fiddly Zod stuff to get this working — should get cleaner once we remove `entries` support.


docs/src/content/docs/guides/pages.mdx
```diff
@@ -106,13 +106,35 @@ The following properties differ from Markdown frontmatter:
 - The [`editUrl`](/reference/frontmatter/#editurl) option requires a URL to display an edit link.
 - The [`sidebar`](/reference/frontmatter/#sidebar) property is not supported. In Markdown frontmatter, this option allows customization of [autogenerated link groups](/reference/configuration/#sidebar), which is not applicable to pages using the `<StarlightPage />` component.
```
sarah11918, line 107:
```suggestion
- The [`sidebar`](/reference/frontmatter/#sidebar) frontmatter property for customizing how the page appears in [autogenerated link groups](/reference/configuration/#sidebar) is not available. Pages using the `<StarlightPage />` component are not part of a collection and cannot be autogenerated in the sidebar.
```
OK, this might or might not be better, *but* I found it a little weird to read "the sidebar property is not supported" and then immediately see (admittedly, a different property) called "sidebar". 😄 

At the very least, that's why I threw "the sidebar FRONTMATTER property..." in there. Then, tried to distinguish *that* sidebar ("how the page is described in the sidebar") from the property below ("which sidebar to generate for the page")

Just take that how you will!

docs/src/content/docs/guides/pages.mdx
```diff
@@ -106,13 +106,35 @@ The following properties differ from Markdown frontmatter:
...
+##### `sidebar`
+
+**type:** `SidebarEntry[]`  
+**default:** the sidebar generated based on the [global `sidebar` config](/reference/configuration/#sidebar)
+
+Provide a custom site navigation sidebar for this page.
+If not set, the page will use the default global sidebar.
+
+For example, the following page overrides the default sidebar with a link to the homepage and a group of links to different constellations.
+The current page in the sidebar is set using the `isCurrent` property and an optional `badge` has been added to a link item.
```
sarah11918, line 118: The example is really explicit! Nicely done!

I am left wondering whether I can still do things like autogenerate here, though. If so, it might be nice to show another grouping where that happens? If not, a line mentioning that this config should manually list out the entire contents of the sidebar as autogeneration of groups is not available, or something like that, would be helpful!

sarah11918 (commented):
Just a couple of quick comments from me!

HiDeoo (approved):
Finally got the time to play more than a few minutes with the changes. In the end, pretty much of all the logic is in the Zod schema and I have to say, it works exactly like I expected after testing a few scenarios. :thumbsup:

I'm really happy with the result, way nicer API for users.

Unrelated note: I'd like at some point to see if we can slightly improve our Zod error map, specially around unions, it's not that great right now, but definitely in another PR.

docs/src/content/docs/guides/pages.mdx
delucis: Yeah, I can see that! Took the liberty of tweaking the very end of your suggestion so I could use it and give you credit 🙌

delucis:
> I'm really happy with the result, way nicer API for users.

Yay!

> Unrelated note: I'd like at some point to see if we can slightly improve our Zod error map, specially around unions, it's not that great right now, but definitely in another PR.

I actually have an old work-in-progress branch that made some improvements here and this work also reminded me of that. Will try to dust it off and share in case there’s something useful there.

PR #1528, opened 2024-02-18
docs: Add more plugins & tools

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content

#### Description

Following the release of Starlight [`v0.19.0`](https://github.com/withastro/starlight/releases/tag/%40astrojs%2Fstarlight%400.19.0), `starlight-blog` and `starlight-openapi` have now been refactored to Starlight plugins. This PR moves them to the appropriate section in the docs site.

The PR also adds `starlight-obsidian` to the plugin list and `starlight-i18n` to the tools list.

PR #1536, opened 2024-02-20
Add new starlight-ghostcms to Community plugins list

Adammatthiesen:
#### Description

Add new starlight-ghostcms to Community plugins list.

starlight-ghostcms

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -53,6 +53,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		title="starlight-obsidian"
 		description="Publish Obsidian vaults in your Starlight site."
 	/>
+  	<LinkCard
+		href="https://astro-ghostcms.xyz/intro/starlight/install/"
+		title="starlight-ghostcms"
+		description="Add your GhostCMS blog posts alongside your Starlight Docs"
+  	/>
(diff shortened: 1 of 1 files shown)


delucis (commented):
Yay! Thanks @Adammatthiesen — exciting to see this released.

I think this should actually go in the “Community plugins” section instead of the “tools and integrations” section given this is actually a Starlight plugin by the looks of things?

Adammatthiesen:
> Yay! Thanks @Adammatthiesen — exciting to see this released.
> 
> I think this should actually go in the “Community plugins” section instead of the “tools and integrations” section given this is actually a Starlight plugin by the looks of things?

I put it in the same category as `starlight-blog` since it has almost the same functions just using different data as well as I had to make some tweaks to handle some things ghost does differently.

delucis:
> I put it in the same category as `starlight-blog` since it has almost the same functions just using different data as well as I had to make some tweaks to handle some things ghost does differently.

Ah yeah, I think that will move up too now! Until the v0.19 release it was technically not a plugin.

delucis:
Ah hang on a sec. We forgot to merge https://github.com/withastro/starlight/pull/1528

I’ll do that now, you can update branch and then move it!

Adammatthiesen:
Updated.  lol when i first saw you message i was so confused.  glad it got cleared up!

PR #1542, opened 2024-02-21
Improve Zod errors

delucis:
#### Description

- Improves our Zod error map:

	- **Better handling of unions.**
		Previously they logged only “Did not match union” and no further information.
		An example of the new format:
		
		> **0**: Did not match union.
		> > Expected type `{ href: string } | { entries: array }`
		> > Received `{ "label": "Custom link 1", "href": 5 }`
		
	- **Better formatting of expected values.**
		Stringified objects/arrays now have some whitespace around items, e.g. `{ "label": "Text" }` instead of `{"label":"Text"}`. They are also now wrapped in code backticks so they render as code in Astro’s error screens.

	- **Better handling of deeply nested issues.** (Thanks @HiDeoo!)
	
		Union errors that were found deep in a schema previously only reported the top-level path, e.g. `sidebar.0`. We are now able to show the full path, e.g. `sidebar.0.items.1`.

- Refactors all `safeParse()` calls on Zod schemas to use a new `parseWithFriendlyErrors()` utility which bundles up our logic for throwing an error with the error map.

- When throwing errors, uses `AstroError` for nicer formatting:

	Example of Astro’s error screen displaying a nicely formatted error message

	Example of terminal displaying a nicely formatted error message

- Testing error messages is improved with a Vitest [custom snapshot serializer](https://vitest.dev/guide/snapshot.html#custom-serializer), which captures both the `message` and the `hint` fields on `AstroError`. I updated Vitest to 1.3 to be able to configure this for all tests in one place (this feature was added in https://github.com/vitest-dev/vitest/pull/5092).


packages/starlight/utils/error-map.ts
```diff
@@ -38,27 +54,41 @@ export const errorMap: z.ZodErrorMap = (baseError, ctx) => {
...
+							.replace(baseErrorPath, '')
+							.replace(/^\./, '');
+						if ('expected' in issue) {
+							return relativePath ? `${relativePath}: ${issue.expected}` : issue.expected;
+						}
+						return relativePath;
+					})
+					.join('; ');
+				return `{ ${props} }`;
+			});
```
HiDeoo, line 83:
```suggestion
			const expectedShapes: string[] = [];
			for (const unionError of baseError.unionErrors) {
				const expectedShape: string[] = [];
				for (const issue of unionError.issues) {
					// If the issue is a nested union error, show the associated error message instead of the
					// base error message.
					if (issue.code === 'invalid_union') {
						return errorMap(issue, ctx);
					}
					const relativePath = flattenErrorPath(issue.path)
						.replace(baseErrorPath, '')
						.replace(/^\./, '');
					if ('expected' in issue) {
						expectedShape.push(
							relativePath ? `${relativePath}: ${issue.expected}` : issue.expected
						);
					} else {
						expectedShape.push(relativePath);
					}
				}
				expectedShapes.push(`{ ${expectedShape.join('; ')} }`);
			}
```

Just sharing a small idea more than a suggestion before I forgot, I did not yet have the time to play/test it a lot but I wonder if, in the case we detect a nested union issue, we could not just re-invoke the error map to get a more precise error message.

The example I'm thinking of is testing something like:

```
parseStarlightConfigWithFriendlyErrors({
  title: 'Test',
  sidebar: [
    {
      label: 'Example',
      items: [
        { label: 'Nested Example 1', link: '/' },
        { label: 'Nested Example 2', link: true },
      ],
    } as any,
  ],
})
```

The current implementation would indicate `sidebar.0` as the error path instead of `sidebar.0.items.1` which I think would be more helpful.

delucis: Added this and worked well in my test! Thanks @HiDeoo

packages/starlight/utils/error-map.ts
```diff
@@ -38,30 +55,51 @@ export const errorMap: z.ZodErrorMap = (baseError, ctx) => {
...
+				for (const issue of unionError.issues) {
+					// If the issue is a nested union error, show the associated error message instead of the
+					// base error message.
+					if (issue.code === 'invalid_union') {
+						return errorMap(issue, ctx);
+					}
+					const relativePath = flattenErrorPath(issue.path)
+						.replace(baseErrorPath, '')
+						.replace(/^\./, '');
+					if ('expected' in issue && typeof issue.expected === 'string') {
```
ematipico, line 83: nit: usually you want to place non-dynamic regex at the top of a file, so don't create a new one at every iteration

ematipico (approved):
`AstroError` ✨ 🤩

PR #1553, opened 2024-02-23
Update lockfile & patch version of EC to latest

hippotastic:
#### Description

- Just updates the dependency to the latest patch. No change for users.

### DIFF
--- a/.changeset/heavy-carrots-destroy.md
+++ b/.changeset/heavy-carrots-destroy.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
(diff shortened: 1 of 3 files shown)


delucis (approved):
Seems to be working fine, hopefully this was enough to always fix it 🤞

hippotastic:
It should! I hope I can add back the lazy-loading in a later patch again once we can reproduce the race condition.

delucis:
Thanks again for diving into this issue and figuring out. I've added a changeset to this PR so we can release a patch forcing anyone to get the latest EC just in case there are a few projects out there that might get stuck on the buggy version.

PR #1561, opened 2024-02-26
i18n(pt-PT): add "environmental-impact" page

rgilsimoes:
#### Description

- Add new /docs/pt-pt/environmental-impact page in Portuguese (Portugal)

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -136,6 +136,7 @@ export default defineConfig({
 								id: 'Dampak terhadap lingkungan',
 								'zh-CN': '环境影响',
 								'pt-BR': 'Impacto Ambiental',
+								'pt-PT': 'Impacto Ambiental',
 								ko: '환경적 영향',
 								tr: 'Çevre Etkisi',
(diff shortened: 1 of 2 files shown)


delucis (changes requested):
Looks great — thank you @rgilsimoes!

Could you also add a `pt-PT` translation for the sidebar link here, please?

https://github.com/withastro/starlight/blob/235149d05987bf6ecc5c9769f44c5c3b83d04ed8/docs/astro.config.mjs#L138

The English text is “Environmental Impact”.

Fryuni (approved):
LGTM! Thanks @rgilsimoes!

One tip though, we use squash merges and discard the commit messages, so you don't need to keep your own branch with super clean commits.

When you force-push GitHub discards in-progress reviews, including the review markers of what was already looked over, so even if you do want to rebase, organize, and rename your commits it is better to do that after the review iterations.

rgilsimoes:
Thank you for the tip. I was not aware of that.

delucis (approved):
Thanks again @rgilsimoes — and for reviewing @Fryuni 🙌

PR #1568, opened 2024-02-28
Add icon support to the `<TabItem>` component

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1453

This PR adds support for a new optional `icon` prop to the `TabItem` component. This prop can be used to add an icon to a tab to make it easier to visually distinguish between tabs.

image

### DIFF
--- a/docs/src/content/docs/guides/components.mdx
+++ b/docs/src/content/docs/guides/components.mdx
@@ -58,7 +58,8 @@ These components are available from the `@astrojs/starlight/components` package.
 import { Tabs, TabItem } from '@astrojs/starlight/components';
 
 You can display a tabbed interface using the `<Tabs>` and `<TabItem>` components.
-Each `<TabItem>` must have a `label` to display to users and can optionally include an `icon` attribute set to the name of [one of Starlight’s built-in icons](#all-icons).
+Each `<TabItem>` must have a `label` to display to users.
+Use the optional `icon` attribute to include one of [Starlight’s built-in icons](#all-icons) next to the label.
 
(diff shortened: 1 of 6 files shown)


docs/src/content/docs/guides/components.mdx
```diff
@@ -58,24 +58,32 @@ These components are available from the `@astrojs/starlight/components` package.
 import { Tabs, TabItem } from '@astrojs/starlight/components';
 
 You can display a tabbed interface using the `<Tabs>` and `<TabItem>` components.
-Each `<TabItem>` must have a `label` to display to users.
+Each `<TabItem>` must have a `label` to display to users and can optionally include an `icon` attribute set to the name of [one of Starlight’s built-in icons](#all-icons).
```
delucis, line 61: Tiny suggestion to break these into two sentences:

```suggestion
Each `<TabItem>` must have a `label` to display to users.
Use the optional `icon` attribute to include one of [Starlight’s built-in icons](#all-icons) next to the label.
```

delucis (commented):
This looks fantastic! I have no notes (well, one tiny suggestion, but given how many notes I sometimes bother you with 😅) — I was thrown for a second by the indirection and then remembered how the tabs component works 😁

docs/src/content/docs/guides/components.mdx
HiDeoo: Notes are never a bother, always appreciated!

delucis (approved):
I declare this iconic PR, ready for launch! 🚀

PR #1569, opened 2024-02-28
Updates internal github actions to the latest versions

martrapp:
#### Description

Get rid of those warnings:
image

### DIFF
--- a/.changeset/fluffy-trains-rest.md
+++ b/.changeset/fluffy-trains-rest.md
@@ -1,5 +1,0 @@
----
-'@astrojs/starlight': patch
----
-
-Updates internal github actions to the latest versions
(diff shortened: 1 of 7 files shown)


kevinzunigacuellar (approved):
LGTM! Logs are looking nice (no warnings) and majors versions seem to have no breaking changes besides upgrading node 16 to 20.

.changeset/fluffy-trains-rest.md
delucis, line 1: No need for a changeset in this case — none of these impact integration users — so we can remove this.

martrapp:
I kept actions/labeler@v4 as updating to v5 requires a change of the configuration. This will be a PR on its own.

delucis (changes requested):
Thanks so much for the careful updates @martrapp!

Just one note from me.

delucis (approved):
Thanks again @martrapp — LGTM 🌟

PR #1592, opened 2024-03-04
i18n(zh-cn): Update i18n.mdx

liruifengv:
#### Description

- Update i18n.mdx
- Upstream: https://github.com/withastro/starlight/pull/1308


liruifengv:
A little idea, maybe here we can replace the default value for the current language.

Or add the current language default value.

HiDeoo (approved):
Thanks for the update 🙌 

> A little idea, maybe here we can replace the default value for the current language.
> 
> Or add the current language default value.

Good question, I think we can continue the discussion in https://github.com/withastro/starlight/pull/1597 where you brought up a similar idea.

PR #1593, opened 2024-03-04
Translates `fileTree.directory` UI string into simplified Chinese.

liruifengv:
#### Description

- Translates `fileTree.directory` UI string into simplified Chinese.


delucis (approved):
Thanks @liruifengv! 🙌

PR #1596, opened 2024-03-04
Add support for toggling the built-in search modal using `Ctrl+k`

HiDeoo:

- Changes to Starlight code

#### Description

Got a few complaints that the shortcut to trigger the search modal was not working on Windows.
Turns out that we are only listening for `/` and `cmd + k` but not `ctrl + k` which is a bit of a standard for web applications.

This PR adds support for `ctrl + k` to trigger the search modal and aligns the behavior with the [Algolia one](https://github.com/algolia/docsearch/blob/df60b8ebe89c80ce3d00b80d94e95b6fd86edd19/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#L46) (and https://docs.astro.build at the same time).


Fryuni (approved):
I liked the current behavior because it didn't shadow Chrome's search shortcut, which is also Ctrl+K. So Ctrl is for the browser and Meta is for the site.

But I agree that this is very ubiquitous

at-the-vr:
> I liked the current behavior because it didn't shadow Chrome's search shortcut, which is also Ctrl+K. So Ctrl is for the browser and Meta is for the site.

I suppose you will like Ctrl+L 😉 , the PRs great (LGTM by default)

.changeset/happy-eggs-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
```
HiDeoo, line 2:
```suggestion
'@astrojs/starlight': patch
```

Switching to a patch as this was not an intended behavior.

PR #1597, opened 2024-03-04
docs: autogenerate default UI strings list

HiDeoo:

- Something else!

#### Description

While updating the French version of the docs `docs/src/content/docs/guides/i18n.mdx` file, I realized that we can now easily autogenerate this list of default UI strings and avoid a few i18n PRs every time we add a new string to the UI.

This PR adds a new component to the docs listing all the default UI strings in a JSON code block like this is currently done except the content is automatically based on `packages/starlight/translations/en.json`.


kevinzunigacuellar (approved):
Nice shortcut. Code component comes to save the day 🤩

HiDeoo:
> For other language pages, do you think it is necessary for us to display the language ui string here?

That's an interesting question.

I think having the list in the documentation in all languages is a good reference and also avoid discrepancies between the English docs and translated ones.
Altho, while working on this, I did thought about a potential prop for the component and use it to display the version associated to the current locale.

I could not find any recommendations for this kind of case so I'm not sure what would be the best approach: keeping the current approach and always display the English version or display the version associated to the current locale.
Maybe Chris or someone else has a strong opinion on this?

delucis (approved):
Looks great to me @HiDeoo! Lovely QoL improvement and good use of the `<Code>` component.

> Maybe Chris or someone else has a strong opinion on this?

At least currently, the docs text is “These are the English defaults”, and perhaps for translation it is ideal to keep one language as the “source of truth” for these strings, especially as occasionally we are unable to update other languages fully when introducing a new string. I don’t have a very strong opinion though!

PR #1600, opened 2024-03-05
docs: add AstroNvim to showcase

mehalter:
#### Description

This adds [AstroNvim's Documentation](https://docs.astronvim.com) to the showcase


HiDeoo (approved):
Thanks for the contribution to the showcase :raised_hands:

Really nice to see AstroNvim using Starlight :star2:

PR #1604, opened 2024-03-06
ci: add file icons generator workflow

HiDeoo:

- Something else!

#### Description

This PR is a follow-up to #1308 and adds a GitHub Actions workflow to run the Starlight file icons generator and create a PR with the updated icons/definitions if any.

- The workflow runs on a weekly schedule ([`0 0 * * 0`](https://crontab.guru/once-a-week))
- The workflow can be manually triggered
- This uses the [`peter-evans/create-pull-request`](https://github.com/peter-evans/create-pull-request) action to create a PR (which is the one used in the Astro docs [`nightly` workflow](https://github.com/withastro/docs/blob/main/.github/workflows/nightly.yml), just a more recent version)

Regarding permissions, the action mentions:

> For repositories belonging to an organization, this setting can be managed by admins in organization settings under Actions > General > Workflow permissions.

I am assuming that this would be fine in our case.

I'm also not aware of a proper way to test this workflow without first merging this PR.


delucis (approved):
Thank you @HiDeoo! And sorry for the wait. We have permissions set up for this correctly as far as I can tell, so will try merging and running. I guess until some upstream changes happen we can’t test the full flow, but we’ll at least see if the basics run OK.

delucis:
First run completed fine ✅ 

https://github.com/withastro/starlight/actions/runs/8601128407/job/23567630110

PR #1611, opened 2024-03-07
i18n(nb): update ui strings and init docs translations

Johannes-Andersen:
#### Description

This PR adds a few missing translations for the UI components `fileTree.directory` and `aside.*`

Since the project supports nb, I started work on also translating the docs into nb as well
It also sorts some of the translation objects to make it easier to find the translation I was looking for


.github/labeler.yml
at-the-vr, line 1: can you bring back older positions for the labels and append the `nb` label at the end like how the ukrainian label is in https://github.com/withastro/starlight/blob/main/.github/labeler.yml

docs/astro.config.mjs
at-the-vr, line 1: can you append `nb` related strings at the end and bring other strings back to their original position as in https://github.com/withastro/starlight/blob/main/docs/astro.config.mjs

docs/lunaria.config.json
at-the-vr, line 1: Similar append suggestion wrt https://github.com/withastro/starlight/blob/main/docs/lunaria.config.json

docs/package.json
at-the-vr, line 1: Similar append suggestion wrt https://github.com/withastro/starlight/blob/main/docs/package.json

at-the-vr (commented):
Hey Johannes, great work with this PR. I have not reviewed it with the content itself since I can't understand Norwegian but I have left some suggestions just for some "ordering" if you will 😅

PR #1614, opened 2024-03-11
Add Polish UI translations

kpodurgiel:
#### Description

This PR adds Polish translations and fixes one minor typo in the i18n schema.

---

I've been on a fence with a few translations, here's what I am choosing and why:

* `search.shortcutLabel`: `(Naciśnij /, aby wyszukać)` – the comma after the slash looks weird, but that's the rule in Polish. I've been thinking about:
  * `Naciśnij / i szukaj` ("Press / and serach"), but it's not better, as there are two single characters next to each other.
  * `Aby wyszukać, naciśnij /` – I can't tell if it's less or more visually annoying, and it sounds weird. It also conveys the meaning later in the sentence.
* `search.devWarning`: `uruchomić aplikację` literally means "run the app". "previewing the site" doesn't translate well, and I think it could cause confusion.
* `page.previousLink`: I added "page" to the word "Previous", as in Polish we decline adjectives according to nouns they go with, and leaving just "Poprzednia" / "Poprzedni" / "Poprzednie" implies the gender anyways, and it's rather rare to see it on the Polish websites. The same goes for `nextLink`.
* `aside.danger`: That was the worst to translate, because direct translations of _niebezpieczeństwo_ and _zagrożenie_ are both strongly charged words (prime definitions regard a threat of hurting or offing someone), so I went with "ważne" – "important", which I've seen in other wikis to be used for aside red sections as well.


packages/starlight/translations/pl.json
```diff
@@ -0,0 +1,30 @@
+{
+  "skipLink.label": "Przejdź do głównej zawartości",
+  "search.label": "Szukaj",
+  "search.shortcutLabel": "(Naciśnij /, aby wyszukać)",
+  "search.cancelLabel": "Anuluj",
+  "search.devWarning": "Wyszukiwanie jest dostępne tylko w buildach produkcyjnych. Spróbuj zbudować i uruchomić aplikację, aby przetestować lokalnie.",
```
HiDeoo, line 6:
```suggestion
  "search.devWarning": "Wyszukiwanie jest dostępne tylko w buildach produkcyjnych. \nSpróbuj zbudować i uruchomić aplikację, aby przetestować lokalnie.",
```

The newline seems to be missing.

packages/starlight/translations/pl.json
```diff
@@ -0,0 +1,30 @@
...
+  "themeSelect.accessibleLabel": "Wybierz motyw",
+  "themeSelect.dark": "Ciemny",
+  "themeSelect.light": "Jasny",
+  "themeSelect.auto": "Auto",
```
HiDeoo, line 10: Just to double-check, this is expected to stay the same as the English version and something missed?

packages/starlight/translations/pl.json
```diff
@@ -0,0 +1,30 @@
...
+  "languageSelect.accessibleLabel": "Wybierz język",
+  "menuButton.accessibleLabel": "Menu",
```
HiDeoo, line 12: Same question, only to confirm this is expected.

packages/starlight/translations/pl.json
```diff
@@ -0,0 +1,30 @@
...
+  "page.editLink": "Edytuj stronę",
+  "page.lastUpdated": "Ostatnia aktualizacja:",
+  "page.previousLink": "Poprzednia strona",
+  "page.nextLink": "Następna strona",
+  "404.text": "Nie znaleziono. Sprawdź URL lub użyj wyszukiwarki.",
+  "aside.note": "Notatka",
+  "aside.tip": "Wskazówka",
+  "aside.caution": "Uwaga",
+  "aside.danger": "Ważne",
+  "fileTree.directory": "Folder",
```
HiDeoo, line 26: And again, the same question to make sure this is not something that was missed.

HiDeoo (changes requested):
Thanks for your contributions :raised_hands: I really appreciate you taking the time and explaining the translations that were a bit tricky and the decisions you made.

As a non-native speaker, this looks pretty good to me. I only left a few questions just to make sure that some translations are indeed expected to be identical to the English version.

Note that when approved, the PR will only be merged before a new version is released so that the documentation is only updated when the new language is available.

packages/starlight/translations/pl.json
kpodurgiel: Correct, _Automatic_ – _Automatyczny_.

packages/starlight/translations/pl.json
kpodurgiel: Correct, a borrowing from French.

packages/starlight/translations/pl.json
kpodurgiel: Correct, as well. This was originally "Directory" ;)

HiDeoo (approved):
Thanks for fixing the line ending issue and taking the time to go over all my questions :thumbsup: 
 
Approving now so it's ready to merge when the next release will be going out.

.changeset/cold-flowers-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Add Polish UI translations
```
HiDeoo, line 5:
```suggestion
Adds Polish UI translations
```

We use the present tense for the changeset verb.

delucis (approved):
Thanks for adding these @kpodurgiel and thanks for the thorough review @HiDeoo 🙌

PR #1617, opened 2024-03-14
Add runs-on.com to showcase-sites.astro

crohr:
Hello, and thanks for making Starlight! I've been looking for a while for a nice documentation website template, with support for landing pages + blog, and Starlight delivers just that!

The screenshot is from the landing page of [runs-on.com](https://runs-on.com) (which uses the starlight `splash`), but let me know if I should reference another page instead.

I'd really like to be added to the showcase if possible :)


HiDeoo (changes requested):
Thanks for your contribution to the showcase, amazing addition :star_struck:

One tiny detail, we usually use `800x450` images for the showcase (we are in the process of documenting this process). Would you mind updating the image to that size, e.g. like this one:

runs-on com

crohr:
@HiDeoo thanks! I think it should be good now.

HiDeoo (approved):
Thanks for updating the image, everything looks great! :tada:

PR #1618, opened 2024-03-14
Update showcase-sites.astro

fl0wo:
Add DipSway starlight website as showcase example


HiDeoo (approved):
Thanks for the new addition to the showcase :raised_hands: Beautiful work on customizing Starlight!

One tiny detail if you don't mind: on your index page, the current page title visible in a tab is currently `The only <span style="color:#C702F5">Crypto Signal Endpoint</span> you'll ever need. | DipSway API` as it's based on the Starlight page title from the frontmatter which I assume contains HTML.

image


This can be something that can be fixed by manually specifying the `title` to use in the `head` section of your page picked up by browsers and removing the `<span>` tag from it:

```yml
head:
  - tag: title
    content: The only Crypto Signal Endpoint you'll ever need. | DipSway API
```

PR #1619, opened 2024-03-15
removed the default outline style of select:focus for theme-select an…

mollypan:
#### Description

- the theme-select and language-select which on the top-right nav part of astro doc has a default blue outline after clicking
- remove the outline style might be better
### before
image

### after
image

### DIFF
--- a/.changeset/clear-select-outline.md
+++ b/.changeset/clear-select-outline.md
@@ -2,9 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
-Commit Message:
-- removed the default outline style of select:focus for theme-select and language-select
-
-File Changes:
-- packages/starlight/components/Select.astro
-- packages/starlight/components/LanguageSelect.astro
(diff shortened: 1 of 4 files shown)


HiDeoo (commented):
Thanks for your pull request.

The default outline is applied as part of the browser's user agent stylesheet, e.g. in Chrome:

```css
:focus-visible {
  outline: -webkit-focus-ring-color auto 1px;
}
```

The change you're proposing would remove the default outline style which would make the component less accessible as outlined in the [W3C's Web Content Accessibility Guidelines](https://www.w3.org/TR/WCAG21/#focus-visible):

> Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.

This means that when a user interacts with the component using the keyboard for example, there should be a visible indication of the focused state which would be removed by this change.

Was this change purely cosmetic or was there another reason for removing the default outline style?

PR #1620, opened 2024-03-18
feat: Support object title for multiple language

emjio:
#### Description

The [`title`](https://starlight.astro.build/reference/configuration/#title-required) config can be translate by Object type.

title object key must be `BCP-47 tag` e.g. "en", "ar", or "zh-CN"

```ts
// title as string
export default defineConfig({
	integrations: [
		starlight({
			title: 'Starlight',
		}),
	],
});
// translate title by object
export default defineConfig({
	integrations: [
		starlight({
			title: {
                            en: 'Starlight',
                            'zh-CN': '星光',
                          },
		}),
	],
});

```

Title must have a key for the default language

```ts
// below code Will throw a error
export default defineConfig({
	integrations: [
		starlight({
			title: {
                            'zh-CN': '星光',
                          },
                         defaultLocale: 'en'
		}),
	],
});

```

#### Docs update

1. Update the type of title and add a new usage code example in ```docs/src/content/docs/configuration.mdx```
2. Description of the new siteTitle field  in ```docs/src/content/docs/reference/overrides.md```
3. Multi-language instructions for the new title in ```docs/src/content/docs/guides/i18n.mdx```


docs/astro.config.mjs
```diff
@@ -10,7 +10,7 @@ export const locales = {
 	fr: { label: 'Français', lang: 'fr' },
 	it: { label: 'Italiano', lang: 'it' },
 	id: { label: 'Bahasa Indonesia', lang: 'id' },
-	'zh-cn': { label: '简体中文', lang: 'zh-CN' },
+	'zh-CN': { label: '简体中文', lang: 'zh-CN' },
```
liruifengv, line 13:
```suggestion
	'zh-cn': { label: '简体中文', lang: 'zh-CN' },
```
keep the lowercase

docs/astro.config.mjs
```diff
@@ -34,7 +34,24 @@ export default defineConfig({
 	trailingSlash: 'always',
 	integrations: [
 		starlight({
-			title: 'Starlight',
+			title: {
```
liruifengv, line 37: Keep starlight docs's title as string.

packages/starlight/utils/user-config.ts
```diff
@@ -33,8 +33,9 @@ const LocaleSchema = z.object({
 
 const UserConfigSchema = z.object({
 	/** Title for your website. Will be used in metadata and as browser tab title. */
-	title: z
-		.string()
+	title:z.object({
+		// At least defind root lang.
+		root: z.string()}).catchall(z.string())
```
liruifengv, line 38: We need to support both string and object

liruifengv (commented):
Thanks for your work about this feature. I leave some suggestion.

- The title of the starlight document does not need to be changed.
- The `title` config need to support both string and object type.
- Update the Header's UI title in [SiteTitle.astro](https://github.com/withastro/starlight/blob/main/packages/starlight/components/SiteTitle.astro)
- You can add a changeset by run `pnpm exec changeset`
- You need update the docs in [configuration](https://starlight.astro.build/reference/configuration/#title-required)

liruifengv (commented):
Thank you for your continuous updates. I changed some codes to implement the basic functions and submitted them.

delucis (changes requested):
Thank you for the PR @emjio and your help on this @liruifengv!

I think we can make some improvements to this to simplify things a bit.

1. Instead of using `TitleTransformConfigSchema` separately every time `createTitle()` is called, we should run it once on the user config schema directly, similar to how we  `transform()` for the `locales` object there:

   https://github.com/withastro/starlight/blob/f264730c705be06caaf6728c40069336b01d064e/packages/starlight/utils/user-config.ts#L213-L266

2. Instead of requiring both `Head.astro` and `SiteTitle.astro` to use `createTitle()` directly, we should generate the correct title in our `generateRouteData()` function here:

   https://github.com/withastro/starlight/blob/f264730c705be06caaf6728c40069336b01d064e/packages/starlight/utils/route-data.ts#L36-L55
   
   I’d suggest a new entry there called `siteTitle`. That way components like Head and SiteTitle can access `Astro.props.siteTitle` and get the title for the current language without needing extra logic. For users extending or overriding our default components this means they don’t also need access to any internal logic like `createTitle()`.

3. This will need some docs updates, both for the change to the config shape (in [configuration reference](https://starlight.astro.build/reference/configuration/#title-required)) and, if we follow my suggestions, to add the new `siteTitle` entry to the props documentation (in [overrides reference](https://starlight.astro.build/reference/overrides/#props)). I guess something could also go in the i18n guide if we wanted too.

Thank you both again and let me know if any of these points needs any clarification or more information!

emjio:
Thank you @liruifengv and @delucis for taking the time to review this PR. I have completed the feature incorporating all of your suggestions.  Once again, thank you for your valuable feedback.

docs/src/content/docs/guides/i18n.mdx
```diff
@@ -13,15 +13,20 @@ Starlight provides built-in support for multilingual sites, including routing, f
...
    import starlight from '@astrojs/starlight';
 
    export default defineConfig({
    	integrations: [
    		starlight({
-   			title: 'My Docs',
+			// Set your siteTitle for multiple languages.
+   			title: {
+				en: 'My Docs',
+				'zh-cn': '我的文档',
```
liruifengv, line 27:
```suggestion
				'zh-CN': '我的文档',
```

The title object key must be BCP-47 tag e.g. "en", "ar", or "zh-CN" now. May be wo should normalize it for support it can be `zh-cn`、`pt-br`? @delucis

delucis (approved):
Thanks again for your patience with this PR @emjio! I just revised the documentation changes and this PR is ready to merge in our next release.

Thank you for adding this very helpful feature — we really appreciate you taking the time 🚀 

Thanks also to @liruifengv for championing and helping review this idea 💖

liruifengv (approved):
LGTM. Congratulations！

PR #1625, opened 2024-03-19
i18n(es): Update `plugins.mdx`

Waxer59:
#### Description

 `plugins.mdx` https://github.com/withastro/starlight/commit/8181a90d19b2748caf424e0c9b08fc3d40aec53f


HiDeoo:
Damn, I totally did not see this PR in the list before merging #1636 which was opened after this one. Totally my bad.

I notice that the translation is slightly different between the two PRs, do you think one is better than the other? If yes, we could fix the conflict and keep this one. Otherwise, we could fix the conflict and merge the PR anyway so you get your contribution too as this is totally my fault.

Waxer59:
The translation of the PR https://github.com/withastro/starlight/pull/1636 is perfect, it would not be necessary to change it.

HiDeoo (approved):
Perfect, thanks for the update and sorry again for not spotting your PR before.

PR #1627, opened 2024-03-19
Fix astro.config.mjs example syntax

alexcochran:
#### Description

This PR just fixes a small error in an `astro.config.mjs` example that I discovered as I was using these docs.


at-the-vr:
Solid suggestion Alex 😄 I wonder if you can cover this fix for remaining languages in starlight docs?

HiDeoo (approved):
Thanks for spotting this and submitting a PR, super appreciated! :raised_hands:

PR #1633, opened 2024-03-20
fix plugin hooks names

liruifengv:
#### Description

`plugin` seems was a typo? should be `setup`?


HiDeoo (approved):
Oh, I missed this one in my notifications, thought this was already merged. Thanks for tackling this! :thumbsup:

PR #1640, opened 2024-03-20
Add basic ESLint config, lint script and lint CI job

hippotastic:
#### Description

- Adds ESLint to the Starlight repo to help catch non-recommended code patterns and bugs. It can be run locally using `pnpm lint` from the repo root.
- Adds a lint job to the CI workflow that should generate annotations on PRs to make it easier to see any found issues in context.


liruifengv:
How about biome?

hippotastic:
> How about biome?

Good question! To be honest, I didn't try Biome yet, but I've heard good things about it. I was simply sticking with ESLint as it's part of my default setup I use for all projects and as it's also used in other Astro repos. Personally, I'm open for adopting Biome if it provides an experience for contributors that is on par with ESLint and finds at least the same issues.

package.json
```diff
@@ -5,6 +5,8 @@
   "description": "",
   "scripts": {
     "build:examples": "pnpm --no-bail --workspace-concurrency 1 --filter '@example/*' build",
+    "lint": "eslint .",
+    "lint:ci": "eslint . --max-warnings 0",
```
HiDeoo, line 9: Not sure why we may need 2 commands? I think we could only get away with `eslint . --max-warnings 0` and if a user explicitly runs it locally, he gets the same result as in CI. If I am not missing anything, the only reason to have 2 commands could be if we had git hooks and we would not want to prevent users from pushing non valid code?

HiDeoo (commented):
Did not have time yet to deep dive in the reported errors individually so it's mostly a general review of the approach and the config but I thought I would still share it to get the discussion started.

One last question, I see the "Unchanged files with check annotations" in GitHub:
- How does it work?
- Why is it only reporting so few errors?

.eslintrc.js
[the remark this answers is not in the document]
hippotastic: Done

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+    browser: true,
+    node: true,
+  },
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module',
+    tsconfigRootDir: __dirname,
+    project: ['./tsconfig.json', './docs/tsconfig.json'],
+  },
+  ignorePatterns: ['**/node_modules/*', '**/dist/*'],
```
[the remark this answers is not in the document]
hippotastic: Good question! I have the same redundancy in my projects, time to remove it there as well :)

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+  rules: {
+    'no-only-tests/no-only-tests': 'warn',
+    'no-mixed-spaces-and-tabs': 'off',
```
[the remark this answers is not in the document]
hippotastic: Done. I actually use the config in my own projects but thought it did more than just disabling the default Prettier rules, so I manually removed it from this one. Added it back.

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+      extends: ['plugin:@typescript-eslint/recommended-type-checked'],
+      rules: {
+        '@typescript-eslint/no-unused-vars': [
+          'warn',
+          {
+            argsIgnorePattern: '^_',
+            destructuredArrayIgnorePattern: '^_',
+            varsIgnorePattern: '^_',
+          },
+        ],
```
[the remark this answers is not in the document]
hippotastic: This is not just to reduce the level. By default, there are no ignore patterns. Other Astro repos also use the same config for this setting and this is also the one I use.

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+        '@typescript-eslint/no-non-null-assertion': 'off',
```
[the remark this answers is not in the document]
hippotastic: True, removed.

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+        '@typescript-eslint/no-redundant-type-constituents': 'off',
```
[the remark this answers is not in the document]
hippotastic: I'm usually turning this off, but if we want to start as close as possible to the defaults, I'm fine to drop it.

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+        '@typescript-eslint/triple-slash-reference': 'off',
+      },
+    },
+    {
+      files: ['packages/*/__tests__/**/*.ts'],
+      rules: {
+        '@typescript-eslint/no-unsafe-argument': 'off',
+        '@typescript-eslint/no-unsafe-assignment': 'off',
+        '@typescript-eslint/no-explicit-any': 'off',
+        '@typescript-eslint/no-empty-function': 'off',
```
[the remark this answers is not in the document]
hippotastic: Nope, `off` seems to be the default value. Removed it! :)

.eslintrc.js
```diff
@@ -0,0 +1,54 @@
...
+      rules: {
+        '@typescript-eslint/triple-slash-reference': 'off',
+      },
+    },
+    {
+      files: ['packages/*/__tests__/**/*.ts'],
+      rules: {
+        '@typescript-eslint/no-unsafe-argument': 'off',
+        '@typescript-eslint/no-unsafe-assignment': 'off',
+        '@typescript-eslint/no-explicit-any': 'off',
```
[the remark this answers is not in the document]
hippotastic: I'd be fine with removing this. It's 13 additional issues found by ESLint, so really not that much.

hippotastic:
Thank you for the great suggestions! I commented on all of them and applied most of them.

> One last question, I see the "Unchanged files with check annotations" in GitHub:
> 
> * How does it work?
> * Why is it only reporting so few errors?

To my knowledge, this is a beta GitHub feature that's using an error matcher to automatically create annotations from the errors contained in the ESLint CLI output. I don't know why it's only reporting a subset of errors. Maybe the output is limited by default, or the error matcher doesn't recognize some of the ESLint output yet. I wanted to use a different approach with a JSON output file and a separate annotation step at first, but this doesn't work on PRs coming in from forks due to security reasons, so I had to remove this again and not GitHub only applies its default matching logic.

HiDeoo:
> To my knowledge, this is a beta GitHub feature that's using an error matcher to automatically create annotations from the errors contained in the ESLint CLI output. I don't know why it's only reporting a subset of errors. Maybe the output is limited by default, or the error matcher doesn't recognize some of the ESLint output yet. I wanted to use a different approach with a JSON output file and a separate annotation step at first, but this doesn't work on PRs coming in from forks due to security reasons, so I had to remove this again and not GitHub only applies its default matching logic.

I see, thanks for the explanation. I was wondering if you had something special in place for this or not.

> How about biome?

I actually just did a test right now, I'm still amazed how crazy fast it is and how much easier the configuration is. By turning off only `style` rules, we get pretty much to the same number of errors as ESLint, altho unfortunately, the errors that started the idea of this PR are not reported yet. Maybe in the future 🤞 

---

I think configuration and setup wise, we are pretty much close to the minimum of what we would need. What do you think could be a good next step Hippo? Maybe making a list of all errors reported and figuring out a smaller list of obvious rules we want enabled as they are very valuable? And later on discuss the remaining ones?

hippotastic:
> > How about biome?
> 
> I actually just did a test right now, I'm still amazed how crazy fast it is and how much easier the configuration is. By turning off only `style` rules, we get pretty much to the same number of errors as ESLint, altho unfortunately, the errors that started the idea of this PR are not reported yet. Maybe in the future 🤞

Coincidentally I also just did an exploration into Biome! :) I was able to configure it in a way that reports the errors we were looking for. I created #1649 as an alternative to this PR so we can compare.

> I think configuration and setup wise, we are pretty much close to the minimum of what we would need. What do you think could be a good next step Hippo? Maybe making a list of all errors reported and figuring out a smaller list of obvious rules we want enabled as they are very valuable? And later on discuss the remaining ones?

My preference would be to decide about the linter to use (ESLint vs Biome), and then go through the errors and actually fix them instead of turning off any further rules. :) In my experience this doesn't take long. Most fixes are straightforward, and there may just be a bit of discussion about how to fix cases in which we have multiple options that all seem to make sense.

HiDeoo:
> I was able to configure it in a way that reports the errors we were looking for.

Ah maybe we are not exactly talking about the exact same error. For me, [`useAwait`](https://biomejs.dev/linter/rules/use-await/) only cares about an `async` function which lacks an `await` expression.
If we have a test similar to the following one which is reported by both this PR and the biome PR:

```ts
test("validates something", async () => {
  expect(async () => await doSomething()).rejects.toThrowError(/invalid/);
});
```

The big difference for me between `useAwait` and `@typescript-eslint/no-floating-promises` is that if my code was instead this one:

```ts
test("validates something", async () => {
  await expect(async () => await doSomething1()).rejects.toThrowError(/invalid 1/);
  expect(async () => await doSomething2()).rejects.toThrowError(/invalid 2/);
});
```

My test does contain an `await` so `useAwait` will not report any error. But `@typescript-eslint/no-floating-promises` will report an error for the second `expect` call because it is not awaited.

hippotastic:
Good catch! Yeah, the `no-floating-promises` rule doesn't seem to have an equivalent in Biome yet. Also, that the reported issues are not being picked up by GitHub's problem matcher is another downside of Biome right now.

We could just start with ESLint and maybe migrate to Biome once these were fixed?

HiDeoo:
> We could just start with ESLint and maybe migrate to Biome once these were fixed?

I think that's a good approach.

> go through the errors and actually fix them instead of turning off any further rules

I personally did not have the time yet to review all the reported errors to make sure they all are real issues in their context as Chris mentioned, make sense for us and would not be too annoying on the long run so if you prefer fixing them all first and we can decide in a later review if we want to disable some rules because the changes are too annoying, I'm fine with that 👍

HiDeoo:
I still did not get a chance to give this a thorough review but I'd like to point the following recent news:

- ESLint v9 has now been released.
- The new flat config format is now the default one.
- The old format (used in this PR) is now considered deprecated.
- The VSCode ESLint extension is in `pre-release` stage of a new major which already deprecate `eslint.experimental.useFlatConfig` in favor of using flat config by default (with a `eslint.useFlatConfig` setting to opt-out).
- The plugins used in this PR are all already flat config [compatible](https://github.com/typescript-eslint/typescript-eslint/issues/8211#issuecomment-2044094079) or work out of the box with the new format.
- Blu is already looking into updating Astro setup to v9

Considering all this, we should consider if it may be wise to switch to the new format before landing this PR to avoid having to do it again later and having to introduce in this PR changes to use the now deprecated format (the CLI would need env vars to opt-out of the new format, the VSCode extension would need to be configured to use the old format, etc.)

We may need to wait a bit for the VSCode ESLint extension to release its new major, in the meantime some plugins will release new versions with minor fixes for v9 unrelated changes, but except for that, this should be a pretty easy change (and something I can help if needed, done it already in a few projects including my shareable config that I use in all my projects).

HiDeoo:
Just a small update to mention that I did not forget about this PR. I still have a local copy of this branch that I update regularly.

All the tasks from my previous message are completed, all the remaining errors are fixed and the last remaning task is waiting for the next major version of `typescript-eslint` which [should land](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta) very soon. When available, I'll do a last update to the branch and push all the changes.

delucis:
> Just a small update to mention that I did not forget about this PR. I still have a local copy of this branch that I update regularly.

Do you want to push that? Or should we close this do you think @HiDeoo? `typescript-eslint` v8 [was released](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8), so I guess we could fix things if we wanted.

HiDeoo:
> Do you want to push that? Or should we close this do you think @HiDeoo? `typescript-eslint` v8 [was released](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8), so I guess we could fix things if we wanted.

I'll update my local branch using a flat config with the latest deps and see where we stand and report back :thumbsup:

HiDeoo:
Finally got the time to get back to this PR and update it, sorry for the extra super long delay!

Here is a summary of the changes I made:

- Updated the branch which was quite old
- Updated ESLint and all related packages
- Switched to a [flag config](https://eslint.org/docs/latest/use/configure/migration-guide)
- Reverted all changes previously made to the PR to ensure some changes due to previous ESLint errors were still relevant now (turns out some of them were not so this was not entirely a huge waste of time ^^)
- Disabled a bit more rules and entirely commented each change in the ESLint config
- Fixed all the remaining errors
- Remove `eslint-plugin-no-only-tests` as Vitest defaults to error for this on CI and we do the same for Playwright
- Add the ESLint extension to the recommendations in `.vscode/extensions.json`

I did not add `eslint-plugin-astro` as we could do that at a later stage when we have more experience with maintaining the codebase with ESLint being part of it and also at the moment, it does not yet support [`projectService`](https://typescript-eslint.io/packages/parser/#projectservice).

I don't think me updating the branch means _"ok, we should merge this"_ but it's more of a _"ok, let's see now what this would look like with everything updated, are we happy with this and all the changes, do we need to disable more rules, etc."_. I have been so focused on fixing errors that I also think I need some time to step back and look at all the changes too.

Any feedbacks and opinions are welcome of course.

packages/starlight/__e2e__/components.test.ts
```diff
@@ -174,7 +174,7 @@ test.describe('tabs', () => {
 		await expectSelectedTab(pkgTabsB, 'pnpm', 'another pnpm command');
 		await expectSelectedTab(pkgTabsC, 'pnpm', 'another pnpm command');
 
-		page.reload();
+		await page.reload();
```
delucis, line 177: At the very least fixing this seems valuable! 👀

packages/starlight/__tests__/basics/i18n.test.ts
```diff
@@ -13,7 +13,8 @@ describe('pickLang', () => {
 	});
 
 	test('returns undefined for unknown languages', () => {
-		expect(pickLang(dictionary, 'ar' as any)).toBeUndefined();
+		// @ts-expect-error - Testing unknown language
+		expect(pickLang(dictionary, 'ar')).toBeUndefined();
```
delucis, line 17: I like all the refactors like this — much more explicit about what’s going on 👍

packages/starlight/expressive-code.mjs
```diff
@@ -15,7 +15,7 @@
 
 export * from 'astro-expressive-code';
 
-// @ts-ignore - Types are provided by the separate `expressive-code.d.ts` file
+// @ts-expect-error - Types are provided by the separate `expressive-code.d.ts` file
```
delucis, line 18: Are we sure there are no scenarios where `expect-error` could cause issues for users here? i.e. in some configuration scenarios, this might no longer be an error causing the `expect-error` check to fail? (e.g. with `noImplicitAny` or `strict` disabled in `tsconfig.json`)

Otherwise could need `// eslint-disable-next-line @typescript-eslint/ban-ts-comment` like in [`packages/starlight/utils/translations.ts`](https://github.com/withastro/starlight/pull/1640/files#diff-2e1c8ef06b82cc2f20ae7c5ac0ad10292fd2656a190d53ac21a2412a47b75d31).

packages/starlight/integrations/heading-links.ts
```diff
@@ -56,7 +56,7 @@ export default function rehypeAutolinkHeadings(
 				{
 					type: 'element',
 					tagName: 'a',
-					properties: { class: 'sl-anchor-link', href: '#' + node.properties.id },
+					properties: { class: 'sl-anchor-link', href: '#' + String(node.properties.id) },
```
delucis, line 59: Hah, I _guess_ the explicit coercion is better? Not something that really bothers me in the existing version of the code, but don’t mind too much either way.

packages/starlight/utils/collection.ts
```diff
@@ -1,8 +1,8 @@
 import { resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-const collectionNames = ['docs', 'i18n'] as const;
-export type StarlightCollection = (typeof collectionNames)[number];
+const _collectionNames = ['docs', 'i18n'] as const;
+export type StarlightCollection = (typeof _collectionNames)[number];
```
delucis, line 5: Is there a reason not to convert this just to a type if we’re not using the tuple anywhere? (I guess maybe we used to use the tuple in the past and this is a left over artifact of that.)

```suggestion
export type StarlightCollection = 'docs' | 'i18n';
```

packages/starlight/utils/createTranslationSystem.ts
```diff
@@ -42,7 +42,7 @@ export function createTranslationSystem<T extends i18nSchemaOutput>(
 	}
 
 	const i18n = i18next.createInstance();
-	i18n.init({
+	void i18n.init({
```
delucis, line 45: Is there a rule requiring `void` for any side-effect function call like this? Seems a bit unnecessary unless there’s an argument for it I’m missing?

(In the past I’ve usually only ever used `void` in scenarios like `() => void fn()` to shield the outer function from returning the inner function call’s returned value.)

delucis (commented):
Thanks for updating the PR @HiDeoo! Left a few tiny notes, but didn’t spot anything super major I don’t think.

One question: this doesn’t include [`eslint-plugin-astro`](https://ota-meshi.github.io/eslint-plugin-astro/). Is that something you think we should consider too for consistency? (Not essential to include _in_ this PR I don’t think necessarily, unless it might inform a bit more which rules to enable/disable.)

A question for future us would presumably be whether we ever move to Biome for linting and formatting once full Astro support is available, but that could be a way off.

Otherwise, I think I’d find it hard to evaluate the chosen rules too much before actually working with them for a bit to see if any are annoying in practice (I hope not — there aren’t _too_ many instances of code changing or ignore comments in this PR to make me worry) and to get a feel for where it helps catch mistakes. There are definitely some good examples in this PR of where it can help fix and clean things up, so that’s promising.

packages/starlight/utils/createTranslationSystem.ts
HiDeoo: This one I need to investigate a bit more, but with `typescript-eslint`, the `void` operator is only used here to explicitly mark a Promise as intentionally not awaited. This is the case here as we are explicitly passing it the `resources` to initialize with and not relying on any loading from the library, which also allows us to keep the entire chain synchronous and avoid having to switch everywhere to `await useTranslations()` for example.

Depending on where we land at the end, this part will probably need to be updated, at least to comment it, or switch to a different approach.

HiDeoo:
> One question: this doesn’t include [`eslint-plugin-astro`](https://ota-meshi.github.io/eslint-plugin-astro/). Is that something you think we should consider too for consistency? (Not essential to include _in_ this PR I don’t think necessarily, unless it might inform a bit more which rules to enable/disable.)

Indeed, this is something I commented on in [this message](https://github.com/withastro/starlight/pull/1640#issuecomment-2880248184): 

> I did not add `eslint-plugin-astro` as we could do that at a later stage when we have more experience with maintaining the codebase with ESLint being part of it and also at the moment, it does not yet support [`projectService`](https://typescript-eslint.io/packages/parser/#projectservice).

I think it's highly likely that we will need to tweak rules when adding it, but I think it's fine to do that in the future, once we get used to the current setup, maybe refine it based on usage, and maybe once the plugin supports `projectService` as well (we can use a different approach for that last point but `projectService` is the new recommended approach by `typescript-eslint`).

> A question for future us would presumably be whether we ever move to Biome for linting and formatting once full Astro support is available, but that could be a way off.

I think that would be the ultimate goal (which was even tested in https://github.com/withastro/starlight/pull/1649) but we would indeed need full Astro support and also `no-floating-promises` as I commended on in [this message](https://github.com/withastro/starlight/pull/1640#issuecomment-2010709725), as this rule was part of what initially triggered the initiative.

HiDeoo:
Just updated the PR, updated some new dependencies introduced in this PR, fixed the new issues following the branch update, and addressed the last review (the most notable change is `createTranslationSystem` now being async as my [assumption](https://github.com/withastro/starlight/pull/1640#discussion_r2097192609) about `i18n.init()` was based on older i18next versions and no longer valid).

packages/starlight/expressive-code.mjs
```diff
@@ -15,7 +15,9 @@
 
 export * from 'astro-expressive-code';
 
-// @ts-ignore - Types are provided by the separate `expressive-code.d.ts` file
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore - Types are provided by the separate `expressive-code.d.ts` file and this may not be
+// a type error for some users runnign `tsc --noEmit` depending on their `tsconfig.json` settings.
```
delucis: Typo here:

```suggestion
// a type error for some users running `tsc --noEmit` depending on their `tsconfig.json` settings.
```

delucis (approved):
Changeset looks good! I think the time has come.

Thanks @hippotastic for kicking this off in the distant mists of time and to @HiDeoo for forging ahead to take us to the promised lint.

See you in 6 months when we rewrite it to use Biome 😁

hippotastic:
> Thanks @hippotastic for kicking this off in the distant mists of time and to @HiDeoo for forging ahead to take us to the promised lint.
> 
> See you in 6 months when we rewrite it to use Biome 😁

Hahaha, I just checked the date when I first opened this - it really has been a while! Amazing to see how the rough 4-day draft turned into such a solid setup. Huge thanks to @HiDeoo for taking the torch and shaping it into something that truly fits Starlight's codebase.

Can't wait for `noFloatingPromises` to leave Biome V2's nursery group, then we might actually finally revisit #1649. 😄 And given the solid setup we now have thanks to you, it hopefully won't take another year.

PR #1643, opened 2024-03-20
i18n(it): Update css-and-tailwind.mdx

casungo:


docs/src/content/docs/it/guides/css-and-tailwind.mdx
```diff
@@ -87,112 +91,116 @@ yarn create astro --template starlight/tailwind
...
-   	content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
-   	plugins: [starlightPlugin()],
-   };
-   ```
+    /** @type {import('tailwindcss').Config} */
+    export default {
+    	content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
+    	plugins: [starlightPlugin()],
+    };
+    ```
```
HiDeoo, line 201: Are you running a specific formatter or a different one than the one configured in the repo? All the changes to these lines will be reverted when Prettier will run (this can be checked using the `pnpm format` command).

Not that this would break anything in this specific case but this makes the diff larger and less readable during the review so I thought I would ask.

casungo: mhh no, this is very strange. my current workflow is:

1. i sync the files from main
2. i check the i18n tracking site to see what changed and how
3. do a diff between the "en" file and the "it" file
4. update the "it" file

should i resolve this in a commit?

HiDeoo: Weird, I would guess your editor apply that at the end of the process. Maybe it would be worth checking if it's not configured to use another formatter or has a settings to ignore the project config or something like that. 
In this case, no need to revert as I checked locally and running `pnpm format` will revert the entire changes and not break anything so we are good to go.

HiDeoo (approved):
Thanks for the update 👍

PR #1646, opened 2024-03-20
i18n(it): Update overriding-components.mdx

casungo:


HiDeoo (approved):
Thanks for updating this file, looks perfect to me :thumbsup:

PR #1649, opened 2024-03-20
Add Biome config, lint script and lint CI job

hippotastic:
#### Description

- This is an alternative to #1640 which uses Biome instead of ESLint.


hippotastic:
Here are the differences I noticed vs ESLint:
- Simpler to configure
- Much faster
- The error messages generated by the CLI are awesome, include relevant code snippets and provide actionable recommendations
- The CLI output doesn't seem to be picked up by GitHub's error matcher :(

liruifengv:
Therel has a biome GitHub Action setup, I don't know if it works.
https://biomejs.dev/recipes/continuous-integration/

hippotastic:
Closed for now after internal discussion, we'll start with ESLint and revisit a migration to Biome once the cons have been fixed in Biome. We love it though, really fast and promising!

PR #1674, opened 2024-03-22
Allow opening social links in new tab

ibdul:
before this change, clicking on the social icons opens the link in the same tab, but it is a better experience to keep the docs in its tab and then open the link in a blank tab.

### DIFF
--- a/packages/starlight/components/SocialIcons.astro
+++ b/packages/starlight/components/SocialIcons.astro
@@ -12,7 +12,7 @@ const links = Object.entries(config.social || {}) as [Platform, SocialConfig][];
 	links.length > 0 && (
 		<>
 			{links.map(([platform, { label, url }]) => (
-				<a href={url} rel="me" class="sl-flex">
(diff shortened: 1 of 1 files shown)


HiDeoo:
Thanks for your contribution.

The current behavior is actually an intentional design decision. We avoid opening links in new tabs by default as shown in various UX researches like this [one](https://www.nngroup.com/articles/new-browser-windows-and-tabs/).

In general, it's best to avoid forcing a user to open something in a new tab and let them decide how they want to open a link, e.g. a basic click = same window, cmd/ctrl + click or middle click = new window/tab.

Thanks to [component overrides](https://starlight.astro.build/guides/overriding-components/), users can still explicitly opt-in to open links in new tabs even though we do not recommend it as the default behavior.

Considering the above, I'm going to close this PR. If you have any questions or concerns, feel free to reach out on [Discord](https://astro.build/chat/) or open a discussion.

ibdul:
alright cool. makes sense. well, a worthy first try

PR #1687, opened 2024-03-28
i18n(zh-tw): translate `fileTree.directory` UI string into Traditional Chinese

mingjunlu:
#### Description

- This PR translates `fileTree.directory` UI string into Traditional Chinese (zh-TW)


delucis (approved):
Thank you for the update @mingjunlu!

PR #1606, opened 2024-03-06
docs: add showcase section to the contributing guide

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content
- Something else!

#### Description

This PR adds a new section to the contributing guide, which explains how to add a showcase entry to the Starlight docs site which is something that was never properly documented before.

I've also added a link to this new section in `docs/src/content/docs/resources/showcase.mdx` so that it's easier to find.

PR #1692, opened 2024-03-29
Remove duplicate from showcase example in `contributing.md`

HiDeoo:

- Something else!

#### Description

When modifying https://github.com/withastro/starlight/pull/1606 during Talking & Doc'ing and playing with URLs, I did not realize that we ended up with an example adding a duplicate entry to the showcase.

This PR updates the example to use a different URL based on the RFC for [Reserved Example Second Level Domain Names](https://www.rfc-editor.org/rfc/rfc2606.html#section-3).


HiDeoo:
Gonna go ahead and mentally approve myself on this one considering the scope of the changes 🫣

PR #1702, opened 2024-04-02
Add SudoVanilla to showcase

SudoVanilla:


HiDeoo (changes requested):
Thanks for your contribution to the showcase, really nice customizations! 🙌 

One small note, would you mind moving the new entry to the end of the list? We append new entries to the end of the list as documented in the [contributing guide](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#showcase). Thanks!

HiDeoo (approved):
Thanks for the update, looks perfect to me 🌟

PR #1703, opened 2024-04-02
Fix nested aside title rendering issue

### LINKED ISSUE #1699: Bug: Nested Asides

0.21.2

4.5.12

npm

Windows

Chrome, Firefox

### Describe the Bug

Nested `asides` start breaking, they either switch up their content, or just void the title/content completely.

## Bugged Behaviour:

|  Current (bugged) behaviour  |
| :--------------------------------------------------------------: |
|                    Current (bugged) behaviour                    |
|       Mixed up the title of the first aside, voided the rest     |

## Intended Behaviour:

|      Intended behaviour      |
| :--------------------------------------------------------------: |
|                        Intended behaviour                        |
|                    Everything in correct order                   |

## Code Snippet:

```md
:::::note["This should be the title of the first aside!"]

This is the content of the first note.

::::tip["This should be the title of the second aside!"]

This should be the content of the second note.

:::danger["This should be the title of the third aside!"]

This should be the content of the third note.

:::

::::

:::::
```

https://stackblitz.com/edit/starlight-asides

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1699

This PR fixes an issue with nested asides using custom titles where nested custom titles would not be rendered and the outer most custom title would be the one of the inner most aside.

This behavior was caused by the assumption that `remark-directive` stores the custom title of a container in a paragraph node added to its children. To remove such node, the current code would use `unist-util-remove` to remove this specific paragraph. However, this approach is incorrect as `unist-util-remove` iterates recursively through all children of a node, which means that if a nested aside had a custom title, the paragraph node containing its custom title would be removed by the outer aside.

The [spec](https://github.com/syntax-tree/mdast-util-directive#containerdirective-1) of `mdast-util-directive` is also a bit more precise about the structure of a container with a custom title. It states that:

> The phrasing in the label is, when available, added as a paragraph with a directiveLabel: true field, as the head of its content.

As the label would always be a paragraph added as the head of the content of a container, we can entirely remove the iteration step and directly remove the first child of a container if it is a paragraph with a `directiveLabel` field set to `true`.


packages/starlight/__tests__/remark-rehype/asides.test.ts
```diff
@@ -106,6 +106,35 @@ Nested tip.
...
+	const labels = [...res.code.matchAll(/aria-label="(?<label>[^"]+)"/g)].map(
+		(match) => match.groups?.label
+	);
+	expect(labels).toMatchInlineSnapshot(`
+		[
+		  "Caution with a custom title",
+		  "Note",
+		  "Tip with a custom title",
+		]
+	`);
```
HiDeoo, line 134: Even if the snapshot can be used to ensure custom titles are properly rendered, I added this small expectation to make it easier to glance at what the expected output is instead of having to read the snapshot file.

delucis (approved):
Looks great! Thank you for the quick fix @HiDeoo 🙌 

Let everyone go forth and populate their docs with dubious deeply nested asides!

PR #1705, opened 2024-04-02
Redesign experiments

HiDeoo:

- Something else!

#### Description

This PR is an experiment to play with various changes suggested in this [discussion](https://github.com/withastro/starlight/discussions/1701).

This should **never** be merged. It's just a playground to quickly see how things look and feel using a lot of non-ideal changes just to get things done quickly.

See the discussion for more details or to provide feedback.


HiDeoo:
I am going to close this PR, this was only used in the purpose of a discussion and the preview should still be available if ever needed.

PR #1706, opened 2024-04-02
fix: minor type issues in starlight

jorenbroekema:
#### Description

- Closes # N/A didn't create an issue since the PR is so tiny
- What does this PR change? Fixes some types

Here's the type issues I was getting, skipLibCheck unfortunately doesn't work for TS files (only .d.ts) so this is a real blocker when trying to lint/compile my project which uses starlight.
image


packages/starlight/integrations/shared/pathToLocale.ts
```diff
@@ -17,7 +17,7 @@ export function pathToLocale(
 		astroConfig,
 	}: {
 		starlightConfig: { locales: StarlightConfig['locales'] };
-		astroConfig: { root: URL; srcDir: URL };
+		astroConfig: { root: URL | string; srcDir: URL };
```
HiDeoo, line 20:
```suggestion
		astroConfig: { root: AstroConfig['root']; srcDir: AstroConfig['srcDir'] };
```

I guess a more appropriate fix would be something similar to what is done for [`AsidesOptions`](https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/asides.ts#L23) and reference the Astro types (with `AstroConfig` being from `import type { AstroConfig } from 'astro';`).

Altho, this would not fix the issue as it's typed as `URL` in Astro. I think the issue is located in Expressive Code which [defines it](https://github.com/expressive-code/expressive-code/blob/main/packages/astro-expressive-code/src/astro-config.ts#L26) as `URL | string`. I guess a fix would need to happen there too (and maybe refactor this entire type to rely on Astro's types, cherry pick what is needed, make some bits partial or use a deep partial type helper, etc. instead of duplicating them manually?).

cc @hippotastic

packages/starlight/user-components/rehype-file-tree.ts
```diff
@@ -168,7 +168,7 @@ function getFileIcon(fileName: string) {
 
 /** Return the icon name for a file based on its file name. */
 function getFileIconName(fileName: string) {
-	let icon = definitions.files[fileName];
+	let icon = definitions.files[fileName] as string | undefined;
```
HiDeoo, line 171: This is a weird one (or I am missing something obvious right now ^^). Maybe this only happens when type-checked from outside of Starlight because inside, the type is properly inferred to `string | undefined`.

SCR-20240404-opgb

Does switching to `let icon: string | undefined = definitions.files[fileName];` fix the issue on your end?

HiDeoo (commented):
Thanks for your contribution, super appreciated!

Left a few comments/questions inline to make sure we properly understand the issues and fix them properly but this is clearly something that needs to be fixed. Thanks for opening up a PR.

packages/starlight/user-components/rehype-file-tree.ts
jorenbroekema: Ah yeah that fixes it too, probably nicer than a typecast.

The issue is that `let icon = definitions.files[fileName];` infers the type as `string` but 2 lines below it `icon = getFileIconTypeFromExtension(fileName);` gives a type error because `getFileIconTypeFromExtension` return type is `string | undefined`, which is correctly guarded against on the line below it.

I'm guessing somewhere I'm getting a different version of something installed meaning I get the type error but you don't, could be a different typescript version maybe, but it looks to me like this type error should definitely show on your end too.

packages/starlight/integrations/shared/pathToLocale.ts
jorenbroekema: If there's a way where we can make the type more accurate and not break in the future, I'm all for it!
I am personally not super familiar with astro/starlight types, nor zod, so I don't have a strong opinion here.

HiDeoo: > It might make sense to change your publishing strategy in light of this

Definitely, there is this [discussion](https://github.com/withastro/starlight/discussions/983) summarizing the issue but not something that has been tackled yet.

> If there's a way where we can make the type more accurate and not break in the future, I'm all for it!

Yeah, I think the change you suggested (with the above suggestion) and a fix on the Expressive Code side (once confirmed by Hippo) would be the best way to go forward.

packages/starlight/user-components/rehype-file-tree.ts
HiDeoo: I would guess this may be due to the TS configuration based on [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) which makes it infer as `string | undefined` in the Starlight codebase but `string` if this option is not enabled.

But in the end, this boils down to the same issue as above, where this should not matter if not published as TypeScript. In the meantime, switching to `let icon: string | undefined = definitions.files[fileName];` should workaround the issue.

packages/starlight/integrations/shared/pathToLocale.ts
hippotastic: Just FYI, the type has been fixed in Expressive Code v0.34.0.

packages/starlight/user-components/rehype-file-tree.ts
HiDeoo:
```suggestion
	let icon: string | undefined = definitions.files[fileName];
```

As this fixes the issue, I guess we can now workaround the problem with this syntax.

packages/starlight/integrations/shared/pathToLocale.ts
HiDeoo: > Just FYI, the type has been fixed in Expressive Code v0.34.0.

Amazing 🙌 

I guess applying the two suggested changes + the EC update would then fix the issues.

jorenbroekema: Done :) amended my commit and pushed

HiDeoo: Thanks a lot :raised_hands: If you dont't mind, I think we should remove the EC bump from this PR as not strictly necessary for this PR and it also looks like there is another [PR](https://github.com/withastro/starlight/pull/1720) opened for this as it looks it may requires extra changes and may contain some breaking changes by quickly skimming through it.

jorenbroekema: Makes sense, didn't think of the fact that pre-v1.0.0 minor updates would usually contain breaking changes

delucis (approved):
Thank you for these fixes and your responsiveness making changes @jorenbroekema! Appreciate the contribution 🌟

PR #1709, opened 2024-04-04
Fix translation issue for sites with a single non-root language different from English

### LINKED ISSUE #1556: UI text not translated when creating a multilingual site with only one locale

0.20.1

4.4.4

npm

Mac

Brave

### Describe the Bug

If I create a multilingual site with only one non-English locale like this, the UI text (like "On this page") is not translated, but always English instead.

```js
defaultLocale: 'de',
locales: {
    de: {
        lang: 'de',
        label: 'Deutsch'
    }
},
```

As soon as I add another locale, for example `fr`, the UI text of the `de` locale is correctly translated to German.

Why am I trying to create a multilingual site with only one locale in the first place? I want to prepare for translations in the future.

https://stackblitz.com/edit/github-4oxm53?file=astro.config.mjs

You have to visit `/de/` in Stackblitz manually. For some reason, the redirect from `/` to `/de/` doesn't seem to work in Stackblitz - in my local installation, it does.

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1556

This pull request fixes a UI strings translation issue for sites configured with a single non-root language different from English. In such case, the UI strings were not being translated correctly and were being displayed in English.


packages/starlight/utils/user-config.ts
delucis, line 1: I’m wondering if the more logically “correct” fix here is actually in the second return statement rather than changing the `if()` block.

```diff
-				...locales?.root,
+				...locales?.[defaultLocale || 'root'],
```

That helps keep the separation of the first block handling genuinely multilingual sites with the default return handling the monolingual case, even if that default is not `root`.

But maybe there’s something I’m not accounting for?

delucis (commented):
Thank you for tackling this @HiDeoo! 💖

packages/starlight/utils/user-config.ts
HiDeoo: > But maybe there’s something I’m not accounting for?

I don't think so, at least, I did not see the separation this way. I saw it more as "having a `locales` option configured" vs "not having one".

One benefit of the approach of the PR is we get the validation out-of-the-box just like when we use the `locales` option. For example, if we take the following configuration:

```
locales: {
  fr: { label: 'English', lang: 'fr' }
},
```

This configuration is technically invalid because there is no root locale and no `defaultLocale` set but this would not error out with the suggested approach and the site would be in English.

With the current approach of the PR, we get:

> Could not determine the default locale. Please make sure `defaultLocale` in your Starlight config is one of "fr"

delucis: Ah yes, I didn’t really think of people misconfiguring this — just the `defaultLocale: 'fr', locales: { fr: ... }` case. A better error message is definitely preferable.

I guess why I considered my suggestion as “logically” correct is that the two returned objects are for two different scenarios — monolingual and multilingual — and I’m privileging output over input shape here as the schema is designed to normalize for output usage and not necessarily to reflect the input shape (not sure if I’m even making sense here, but that’s how it is in my head).

It’s mostly subtleties though, like `locales` being `undefined` in monolingual sites. The one obvious place where you had to adapt was `isMultilingual` no longer being defined statically. I think in practice that doesn’t change anything in Starlight code, because `isMultilingual` is used as the discriminator to see if `locales` is defined and this doesn’t change that (only that now sometimes `locales` is defined even when `isMultilingual` is false).

HiDeoo: > It’s mostly subtleties though, like locales being undefined in monolingual sites.

That's interesting, I saw this specific case as being monolingual while having a locale defined (which reflects in the tests I wrote). So interesting to see how our minds can see things differently while looking at the same problem.

I guess we can hoist or refactor the default locale validation, switch to the approach you suggested (and update the tests as the locale would now be `undefined`). I guess that would make more sense if we want to favor the the output like you mentioned rather than the input (even tho my head is still trying to battle this somehow ^^). Does that sound good to you?

delucis: I mean, the current fix is better than the bug! 😄 

I’d personally like that, but not if it feels wrong to you. I generally think of the schema as a processing layer that outputs a clean and consistent output shape for our more complex APIs so that internal code can do as little work as possible to understand it. So that’s what I mean by “privileging output over input” — the schema + transform is the glue that converts a user-friendly API shape to an internal code-friendly shape. That said, I’m not sure the processed i18n config is as clean as it could be — there’s a lot of `lang ?? locale` type stuff all over our codebase 😅

HiDeoo: Oh no, definitely does not feel wrong otherwise I would have not suggested to change it. It's just the fact that I saw that in a different way at first ^^

Not gonna have the time to refactor that today but will get to it asap :thumbsup:

HiDeoo: Ok, just updated the PR with the following:

- Use the suggested approach
- Hoist the validation step to always get a meaningful error message when the default locale cannot be inferred
- `defaultLocale.locale` is no longer always `undefined`, as in this specific case we need to keep a reference to it (as it's not a key in `locales` compared to the previous approach) so that `slugToLocale()` can use it if needed

Note that I did not add new tests for this change as the previous attempt had slugs tests which broke when switching to the new approach and got fixed with the new `defaultLocale.locale` and `slugToLocale()` changes.

HiDeoo: Just updated the PR to revert the refactor. I also added a bit of comments to describe a bit more the 2 branches, but not sure if it helps a lot (but like we discussed on Discord, this may need to be revisited to improve the output format to better handle all the possible cases and simplify the internal logic).

delucis (approved):
Thanks for the back and forth on this one @HiDeoo! LGTM! 🚀

sschultze:
Thank you @HiDeoo , this works perfectly!

PR #1713, opened 2024-04-05
docs: sort icons alphabetically

HiDeoo:

- Something else!

#### Description

This pull request address this [Discord message](https://discord.com/channels/830184174198718474/1070481941863878697/1225768815397310574):

> Would you be open to sorting the icons alphabetically? I surf on mobile and tablet a lot and not easy to find a specific icon when it's not sorted. Find functionality is too many steps on mobile

Until we get a better story for icons in the documentation, some search feature, etc., I think this is a good idea. This PR sorts the icons alphabetically (always using the same locale to ensure a consistent order in all languages which I think makes sense in this specific case).

### DIFF
--- a/docs/src/components/icons-list.astro
+++ b/docs/src/components/icons-list.astro
@@ -2,7 +2,9 @@
 import { Icon } from '@astrojs/starlight/components';
 import { Icons } from '../../../packages/starlight/components/Icons';
 
-const icons = Object.keys(Icons) as (keyof typeof Icons)[];
+const icons = [...(Object.keys(Icons) as (keyof typeof Icons)[])].sort((a, b) =>
+	a.localeCompare(b, 'en')
+);
 ---
 
 <div class="icons-grid">


delucis (commented):
Thanks @HiDeoo! This is a bit of a tricky one. While alphabetical order superficially seems nice, I do have a few concerns:

- It’s a bit inconvenient to have to skip all the `seti:` icons to get to the icons later in the alphabet.
- Alphabetical sorting breaks up logically associated groups (although these are very implicit currently). For example, the down-, right-, and up-caret icons are currently displayed together but scattered through the list when sorted alphabetically. (There’s an argument that maybe they should have been named `caret-down`, `caret-right`, etc.) More conceptually, the “brand” icons are currently mostly together, but this change breaks them up.

So I guess, I’m uncertain this change is a net positive or whether there are quick wins here. I have an old branch where I started work on using Unicons’ category metadata to group icons that might be better to dust off.

HiDeoo:
This is indeed tricky.

By looking at various popular icons set libraries, I guess the most common approach seems to be:

- all by alphabetical order by default
- category list always visible
- possibility to filter by category (and still keep the alphabetical order)
- search feature

Considering we only have a list right now with no visible categories, I think I would still categorize this as a better temporary improvement.
Altho, I'm not against skipping this entirely and aiming for a more robust solution right away (I could also take a look at your old branch if it's not something you have the time to prioritize right now).

HiDeoo:
I am going to close this PR, it did not get any traction so I think it is safe to assume we can take the time to skip it and later implement it in a more efficient way.

PR #1637, opened 2024-03-20
docs: Use awesome `Steps` component

liruifengv:
#### Description

- Use awesome `Steps` component

PR #1717, opened 2024-04-07
i18n(ja): Use Steps component

morinokami:
#### Description

- Add Steps component to match https://github.com/withastro/starlight/pull/1637.


delucis (approved):
Thank you for the update @morinokami! Nice to see the `<Steps>` component working well alongside text in Japanese 🥳

PR #1723, opened 2024-04-08
Tab aria selected

OliverSpeir:
## What

Instead of removing the aria-selected attribute, set it to false for the tabs that are not selected.

## Why

When using VoiceOver (built in macos screenreader)  if a tab does not have aria-selected="false" it is read as "selected", so right now starlight tabs are all read as selected making it confusing to know which one is actually selected

### DIFF
--- a/.changeset/late-onions-jog.md
+++ b/.changeset/late-onions-jog.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Instead of removing aria-selected attribute for non selected tabs, add aria-selected="false"
+Fixes accessibility by using `aria-selected="false"` for inactive tabs instead of removing `aria-selected="true"` in the tablist of Starlight’s `<Tabs>` component
--- a/.changeset/late-onions-jog.md
+++ b/.changeset/late-onions-jog.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Instead of removing aria-selected attribute for non selected tabs, add aria-selected="false"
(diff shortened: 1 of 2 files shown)


.changeset/late-onions-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Instead of removing aria-selected attribute for non selected tabs, add aria-selected="false"
```
delucis, line 5: Small suggestion to clarify the changeset (which component this effects) and to use our standard verb form of third-person instead of imperative:

```suggestion
Fixes accessibility by using `aria-selected="false"` for inactive tabs instead of removing `aria-selected="true"` in the tablist of Starlight’s `<Tabs>` component
```

delucis (commented):
Thank you @OliverSpeir!

MDN’s [`aria-selected` docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#tab) also says:

> All inactive tabs in the tablist should have `aria-selected="false"` set.

So sounds like is indeed the correct behaviour. Appreciate the fix 💖

delucis (approved):
> To my knowledge, omitting `aria-selected=false` entirely was often a practice used to reduced the noise with some screen readers

Ah, that is interesting. Thanks for sharing @HiDeoo. Something to monitor then, just in case this change actually makes for a worse experience. But definitely seems like an improvement over all tabs being announced as selected.

Thanks again @OliverSpeir 🙌

PR #1728, opened 2024-04-08
Adds a Pkl file format icon

delucis:
#### Description

- As requested in https://github.com/withastro/starlight/discussions/1599
- Uses `opacity` attribute to reproduce the two-tone quality of the original icon.
- Ran into an interesting issue with our standard `<path />` style where with two paths, the second path ended up rendered _inside_ the first: `<path><path /></path>`. Solved it for now by using a full `</path>` to close the first path.
- Updated our file tree code to also understand the `.pkl` extension and map it to the new icon:
   several file tree entries including a config.pkl file with matching icon


HiDeoo (approved):
Looks perfect to me on both themes, and even when used with the `<Icon />` component 🌟

PR #1729, opened 2024-04-09
docs: add caution about scripts processing when using head config

### LINKED ISSUE #1704: Local Typescript files are not working

0.21.2

4.5.10

npm

Linux Ubuntu 20 LTS

Chrome

### Describe the Bug

It is not possible to include Typescript files into the head of starlight projects.

I've tried quite a few methods:

```js
{
  integrations: [
    starlight({
      head: [
        {
          tag: 'script',
          content: `import '/src/load-playground.ts';`,
        },
      ],
    })
  ]
}
```

The above complains about not allowing to use import statements because it's not a module, however in the astro docs https://docs.astro.build/en/guides/client-side-scripts/#script-processing it clearly says that imports to local files should be working regardless of the missing type="module"

I also tried:
```js
{
  integrations: [
    starlight({
      head: [
        {
          tag: 'script',
          attrs: {
            type: 'module',
            src: '/src/load-playground.ts',
          },
        },
      ],
    })
  ]
}
```
Which is working locally when running `astro dev`, but the script isn't processed/bundled at all when running `astro build` and `astro preview`, so when deploying my site, it's not working.

https://stackblitz.com/edit/github-gw67mn?file=astro.config.mjs,src%2Fload-playground.ts

### Participation

jorenbroekema:
#### Description

- Closes #1704 
- What does this PR change? Adds some documentation about scripts processing when using the `head` config option, and what to do if you need scripts to be processed by Astro


docs/src/content/docs/reference/configuration.mdx
```diff
@@ -444,6 +444,11 @@ starlight({
 });
 ```
 
+:::caution
+Scripts are placed into head using the `set:html` directive which is similar to using `is:inline` in that it [precludes these scripts from being processed by Astro](https://docs.astro.build/en/guides/client-side-scripts/#opting-out-of-processing).
+This means that if you want to include a local script and you need this to be processed, you'll probably need to [override or extend the Head component](/guides/overriding-components/#reuse-a-built-in-component).
+:::
```
delucis, line 450: Instead of a caution, I think we can add a clear description of how this _does_ work (rather than specifying how it doesn’t):

```suggestion
Entries in `head` are converted directly to HTML elements and do not pass through Astro’s script or style processing.
If you need to import local assets like scripts, styles, or images, [override the Head component](/guides/overriding-components/#reuse-a-built-in-component).
```

This also helps us make this a bit simpler because we don’t need to explain too much all these other things which `head` _isn’t_, instead focusing on the simpler case of what it is.

delucis (changes requested):
Thank you for the PR @jorenbroekema! Super helpful to have suggestions like this and know which parts of our docs tripped people up. Left one suggestion.

docs/src/content/docs/reference/configuration.mdx
jorenbroekema: Done, I did embed a link to the Astro style and script docs in case users reading this will want to learn what this style and script processing is all about

delucis (approved):
Perfect — thank you!

PR #1734, opened 2024-04-10
Refactor `<ThemeSelect>` script to avoid double media query event listener

delucis:
#### Description

- Follow up to #1731 
- That PR introduces a media query event listener so that the colour theme can stay in sync with the system theme when using the default `auto` theme. But because we often have two `<ThemeSelect>` components on each page (nav bar and mobile menu), this meant a double listener, each updating both select menus.
- To solve this I pulled most of the logic out of the `StarlightThemeSelect` class, which wasn’t actually instance-specific, so the media query event listener can just run once at the top level.
- The diff is a bit overwhelming but it’s mainly moving the private class methods up to top-level constants. The media query gets moved from the custom element constructor to the top level so we only have one.

### DIFF
--- a/.changeset/sweet-vans-carry.md
+++ b/.changeset/sweet-vans-carry.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Refactors `<ThemeSelect>` custom element logic to improve performance
--- a/packages/starlight/components/ThemeSelect.astro
+++ b/packages/starlight/components/ThemeSelect.astro
@@ -28,65 +28,50 @@ const { labels } = Astro.props;
 <script>
 	type Theme = 'auto' | 'dark' | 'light';
 
-	class StarlightThemeSelect extends HTMLElement {
-		/** Key in `localStorage` to store color theme preference at. */
-		#key = 'starlight-theme';
+	/** Key in `localStorage` to store color theme preference at. */
+	const storageKey = 'starlight-theme';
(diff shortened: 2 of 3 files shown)


HiDeoo (approved):
Thanks, this looks like a thorough refactor to me and my local tests did not spot any issues. Looks perfect to me :thumbsup:

Unrelated to this change, we should probably do a follow-up to the follow-up to fix the `Property 'StarlightThemeProvider' does not exist on type 'Window & typeof globalThis'.ts(2339)` TS error, I would guess it's just missing an export for `declare` in `packages/starlight/global.d.ts`.

delucis:
> Unrelated to this change, we should probably do a follow-up to the follow-up to fix the `Property 'StarlightThemeProvider' does not exist on type 'Window & typeof globalThis'.ts(2339)` TS error, I would guess it's just missing an export for `declare` in `packages/starlight/global.d.ts`.

I could do that in this PR I think — you’re right that `export declare ...` fixes the error on my end.

techfg:
@delucis  - Great stuff here and thank you!  I had mis-interpreted the previous logic when making the media change mod in that the I thought the double event listener issue already existed for the `select` itself but realize now that the event listener that existed on the select itself wasn't doubled up since it was instance specific so my change did introduce the new bug - it didn't exist prior.  Sorry I overlooked that originally and again, appreciate you getting this cleaned-up!

PR #1735, opened 2024-04-10
Adds custom styles for `<details>` and `<summary>` elements in Markdown content

HiDeoo:

- Changes to Starlight code

#### Description

This PR adds custom styles for `<details>` and `<summary>` elements in Markdown content, adresssing the [request](https://github.com/withastro/starlight/discussions/1491#discussioncomment-8889999) in the [built-in components discussion](https://github.com/withastro/starlight/discussions/1491).

Preview playground page: https://starlight-git-fork-hideoo-hd-style-details-astrodotbuild.vercel.app/reference/details/

The approach used does not add a new `<Details>` component but relies on the native `<details>` and `<summary>` HTML elements, similar to the [Docusaurus approach](https://docusaurus.io/docs/markdown-features#details).

The current design is based on the [Astro Docs](https://docs.astro.build/en/tutorial/0-introduction/1/) custom implementation but the implementation is not exactly identical as I noticed a few minor issues while implementing it (some margins inconsistencies and some border being offset by 1px when toggling the details). 
This PR also adds support for RTL languages, `prefers-reduced-motion` and some minor color adjustments to match the Starlight theme.

One point to note is also the usage of `<details>` in asides, where the styling deviates from Astro Docs as it can lead to some readability issues. I implemented custom variations to address this, but I don't think it looks the best as I did not introduce any new color (which may be needed to make it more appealing) and I'm also not the best when it comes to design decisions.

I reviewed the accessibility of the implementation using this [checklist](https://designsystem.digital.gov/components/accordion/accessibility-tests/) and testing with screen readers but relying on the native elements ensures most of the concerns are addressed out of the box.

Any feedback is welcome!


HiDeoo:
Following a Talking and Doc'ing session (April 11, 2024), some very interesting feedback was received from the participants, thank you all for your time and feedback!

One particular piece of feedback from @kevinzunigacuellar  was that it may be difficult to follow if a user is still in a `<details>` or not , specially in nested `<details>`. Discussions about this included the idea of adding some borders to the `<details>` to make it more clear. Here are some screenshots of me playing around with the idea, nothing is final of course and the goal is just to discuss the idea.

### Borders around all `<details>`

Personal opinion: I think this adds a lot of visual noise and importance to the `<details>`.

Borders around all details

### Borders only around nested `<details>`

Borders only around nested details

### Slightly dimmed borders around nested `<details>`

Slightly dimmed borders around nested details

HiDeoo:
Sharing here another great approach to the design by @delucis 

image

image

The preview is available here: https://starlight-git-chris-details-dabbling-astrodotbuild.vercel.app/reference/details/

HiDeoo:
I updated the PR to use Chris's design. I personally prefer it but as it's now part of the history, we can easily switch between the two if ever needed. My main remaining concern may be the visibility of the left border in light mode.

In a separate commit, I also added a suggestion from Sarah to move the marker to the left to emphasize that these are clickable which was one of the main feedback from the previous post. Same, this can be easily reverted now during iterations.

HiDeoo:
Updated the PR and tweaked border colors: the `<details>` border color is now more "vivid" compared to the one from a blockquote or a `<Steps>` component to emphasize that it's an interactive element. Hovering and the opened state borders are also more "vivid" now to match this change.

Hovering an opened `<details>` element will now also change the border color to a different color to still indicate that it's clickable.

And finally, I fixed a marker position issue with long summaries due to MDX wrapping the summary in a `<p>` tag.

packages/starlight/style/markdown.css
```diff
@@ -113,3 +113,73 @@
...
+	padding-inline-start: 0.5rem;
+}
+.sl-markdown-content details[open] > summary:not(:where(.not-content *)) {
+	margin-bottom: 1rem;
+}
+
+/* <summary> marker styles */
+.sl-markdown-content summary:not(:where(.not-content *))::marker,
+.sl-markdown-content summary:not(:where(.not-content *))::-webkit-details-marker {
+	display: none;
```
subframe7536, line 145: `display: none` is invalid for summary marker, please add `list-style: none` instead

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -361,6 +361,31 @@ A code block’s optional title can be set either with a `title="..."` attribute
   ```
   ````
 
+## Details
+
+Details (also known as “disclosures” or “accordion”) are useful to conceal information that is not immediately relevant to the user.
+The user can reveal further information with an interaction to toggle the visibility of the content.
```
delucis, line 367: This sentence was maybe saying the same thing twice a bit: “reveal further information” and “toggle the visibility of the content”.

I think we can maybe boil it down to this?

```suggestion
Users can click a short summary to expand and view the full content.
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -361,6 +361,31 @@ A code block’s optional title can be set either with a `title="..."` attribute
...
+
+Starlight does not provide a custom Markdown syntax for details, but you can use the standard HTML `<details>` and `<summary>` elements to create them while authoring your Markdown content.
```
delucis, line 369: I think we can describe the positive “what to do” here without mentioning what we don’t do. I do not _love_ “widget” as a word here, but it does appear to be what MDN uses, so maybe it’s OK? (And does work better than “element”, “component”, or “pattern”… any other ideas?)

```suggestion
Use the standard HTML [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) and [`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) elements in your Markdown content to create a disclosure widget.
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -361,6 +361,31 @@ A code block’s optional title can be set either with a `title="..."` attribute
...
+
+You can nest any other Markdown content types inside a `<details>` element.
```
delucis, line 371: Maybe “syntax” works better here?

```suggestion
You can nest any other Markdown syntax inside a `<details>` element.
```

docs/src/content/docs/guides/authoring-content.md
```diff
@@ -361,6 +361,31 @@ A code block’s optional title can be set either with a `title="..."` attribute
...
+</details>
+
+```md
+<details>
+<summary>Where and when the Andromeda constellation is best visible?</summary>
+
+The [Andromeda constellation](<https://en.wikipedia.org/wiki/Andromeda_(constellation)>) is best visible in the night sky during the month of November at latitudes between `+90°` and `−40°`.
+
+</details>
+```
```
delucis, line 387: Guess we need to make sure the demo content is also just right 😅 

Just a small English grammar detail here:

````suggestion
<details>
<summary>Where and when is the Andromeda constellation most visible?</summary>

The [Andromeda constellation](<https://en.wikipedia.org/wiki/Andromeda_(constellation)>) is most visible in the night sky during the month of November at latitudes between `+90°` and `−40°`.

</details>

```md
<details>
<summary>Where and when is the Andromeda constellation most visible?</summary>

The [Andromeda constellation](<https://en.wikipedia.org/wiki/Andromeda_(constellation)>) is most visible in the night sky during the month of November at latitudes between `+90°` and `−40°`.

</details>
```
````

delucis (changes requested):
Thanks again for the work on this @HiDeoo!

I’ll summarise again the couple of changes that we discussed and I pushed directly:
- border is now 2px instead of 1px, matching the `<Tabs>` component thickness
- hover styling only applies when hovering `<summary>` on expanded `<details>` — this helps avoid a slightly chaotic/flickering feeling with the border colour changing rapidly as user’s move over details
- hover colour is now a more bold accent like `<Tabs>` uses for selected tabs

Other than that I think we’re more or less there! I left a couple of notes on the docs and some CSS, but then I think we should be good to clean up the test page.

We’ll probably also need to bump up the CSS size-limit budget.

sarah11918 (approved):
Approving for docs!

delucis (approved):
Thank you for working through the different iterations and rolling with the changes I kept throwing at you @HiDeoo! 🎉 

It will be nice to see these updated and in-use in Astro’s docs soon too.

PR #1759, opened 2024-04-15
test: fix Windows path separator test issues

HiDeoo:

- Something else!

#### Description

While testing another PR on Windows, I noticed that some tests where failing. They are all related to tests with regex matching paths where the path separators were invalid and `:` was sometimes missing for `X:` drives.


delucis (approved):
Thanks for the fix!

PR #1751, opened 2024-04-12
docs: rewrite `guides/sidebar` examples to be more generic

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content

#### Description

While working #1750, I realized that the "Sidebar Navigation" guide was present in a lot of search results for unrelated queries (and will be even more after the `pagefind` v1.1.0 update). This is because the sidebar examples embedded in the guide are re-using the Starlight documentation site's sidebar making it appear in search results for queries like "component", "CSS", etc.

This pull request rewrites the guide examples to be more generic and use constellations/stars/moons instead that are being used in other guides.

- Previous example links were actually linking to said pages, now they do not redirect anywhere (this uses an invalid anchor - `#` scroll to the top of the page and others solutions involve JS)
- The "Internationalization" section involves Brazilian Portuguese translations, I linked the reference I used for the translations but this should be reviewed by a native speaker

PR #1760, opened 2024-04-15

jsparkdev:
#### Description

- update `sidebar.mdx`
- #1751

### DIFF
--- a/docs/src/content/docs/ko/guides/sidebar.mdx
+++ b/docs/src/content/docs/ko/guides/sidebar.mdx
@@ -488,7 +488,7 @@ starlight({
 				{ label: '용골자리', link: '' },
 				{ label: '켄타우루스', link: '' },
 				{
-					label: 'seasonal,
+					label: 'seasonal',
 					collapsed: true,
 					items: [{ label: '안드로메다', link: '' }],
 				},
--- a/docs/src/content/docs/ko/guides/sidebar.mdx
+++ b/docs/src/content/docs/ko/guides/sidebar.mdx
@@ -19,11 +19,11 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
 - src/
   - content/
     - docs/
-      - guides/
-        - components.md
-        - i18n.md
-      - reference/
-        - configuration.md
+      - constellations/
+        - andromeda.md
+        - orion.md
+      - stars/
+        - betelgeuse.md
 
 </FileTree>
 
@@ -32,17 +32,15 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
 <SidebarPreview
 	config={[
 		{
-			label: 'guides',
+			label: 'constellations',
 			items: [
-				{ label: 'Components', link: '/guides/components/' },
-				{ label: 'Internationalization (i18n)', link: '/guides/i18n/' },
+				{ label: 'Andromeda', link: '' },
+				{ label: 'Orion', link: '' },
 			],
 		},
 		{
-			label: 'reference',
-			items: [
-				{ label: 'Configuration Reference', link: '/reference/configuration/' },
-			],
+			label: 'stars',
+			items: [{ label: 'Betelgeuse', link: '' }],
 		},
 	]}
 />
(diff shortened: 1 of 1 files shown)


HiDeoo (commented):
Thanks for the update, this is looking really great.

Small question: I noticed you did not translate the new examples (they can safely be translated). Is this intentional? Do you think it would be better to leave them in English? Is this a domain-specific where English is more common? For example, if I am not wrong, Andromeda is [안드로메다자리](https://ko.wikipedia.org/wiki/%EC%95%88%EB%93%9C%EB%A1%9C%EB%A9%94%EB%8B%A4%EC%9E%90%EB%A6%AC) and we have more generic terms like "stars", "constellations", "moons", etc. that I guess could be translated?

 Just a question of course, you definitely know better than me what is best in this case so feel free to tell me if I am wrong and things are better this way.

jsparkdev:
Thank you!

I translated the content that I thought was natural to translate.

I left badges containing `"New"`, `"Outdated"`, and `"Stub"` as is because I thought it would be better not to translate them.

docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -32,17 +32,15 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
 <SidebarPreview
 	config={[
 		{
-			label: 'guides',
+			label: '별자리',
```
HiDeoo, line 35:
```suggestion
			label: 'constellations',
```

The default sidebar will use the folder names and as they are not translated in this case, this would be just `constellations`.

docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -32,17 +32,15 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
...
+				{ label: '안드로메다', link: '' },
+				{ label: '오리온', link: '' },
 			],
 		},
 		{
-			label: 'reference',
-			items: [
-				{ label: 'Configuration Reference', link: '/reference/configuration/' },
-			],
+			label: '별',
```
HiDeoo, line 42:
```suggestion
			label: 'stars',
```

Same as above

docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -172,18 +169,13 @@ starlight({
...
+				{ label: '켄타우루스', link: '' },
 				{
-					label: '고급',
-					items: [
-						{
-							label: '프로젝트 구조',
-							link: '/ko/guides/project-structure/',
-						},
-					],
+					label: '계절별',
```
HiDeoo, line 177:
```suggestion
					label: 'seasonal',
```

For nested folders in autogenerated group, the folder name will be used.

docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -490,17 +482,15 @@ starlight({
...
+			label: '별자리',
 			collapsed: true,
 			items: [
-				{ label: '컴포넌트', link: '/ko/guides/components/' },
-				{ label: '국제화 (i18n)', link: '/ko/guides/i18n/' },
+				{ label: '용골자리', link: '' },
+				{ label: '켄타우루스', link: '' },
 				{
-					label: '고급',
+					label: '계절별',
```
HiDeoo, line 491:
```suggestion
					label: 'seasonal,
```

This is also a folder in an autogenerated group, so the folder name would be used.

docs/src/content/docs/ko/guides/sidebar.mdx
```diff
@@ -527,16 +517,14 @@ starlight({
...
-			label: '가이드',
+			label: '별자리',
 			items: [
-				{ label: '컴포넌트', link: '/ko/guides/components/' },
-				{ label: '국제화 (i18n)', link: '/ko/guides/i18n/' },
+				{ label: '용골자리', link: '' },
+				{ label: '켄타우루스', link: '' },
 				{
-					label: '고급',
+					label: '계절별',
```
HiDeoo, line 525:
```suggestion
					label: 'seasonal',
```

Same as above.

HiDeoo (changes requested):
Thanks a lot for taking the time to update the translation, this looks amazing and what you did not change makes total sense to me.

I left a few suggestions for the case of no config or folders in autogenerated groups where the folder names are used (as we have no API at the moment to properly translate them) but once fixed, this is ready to go :rocket:

docs/src/content/docs/ko/guides/sidebar.mdx
jsparkdev: Thank you!

HiDeoo (approved):
Amazing, thanks for your patience, this looks great 🌟 (I pushed a tiny commit to add a missing quote introduced by one of my suggestions 🥲)

PR #1761, opened 2024-04-16
i18n(zh-cn): Update `sidebar.mdx`

huyikai:
#### Description

Update `sidebar.mdx`

Upstream #1751

### DIFF
--- a/docs/src/content/docs/zh-cn/guides/sidebar.mdx
+++ b/docs/src/content/docs/zh-cn/guides/sidebar.mdx
@@ -177,7 +177,7 @@ starlight({
 				{ label: '船底座', link: '' },
 				{ label: '半人马座', link: '' },
 				{
-					label: '周期',
+					label: 'seasonal',
 					items: [{ label: '仙女座', link: '' }],
 				},
 			],
--- a/docs/src/content/docs/zh-cn/guides/sidebar.mdx
+++ b/docs/src/content/docs/zh-cn/guides/sidebar.mdx
@@ -419,7 +419,7 @@ starlight({
 <SidebarPreview
 	config={[
 		{
-			label: 'Constelação',
+			label: '星座',
 			items: [
 				{ label: '仙女座', link: '' },
 				{ label: '天蝎座', link: '' },
--- a/docs/src/content/docs/zh-cn/guides/sidebar.mdx
+++ b/docs/src/content/docs/zh-cn/guides/sidebar.mdx
@@ -34,13 +34,13 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
 		{
 			label: 'constellations',
 			items: [
-				{ label: 'Andromeda', link: '' },
-				{ label: 'Orion', link: '' },
+				{ label: '仙女座', link: '' },
+				{ label: '猎户座', link: '' },
 			],
 		},
 		{
 			label: 'stars',
-			items: [{ label: 'Betelgeuse', link: '' }],
+			items: [{ label: '参宿四', link: '' }],
 		},
 	]}
 />
@@ -60,8 +60,8 @@ import SidebarPreview from '~/components/sidebar-preview.astro';
 ```js "label:" "link:"
 starlight({
 	sidebar: [
-		// 指向 Ganymede moon 页面的链接。
-		{ label: 'Ganymede', link: '/moons/ganymede/' },
+		// 指向木卫三卫星页面的链接。
+		{ label: '木卫三', link: '/moons/ganymede/' },
 		// 指向 NASA 网站的外部链接。
 		{ label: 'NASA', link: 'https://www.nasa.gov/' },
 	],
@@ -72,7 +72,7 @@ starlight({
 
 <SidebarPreview
 	config={[
-		{ label: 'Ganymede', link: '' },
+		{ label: '木卫三', link: '' },
 		{ label: 'NASA', link: 'https://www.nasa.gov/' },
 	]}
 />
@@ -89,19 +89,19 @@ starlight({
 ```js /^\s*(label:|items:)/
 starlight({
 	sidebar: [
-		// 一个名为 "Constellations" 的链接分组
+		// 一个名为 "星座" 的链接分组
 		{
-			label: 'Constellations',
+			label: '星座',
 			items: [
-				{ label: 'Carina', link: '/constellations/carina/' },
-				{ label: 'Centaurus', link: '/constellations/centaurus/' },
-				// seasonal constellations 的嵌套链接分组。
+				{ label: '船底座', link: '/constellations/carina/' },
+				{ label: '半人马座', link: '/constellations/centaurus/' },
+				// 星座周期的嵌套链接分组。
 				{
-					label: 'Seasonal',
+					label: '周期',
 					items: [
-						{ label: 'Andromeda', link: '/constellations/andromeda/' },
-						{ label: 'Orion', link: '/constellations/orion/' },
-						{ label: 'Ursa Minor', link: '/constellations/ursa-minor/' },
+						{ label: '仙女座', link: '/constellations/andromeda/' },
+						{ label: '猎户座', link: '/constellations/orion/' },
+						{ label: '小熊座', link: '/constellations/ursa-minor/' },
 					],
 				},
 			],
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/zh-cn/guides/sidebar.mdx
```diff
@@ -423,10 +419,10 @@ starlight({
 <SidebarPreview
 	config={[
 		{
-			label: '指南',
+			label: 'Constelação',
```
liruifengv, line 422:
```suggestion
			label: '星座',
```

liruifengv (approved):
LGTM. Thanks!

HiDeoo:
Thanks for the update and the review.

Small question: I see that labels like "constellations", "moons", "stars", etc. are not translated and same for constellation names like "Andromeda". Is this on purpose? Do you think it's better to leave them in English? Is this a domain-specific where English is more common? Could something like "Andromeda" be translated to ["仙女座"](https://zh.wikipedia.org/wiki/%E4%BB%99%E5%A5%B3%E5%BA%A7) or this would not make sense?

Of course, this is just a question, you definitely know better than me what is best in this case and I'd like your opinions.

huyikai:
> Thanks for the update and the review.
> 
> Small question: I see that labels like "constellations", "moons", "stars", etc. are not translated and same for constellation names like "Andromeda". Is this on purpose? Do you think it's better to leave them in English? Is this a domain-specific where English is more common? Could something like "Andromeda" be translated to ["仙女座"](https://zh.wikipedia.org/wiki/%E4%BB%99%E5%A5%B3%E5%BA%A7) or this would not make sense?
> 
> Of course, this is just a question, you definitely know better than me what is best in this case and I'd like your opinions.

Hi HiDeoo, nice to communicate with you.  

I didn't translate these contents for the following reasons. 

1. Some of the content involves paths and filenames, which are usually not set in Chinese due to issues such as sorting. Translating them into Chinese would be somewhat counterintuitive. 

2. In my opinion, in an English-speaking environment, these terms in the documentation have a kind of easter egg feel for Starlight. However, as far as I know, the educational foundation in Chinese-speaking environments rarely includes a knowledge system about astronomy or constellations. People are not sensitive to these terms in Chinese, and translating them might actually make it harder to understand.

3. The **most important point** is to differentiate from the i18n content below, so that readers can better understand the use of basics and the use of internationalization.

liruifengv:
> Thanks for the update and the review.
> 
> Small question: I see that labels like "constellations", "moons", "stars", etc. are not translated and same for constellation names like "Andromeda". Is this on purpose? Do you think it's better to leave them in English? Is this a domain-specific where English is more common? Could something like "Andromeda" be translated to ["仙女座"](https://zh.wikipedia.org/wiki/%E4%BB%99%E5%A5%B3%E5%BA%A7) or this would not make sense?
> 
> Of course, this is just a question, you definitely know better than me what is best in this case and I'd like your opinions.

Nope. This is not English domain-specific. I think it would be fun to translate them too. 
For [Default sidebar](https://starlight-git-fork-huyikai-20240416-sidebarmdx-astrodotbuild.vercel.app/zh-cn/guides/sidebar/#default-sidebar), I'm not sure whether if need translate, because we don't usually use Chinese for file paths. Maybe we should keep them in English?

HiDeoo:
Thanks you both for your feedback, really appreciated.

> > Thanks for the update and the review.
> > Small question: I see that labels like "constellations", "moons", "stars", etc. are not translated and same for constellation names like "Andromeda". Is this on purpose? Do you think it's better to leave them in English? Is this a domain-specific where English is more common? Could something like "Andromeda" be translated to ["仙女座"](https://zh.wikipedia.org/wiki/%E4%BB%99%E5%A5%B3%E5%BA%A7) or this would not make sense?
> > Of course, this is just a question, you definitely know better than me what is best in this case and I'd like your opinions.
> 
> Nope. This is not English domain-specific. I think it would be fun to translate them too. For [Default sidebar](https://starlight-git-fork-huyikai-20240416-sidebarmdx-astrodotbuild.vercel.app/zh-cn/guides/sidebar/#default-sidebar), I'm not sure whether if need translate, because we don't usually use Chinese for file paths. Maybe we should keep them in English?

Indeed, and I guess having `/constellations/something` but the label being translated is not as obvious as having `/constellations/something` and the label being "Constellations" as pointed out by @huyikai.

> **The most important point** is to differentiate from the i18n content below, so that readers can better understand the use of basics and the use of internationalization.

What is done in the [French translation](https://starlight.astro.build/fr/guides/sidebar/#internationalisation) for example, is that default labels are translated in French but we keep the original translations in `pt-BR` so readers can still see the internationalization in action. Altho, again, it's easier as having file paths in French is not an issue and fairly common.

---

Considering all this, maybe only "Resources" in this [section](<https://starlight-git-fork-huyikai-20240416-sidebarmdx-astrodotbuild.vercel.app/zh-cn/guides/sidebar/#%E8%87%AA%E5%AE%9A%E4%B9%89-html-%E5%B1%9E%E6%80%A7>) should be translated as this is a safe one, right?

HiDeoo:
Tiny update if this can help the discussion: we had the same discussion for the `ko-KR` PR where the [final result](https://starlight.astro.build/ko/guides/sidebar/) is mostly translated except some very specific terms ("New", "Outdated", "Stub") which did not make sense. Maybe it could help figuring out what is best in this case.

docs/src/content/docs/zh-cn/guides/sidebar.mdx
```diff
@@ -174,15 +172,13 @@ starlight({
...
-				{ label: 'Components', link: '/guides/components/' },
-				{ label: 'Internationalization (i18n)', link: '/guides/i18n/' },
+				{ label: '船底座', link: '' },
+				{ label: '半人马座', link: '' },
 				{
-					label: 'advanced',
-					items: [
-						{ label: 'Project Structure', link: '/guides/project-structure/' },
-					],
+					label: '周期',
```
HiDeoo, line 180:
```suggestion
					label: 'seasonal',
```

In the case of nested folders in autogenerated groups, as we do not have an API for this, the folder name will be used (which is even more obvious now so I think it helps a bit)

docs/src/content/docs/zh-cn/guides/sidebar.mdx
```diff
@@ -487,17 +483,15 @@ starlight({
...
+			label: '星座',
 			collapsed: true,
 			items: [
-				{ label: 'Components', link: '/guides/components/' },
-				{ label: 'Internationalization (i18n)', link: '/guides/i18n/' },
+				{ label: '船底座', link: '' },
+				{ label: '半人马座', link: '' },
 				{
-					label: 'advanced',
+					label: '周期',
```
HiDeoo, line 492:
```suggestion
					label: 'seasonal',
```

Same as the previous suggestion for nested folders in autogenerated groups.

docs/src/content/docs/zh-cn/guides/sidebar.mdx
```diff
@@ -524,16 +518,14 @@ starlight({
...
-			label: 'Guides',
+			label: '星座',
 			items: [
-				{ label: 'Components', link: '/guides/components/' },
-				{ label: 'Internationalization (i18n)', link: '/guides/i18n/' },
+				{ label: '船底座', link: '' },
+				{ label: '半人马座', link: '' },
 				{
-					label: 'advanced',
+					label: '周期',
```
HiDeoo, line 526:
```suggestion
					label: 'seasonal',
```

Same as above.

HiDeoo (changes requested):
Thanks for the update, this is looking really great :thumbsup:

I left a few suggestions for the case of nested folders in autogenerated groups which would use the folder name (this would also becomes more obvious for readers with the translated content that we do not support this yet as nested folders would be using the folder name in English).

Except for that, everything looks in order to me.

huyikai:
> Thanks for the update, this is looking really great 👍
> 
> I left a few suggestions for the case of nested folders in autogenerated groups which would use the folder name (this would also becomes more obvious for readers with the translated content that we do not support this yet as nested folders would be using the folder name in English).
> 
> Except for that, everything looks in order to me.

Thanks for your correction. I submitted your suggestions. Could you please confirm again.

HiDeoo (approved):
Amazing, this is looking really great. Thanks a lot for the updates :raised_hands:

PR #1765, opened 2024-04-18
Fix sidebar badge outline style with Tailwind CSS

### LINKED ISSUE #1764: Badge styles break when refreshing with sidebar entry active

0.21.5

4.3.5

npm

Linux

Firefox and chromium

### Describe the Bug

- using tailwind set up according to the docs
- Create a page that displays a badge in the sidebar
- Go to the page
- Refresh
- The badge style is broken
- Navigate to a different page without a badge
- Refresh
- The badge styling is fixed

live demo https://starlight-lovat.vercel.app/

https://github.com/tgoHQ/knowledgebase

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1764

Outlined sidebar badges looks like this by default in Starlight:

SCR-20240418-kksc

When using Tailwind CSS, the outline style is thicker:

SCR-20240418-kfwo

This is due to the variant (`outline`) being used as a CSS class name which collides with Tailwind CSS [`outline`](https://tailwindcss.com/docs/outline-style) utility adding `	outline-style: solid;`.

This PR renames the class name to `sl-outline` to avoid the collision and get the following result matching the default Starlight style:

SCR-20240418-kgbg

#### Alternative approach

Edit: a potential alternative solution suggested by @martrapp could be to remove entirely the need of a CSS class for this and rely on a selector like `[aria-current='page'] :global(.sl-badge)`. This would mean not exposing the `outline` variant in #1530 (which is the current approach) but we would need to confirm that first.

#### How to test

1. Edit `examples/tailwind/astro.config.mjs` to add a badge to the "Example Guide" sidebar link.
1. Run `pnpm dev` in the `examples/tailwind/` directory.
1. Navigate to `http://localhost:4321/guides/example/`

### DIFF
--- a/.changeset/eight-grapes-watch.md
+++ b/.changeset/eight-grapes-watch.md
@@ -1,0 +1,8 @@
+---
+'@astrojs/starlight': minor
+---
+
+Fixes the outline style of sidebar badges when using Tailwind CSS.
+
+This involves changing the name of an internal CSS class name to avoid a collision with Tailwind CSS utility classes.
+If you were relying on the previous (`.outline`) class name for custom styling, you will need to update your custom CSS to use the new class name (`.sl-outline`).
--- a/packages/starlight/components/Badge.astro
+++ b/packages/starlight/components/Badge.astro
@@ -6,9 +6,10 @@ interface Props {
 	text?: string;
 }
 const { variant = 'default', text } = Astro.props;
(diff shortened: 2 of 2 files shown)


HiDeoo:
Closing as I found a thread mentioning that the outline variant is meant to be internal and not exposed even with the `<Badges>` component. This means that #1530 removing it will close the issue.

PR #1769, opened 2024-04-18
Pass "frontmatter" variable to content renderer

ncjones:
Setting the "frontmatter" variable allows markdoc content to access the frontmatter like `{% $frontmatter.title %}`. This pattern is documented in the Astro markdoc guide:

https://docs.astro.build/en/guides/integrations-guide/markdoc/#access-frontmatter-from-your-markdoc-content


delucis (approved):
Thank you for the PR and your patience on this one.

I added a changeset so this gets added to our change log, but the change seems fair enough to me. I kind of feel like Astro should take care of this by default, but until that’s the case, I think this is fine to add.

PR #1773, opened 2024-04-21
Add TrueCharts to showcases

Ornias1993:
#### Description

Adds TrueCharts.org to Showcases.
Highlighting a significant number of Astro and Starlight components and multiple custom components

It also has thousands of pages of docs to showcase efficiency of Astro+Starlight as a docs backend


HiDeoo (commented):
Thanks for your submission, really great website to add to the showcase 🤩 

I noticed in the [preview](https://starlight-git-fork-ornias1993-patch-1-astrodotbuild.vercel.app/resources/showcase/) that the screenshot felt quite zoomed out compared to the other websites in the showcase.

What do you think about this screenshot instead which still matches the screenshot size but is a bit more zoomed in? I personally think this could make it a bit more recognisable.

truecharts org

Ornias1993:
> Thanks for your submission, really great website to add to the showcase 🤩
> 
> I noticed in the [preview](https://starlight-git-fork-ornias1993-patch-1-astrodotbuild.vercel.app/resources/showcase/) that the screenshot felt quite zoomed out compared to the other websites in the showcase.
> 
> What do you think about this screenshot instead which still matches the screenshot size but is a bit more zoomed in? I personally think this could make it a bit more recognisable.
> 
> truecharts org

Totally agree, good catch!

HiDeoo (approved):
> Totally agree, good catch!

Great, I updated the image to the suggested one and this looks perfect to me :thumbsup:

PR #1777, opened 2024-04-22
Fix TypeScript schema error when using project references

### LINKED ISSUE #1612: Starlight not usable with typescript `declarations: true`

0.21.1

4.3.5

npm

Mac

Chrome

### Describe the Bug

Starlight is not usable if `"declaration": true,` is set in `tsconfig.json`.
This option however is mandatory for projects with `"composite": true`. _(Enable constraints that allow a TypeScript project to be used with project references.)_

The following issue occurs in `src/content/config.ts`

```
The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.(7056)
```

Potentially related to this is:

https://github.com/colinhacks/zod/issues/1040

https://stackblitz.com/edit/github-gx6te3?file=src%2Fcontent%2Fconfig.ts

### Participation

HiDeoo:

- Changes to Starlight code

#### Description

- Closes #1612

The issue in #1612 is related to having [`composite`](https://www.typescriptlang.org/docs/handbook/project-references.html#composite) set to `true` in a user `tsconfig.json`. This option, common for projects using [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), also enables the [`declaration`](https://www.typescriptlang.org/tsconfig#declaration) option, which would generate a TypeScript error in the `src/content/config.ts` file:

```
The inferred type of this node exceeds the maximum length the compiler will serialize.
An explicit type annotation is needed.(7056)
```

Even tho a workaround could be to just disable `composite` specifically in the Starlight project, this could prevent even type validation using `tsc`.

The issue is related to the recursion happening in the [`BaseSchemaWithoutEffects`](https://github.com/withastro/starlight/blob/main/packages/starlight/schema.ts#L111C6-L115) type of the schema.
After spending many hours playing with [`tsc --extendedDiagnostics`](https://www.typescriptlang.org/tsconfig/#extendedDiagnostics), `tsc --generateTrace`, `chrome://tracing`, [`Perfetto`](https://ui.perfetto.dev/), and [`@typescript/analyze-trace`](https://www.npmjs.com/package/@typescript/analyze-trace), I did not find a way to optimize the type to avoid the error while still keeping the exact same behavior and possibilities it offers.

Fortunately, and surprisingly, I noticed that the issue only happens when the `docsSchema()` is called with no argument (which unfortunately is the most common usage with Starlight).
Considering this, this PR introduces a new explicit type on the `docsSchema()` removing a bit of inference and adding a new short-circuit when it is called with no argument to default to `DefaultSchema`.

~~__This PR is a draft as I still need to play with it a bit more and make sure it does not break anything but I want to do that with a clear mind after not spending too much time in front of it.__~~ 

#### How to test?

Simply adding `"composite": true` to the `compilerOptions` of the `docs/tsconfig.json` file and opening `docs/src/content/config.ts` should trigger the error without the changes from this PR.


packages/starlight/schema.ts
```diff
@@ -117,7 +117,9 @@ type BaseSchemaWithoutEffects =
 type BaseSchema = BaseSchemaWithoutEffects | z.ZodEffects<BaseSchemaWithoutEffects>;
 
 /** Type that extends Starlight’s default schema with an optional, user-defined schema. */
-type ExtendedSchema<T extends BaseSchema> = T extends BaseSchema
+type ExtendedSchema<T extends BaseSchema | never = never> = [T] extends [never]
```
HiDeoo, line 120: This [pattern](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) is used to disable distributivity as when conditional types act on a generic type, they become distributive when given a union type which can be the case here.

HiDeoo:
Marking this as ready for review now as I did not find so far anything breaking with the change so happy to get people trying it out and getting feedback.

delucis (approved):
I have only a hazy understanding of this issue, but fully trust you and the code change seems fine to me!

Needs a changeset though — hopefully if we can describe what this fixes simply, it won’t be as complex as the concept of distributivity 😁

HiDeoo:
> Needs a changeset though — hopefully if we can describe what this fixes simply, it won’t be as complex as the concept of distributivity 😁

Took a stab at this, hope this makes any sense :laughing:

delucis (approved):
Changeset looks good to me! No docs in this one, so I think we can merge right away. Thanks @HiDeoo 🎉

PR #1782, opened 2024-04-24
i18n(id) Update components.mdx

yudyananda:
#### Description

Updating outdated and fix small error in translation

### DIFF
--- a/docs/src/content/docs/id/guides/components.mdx
+++ b/docs/src/content/docs/id/guides/components.mdx
@@ -3,8 +3,8 @@ title: Komponen
 description: Menggunakan komponen dalam MDX dengan Starlight.
 ---
 
-Komponen memungkinkan Anda dengan mudah menggunakan kembali bagian UI atau _style_ secara konsisten.
-Contohnya mungkin kartu tautan atau _YouTube embed_.
+Komponen memudahkan Anda menggunakan kembali bagian UI atau _styling_ secara konsisten.
+Contohnya bisa kartu tautan atau penyematan YouTube.
 Starlight mendukung penggunaan komponen dalam file [MDX](https://mdxjs.com/) dan menyediakan beberapa komponen umum yang dapat Anda gunakan.
 
 [Pelajari lebih lanjut tentang membuat komponen di Astro Docs](https://docs.astro.build/en/core-concepts/astro-components/).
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/id/guides/components.mdx
```diff
@@ -163,6 +164,198 @@ import { LinkCard } from '@astrojs/starlight/components';
...
+### File Tree
+
+Gunakan komponen `<FileTree>` untuk menampilkan struktur direktori dengan ikon berkas beserta subdirektori yang dapat diciutkan.
+
+Tentukan struktur file dan direktori Anda dengan [daftar Markdown tidak berurutan](https://www.markdownguide.org/basic-syntax/#unordered-lists) di dalam `<FileTree>`.
+Buat subdirektori menggunakan daftar bersarang atau tambahkan `/` diakhir item untuk merendernya sebagai direktori tanpa konten yang spesifik.
+
+Sintaks berikut dapat digunakan untuk menyesuaikan tampilan pohon file:
+
+The following syntax can be used to customize the appearance of the file tree:
```
HiDeoo, line 277:
```suggestion
```

This is probably something forgotten.

docs/src/content/docs/id/guides/components.mdx
```diff
@@ -163,6 +164,198 @@ import { LinkCard } from '@astrojs/starlight/components';
...
+- Menambah _placeholder_ untuk berkas dan direktori dengan memakai `...` or `…` sebagai nama.
+
+```mdx
+# src/content/docs/example.mdx
+
+import { FileTree } from '@astrojs/starlight/components';
+
+<FileTree>
+
+- astro.config.mjs an **important** file
```
HiDeoo, line 290: Do you think it would make sense to translate the "an important file comment". We did it in other languages so if we can, I think it could help a bit. Let me know what you think.

docs/src/content/docs/id/guides/components.mdx
```diff
@@ -163,6 +164,198 @@ import { LinkCard } from '@astrojs/starlight/components';
...
+</FileTree>
+```
+
+Kode di atas akan menampilkan hasil sebagai berikut di halaman:
+
+import { FileTree } from '@astrojs/starlight/components';
+
+<FileTree>
+
+- astro.config.mjs an **important** file
```
HiDeoo, line 308: Same question as the previous comment, if we translate it above, we need to update this one.

HiDeoo (changes requested):

I left a few comments on the PR but this is looking really great :thumbsup:

Something I noticed in this PR, is that heading level 3 in the "Built-in components" section, e.g. `### Tabs`, `### Cards`, etc., are usually translated as they do not need to match component names. Do you think this is something that would make sense in this specific case? Totally fine if you think it's not necessary or would make the documentation less clear. Just wanted to bring it up in case you think it's a good idea and would like to make the change.

delucis:
Hi @yudyananda are you still interested in updating this page? Let us know if you have any issues with @HiDeoo’s suggested changes!

delucis:
Thanks again for opening this PR @yudyananda — as we haven’t seen any updates recently, I’ll close this PR so it doesn’t get stale, but happy to receive any future PRs if you have time!

PR #1784, opened 2024-04-25
Add `<LinkButton>` component

HiDeoo:

- Changes to Starlight code

#### Description

This PR adds a new `<LinkButton>` component adressing the [request](https://github.com/withastro/starlight/discussions/1491#discussioncomment-8445584) in the built-in components [discussion](https://github.com/withastro/starlight/discussions/1491).

Preview playground page: // TODO(HiDeoo) 

Visually, this is almost identical to the internal `<CallToAction>` component (now removed and replaced by the new component) that was used for actions in the [hero component](https://starlight.astro.build/reference/frontmatter/#hero).

The name is still up for discussion, some design system use `CTALink` or `CallToAction` but this may not be obvious to everyone. Some others use `ButtonLink` but as pointed out by Chris, we already have a `<LinkCard>` component so this uses the same pattern.
Using something like `<Button>` may be confusing as this is a link that should be used for navigation and not a form primitive for example.

#### Remaining tasks

- [ ] Remove demo style (`docs/src/assets/landing.css`)
- [ ] Remove demo page (`docs/src/content/docs/reference/link-button.mdx`)


packages/starlight/style/markdown.css
```diff
@@ -62,10 +62,10 @@
 	font-size: var(--sl-text-h6);
 }
 
-.sl-markdown-content a:not(:where(.not-content *)) {
+.sl-markdown-content a:not(:where(:is(.not-content, .not-content *))) {
 	color: var(--sl-color-text-accent);
 }
-.sl-markdown-content a:hover:not(:where(.not-content *)) {
+.sl-markdown-content a:hover:not(:where(:is(.not-content, .not-content *))) {
```
HiDeoo, line 68: These changes are used to avoid `<LinkButton>` getting Markdown link styles applied to it by specifying the `not-content` class on the link itself and not a wrapping element. 

I'm wondering if this is a change that could be applied to some other places to avoid relying on a wrapping element to avoid Markdown styles :thinking:

HiDeoo:
Following the last Talking & Doc'ing session discussions, I updated the pull request to include the following changes:

- The `<LinkButton>` variant now defaults to `primary` instead of `minimal`.
- The hero component action button variant now defaults to `primary` instead of `minimal`.

packages/starlight/package.json
```diff
@@ -51,10 +51,6 @@
       "types": "./components/Sidebar.astro.tsx",
       "import": "./components/Sidebar.astro"
     },
-    "./components/CallToAction.astro": {
```
lorenzolewis, line 54: May want to call this out in the changeset as well

lorenzolewis:
One additional thing: Not sure if you were basing this off of the [Figma file](https://www.figma.com/design/GrDkcguAR7tSWPFuLVDixq/Theme---Starlight-Docs?node-id=171-534&t=Fe0I26jfeXcKFztj-1) or not, but how about links without any text and only an icon? Is that something you wanted to support in this?

packages/starlight/style/markdown.css
delucis: I’d like to avoid this if we can to avoid introducing inconsistencies.

The main reason `.not-content` applies to children not the element with the class is due to spacing. Consider an example like this:

```html
<p>Lorem ipsum...</p>
<div class="not-content">Some non-content...</div>
<p>Dolor sic...</p>
```

If `not-content` removed the margin styles from the `<div>` in this example, it would end up squished next to the first paragraph.

I have sometimes considered whether we should have separate rules for escaping spacing vs other styles, but I’m not sure that’s particularly easy to understand or communicate.

What _would_ be better would be to either give the Markdown styles a lower specificity (e.g. wrap each selector in `:where()` maybe?) or move forward with the `@layer` refactor. That would make it less painful for user components (or our own in this case) to escape the default styles without the extra `not-content` class. (`not-content` would still be needed for circumstances that benefit from a full reset, but there wouldn’t be the temptation to use it because otherwise `!important` is needed to enforce a component style.)

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,72 @@
...
+	.sl-link-button.primary {
+		background: var(--sl-color-text-accent);
+		border: 1px solid var(--sl-color-text-accent);
+		color: var(--sl-color-black);
+	}
+	.sl-link-button.secondary {
+		border: 1px solid;
+	}
+	.sl-link-button.minimal {
+		outline-offset: 0.25rem;
```
delucis, line 50: Not strictly related to this PR, but noting it here while paying attention to this component: our default colours make focus styling quite hard to see. For example, one of these is focused but the difference is almost imperceptible:

Two blue link buttons. The left button has a focus outline that is so close in colour to the button background that it is almost impossible to see.

The situation is slightly better in dark mode, but not much. I wonder if we could do something with `outline-offset` also for the other variants to help avoid people (and us) hitting this?

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,72 @@
...
+
+<style>
+	.sl-link-button {
+		align-items: center;
+		border-radius: 999rem;
+		color: var(--sl-color-white);
+		display: inline-flex;
+		font-size: var(--sl-text-sm);
+		gap: 0.5em;
+		line-height: 1.7145;
```
delucis, line 36: What’s the story behind this number? I’m assuming it’s higher than the line height for the existing CTA component (`1.1875`) due to `display: inline-flex`, but the number seems a bit random, so would be good to document if possible.

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,72 @@
...
+		padding: 0.5rem 1.125rem;
+		text-decoration: none;
+	}
+
+	.sl-link-button.primary {
+		background: var(--sl-color-text-accent);
+		border: 1px solid var(--sl-color-text-accent);
```
delucis, line 43: I think it might be cleanest to always set a border on links rather than only on primary and secondary variants. That will avoid sizing inconsistencies with the minimal variant.

I’d probably go with `border: 1px solid transparent` on the main button class and then set the colour only in primary and secondary.

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,72 @@
...
+---
+
+<a class:list={['sl-link-button not-content', variant, className]} {...attrs}>
+	{icon && iconPlacement === 'start' && <Icon name={icon} size="1.5rem" />}
+	<slot />
+	{icon && iconPlacement === 'end' && <Icon name={icon} size="1.5rem" />}
+</a>
+
+<style>
+	.sl-link-button {
```
delucis, line 29: Nit: the default new component is 2px taller than the current one (58px vs 56px on the home page in desktop view, 42px vs 40px in mobile view). That sounds like not much, but I definitely feel it as a little bit chunkier, so would be great to keep the height consistent with the current appearance.

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,72 @@
...
+			padding: 1rem 1.25rem;
+		}
+	}
+
+	:global(.sl-markdown-content) .sl-link-button {
+		margin-inline-end: 0.5em;
+	}
+	:global(.sl-markdown-content) .sl-link-button:not(:where(p *)) {
+		margin-block: 1rem;
+	}
```
delucis, line 71: This makes me a bit nervous, but I guess it’s OK 😅 

I don’t love the context-specific styling, but I also thought about it a bit and can see that without this, the default behaviour would feel pretty broken. Other solutions I can think of are also annoying, so we should probably go with this.

docs/src/content/docs/guides/components.mdx
```diff
@@ -222,6 +222,37 @@ import { LinkCard } from '@astrojs/starlight/components';
...
+</LinkButton>
+```
+
+The above code generates the following on the page:
+
+import { LinkButton } from '@astrojs/starlight/components';
+
+<LinkButton href="/getting-started/">Get started</LinkButton>
+<LinkButton href="https://docs.astro.build" variant="secondary" icon="external">
+	Related: Astro
```
delucis, line 253: Noting that this renders as `<a ...><p>Related: Astro</p></a>`

Not sure if there’s anything we could or should do about that? It’s not invalid HTML and I guess doesn’t really have any bad side effects.

delucis (commented):
Thanks again for the PR @HiDeoo and apologies for the belated review — I left some comments!

> how about links without any text and only an icon? Is that something you wanted to support in this?

Regarding this comment from @lorenzolewis, I think it’s OK not to support that for now. I’m not 100% sure what the intended usage was for icon buttons in Figma and they complicate things somewhat — icon-only buttons still need an accessible label etc. Also, might be possible to work around for someone who really needs it:

```jsx
<LinkButton href="/astro" icon="astro">
  <span class="sr-only">Astro</span>
</LinkButton>
```

packages/starlight/style/markdown.css
HiDeoo: Thanks, this makes sense, in this case this was a link color only change but there is no guarantee this would stay the same any way. I think in this case, we should be able to just specify the `color` in the various selectors, e.g. `.sl-link-button.primary`, `.sl-link-button.secondary`, and `.sl-link-button.minimal`.

packages/starlight/user-components/LinkButton.astro
HiDeoo: There was already a `outline-offset: 0.25rem;` introduced in this PR for the minimal variant to avoid having the outline being on top of the text. We can just move this to the base `.sl-link-button` selector and this would fix it for all buttons.

packages/starlight/user-components/LinkButton.astro
HiDeoo: No idea, and this does not even make sense 😅 The best I can come up with would be me testing something at the time randomly and forgetting to remove it. I think we should just be able to go back to the `line-height: 1.1875;` that was used before.

packages/starlight/user-components/LinkButton.astro
HiDeoo: Well, I don't think this is technically true. Current ones are `56px` (`40px` on mobile) if you only use primary and minimal variants. As soon as you introduce the secondary variant, the height becomes `58px` (`42px` on mobile). This PR makes the heights consistent across all variants to avoid the sizing inconsistencies you described in the previous comment. Altho, I picked `58px` as the height for all variants, but we can definitely change this to `56px` if you think that would be better but wanted to make sure you are aware of this before I make the change.

delucis: Ah yeah, good point. Yes, let’s standardise to 56px — the [design reference in Figma](https://www.figma.com/design/GrDkcguAR7tSWPFuLVDixq/Theme---Starlight-Docs?node-id=171-506&t=Mo8nQ4EThzphRaE4-4) is also 56px for all variants (thanks to Figma’s utopian “inside” stroke 😁).

HiDeoo:
Updated the branch and push a few changes that should address all the great feedback, thanks for the review 🙌 

I also temporarily added back the demo page, it was useful to me to get back into the context of the PR so thought I might as well include it in case it can help and I'll remove it later (also unticked the associated todo in the PR body).

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,76 @@
...
+
+	@media (min-width: 50rem) {
+		.sl-link-button {
+			font-size: var(--sl-text-base);
+			padding: 0.9375rem 1.25rem;
+		}
+	}
+
+	:global(.sl-markdown-content) .sl-link-button {
+		margin-inline-end: 0.5em;
```
delucis, line 71: Your demo page was super helpful!

A bit subjective, but what do you think of increasing this slightly? The flex lock-up used on the landing page has a full `2rem` gap between buttons. I tried that here and it’s too much given the context with less vertical whitespace, but I think `1em` does work OK

| 0.5em | 1em |
|---|---|
| image | image |

It’s particularly helpful for the minimal variant, which felt like it was a bit squished up too close to the other links otherwise.

```suggestion
		margin-inline-end: 1em;
```

packages/starlight/user-components/LinkButton.astro
```diff
@@ -0,0 +1,76 @@
...
+	}
+	:global(.sl-markdown-content) .sl-link-button:not(:where(p *)) {
+		margin-block: 1rem;
```
delucis, line 74: Also, is there a reason you opted for `rem` on the vertical axis, but `em` on the horizontal?

delucis (commented):
Thanks for the updates @HiDeoo — left a couple more comments now I had the preview page to play with, but I think it looks pretty close to being ready 🙌

packages/starlight/user-components/LinkButton.astro
HiDeoo: None at all, probably a typo 🙈 Fixed it in the latest commit.

packages/starlight/user-components/LinkButton.astro
HiDeoo: I quite like it too on desktop. On mobile, I'm less confident but I pushed the change and tweak if needed on mobile depending on what you think.

docs/src/content/docs/reference/link-button.mdx
delucis, line 1: Time to delete our demo page friend!

delucis (commented):
I think we’re ready as far as I’m concerned — thank you!

Re: spacing on mobile vs desktop, I see what you mean, but I still think it will look OK in context. One of the slightly tricky things about the demo page is that there are buttons _everywhere_, but hopefully in reality people will only be using one or two 😁

delucis (approved):
I declare this `<LinkButton>` seaworthy — anchors away! 🚢

PR #1788, opened 2024-04-26
Add translation for copy button

dragomano:
Add Russian translation for Copy button on docs site


delucis (commented):
Thank you for the contribution @dragomano! I will actually move these across to Starlight’s core package so all users can benefit.

delucis (approved):

PR #1811, opened 2024-05-01
Fix `<Tabs>` sync issue with inconsistent use of `icon` on `<TabItem>` components

HiDeoo:
#### Description

This PR fixes an issue with synced tabs when `<Tabs>` sharing the same `syncKey` have `<TabItem>` components that inconsistently use `icon` or not.

The issue is due to the fact that labels (which need to be identical for tabs to sync) were coming from the `textContent` property on the tab link which gets the content of all elements. In the presence of an icon, the `textContent` property could potentially include extra spaces due to the presence of the SVG icon. 
This pull request fixes the issue by using the `innerText` property instead of `textContent` to get the label text as it only returns the rendered text content, which is the exact label text we want to compare.


delucis (approved):
Thanks for the fix @HiDeoo! LGTM 🚀

PR #1812, opened 2024-05-01
Add missing `siteTitle` property to Starlight pages route data

HiDeoo:
#### Description

This PR adds to Starlight pages route data the missing `siteTitle` property introduced in #1620.

It also adds a new test to ensure the root shape of the route data for a Starlight page is matching the one from a regular page. I'm not particularly happy with the test, and should ultimately be catched by properly running type-checking on the project but until then, it's better than nothing and would have caught this issue.

### DIFF
--- a/.changeset/six-ravens-tap.md
+++ b/.changeset/six-ravens-tap.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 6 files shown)


delucis (approved):
Looks good to me! Thanks for fixing this so quickly @HiDeoo 🙏

PR #1825, opened 2024-05-02
i18n(ru): update translations

dragomano:
Updates some translations:

- ru/getting-started.mdx
- ru/guides/components.mdx
- ru/guides/css-and-tailwind.mdx
- ru/guides/customization.mdx
- ru/guides/pages.mdx
- ru/reference/configuration.mdx
- ru/reference/frontmatter.md
- ru/reference/overrides.md
- ru/resources/community-content.mdx


docs/src/content/docs/ru/guides/i18n.mdx
```diff
@@ -148,16 +143,42 @@ Starlight предполагает, что вы создадите эквива
...
+    starlight({
+-     title: 'Моя документация',
++     title: {
++       en: 'My Docs',
++       ru: "Моя документация",
++     },
+      defaultLocale: 'ru',
+      locales: {
+        en: { label: 'English' },
+        ru: { label: 'Русский', lang: 'ru' },
```
HiDeoo, line 167: I think in this case, it would better to keep the `zh-CN` example and replace `en` by `ru`. This is what we've been doing so far in other translations and the goal is to keep an example with a BCP 47 tag with a regional subtag as they can be tricky with `zh-CN` vs `zh-cn`.

packages/starlight/translations/ru.json
HiDeoo, line 1: Would you mind removing these changes from this PR and moving them to a new PR?

To explain the rationale, these UI string translations are shipped to the users, so they would need a changeset so that the change is described to users in our changelog. Considering the size of the changes to this file, I think it would be easier to move them to a new PR so that we can merge the documentation changes in this PR right away and the UI string changes in a new one.

Let me know if that sounds ok to you.

HiDeoo (changes requested):
Thanks for another amazing translation :raised_hands:

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). Left 2 tiny suggestions, but everything else looks perfect to me.

HiDeoo (approved):
Amazing, thanks for the fast updates 👍

PR #1827, opened 2024-05-02
Add type checking job to the CI workflow

HiDeoo:
#### Description

Following a Discord conversation, this PR adds a type checking job to the CI workflow to catch type errors in the codebase even tho we are not bundling the code yet.

### DIFF
--- a/packages/tailwind/vitest.config.ts
+++ b/packages/tailwind/vitest.config.ts
@@ -4,11 +4,13 @@ export default defineConfig({
 	test: {
 		coverage: {
 			reportsDirectory: './__coverage__',
-			thresholdAutoUpdate: true,
-			lines: 100,
-			functions: 100,
-			branches: 100,
-			statements: 100,
+			thresholds: {
+				autoUpdate: true,
+				lines: 100,
+				functions: 100,
+				branches: 100,
+				statements: 100,
+			},
 		},
 	},
 });
--- a/packages/tailwind/vitest.config.ts
+++ b/packages/tailwind/vitest.config.ts
@@ -6,10 +6,10 @@ export default defineConfig({
 			reportsDirectory: './__coverage__',
 			thresholds: {
 				autoUpdate: true,
-				lines: 100,
+				lines: 94,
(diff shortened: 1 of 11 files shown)


HiDeoo:
The unit test failure is related to the Tailwind package which does not match the coverage thresholds expected. This used to pass before the configuration was invalid.

packages/tailwind/vitest.config.ts
```diff
@@ -4,11 +4,13 @@ export default defineConfig({
...
-			lines: 100,
-			functions: 100,
-			branches: 100,
-			statements: 100,
+			thresholds: {
+				autoUpdate: true,
+				lines: 100,
+				functions: 100,
+				branches: 100,
+				statements: 100,
```
delucis, line 12: I think it’s fine to drop these down to match the current state — we can aim to get back to full coverage later.

```suggestion
				lines: 94,
				functions: 100,
				branches: 85,
				statements: 94,
```

tsconfig.typecheck.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["**/dist/**", "**/__coverage__/**", "packages/starlight/components.ts"]
```
delucis, line 3: Could it make sense to exclude these in the default tsconfig instead of adding the custom one?

delucis (approved):
Thank you for tackling this @HiDeoo — looks great. Will be a nice way to catch details that have been slipping through recently.

Left a couple of small comments, but basically LGTM! 🌟

tsconfig.typecheck.json
HiDeoo: The issue is that we only want to exclude `packages/starlight/components.ts` during type-checking and not while working on Starlight with for example the Astro extension in Visual Studio Code.
`**/dist/**"` and `**/__coverage__/**` are already part of the default tsconfig but unfortunately, when using `extends` in a `tsconfig.json` file, arrays are not merged.

PR #1613, opened 2024-03-08
Add support for draft documentation pages

HiDeoo:
#### What kind of changes does this PR include?

- Changes or translations of Starlight docs site content
- Changes to Starlight code

#### Description

- Closes #990

This PR adds a new `draft` frontmatter option to exclude a page from production builds.

During development, draft pages can be accessed but a new notice will be displayed at the top of the page.

image

PR #1828, opened 2024-05-03
Enable `BASE_URL` tests

HiDeoo:
#### Description

Following the merge of #1613, this PR enables the `BASE_URL` tests we had in the codebase using a similar approach as the one used to stub the `MODE` to `production` in the tests.

### DIFF
--- a/packages/starlight/__tests__/basics/slugs.test.ts
+++ b/packages/starlight/__tests__/basics/slugs.test.ts
@@ -95,10 +95,7 @@ describe('urlToSlug', () => {
 		);
 	});
 
-	// It is currently not possible to test this as stubbing BASE_URL is not supported due to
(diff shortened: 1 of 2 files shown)


packages/starlight/__tests__/basics/slugs.test.ts
```diff
@@ -95,10 +95,7 @@ describe('urlToSlug', () => {
 		);
 	});
 
-	// It is currently not possible to test this as stubbing BASE_URL is not supported due to
```
HiDeoo, line 98: Note that this test does not need the whole `resetModules()` → `stubEnv()` → `import()` dance as contrary to `base.ts`, `BASE_URL` is not used/evaluated at the top-level scope of the module but in the scope of the function under test.

delucis (approved):
Thanks for remembering we had these stubbed out and realising these tests could now work! Lovely work @HiDeoo 💖

PR #1831, opened 2024-05-06
Improve type checking job

HiDeoo:
This PR is a follow-up to #1827 as I forgot that `astro check` checks all the files from a TypeScript project and not just Astro ones. The changes switch from `tsc` to `astro check` in the type checking job:

- We get the same type checking as before plus Astro files
- We no longer need to exclude `components.ts` so we can entirely remove the `tsconfig.typecheck.json` file

This PR also fixes issues reported by `astro check` in the Astro files and also addresses the warnings emitted so that we have a clean type checking job:

- The EC related change is just replacing a deprecated import by its new one ([reference](https://github.com/expressive-code/expressive-code/blob/21fdaa441c89d6a6ac38f5d522b6b60741df2f5d/packages/%40expressive-code/core/src/helpers/ast.ts#L3-L4))
- Regarding the Tailwind test related one, did not even know TS emitted something for this (and this [cannot be disabled](https://github.com/microsoft/TypeScript/issues/28825)) but I guess, meh, doesn't hurt and at least the output is clear…


.gitignore
```diff
@@ -18,4 +18,7 @@ pnpm-debug.log*
 __coverage__/
 
 # Vercel output
-.vercel
\ No newline at end of file
+.vercel
+
+# Created by @astrojs/check
+/src/
```
delucis, line 24: Huh, do you know why `@astrojs/check` is outputting stuff? Might be a bug.

delucis (approved):
Looks great @HiDeoo! Thank you. Have one query, but looks good.

.gitignore
HiDeoo: The file is `src/env.d.ts` which contain `/// <reference types="astro/client" />` so I would guess that `astro check` would run something similar to `astro sync` from where it's invoked which would generate this file at the root in our case.

But that's a guess so we may need to ask for a more definitive answer.

HiDeoo: This is indeed the case: [reference](https://discord.com/channels/830184174198718474/878341752799494154/1237151910385942691)

PR #1838, opened 2024-05-07
Add more diagnostic help to error messages thrown by `<Steps>`

delucis:
#### Description

- Follow up to some feedback from @yanthomasdev that it was hard to figure out **which** `<Steps>` component was erroring in builds

- The `<Steps>` component now logs the HTML it receives as part of the hints in error messages:

   Astro error overlay showing the new error message including HTML

- This is still relatively verbose and not 1:1 with what is authored because `<Steps>` receives the HTML rendered by Astro instead of the raw Markdown an author wrote, but is hopefully still helpful for locating which instance of `<Steps>` is causing issues.

#### Notes

- @HiDeoo also suggested using a global remark plugin to check structure and be able to error with proper positional information in the authored Markdown. This could indeed be nicer, but would not work in Markdoc and would add extra overhead to all content processing, vs. this approach which only runs the error logic on the children of `<Steps>`.

- We can consider potential refinements to this output in the future if we hear from people that this verbose logging is too overwhelming.

### DIFF
--- a/.changeset/breezy-cats-taste.md
+++ b/.changeset/breezy-cats-taste.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Adds extra information to the errors thrown by the `<Steps>` component to help locate misformatted code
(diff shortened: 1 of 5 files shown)


HiDeoo (approved):
Played with a few broken scenarios locally and even on an Astro docs branch with a few steps errors, and it definitely helps a lot already imo. Great improvement 🌟

PR #1845, opened 2024-05-09
Add social share component

ibdul:
#### Description
This PR includes additions to allow the user share the page they are viewing with social buttons
before:
Screenshot from 2024-05-09 14-55-36

after:
Screenshot from 2024-05-09 15-00-36
Screenshot from 2024-05-09 15-01-32

### DIFF
--- a/docs/src/content/docs/reference/overrides.md
+++ b/docs/src/content/docs/reference/overrides.md
@@ -330,6 +330,12 @@ The default implementation uses the page’s [`banner`](/reference/frontmatter/#
 
 Layout component used to wrap sections of the main content column.
 
+#### `SocialShare`
+
+**Default component:** [`SocialShare.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/SocialShare.astro)
+
+Component that renders the Social Share buttons
+
 #### `PageTitle`
 
 **Default component:** [`PageTitle.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/PageTitle.astro)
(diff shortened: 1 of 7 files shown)


ibdul:
I must have forgotten to remove the docs label. it is a core update, has nothing to do with the docs (except that the changes show up there as well, hehe)

HiDeoo:
:wave: Thanks for this pull request and your contribution to Starlight.

We usually follow a specific process for contributions/new features to Starlight starting from a [discussion](https://github.com/withastro/starlight/discussions) to understand a problem or feature, identifying the best solution, and then only implementing it. This helps us ensure that the feature is a good fit for Starlight and that it is implemented in a way that is consistent with the rest of the project.
In the future, always feel free to open a [discussion](https://github.com/withastro/starlight/discussions) before starting to work on a pull request to see if your idea is something that might make sense as part of Starlight or is better as external content. We’re always happy to provide guidance and you can also get feedback from the community.

Regarding the specific of this pull request, I don't think this a common enought pattern to include the feature in Starlight itself, also considering that some documentation authors may not want such a sharing feature in their documentation.

On the other hand, I think this would make a great third-party plugin that users can install if they want such feature to share their documentation. If you're interested in creating such a resource, a good starting point could be this [plugin](https://github.com/HiDeoo/starlight-image-zoomhttps://github.com/HiDeoo/starlight-image-zoom) adding a zoom feature to documentation images.

We will also be happy to share such a resource with the community using our [plugins and integrations showcase](https://starlight.astro.build/resources/plugins/) or [community content page](https://starlight.astro.build/resources/community-content/) for any guides you might create.

PR #1846, opened 2024-05-09
Update @astrojs/mdx to v3

delucis:
#### Description

- Updates the version of `@astrojs/mdx` bundled with Starlight to the new v3 release, which comes with several performance improvements
- Enables the `optimize` option to MDX for even faster builds.
- Also updates the peer dependency version of Astro, which is now ~~^4.8.0~~ ^4.8.6
- Taking the Starlight docs as a benchmark, it looks like builds are roughly 40% faster for us. Will be curious to see how things improves for other people!

Some other small changes I needed to make:

- ~~Updated Vitest to 1.4.0 — there were some compatibility issues with the upgraded Vite and the older Vitest version we were using. (Did not update to latest Vitest as that was also causing another issue we can investigate later.)~~ Superseded by #1861 
- Fixed a type issue due to a content collection types change — based on HiDeoo’s change here: https://github.com/withastro/starlight/pull/1841/files/c193d55f0046553772bfa540f15d956290f258e4#r1593616510
- ~~Ignored some more `.d.ts` files being generated during tests~~ Turned out that was only happening with Vitest 1.6.0 — reverted.


HiDeoo (approved):
Tested a few scenarios locally and could not find any specific issue, this is looking like a pretty amazing update to me :star2:

ematipico (approved):
Great! Looking forward to the next release

bluwy (approved):
Missed this PR. Looking forward to this!

delucis:
Given Astro 4.8 seems to break script hoisting MDX and this forces an upgrade to 4.8, we _might_ need to hold off merging this until https://github.com/withastro/astro/issues/11063 is fixed?

bluwy:
I've sent https://github.com/withastro/astro/pull/11084 that fixes it.

delucis:
Thank you for the quick fix @bluwy! 💖

PR #1852, opened 2024-05-11
i18n(zh-cn): Update `components.mdx` translation

hexWars:
An error will occur when the indentation is incorrect.

#### Description

- Closes # null
- What does this PR change? Some minor formatting errors.


HiDeoo (approved):
Great catch, thanks for the fix, really appreciated :raised_hands:

PR #1857, opened 2024-05-13
 i18n(tr): update ui translations

tarikcoskun:
#### Description

Updated missing UI translations for Turkish.


packages/starlight/translations/tr.json
```diff
@@ -3,7 +3,7 @@
   "search.label": "Ara",
   "search.shortcutLabel": "(Aramak için / bas)",
   "search.cancelLabel": "İptal",
-  "search.devWarning": "Arama yalnızca üretim derlemelerinde kullanılabilir. \nYerel bilgisayarınızda test etmek için siteyi derleyin ve önizleme yapın.",
+  "search.devWarning": "Arama yalnızca canlı sürümde kullanılabilir. \nYerel bilgisayarınızda test etmek için siteyi derleyin ve önizleme yapın.",
```
delucis, line 6: Are you sure the previous version is incorrect?

In English the text here is “Search is only available in production builds” and at least according to Google Translate that is what the current version says vs “live version”.

packages/starlight/translations/tr.json
```diff
@@ -18,11 +18,11 @@
...
   "page.nextLink": "Sonraki",
-  "page.draft": "This content is a draft and will not be included in production builds.",
-  "404.text": "Sayfa bulunamadı. URL'i kontrol edin ya da arama çubuğunu kullanmayı deneyin.",
-  "aside.note": "Note",
-  "aside.tip": "Tip",
-  "aside.caution": "Caution",
-  "aside.danger": "Danger",
-  "fileTree.directory": "Directory"
+  "page.draft": "Bu içerik taslaktır ve canlı sürümde bulunmayacaktır.",
+  "404.text": "Sayfa bulunamadı. Bağlantıyı kontrol edin ya da arama çubuğunu kullanmayı deneyin.",
```
delucis, line 22: I’m guessing this should also maybe not change, and reflect the English text which uses “URL” rather than “link”?

```suggestion
  "404.text": "Sayfa bulunamadı. URL'i kontrol edin ya da arama çubuğunu kullanmayı deneyin.",
```

delucis (commented):
Thank you for your PR @tarikcoskun! I had a couple of questions about two translations you changed.

Maybe @erkanaltinors would like to take a look as someone who as helped us with Turkish translations in the past.

packages/starlight/translations/tr.json
beyazit: If the goal is to translate the word meaning exactly without considering the context, the previous version is correct. However, it conveys the wrong message. The new translation is more understandable as slang used by Turkish developers.

packages/starlight/translations/tr.json
erkanaltinors: We can keep as suggested change from @delucis. "URL" is used in Turkish as abbreviation, too. "Bağlantı" refers to "link" word so they have different meanings.

packages/starlight/translations/tr.json
erkanaltinors: There is no translation option for "production builds" in Turkish to complete the meaning perfectly. So we should consider the sentences together to keep meaning as much as possible.

Second sentence points out the "preview" stage. "Üretim derlemeleri" defines the stage while "canlı sürüm" refers to result and ignores the "preview" stage.

delucis: Thank you for the feedback @erkanaltinors! Sounds like we should stick with the current translation in this case.
```suggestion
  "search.devWarning": "Arama yalnızca üretim derlemelerinde kullanılabilir. \nYerel bilgisayarınızda test etmek için siteyi derleyin ve önizleme yapın.",
```

delucis (approved):
Thank you again for contributing these translations @tarikcoskun! Really appreciate it.

Thanks also to @erkanaltinors for reviewing 💖

PR #1861, opened 2024-05-14
Update Vitest to `1.6.0`

HiDeoo:
~~**This PR is targetting #1846.**~~

#### Description

This PR updates the Vitest version used in the monorepo to the latest version instead of the initial `1.4.0` bump due to a test failure.

**Why was this test failing when upgrading to at least Vitest `1.5.0`?**

- Vitest `1.5.0` introduced a [change](https://github.com/vitest-dev/vitest/pull/5476) to the workspace feature current working directory (`cwd`) making it the config directory.
  - Using version `1.4.0`, the current working directory would be `packages/starlight/`
  - Using version `1.5.0`, the current working directory would be `packages/starlight/__tests__/i18n-non-root-single-locale/` for the `i18n-non-root-single-locale` test suite.
- Our tests use `getViteConfig()` which will internally call [`resolveConfig()`](https://github.com/withastro/astro/blob/ad9227c7d1474881fac9b1db15aa7b5a888b42b8/packages/astro/src/config/index.ts#L37) which calls [`resolveRoot()`](https://github.com/withastro/astro/blob/ad9227c7d1474881fac9b1db15aa7b5a888b42b8/packages/astro/src/core/config/config.ts#L190) with an `undefined` argument which would end up using [`process.cwd()`](https://github.com/withastro/astro/blob/ad9227c7d1474881fac9b1db15aa7b5a888b42b8/packages/astro/src/core/config/config.ts#L73C35-L73C48).
- With Astro now using `packages/starlight/__tests__/i18n-non-root-single-locale/` as the root, `packages/starlight/__tests__/i18n-non-root-single-locale/src/content/i18n/fr.json` would be automatically loaded when we call `getCollection()` which was not the case before.

**Why is only the `i18n-non-root-single-locale` test suite failing and not the `i18n` test suite which uses a similar setup?**

- The `i18n` test suite is not impacted because the default locale is `en` but is configured with a `lang` using a regional subtag (`en-US`) which means the user config overrides needs to be in for `en-US` and thus not being impacted by `en.json` being loaded.

**How is this PR fixing the issue?**

- This PR updates the `i18n-non-root-single-locale` test suite to use the `fr-CA` for the `lang` of the default locale to use the same setup as the `i18n` test suite.

**What alternatives were considered?**

- Using Astro 4.8.0 `getViteConfig()` new second argument to specify the `root` in the inline Astro config to fallback to the previous behavior and avoid using `process.cwd()` but I think this is wrong to just assume the CWD is not the root of the test suite.
- Moving out the `packages/starlight/__tests__/i18n-non-root-single-locale/translations-fs.test.ts` to another dedicated suite (where we do not mix real CC files and CC mocking). I think this is a valid one but we would need I think to do the same for the `i18n` test suite which is not failing. I think it's better to have a consistent setup for all the test suites but I'm open to feedback on this.

**Why are the `.gitignore` changes in this PR required?**

By Astro now using the proper root, and `getViteConfig()` somehow triggering someting close to `astro sync`, we end up with all the extra files in each test suites instead of just `packages/starlight/` like before when this folder was considered the CWD.

### DIFF
--- a/packages/starlight/.gitignore
+++ b/packages/starlight/.gitignore
@@ -1,2 +1,4 @@
 # Astro generates this during tests, but we want to ignore it.
 src/env.d.ts
+__tests__/**/env.d.ts
+__tests__/**/types.d.ts
--- a/packages/starlight/__tests__/i18n-non-root-single-locale/config.test.ts
+++ b/packages/starlight/__tests__/i18n-non-root-single-locale/config.test.ts
@@ -2,7 +2,7 @@ import config from 'virtual:starlight/user-config';
 import { expect, test } from 'vitest';
 
 test('test suite is using correct env', () => {
-	expect(config.title).toMatchObject({ fr: 'i18n with a non-root single locale' });
+	expect(config.title).toMatchObject({ 'fr-CA': 'i18n with a non-root single locale' });
 });
 
 test('config.isMultilingual is false with a single locale', () => {
@@ -11,7 +11,7 @@ test('config.isMultilingual is false with a single locale', () => {
 });
 
 test('config.defaultLocale is populated from default locale', () => {
-	expect(config.defaultLocale.lang).toBe('fr');
+	expect(config.defaultLocale.lang).toBe('fr-CA');
 	expect(config.defaultLocale.dir).toBe('ltr');
 	expect(config.defaultLocale.locale).toBe('fr');
 });
--- a/packages/starlight/__tests__/i18n-non-root-single-locale/routing.test.ts
+++ b/packages/starlight/__tests__/i18n-non-root-single-locale/routing.test.ts
@@ -19,11 +19,11 @@ test('route slugs are normalized', () => {
 test('routes for the configured locale have locale data added', () => {
 	for (const route of routes) {
 		if (route.id.startsWith('fr')) {
-			expect(route.lang).toBe('fr');
+			expect(route.lang).toBe('fr-CA');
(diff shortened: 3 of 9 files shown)


delucis (approved):
Amazingly thorough investigation and documentation of the fix @HiDeoo! This all makes sense to me and I’m happy to go with this approach as you suggested.

One question: do you think it makes most sense to merge this into the MDX v3 branch or is it better to target `main` and then I can rebase the MDX v3 branch on top of these changes?

HiDeoo:
> One question: do you think it makes most sense to merge this into the MDX v3 branch or is it better to target `main` and then I can rebase the MDX v3 branch on top of these changes?

I initially targeted the MDX v3 branch as it's the one that needed the Vitest update (due to Vite 5.1 changes with the Astro bump) but technically, I guess this could also makes sense to target `main` as it kinda ends up being an unrelated issue :thumbsup: Up to you, I'm fine with either way, let me know if you want me to update it so that it can target `main`.

delucis:
Let's target `main` then if you have time to do that — helps us avoid bundling up too many changes in one PR. Thank you!

HiDeoo:
This should be good, also pushed an empty commits to re-run all the workflows as I think there is still no event so that the workflows can be triggered after the base branch change.

delucis (approved):
LGTM! Thanks for cleaning that up @HiDeoo 🚀

PR #1862, opened 2024-05-14
docs: improve `<Icon>` documentation for CSS attributes

HiDeoo:
#### Description

- Closes #1856 and specifically this [comment](https://github.com/withastro/starlight/discussions/1856#discussioncomment-9404325)

This PR is an attempt to improve the documentation for the `<Icon>` component and making it more obvious that the component accepts for `size` and `color` any valid CSS value.


docs/src/content/docs/guides/components.mdx
```diff
@@ -419,19 +419,22 @@ import IconsList from '~/components/icons-list.astro';
...
 import { Icon } from '@astrojs/starlight/components';
 
 <Icon name="star" color="goldenrod" size="2rem" />
+<Icon name="rocket" color="var(--sl-color-accent)" />
 ```
 
 <Icon name="star" color="goldenrod" size="2rem" />
+<Icon name="rocket" color="var(--sl-color-accent)" />
```
delucis, line 437: Tiny nit, but in dark mode the blue accent has slightly low contrast so I might prefer the text accent colour:

````suggestion
<Icon name="rocket" color="var(--sl-color-text-accent)" />
```

<Icon name="star" color="goldenrod" size="2rem" />
<Icon name="rocket" color="var(--sl-color-text-accent)" />
````

delucis (approved):
Thanks for clarifying this @HiDeoo!

One thing on my mind recently as we expand components is that we don’t have a standard, thorough way to document the props and prop types for our components. One thought I’ve had is breaking these up into one page per component to give a bit more space to have a dedicated props section listing them with detailed descriptions.

HiDeoo:
> One thing on my mind recently as we expand components is that we don’t have a standard, thorough way to document the props and prop types for our components. One thought I’ve had is breaking these up into one page per component to give a bit more space to have a dedicated props section listing them with detailed descriptions.

Definitely, specially with the new components coming in. I just tried locally having all the new components documentation and the page is getting fairly big. This would also allow us to have more code snippets/examples for more props.
Maybe some page would be shorter like `<Steps>` but I think this would still be fine.

Do you already have something in mind for this? Maybe a new "Components" folder?

delucis:
> Do you already have something in mind for this? Maybe a new "Components" folder?

Yes, I think probably a Components group in the sidebar with an initial “Using components” guide followed by a page for each built-in component. The other option _might_ be a nested “Built-in components” group below the general guide that is collapsed by default?

And then maybe research some design system documentation practices, e.g. each page structure might look like:

1. Brief 1–2 sentence description of the component
2. Visual example of the component / code (maybe inside a specially designed container to make it visually distinct)
3. Props reference

PR #1871, opened 2024-05-15
Add BlueSky social icon

delucis:
#### Description

- Closes #1853 
- Adds an icon for the BlueSky social network
- Also adds it as a social link option
- I decided to capitalise “Sky” in the API and accessible label to reflect the common understanding of the brand name as two words — “blue sky” — even though it appears canonically they use the unusual “Bluesky” casing (see [some discussion on Mastodon](https://elk.zone/m.webtoo.ls/@swithinbank/112445218129925762)).


at-the-vr (approved):
Tested with GitPod, works 💯 LGTM 👍

PR #1874, opened 2024-05-15
Feat: Autogenerate sidebar labels

lorenzolewis:
#### Description

- Resolves https://github.com/withastro/starlight/discussions/1630
- Now using a sidebar entry with the key `slug` to specify the file will dynamically pull in that file's title
- ~~Update any relevant Starlight docs entries to move their translations out of the config and to the respective content files~~ These will be done in a separate follow-up PR to not trigger i18n

#### TODOs

- [x] Do the test updates look alright? I added on to the current tests instead of breaking out into a separate one, but can do that if wanted
- [x] Changeset
- [x] Docs updates


andrii-bodnar:
Wow! Very nice feature, it will greatly improve the DX!

I also faced this and it feels uncomfortable especially when the sidebar contains autogenerated groups and some manually added items. In this case, the autogenerated pages define their sidebar title in the Front Matter section. At the same time, the manual items should define it in the sidebar configuration.

This is also bad for i18n, because the labels for manually added items should be localized separately using the `translations` configuration.

Waiting for this to ship! 🚀

delucis (approved):
Thanks again for kicking this off @lorenzolewis! Sorry it took a long time for me to get to this. I hope you don’t mind too much that I ended up diving in and taking over this PR, but my brain really needed to be in the code to get to grips with it, so sorry for not providing a helpful review.

Quick overview of some of my changes:

- Tweaked the docs in a few places
- Added a shorthand version of this, so for the simple case `sidebar: ['slug-one', 'slug-two', 'slug-three']` is all you need, without the `{ slug: 'slug-one' }` object syntax.
- Added a bunch more tests for some edge cases
- Faffed about with naming variables 😁 
- Split out the error messages so that the part about leading/trailing slashes only displays when a user actually provided a slug with leading/trailing slashes
- Fixed up a test config so that i18n tests for the sidebar were working correctly
- Used `slug` instead of `link` in our starter templates

One thing I’m still thinking about is if we need to rejig the sidebar guide some more. Now that this new format is available, I’d say that using `slug` is probably now the **recommended** path for internal links as it will also help you catch broken links. I’ve moved the internal links docs up top and used `slug` or the shorthand for all examples of internal links for now.

Nevertheless, awesome work! I’d been saying for ages we should do this and then you went and did it 👏

lorenzolewis:
Thanks for diving in, @delucis , and feel free to run wild with any of my PRs anytime 😂

RE: usage in the sidebar guide, I think it's covered pretty well with what you've written. My only though overall for documentation is that it may be helpful to specify this is for files in `src/content` as "internal" could also be considered for thinks like assets, public files, and non-content collection pages.

delucis:
> My only though overall for documentation is that it may be helpful to specify this is for files in `src/content` as "internal" could also be considered for thinks like assets, public files, and non-content collection pages.

Yes, I also considered that. That’s why I used “internal **docs** pages”, but agree that being more explicit could make sense! Let me have a quick try.

lorenzolewis:
LGTM!

kmalloy24:
I was just looking for this functionality this morning, perfect timing, and great work! Thank you 🙏

PR #1877, opened 2024-05-16

jsparkdev:
#### Description

- update `components.mdx`
- #1862


delucis (approved):
Thanks for the update @jsparkdev — looks good to me! 🚀

PR #1881, opened 2024-05-16
Fix `.pnpm-store` in devcontainer setup

### LINKED ISSUE #1880: `.pnpm-store` in workspace directory

*

*

pnpm

VSCode Development Container

*

### Describe the Bug

Not sure if this is the correct place for this issue.

To reproduce:
- Clone the repository locally
- Open it in VSCode
- Click on "Reopen Repository in Container"

In the development container is configured to install `pnpm` and run `pnpm install` automatically as a `postCreateCommand`.

When running in a VSCode devcontainer, we cannot access the user's local pnpm-store so a `.pnpm-store` directory is created in the workspace. Since this is not ignored, git wants to commit the `.pnpm-store`

https://github.com/avanderbergh/starlight

### Participation

avanderbergh:
#### Description

Closes #1880 

This moves the pnpm store to the home directory so that it does not appear in the workspace.
It also added the `--frozen-lockfile` flag to the `pnpm install` in the "postCreateCommand" to avoid generating a new lockfile.


HiDeoo:

Sharing a few thoughts based on me quickly trying the changes:

- It looks like our devcontainer setup fails to work for me on my current setup (or maybe for all Apple Silicon?) due to https://github.com/puppeteer/puppeteer/issues/6622 but I did not get the time to debug this issue yet (this can be temporarily workaround for testing purposes using `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1`)
- I don't think the `--frozen-lockfile` should be necessary in this case, and an updated lockfile sounds like an issue. It looks like this is due to a pnpm version mismatch and the package.json `packageManager` field not being respected. I wonder if it would be possible and maybe easier to switch to corepack for the pnpm installation and ensure we get the expected version?

avanderbergh:
Hey @HiDeoo,

Yeah, I was also thinking about the mismatched "packageVersion".

So, instead of updating pnpm in the `Dockerfile`, I just uninstall it there, and then enable the corepack version in the `postCreateCommand` where the `package.json` can be read.

HiDeoo:
Sorry for the delayed answer and I wanted to do a bit of research regarding best practices for this, what are other repos doing, etc.

From what I can see, most repos actually just `.gitignore` the `.pnpm-store` directory, e.g. [Astro](https://github.com/withastro/astro/blob/main/.gitignore#L17), [Nuxt](https://github.com/nuxt/cli/blob/main/.gitignore#L6), etc.
Altho, for me this solution is still not enough to get an error-free devcontainers.

After a few experimentations, I came up with the following `devcontainer.json` and `Dockefile` that seems to work well for me:

```dockerfile
FROM node:18

COPY welcome-message.txt /usr/local/etc/vscode-dev-containers/first-run-notice.txt
```

```json
{
  "name": "Contribute to Starlight",
  "build": {
    "dockerfile": "Dockerfile"
  },
  "postCreateCommand": "corepack enable pnpm && pnpm config set store-dir /home/node/.pnpm-store && PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 pnpm install",
  "waitFor": "postCreateCommand",
  "customizations": {
    "codespaces": {
      "openFiles": ["CONTRIBUTING.md"]
    },
    "vscode": {
      "extensions": ["astro-build.astro-vscode", "esbenp.prettier-vscode"]
    }
  }
}
```

To summarize the changes:

- Changed the base image to `node:18`: not quite sure how the devcontainers base image is setup, but I'm not a huge fan of having to do `sudo corepack …`
- Moved the `store-dir` to the root of the `node` user home directory, not sure we need to replicate the default nesting
- Added `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1` before the `pnpm install` command.
  - This skips the download of the Chromium browser that Puppeteer needs. This is a bit annoying but I think it may be at the moment the best solution to install on all platforms without any issues.
  - I tried to use an [approach similar to Astro](https://github.com/withastro/astro/blob/main/.devcontainer/Dockerfile#L9-L11) to install Chrome, which seems to be the recommended approach, but I couldn't get it to work for me, the Chrome download URL leads to a 404 with my architecture.
  - Considering Puppeteer is only needed for the a11y tests, I think it's fine to skip the download for now when using a devcontainer and re-evaulate this when we have a better solution.

What do you think of these changes? Would you be able to test them on your end and see if they work for you as well?

avanderbergh:
Hi @HiDeoo,

Thanks for checking that 🙏.

## Base image

The configuration for the `devcontainer` base image can be found here: https://github.com/devcontainers/images/tree/main/src/javascript-node/.devcontainer

**Dockerfile**
```Dockerfile
ARG VARIANT=22-bookworm
FROM node:${VARIANT}

ARG USERNAME=node
ARG NPM_GLOBAL=/usr/local/share/npm-global

# Add NPM global to PATH.
ENV PATH=${NPM_GLOBAL}/bin:${PATH}

RUN \
    # Configure global npm install location, use group to adapt to UID/GID changes
    if ! cat /etc/group | grep -e "^npm:" > /dev/null 2>&1; then groupadd -r npm; fi \
    && usermod -a -G npm ${USERNAME} \
    && umask 0002 \
    && mkdir -p ${NPM_GLOBAL} \
    && touch /usr/local/etc/npmrc \
    && chown ${USERNAME}:npm ${NPM_GLOBAL} /usr/local/etc/npmrc \
    && chmod g+s ${NPM_GLOBAL} \
    && npm config -g set prefix ${NPM_GLOBAL} \
    && su ${USERNAME} -c "npm config -g set prefix ${NPM_GLOBAL}" \
    # Install eslint
    && su ${USERNAME} -c "umask 0002 && npm install -g eslint" \
    && npm cache clean --force > /dev/null 2>&1

# [Optional] Uncomment this section to install additional OS packages.
# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
#     && apt-get -y install --no-install-recommends <your-package-list-here>

# [Optional] Uncomment if you want to install an additional version of node using nvm
# ARG EXTRA_NODE_VERSION=10
# RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"

# [Optional] Uncomment if you want to install more global node modules
# RUN su node -c "npm install -g <your-package-list-here>"
```

So it sets up nvm and configures permissions for npm global installs.

Here's the config for that base image.

**devcontainer.json**
```jsonc
{
	"name": "Node.js",
	"build": {
		"dockerfile": "Dockerfile"
	},
	"features": {
		"ghcr.io/devcontainers/features/common-utils:2": {
			"installZsh": "true",
			"username": "node",
			"upgradePackages": "true"
		},
		"ghcr.io/devcontainers/features/node:1": {
			"version": "none"
		},
		"ghcr.io/devcontainers/features/git:1": {
            "version": "latest",
            "ppa": "false"
        }
	},

	// Configure tool-specific properties.
	"customizations": {
		// Configure properties specific to VS Code.
		"vscode": {
			// Add the IDs of extensions you want installed when the container is created.
			"extensions": [
				"dbaeumer.vscode-eslint"
			]
		}
	},

	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	// "forwardPorts": [],

	// Use 'postCreateCommand' to run commands after the container is created.
	// "postCreateCommand": "yarn install",

	// Set `remoteUser` to `root` to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
	"remoteUser": "node"
}
```

I notice in "features" it includes [common-utils](https://github.com/devcontainers/features/tree/main/src/common-utils), [node](https://github.com/devcontainers/features/tree/main/src/common-utils) & [git](https://github.com/devcontainers/features/tree/main/src/git).

In the [node feature install.sh]:380(https://github.com/devcontainers/features/blob/ecbfd50952e513db872d8d3380e069ccf74a70a8/src/node/install.sh#L380), I see is where pnpm is installed via `npm install -g pnpm`.

So, by using the `node:18` base image, we lose quite a lot, like `zsh`, latest `git`, non-root user, `node-gyp`, ...

We could of course, add those features to *our* devcontainer, but that would probably increase the build time and add extra complexity?

## Moved the store-dir

👍

## Added PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1

👍

---

I think `sudo corepack enable...` is an OK trade-off against always connecting to the container as `root`? 🤔

HiDeoo (approved):
Thanks for adding all this context. This is definitely a huge help as I'm personally not a devcontainers user and I was mostly seeing this from someone only using Docker to deploy applications.

Considering all this, I think you made the right call. I was even gonna ask if we needed to uninstall pnpm but you already linked to the devcontainer feature showing where they install it so everything makes sense. This new version will be a huge improvement for devcontainers users and it's not like we're deploying a Dockerfile in production, if we get more feedback, we can always iterate and improve it.

Thanks a lot for your contribution and patience with me understanding how all this works :raised_hands:

avanderbergh:
@HiDeoo Happy to be able to help out on such a great project! 🌟 🌟 🌟

PR #1885, opened 2024-05-17
css: move overflow wrap rule to `markdown.css`

julien-deramond:
⚠️ This PR is a draft

---

## Description

This PR is a follow-up of what's been discussed in https://github.com/withastro/starlight/pull/1736#pullrequestreview-2060043908

> I do wonder if our generic set-up for this in reset.css ought really also to be moved to markdown.css:
> [...]
> In a quick check, I think the only element this currently applies to which is outside of the Markdown content is the page title.
> But that can happen as follow-up work to keep this PR focused on the specific fix.

Basically, we move the `p ,h1, h2, h3, h4, h5, h6, code { overflow-wrap: anywhere; }` rule to the `markdown.css`.

Temporarily, the red code shows the elements impacted by the previous rule, and the green one shows the elements impacted by the new rule.

### DIFF
--- a/packages/starlight/style/reset.css
+++ b/packages/starlight/style/reset.css
@@ -40,7 +40,7 @@ h4,
 h5,
 h6,
 code {
-	overflow-wrap: anywhere;
+	background-color: rgba(255, 0, 0, 0.2);
 }
 
 code {
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -1,5 +1,5 @@
 ---
-title: Getting Started
+title: Getting Started xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxx xxxx
 description: Learn how to start building your next documentation site with Starlight by Astro.
 ---
 
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -1,3 +1,15 @@
+.sl-markdown-content p,
+.sl-markdown-content h1,
+.sl-markdown-content h2,
+.sl-markdown-content h3,
+.sl-markdown-content h4,
+.sl-markdown-content h5,
+.sl-markdown-content h6,
+.sl-markdown-content code {
+	background-color: rgba(0, 255, 0, 0.2);
+	overflow-wrap: anywhere;
+}
+
 .sl-markdown-content
 	:not(a, strong, em, del, span, input, code)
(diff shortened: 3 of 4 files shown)


packages/starlight/style/reset.css
```diff
@@ -40,7 +40,7 @@ h4,
 h5,
 h6,
 code {
-	overflow-wrap: anywhere;
+	background-color: rgba(255, 0, 0, 0.2);
```
julien-deramond, line 43: I've been through the entire doc, the only elements that are displayed in red (old rule) are the page main titles, and the "On this page" area (see https://starlight-git-fork-julien-deramond-main-jd-dc421d-astrodotbuild.vercel.app/getting-started/).

Screenshot 2024-05-17 at 09 26 05

The rest in green works well.

One solution would probably be to keep that in `markdown.css` and add specific rules for this main title and "On this page" area. What do you think @delucis? Want me to try this approach in this PR?

trueberryless:
Hi 👋 

Sorry for the super late reply. It has been some time. You probably also don't remember that this PR existed 😅 

I investigated the current situation and revisited the idea of this PR. I think we should close this because in #3491 we changed the value of `overflow-wrap` from `"anywhere"` to `"break-word"` which together with the idea of this PR would break multiple sections which are temporarily marked with a red background. To give some examples, here are some comparisons:

⛔ undesired behaviour
✅ desired behaviour

| Description  | overflow in `markdown.css` ⛔ | overflow in `reset.css` ✅ |
|---|---|---|
| Page Title  | image  | image  |
| Table of Contents  | image  | image  |
| Hero section (no actual difference)  | image  | image  |

I guess to continue the idea behind the PR, we could move the `overflow-wrap: break-word;` styles for Markdown in `markdown.css` and apply `overflow-wrap: break-word;` to the other items individually (title and table of content - hero section doesn't make a difference) to keep the styles structured. What do you think? Also pinging @delucis if you have time to take a look into this.

delucis:
Thanks for going through and checking those @trueberryless! On reflection — after quite some time has passed 😅 — I think maybe it’s OK to leave these styles where they are so they apply broadly? It probably doesn’t matter too much either way, so the easiest thing is to do nothing 😀 

Thanks again to @julien-deramond for kicking off the discussion — but I agree we can close this one for now.

julien-deramond:
Two years already 😅 Totally forgot about this one.

> It probably doesn’t matter too much either way, so the easiest thing is to do nothing 😀

Yeah, definitely. Thank you both!

PR #1892, opened 2024-05-17
Micro-optimization: clean up `<details>` CSS

delucis:
#### Description

- I realised we could simplify the CSS for the `<details>` element a little bit by avoiding to repeat the hover selector logic when setting the custom colours inside asides
- Also makes it a bit easier for customisation. For example, in Astro Docs I wanted to add [an extra custom colour](https://github.com/withastro/docs/pull/8322/commits/1cd6c34daf49e76c3cc1b0579c561d3b2e8db3cd) where we have `<details>` in a purple container. Currently that requires duplicating (again) the `:hover` selectors. With this change it can be done just with the two custom properties.

### DIFF
--- a/.changeset/silly-tools-drop.md
+++ b/.changeset/silly-tools-drop.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Internal refactor: simplify some CSS for the `<details>` element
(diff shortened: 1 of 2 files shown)


HiDeoo (approved):
Excellent catch, looks like a great change to me 👍

PR #1899, opened 2024-05-19
i18n(ko-KR): update `authoring-content.md`

jsparkdev:
#### Description

- update `authoring-content.md`
- #1735


delucis (approved):
Thank you for the update @jsparkdev! Checked the page and it looks like it’s rendering as expected 🎉

PR #1900, opened 2024-05-19
Add `starlight-multi-sidebar` to `plugins.mdx`

lorenzolewis:
#### Description

- Add https://github.com/lorenzolewis/starlight-multi-sidebar to `plugins.mdx`

PR #1907, opened 2024-05-20

jsparkdev:
#### Description

- update `plugins.mdx`
- #1900


delucis (approved):
Thank you @jsparkdev! So cool that we’re keeping the Korean docs 100% updated 💖

PR #1913, opened 2024-05-21
Use `directory`-style path formatting with `build.format: 'preserve'`

### LINKED ISSUE #1781: Starlight is incompatible with the Astro build.format preserve option

0.21.5

4.6.3

pnpm

Mac

Chrome

### Describe the Bug

when I add the options 
```
    trailingSlash: 'always',
    build: {
      format: 'preserve',
    },
```
to an existing otherwise working starlight project, it stops working with the error
```
13:16:42 [ERROR] Cannot read properties of undefined (reading 'addBase')
  Stack trace:
    at formatPath (/home/projects/withastro-starlight-7ibewh/node_modules/@astrojs/starlight/utils/createPathFormatter.ts:25:25)
    [...] See full stack trace in the browser, or rerun with --verbose.
```

https://stackblitz.com/edit/withastro-starlight-7ibewh?file=astro.config.mjs

### Participation

delucis:
#### Description

- Fixes #1781
- Astro v4.3 added the `build: { format: 'preserve' }` config option, which we haven’t been handling up until now. 
- I tested using this option and Starlight still produces files like `guides/example/index.html` for all its pages. Given this, it seemed safest to treat this the same way we treat `build: { format: 'directory' }` and render links as `guides/example/`.

### DIFF
--- a/.changeset/curvy-pots-ring.md
+++ b/.changeset/curvy-pots-ring.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes support for Astro’s `build: { format: 'preserve' }` configuration option
--- a/packages/starlight/utils/createPathFormatter.ts
+++ b/packages/starlight/utils/createPathFormatter.ts
@@ -12,15 +12,18 @@ interface FormatPathOptions {
 	trailingSlash?: AstroConfig['trailingSlash'];
 }
 
+const defaultFormatStrategy = {
+	addBase: pathWithBase,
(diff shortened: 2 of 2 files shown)


HiDeoo (approved):
To be perfectly honest, I don't think I've ever used `preserve` but from the docs, your explanations and the few tests I've done locally, I think the change makes sense to me :thumbsup:

delucis:
Yeah. Maybe one day we should really have a Starlight `linkFormat` option or something, but as far as I can tell it makes no difference for injected routes like ours.

PR #1970, opened 2024-06-06

jsparkdev:
#### Description

- update `components.mdx`
- #1530

The badge's content has not been intentionally translated to convey its exact meaning.

### DIFF
--- a/docs/src/content/docs/ko/guides/components.mdx
+++ b/docs/src/content/docs/ko/guides/components.mdx
@@ -418,6 +418,36 @@ import { Steps } from '@astrojs/starlight/components';
 
 </Steps>
 
+### 배지
+
+import { Badge } from '@astrojs/starlight/components';
+
+상태나 라벨과 같은 작은 정보를 표시하려면 `<Badge>` 컴포넌트를 사용하세요.
+
+표시하려는 콘텐츠를 `<Badge>` 컴포넌트의 `text` 속성에 전달하세요.
+
+기본적으로 배지는 사이트의 테마 강조 색상을 사용합니다. 내장 배지 색상을 사용하려면 `variant` 속성의 값을 `note` (파란색), `tip` (보라색), `danger` (빨간색), `caution` (주황색), 'success'(녹색) 중 하나로 설정하세요. .
+
+`size` 속성 (기본값: `small`)은 배지의 텍스트 크기를 제어합니다. 더 큰 배지를 표시하기 위해 `medium` 및 `large` 옵션도 사용할 수 있습니다.
+
+맞춤 설정이 더 필요하다면 사용자 정의 CSS와 함께 `class` 또는 `style`과 같은 다른 `<span>` 속성을 사용하세요.
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/ko/guides/components.mdx
```diff
@@ -418,6 +418,36 @@ import { Steps } from '@astrojs/starlight/components';
...
 
+### 배지
+
+import { Badge } from '@astrojs/starlight/components';
+
+상태나 라벨과 같은 작은 정보를 표시하려면 `<Badge>` 컴포넌트를 사용하세요.
+
+표시하려는 콘텐츠를 `<Badge>` 컴포넌트의 `text` 속성에 전달하세요.
+
+기본적으로 배지는 사이트의 테마 강조 색상을 사용합니다. 내장 배지 색상을 사용하려면 `variant` 속성의 값을 `note` (파란색), `tip` (보라색), `danger` (빨간색), `caution` (주황색), 'success'(녹색) 중 하나로 설정하세요. .
```
HiDeoo, line 429: Just to double check, is the ". ." at the end expected or should it be "." only?

docs/src/content/docs/ko/guides/components.mdx
```diff
@@ -418,6 +418,36 @@ import { Steps } from '@astrojs/starlight/components';
...
+<Badge text="Starlight" variant="note" size="large" />
+<Badge text="Custom" variant="success" style={{ fontStyle: 'italic' }} />
+```
+
+위의 코드는 페이지에 다음을 생성합니다.
+
+<Badge text="New" variant="tip" size="small" />
+<Badge text="Deprecated" variant="caution" size="medium" />
+<Badge text="Starlight" variant="note" size="large" />
+<Badge text="Custom" variant="success" style={{ fontStyle: 'italic' }} />
```
HiDeoo, line 449: Note that the `text` can be safely translated, would it makes sense to do so in this case or it's better to keep them in English?

Totally your call, you know what's best so let me know.

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). I left 2 small questions but except for that, everything looks good to me.

docs/src/content/docs/ko/guides/components.mdx
jsparkdev: Thanks for the good point. It has been modified.

docs/src/content/docs/ko/guides/components.mdx
jsparkdev: I intentionally did not translate it as written in the text.

I could have translated the message on the badge, but I think it's better not to.

HiDeoo: Perfect, thanks for the clarification 👍

HiDeoo (approved):
Thanks for the fast update, looks perfect now :rocket:

PR #1958, opened 2024-06-05
Add kudos option to add “Built with Starlight” badge to site footers

delucis:
#### Description

- Adds a new config option to enable a “Built with Starlight” badge in the footer of your site

PR #1973, opened 2024-06-06
i18n(ko-KR): update `configuration.mdx`

jsparkdev:
#### Description

- update `configuration.mdx`
- #1958 
- #1530

### DIFF
--- a/docs/src/content/docs/ko/reference/configuration.mdx
+++ b/docs/src/content/docs/ko/reference/configuration.mdx
@@ -212,7 +212,8 @@ type SidebarItem = {
 ```ts
 interface BadgeConfig {
 	text: string;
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/ko/reference/configuration.mdx
```diff
@@ -571,3 +572,13 @@ starlight({
 ```
 
 나만의 플러그인을 만드는 방법에 대한 자세한 내용은 [플러그인 참조](/ko/reference/plugins/)를 확인하세요.
+
+### `credits`
+
+사이트 바닥글에 “Built with Starlight” 링크 표시를 활성화합니다.
```
HiDeoo, line 578: Just adding a note that the text could be updated later when the [associated UI string](https://github.com/withastro/starlight/blob/main/packages/starlight/translations/ko.json#L28) is translated.

PR #1976, opened 2024-06-06
i18n(ko-KR): update `configuration.mdx` and `ko.json`

jsparkdev:
#### Description

- update `configuration.mdx` and `ko.json`
- Newly added UI string translation in `ko.json` file
- Configuration update using that UI string


jsparkdev:
Thanks for the good point!

PR #1977, opened 2024-06-06
i18n(fr): improve root locale related content translation

HiDeoo:
#### Description

While reviewing another PR, I noticed that the French version of the documentation related to a root locale was confusing.

It was mostly referred to as "racine locale" which sounds like a "root" (the part of a plant) that would be originating from a specific place (the location). The change proposed here is to use "locale racine" instead, which avoids the confusion and is also consistent with the [version](https://starlight.astro.build/fr/reference/configuration/#locale-racine) already used in the configuration reference.


delucis (approved):
Happy you got to the `root` of these docs 💖

PR #1993, opened 2024-06-10
Fix invalid Astro i18n config with a single root locale

### LINKED ISSUE #1975: Status 404 for existing pages

0.24.1

4.10.1

pnpm

Windows

Firefox

### Describe the Bug

Startlight 0.23.4:
sshot-10

Starlight 0.24.0/0.24.1:
sshot-17

In both cases, the pages exist and are displayed.

https://stackblitz.com/edit/github-ztcekb

### Participation

HiDeoo:
#### Description

- Closes #1975

This PR fixes an issue in the generated Astro i18n config when using a single root locale. The issue boils down to 2 problems:

- In our schema, we branch when we have a multilingual site or monolingual site with only 1 non-root locale. Altho, the existing check (`configuredLocales.length >= 1`) was not singling out a website with a single root locale.
- In the Astro i18n config generation process, when the config `locales` is not defined (which is the case when for a site with only 1 root locale configured), the locales was always being the built-in default locale instead of the configured `defaultLocale`.

To test these changes, I added an entirely new test suite with a configuration using only a single root locale which is one suite we were missing.


delucis (approved):
Thank you for the fix @HiDeoo!

PR #1994, opened 2024-06-11
i18n(es): update `authoring-content`

dreyfus92:
#### Description

Updates `authoring-content` based on this commit 1a9ab50d458d6274994ffe66a23fe7a30681337a


docs/src/content/docs/es/guides/authoring-content.md
```diff
@@ -360,6 +360,31 @@ Un título opcional del bloque de código se puede establecer con un atributo `t
   ```
   ````
 
+## Details
```
HiDeoo, line 363: Just to double check, is it expected to have the same heading as the English version or should this be translated too like the other headings?

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). Left 1 tiny question but except for that, everything looks perfect to me.

docs/src/content/docs/es/guides/authoring-content.md
dreyfus92: Yeah, as far as I know it is. I went to MDN's docs for it. Here is the [example]( https://developer.mozilla.org/es/docs/Web/HTML/Element/details)

HiDeoo (approved):
Thanks for the clarification :thumbsup:

PR #2004, opened 2024-06-12
Remove outdated Badge component export

liruifengv:
#### Description

Badge component is user component now. Remove outdate Badge component export


.changeset/green-files-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Remove outdate Badge component export
```
delucis, line 5: Thanks for catching this @liruifengv! I think we can tweak this a little bit for clarity:

```suggestion
Removes an outdated export in `package.json`
```

PR #1121, opened 2023-11-23
Feat: Add sidebar state persistence

kitschpatrol:
#### Description

- Closes #971
- Sidebar now persists its scroll position and expanded / contracted section state across page navigations. It also scrolls the active sidebar item into view if necessary.

Builds on an initial implementation by @duncanwerner.

I grabbed the most salient bits of the issue thread below:

#### Feature criteria

- Respect the default "collapsed" state of the sidebar specified by SidebarItem config on first page navigations.
- Retain user's sidebar scroll position across page navigations.
- Retain user's sidebar sub-section detail expand / collapsed states across page navigations.
- Expand all ancestor detail elements of the active item (regardless of user's sidebar state changes).
- Ensure that the active item is visible after a page navigation (regardless of user's sidebar state changes).
- No Flash of Un-Scrolled Content (FLOUSC).

#### Possible improvements

- [ ] Track the initial expand / collapse state (from SidebarItem config) on first load before sessionStorage is set, and then apply this as the default if users don't actively change expand / collapse state of sections. (For example, the page is defined with all sections collapsed, the user navigates through several sections using the "next" buttons, this would collapse sections in the sidebar behind them once they're out of that section.)
- [ ] Maybe consider an intersection observer instead `getBoundingClientRect()` to determine if scrolling the sidebar is necessary. (Note that trusting `scrollIntoView()` on its own without checking bounds first can result in unnecessary scrolling.)
- [ ] Allow some padding when auto-scrolling a sidebar item into view. Currently, the sidebar scrolls the minimum amount necessary, leaving the item right on the edge of the scroll area.

#### Still to do
- [ ] Check implications of possible [view transitions](https://github.com/withastro/starlight/pull/694) feature.
- [x] Check implications of possible [multiple sidebar](https://github.com/withastro/starlight/discussions/959) feature.
- [ ] Strip comments. (I left these in to clarify any discussion around the PR, I assume we'll want to strip these since inline scripts aren't processed.)
- [ ] Verify that JS used meets [browserlist criteria](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+88%2C+Edge+%3E%3D+88%2C+Firefox+%3E%3D+98%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all).

PR #2012, opened 2024-06-13
Ensure link to current page is visible in sidebar

zadeviggers:
#### Description

When navigating sites with long sidebars (like the Astro docs), it can be quite disorientating when you click on a link and suddenly lose the context of where you are in the sidebar because it disappears. I see this behaviour in Firefox 126 and Chrome canary, so I assume it isn't just a browser quirk.

This PR solves the problem by adding a little script that is injected with the top-level sidebar. The script checks if the sidebar link to the current page is visible, and if it isn't, it is scrolled into view.

I have it set to jump instantly, but it could be jarring if the page loads really slowly. If this proves to be an issue, you could consider changing the behaviour to `"smooth"`. This can be annoying if you see it happen on every page load though.

It seems like Vite isn't processing the script right now, which is probably because of my conditional rendering of it. I  assume we want Vite to process it, but it could also be bad if it causes the JS to be loaded externally. I'd appreciate some guidance here.

### DIFF
--- a/packages/starlight/components/SidebarSublist.astro
+++ b/packages/starlight/components/SidebarSublist.astro
@@ -18,6 +18,7 @@ const { sublist, nested } = Astro.props;
 				{entry.type === 'link' ? (
 					<a
 						href={entry.href}
+						data-sl-current-page-link={entry.isCurrent}
 						aria-current={entry.isCurrent && 'page'}
 						class:list={[{ large: !nested }, entry.attrs.class]}
 						{...entry.attrs}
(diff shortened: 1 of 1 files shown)


delucis:
Thanks for the PR @zadeviggers — nice to see you again 👋 

As we just [chatted about on Discord](https://discord.com/channels/830184174198718474/1070481941863878697/1253015997053669406), I think we’ll probably take something more like the approach in #1121, which has been a bit long running due to my own slowness getting to it, but we’ll get to eventually! Given that, we can probably close this PR.

PR #2007, opened 2024-06-12
[i18nIgnore] Update dependencies

delucis:
#### Description

- Updates dependencies throughout the monorepo
- None of this is really user-facing. Some direct dependencies of the main Starlight package and the Tailwind package are updated, but only `hastscript` is upgraded to a version that was not already covered by our version range and that is only a tiny [non-breaking “major”](https://github.com/syntax-tree/hastscript/releases/tag/9.0.0).
- Main aim of this PR is just getting all our testing and development versions freshened up
- Due to the Prettier updates, there are some formatting changes. I’ve added those to this PR by running `pnpm format` locally to ensure we can `i18nIgnore` them for Lunaria and test any changes to code still work.
- #2009 complements this with some more updates

#### Majors
- `@astrojs/check`: 0.5.10 => 0.7.0
- `@astrojs/markdown-remark`: 4.2.1 => 5.1.0 (used for testing)
- `@astro-community/astro-embed-youtube`: 0.4.4 => 0.5.2
- `hastscript`: 8.0.0 => [9.0.0](https://github.com/syntax-tree/hastscript/releases/tag/9.0.0)
- `prettier-plugin-astro`: 0.13.0 => 0.14.0
- `starlight-links-validator`: 0.7.1 => 0.9.0
- `size-limit` & `@size-limit/file`: 8.2.4 => 11.1.4

#### Minors
- `astro`: 4.8.6 => 4.10.2
- `@astrojs/mdx`: 3.0.0 => 3.1.0
- `@astrojs/sitemap`: 3.0.5 => 3.1.5
- `@playwright/test`: 1.43.1 => 1.44.1
- `prettier`: 3.0.0 => 3.3.2
- `tsx`: 4.7.1 => 4.15.2

#### Patches
- `astro-expressive-code`: 0.35.2 => 0.35.3
- `start-server-and-test`: 2.0.0 => 2.0.4
- `tailwindcss`: 3.4.1 => 3.4.4
- `postcss`: 8.4.33 => 8.4.38
- `@types/hast`: 3.0.3 => 3.0.4
- `@types/mdast`: 4.0.3 => 4.0.4

PR #2026, opened 2024-06-18
Git ignore Playwright test results

HiDeoo:
#### Description

In #2007, Playwright was updated from the version `1.43.1` to the version `1.44.1`. The version [`v1.44.0`](https://github.com/microsoft/playwright/releases/tag/v1.44.0) adds a new `--last-failed` CLI option to run only tests that failed in the previous run. To support this feature, Playwright saves some state in `packages/starlight/test-results/.last-run.json`.

This PR adds the `test-results/` directory to the `.gitignore` file (which could technically have been done since we started using Playwright but we don't have any tests generating artifacts yet).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ pnpm-debug.log*
 # Vitest
 __coverage__/
 
+# Playwright
+test-results/
+
 # Vercel output
 .vercel
 


delucis (approved):
LGTM — thanks @HiDeoo!

PR #2031, opened 2024-06-18
Error for sidebar entries that match multiple entry schemas

### LINKED ISSUE #2019: Conflicting sidebar options are silently ignored

0.24.2

v4.10.2

pnpm (indifferent for this)

MacOS (indifferent for this)

Chrome (indifferent for this)

### Describe the Bug

The sidebar allows for either a link or a list of child items, not both. But if both are provided, the user is not informed that this is a problem, and the child items are simply ignored.

```ts
{
	label: 'Parent',
	link: '/parent',
	items: [
		{
			label: 'Child',
			link: '/parent/child',
		},
	],
}
```

The items won't show up on the sidebar, which might lead the user to try to find a problem within the items option.

https://stackblitz.com/edit/github-f4x149?file=astro.config.mjs

### Participation

delucis:
#### Description

- Closes #2019 
- Previously, a sidebar entry was parse loosely, e.g. `{ label: 'Foo', link: '/bar/', items: [] }` matched the `SidebarLinkItemSchema` (it had a `label` and a `link`) and was considered valid with `items` being ignored.
- This change makes those schemas strict so they may _only_ contain known keys, making the above example an error.

#### Considerations

- The Zod error map message is not _super_ helpful (you can see it in the new tests added), but does at least show the index of the item causing the issue.
- This might make extending the sidebar schema more difficult if people are doing that? IIRC you can’t do `z.object({}).strict().extend({ ... })`, because the `.strict()` is already baked in at that point. In practice, extending the sidebar schema may already be too difficult for anyone to be doing it? Still, would want some feedback on whether this is a change we definitely want, or if we’re happy with the loose parsing and to mark #2019 as working as expected.


HiDeoo:
> * This might make extending the sidebar schema more difficult if people are doing that? IIRC you can’t do `z.object({}).strict().extend({ ... })`, because the `.strict()` is already baked in at that point.

Aren't you thinking of  `transform()`, `refine()` and `superRefine()` which returns a type with `ZodEffects` which would require you to use `and()`? (which would also no longer be true in Zod 4)

Altho, I have to say I've never seen anyone even trying to extend this schema personally so I think the slightly better DX (knowing the index of the sidebar item with an issue is already a good addition imo) makes sense.

PR #2038, opened 2024-06-19
i18n(ru): translate credits

dragomano:
Update:

* `ru/reference/configuration.mdx`
* `packages/starlight/translations/ru.json`


delucis (approved):
Thanks @dragomano!

I also added a changeset so that the change to `ru.json` will be published in our next release.

PR #2054, opened 2024-06-26
Fix usage of `<StarlightPage>` with a custom `srcDir` configuration

HiDeoo:
#### Description

- Closes https://github.com/HiDeoo/starlight-blog/issues/60

When using a custom page with the `<StarlightPage>` component, the user docs schema is retrived using the virtual module `virtual:starlight/collection-config` which attempts to import `/src/content/config.ts`. Altho, this does not consider the user having a custom [`srcDir`](https://docs.astro.build/en/reference/configuration-reference/#srcdir).

In such cases, trying to build the project would result in the following error:

> [vite]: Rollup failed to resolve import "/src/content/config.ts" from "virtual:starlight/collection-config".

This PR fixes the issue by taking the `srcDir` into account when importing the user docs schema.

I could not imagine a proper way to test this using a unit test so I ended up adding an e2e one instead which would fail to even build the fixture project before the fix.

### DIFF
--- a/.changeset/tidy-brooms-complain.md
+++ b/.changeset/tidy-brooms-complain.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue when using the `<StarlightPage>` component in a custom page with a user-defined `srcDir` configuration.
--- a/packages/starlight/__e2e__/collection-config.test.ts
+++ b/packages/starlight/__e2e__/collection-config.test.ts
@@ -1,0 +1,12 @@
+import { expect, testFactory } from './test-utils';
+
+const test = await testFactory('./fixtures/custom-src-dir/');
+
+test('builds a custom page using the `<StarlightPage>` component and a custom `srcDir`', async ({
+	page,
+	starlight,
+}) => {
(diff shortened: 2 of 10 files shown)


delucis (approved):
LGTM! End-to-end test here is actually not a bad idea in general as a basic sanity check for `srcDir` support.

The only thing I’m wondering is if we should also be running E2E tests on Windows? I never remember with file path stuff what we’re supposed to do, so for something like this change, Windows E2E might add some confidence if they weren’t too slow.

HiDeoo:
> The only thing I’m wondering is if we should also be running E2E tests on Windows?

Yeah, I'll enable them.

This is also the approach we would have ended up with very soon anyway as the SSR PR includes atm some "functional" tests but it's just a few tests which does not really justify a new kind of tests. The idea will be to improve our E2E setup and  refactor them as E2E tests, using `@playwright/test` for the runner and just not ending up using a browser instance during these tests.

delucis:
Will merge this without the extra E2E tests — at worst it fixes stuff for users on some platforms and we can add the Windows E2E later. Thanks again @HiDeoo!

PR #2066, opened 2024-06-29
Run E2E tests on Windows

HiDeoo:
#### Description

This PR is a follow-up to [this conversation](https://github.com/withastro/starlight/pull/2054#pullrequestreview-2146233748) and enables E2E tests on Windows.


delucis (approved):
LGTM! Classic case of the same tests taking more than three times as long on Windows 😅

PR #2077, opened 2024-07-03
Tweak CI workflow

HiDeoo:
This pull request apply some changes to the CI GitHub Actions workflow.

First, it updates all Linux jobs to use `ubuntu-latest` as the base image as before this was a mixed bag of `ubuntu-20.04` and `ubuntu-latest`.

Second, it adds a new job using [`dorny/paths-filter`](https://github.com/dorny/paths-filter) (recommended by Chris, thanks a lot) to detect changes in various directories (`docs/` and `packages/` at the moment). Later on, this new job can be used a a dependency for other jobs to only run them when changes are detected in the specified directories.

Modified jobs:

- `unit-test`: updated to only run when changes are detected in the `packages/` directory.
- `e2e-test`: updated to only run when changes are detected in the `packages/` directory.
- `links`: updated to only run when changes are detected in the `docs/` directory.

Unmodified jobs:

- `type-check`: this runs on the entire monorepo and we can introduce an error from a change in the `docs/` or `packages/` directory.
- `pa11y`: we can introduce an accessibility error from a change in the documentation (a custom component for example) or in Starlight itself.
- `windows-smoke`: I guess we can break this from either the `docs/` or `packages/` directory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
     name: Run unit tests
     needs: changes
     if: ${{ needs.changes.outputs.packages == 'true' }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v3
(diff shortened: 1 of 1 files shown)


delucis (approved):
Looks great! Thanks @HiDeoo — will be nice to avoid some unneeded CI runs.

PR #2079, opened 2024-07-03
test: extend default `coverage.exclude`

AriPerkkio:
#### Description

  - Instead of manually copying Vitest's default values, use the default values from their public API: 
    - > This option overrides all default options. Extend the default options when adding new patterns to ignore:
    - https://vitest.dev/config/#coverage-exclude

  - Nope


delucis (approved):
Oh nice! I hadn’t realised this was available as an import. Thanks for making a PR 🌟

PR #2083, opened 2024-07-03
docs: add robots.txt for sitemap

ghostdevv:
#### Description

- ~~Closes # ~~

- What does this PR change?

Adds a note about `robots.txt` to the sitemap docs, and adds a robots.txt for the starlight docs.

Before:

image

After:

image


delucis (commented):
Thanks for the PR @ghostdevv! I think it could make sense to simply link to Astro’s docs for this, so we don’t have to duplicate things. I’ve opened https://github.com/withastro/docs/pull/8997 which will give us a better heading to link to and then can update this PR.

ghostdevv:
Updated @delucis

docs/src/content/docs/guides/customization.mdx
```diff
@@ -96,16 +96,7 @@ You can display different versions of your logo in light and dark modes.
...
-
-```js {4}
-// astro.config.mjs
-
-export default defineConfig({
-	site: 'https://stargazers.club',
-	integrations: [starlight({ title: 'Site with sitemap' })],
-});
-```
+Starlight has built-in support for sitemaps, you can read more about [how to enable it here](https://docs.astro.build/en/guides/integrations-guide/sitemap/#usage).
```
delucis, line 99: Thanks @ghostdevv! I should have been clearer, sorry, I think we can still keep the current “how to enable” docs given they are slightly different and the “add link in `<head>`” part is handled by Starlight.

So we can then add a link specifically for `robots.txt`, maybe something like this:

````suggestion
Starlight has built-in support for generating a sitemap. Enable sitemap generation by setting your URL as `site` in `astro.config.mjs`:

```js {4}
// astro.config.mjs

export default defineConfig({
	site: 'https://stargazers.club/',
	integrations: [starlight({ title: 'Site with sitemap' })],
});
```

Learn how to [add a sitemap link to `robots.txt`](https://docs.astro.build/en/guides/integrations-guide/sitemap/#sitemap-link-in-robotstxt) in the Astro Docs.
````

ghostdevv: IMO it feels a little weird to only link for one short instruction which isn't going to change which is why I assumed you meant to link for the entire thing - WDYT? Happy to change if that's the direction you want to go in

delucis: My aim here was to optimize for the most important information: a sitemap is important and setting `site` is needed to enable it, so I like highlighting that in the Starlight docs.

Listing the sitemap in `robots.txt` is not essential (Starlight already links to it in `<head>` for you), so linking for that information feels OK to me.

delucis (approved):
Thanks again for adding this @ghostdevv! 🎉

PR #2085, opened 2024-07-04
 how to disable lastUpdated global config for specific pages

sanabel-al-firdaws:
i don't know if this is needed or not tell me your thoughts


docs/src/content/docs/reference/frontmatter.md
```diff
@@ -206,6 +206,13 @@ title: Page with a custom last update date
 lastUpdated: 2022-08-09
 ---
 ```
+```md
+---
+# src/content/docs/example.md
+title: Page with disabled lastUpdated global config
```
delucis, line 212: I think this title might make more sense here. Using `lastUpdated: false` in a page would only really make sense if the global config is enabled.

```suggestion
title: Page with a hidden last-updated date
```

delucis (commented):
Thanks @sanabel-al-firdaws — I’m not opposed to adding this if this was something you found difficult to understand with the existing documentation. Left a small suggestion.

sanabel-al-firdaws:
> Thanks @sanabel-al-firdaws — I’m not opposed to adding this if this was something you found difficult to understand with the existing documentation. Left a small suggestion.

yub i don't think its necessary to add this ig people can figure it out easily

PR #2086, opened 2024-07-04
fix: avoid badges with `undefined` class

julien-deramond:
#### Description

In the actual Starlight documentation at https://starlight.astro.build/fr/getting-started/, the "New" badge in the sidemenu contains a class named "undefined".

Screenshot 2024-07-04 at 13 44 01

This PR suggests a possible fix in the `SidebarSublist` Astro component directly to create an empty string if the `entry.badge.class` var is undefined.

Please note that:
* I'm not too much familiar with Starlight's codebase, so maybe this fix must be done at another level in the code
* there's maybe another syntax that would be better for this fix

Don't hesitate to nuke this PR or update it directly if there's a better way to fix that. I can also just close it if the quality is not there, and just create an issue instead :)

### DIFF
--- a/.changeset/cyan-snails-laugh.md
+++ b/.changeset/cyan-snails-laugh.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Avoid badges with `undefined` class
--- a/packages/starlight/components/SidebarSublist.astro
+++ b/packages/starlight/components/SidebarSublist.astro
@@ -26,7 +26,7 @@ const { sublist, nested } = Astro.props;
 						{entry.badge && (
 							<Badge
 								variant={entry.badge.variant}
-								class={entry.badge.class ? entry.badge.class : ''}
+								class={entry.badge.class ?? ''}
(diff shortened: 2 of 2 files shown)


delucis:
Thanks @julien-deramond! I remember seeing this bug before and forgot about it.

It’s technically a more general Astro bug, so I opened this issue: https://github.com/withastro/astro/issues/11413

Ideally, we should fix it in Astro so that our usage can remain unchanged (because this bug could impact various Astro users, not only Starlight), but I’ll see what the feedback is on that issue. If it looks like we won’t be able to fix it upstream in a timely manner, we can use this PR in Starlight instead.

delucis:
Proposed a fix: https://github.com/withastro/compiler/pull/1014

delucis:
This should be fixed by https://github.com/withastro/astro.build/pull/1167 in the next release of the compiler.

Thanks again for bringing our attention to it!

julien-deramond:
Great job fixing it upstream! 💪

PR #2087, opened 2024-07-04
Add synced tabs persistence

HiDeoo:
Closes #1851 and follow-up to #953.

This PR adds persistence to the synced tabs feature by relying on `localStorage` to store the synced tabs state.

This pull request is still a draft for 2 reasons:

- There seems to be a flash of invalid tab that can randomly happen once in a while in development mode. I'd like to investigate this further and see if we can avoid it.
- I want to play a bit more with the feature as most of the time so far has been spent on when/how to include the script to restore the synced tabs rather than actually using the feature.


packages/starlight/user-components/Tabs.astro
```diff
@@ -9,8 +9,70 @@ interface Props {
...
+				const starlightTabs = this.closest('starlight-tabs');
+				if (!(starlightTabs instanceof HTMLElement) || typeof localStorage === 'undefined') return;
+				const syncKey = starlightTabs.dataset.syncKey;
+				if (!syncKey) return;
+				let label = '';
+				try {
+					const storedSyncedTabs = localStorage.getItem('starlight-synced-tabs') ?? '{}';
+					const persistedSyncedTabs = JSON.parse(storedSyncedTabs);
+					if (typeof persistedSyncedTabs !== 'object') return;
+					label = persistedSyncedTabs[syncKey];
```
delucis, line 55: Maybe I’m overthinking this but I’m thinking a bit about the storage model. This means there is a single JSON object in `localStorage` with state for all synced items and currently we re-read and re-parse it for every tab instance, right?

Almost feels like we should either:

1. Store this way but only read and parse once

2. Store each key separately (e.g. `localStorage.setItem('starlight-synced-tabs__' + syncKey, label)`) to make a single read simpler (less data per read/write, no JSON parsing)

I guess the second option would also simplify persistence because each `syncKey` only worries about its own state then — no need to handle reading and updating the full state object. The con is a “noisier” `localStorage` with potentially many keys. Would that be a concern do you think?

packages/starlight/user-components/Tabs.astro
```diff
@@ -9,8 +9,70 @@ interface Props {
...
+				} catch {}
+				if (!label) return;
+				const tabs = [...starlightTabs?.querySelectorAll('[role="tab"]')];
+				const tabIndexToRestore = tabs.findIndex((tab) => tab instanceof HTMLAnchorElement && tab.innerText === label);
```
delucis, line 59: I feel like we did discuss this once in the past, but would `textContent` also work for syncing tabs? `innerText` is style aware so is [much slower](https://www.measurethat.net/Benchmarks/Show/3618/0/createtextnode-vs-textcontent-vs-innertext-vs-innerhtml) than `textContent` and in the context of this inline restore script _might_ cause style recalc in a similar way to what we just fixed in #2155.

.changeset/silver-houses-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Adds persistence to synced `<Tabs>` so that user's choices are reflected throughout the entire website.
```
delucis, line 5: Suggesting a small tweak here:

```suggestion
Adds persistence to synced `<Tabs>` so that a user's choices are reflected across page navigations.
```

docs/src/content/docs/guides/components.mdx
```diff
@@ -91,7 +91,7 @@ The code above generates the following tabs on the page:
 
 Keep multiple tab groups synchronized by adding the `syncKey` attribute.
 
-All `<Tabs>` on a page with the same `syncKey` value will display the same active label. This allows your reader to choose once (e.g. their operating system or package manager), and see their choice reflected throughout the page.
+All `<Tabs>` with the same `syncKey` value will display the same active label. This allows your reader to choose once (e.g. their operating system or package manager), and see their choice persisted using [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and reflected throughout the entire website.
```
delucis, line 94: Do you think mentioning the implementation detail of local storage is important here? Might just be a distraction and behaviour is clear enough with this:

```suggestion
All `<Tabs>` with the same `syncKey` value will display the same active label. This allows your reader to choose once (e.g. their operating system or package manager), and see their choice persisted across page navigations.
```

delucis (commented):
Getting an early review in before we go through this on T&D on Thursday. Working great in the deploy preview for me @HiDeoo!

packages/starlight/user-components/Tabs.astro
HiDeoo: Thanks for the early review, super appreciated.

I don't think you're overthinking this at all, and it's actually part of the improvements I mentioned in my previous message. My initial thought so far was closer to your second example but I may try for curiosity in a page with a lot of synced tabs if the first one can make a difference or not (don't think so but still want to test it now).

packages/starlight/user-components/Tabs.astro
HiDeoo: We did discuss it in the context of #1811 where we had to switch from `textContent` to `innerText`. I'll try it for sanity, but I think using `textContent` is this context would have the same issue of #1811.

delucis: Ah, yes, thanks for remembering the context.

Maybe this test is too naive, but I just tried in the [Astro Embed docs](https://astro-embed.netlify.app/getting-started/) which uses your package manager tabs with icons. It looks like the difference is only in leading/trailing whitespace:

```
> tab.textContent
'   npm '
> tab.innerText
'npm'
```

In that case, I guess we could also solve it with `tab.textContent.trim()`?

HiDeoo: Forgot to mention, the recent changes are now all using trimmed `textContent` and I added some comments to make sure we remember why.

packages/starlight/user-components/Tabs.astro
```diff
@@ -159,26 +220,42 @@ const { html, panels } = processPanels(panelHtml);
...
 			if (shouldSync) {
 				newTab.focus();
-				StarlightTabs.#syncTabs(this, newTab.innerText);
+				const label = StarlightTabs.getTabLabel(newTab);
+				if (!label) return;
+				StarlightTabs.#syncTabs(this, label);
 				window.scrollTo({
 					top: window.scrollY + (this.getBoundingClientRect().top - previousTabsOffset),
 				});
+				this.#persistSyncedTabs(label);
```
delucis, line 229: Small nit that maybe it makes more sense to move this (and probably the `getTabLabel()`) inside `#syncTabs()` directly so that the logic is always `#syncTabs() => #persistSyncedTabs()` if we ever wanted to sync tabs from somewhere else for some reason, but it’s not crucial.

delucis (commented):
LGTM from an initial review! Look forward to seeing how testing goes tomorrow.

packages/starlight/user-components/Tabs.astro
HiDeoo: Makes total sense now with a more advanced prototype. Thanks again for the early review :raised_hands:

delucis: Refactor looks much more readable to me now too — nice work!

HiDeoo:
Based on the feedback during the T&D exploration, I've updated the PR:

- We're now using a global symbol to keep track in `Astro.locals` if the restore script has been rendered or not. The goal is to avoid potential clash with user-defined variables in `Astro.locals` (even if unlikely).
- I tried quite a few scenarios (embedded components, delayed components, etc.) to see if we could hit any issue with the restore script. So far, everything seems to work as expected no matter the scenario.

packages/starlight/user-components/Tabs.astro
```diff
@@ -9,8 +9,67 @@ interface Props {
...
+
+if (isSynced) {
+	// @ts-expect-error - See above
+	Astro.locals[didRenderSyncedTabsRestoreScriptSymbol] = true
+}
 ---
 
+{/* Inlined to avoid a flash of invalid active tab. */}
+{shouldRenderSyncedTabsRestoreScript && <script is:inline>
+	(() => {
```
delucis, line 44: Nit: whitespace will be preserved inside this inline script tag once rendered, so could maybe indent by one less tab?

(Also fantasising a bit about a home-made system to minify inline scripts — imagine we had an `inline-scripts/` directory that got loaded, minified, and available as a string each from a virtual module, e.g. `import { tabsRestore } from 'virtual:starlight-inline-scripts';`. Then these scripts could look like `<script is:inline set:html={tabsRestore} />` or something.)

delucis (approved):
LGTM! Left two nits, but don’t have any “real” comments left!

packages/starlight/user-components/Tabs.astro
HiDeoo: Quite the interesting idea, I would guess a bit easier to setup when we have a build system, but definitely interesting 🧠

HiDeoo: Actually, scratch that, I just quickly tried locally using the JS API of esbuild and a Vite plugin and it works, even supports watch mode. Altho, this makes me wonder, could Astro auto minify inline scripts (I guess there are some conditions to this) 🤔

delucis: Yeah, I’ve been chatting about that a bit with the team. Posted a proposal here: https://github.com/withastro/roadmap/discussions/995

PR #2101, opened 2024-07-06
Add passed style for the Table of Contents Links

jumpingpxl:
#### Description

This PR adds a `passed` attribute to the links of the Table of Contents. This allows for further distinction of where exactly on the page you are. 
Additionally, to prevent the links getting too dark, the "default" color has been changed from `--sl-color-gray-3` to `--sl-color-gray-2`.

I've added this because I am porting a page from MkDocs (Material) to Starlight, and this was something that I missed with Starlight.

Before:
Gif of how it was before

After: 
Gif of how it is now

### DIFF
--- a/.changeset/metal-melons-arrive.md
+++ b/.changeset/metal-melons-arrive.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Adds a `passed` style to the Table of Contents links, allowing to further distinct where on the page you are.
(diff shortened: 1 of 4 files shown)


delucis:
Hi @jumpingpxl! Thanks for the PR — I appreciate your interest in contributing.

I’m not 100% confident that we want to make this change. It is a cool effect, but I think it actually makes it less clear what the currently highlighted item is at a quick glance because the contrast between the current item and the items below it is quite low with the items lower on the page arguably being **more** prominent than the current item:

image

For the future, it’s always better to first open a [feature request discussion](https://github.com/withastro/starlight/discussions/categories/feature-requests) before creating a PR to chat through details of an idea and find out if there are other users who would appreciate a change. That also helps the team evaluate the need for work on something.

PR #2102, opened 2024-07-06
Update getting-started.mdx

adepauldami:
The use of asap for starting server [since it's happening almost immediately].


HiDeoo:
Thanks for your contribution.

I don't think adding the "asap" abbreviation is necessary in this specific case and something that we would want to change in this part of the documentation.

If you have any questions or feel like this section could be improved in any way, I think starting a discussion in the [Astro Discord server](https://astro.build/chat/) would be a good starting point to get feedback from the community and discuss the potential changes before submitting a PR.

PR #2105, opened 2024-07-06
Fixes an edge case in custom pagination link processing

### LINKED ISSUE #1886: Inconsistent Behavior of Links in Starlight | Frontmatter Prev and Next

0.22.4

4.8.6

pnpm

Windows

Brave

### Description

I've been experimenting with Starlight for a couple of days and have recently set up a project hosted on GitHub Pages. I've noticed that the behavior of the `Prev` and `Next` links is inconsistent and I'm unsure why this is happening.

### Details

The issue is that sometimes the links work as expected, and other times they don't. It seems that sometimes they need to be prefixed with the Astro config's `base` value, and other times they automatically prefix themselves.

To understand why the behavior changes, I created a fresh template install. In this template, I have three files in an auto-generated directory. Each file uses the exact same link. The link in `three.md` works, but the links in `one.md` and `two.md` do not.

In my actual documentation, I've tried prefixing the frontmatter links with the base value. Everything works with this prefixing, except for one file. I've also noticed odd differences where a link works in one test run and not in another. It seems whether the link is treated as relative or absolute, and whether it needs the base as a prefix, changes unpredictably.

### Steps to Reproduce

1. Create a fresh Starlight template.
2. Create a directory at `/src/content/docs/myfiles` within the new project.
3. Add three files (`one.md`, `two.md`, `three.md`) and add them to the sidebar config as auto-generated directory.
4. Use the same link in each file.

```js
---
title: one
sidebar:
    order: 1
next:
    label: example ref
    link: reference/example/
---

---
title: two
sidebar:
    order: 2
next:
    label: example ref
    link: reference/example/
---

---
title: three
sidebar:
    order: 3
next:
    label: example ref
    link: reference/example/
---
```

6. Observe that the link in `three.md` works, but the links in `one.md` and `two.md` do not.

### Expected Behavior

The links should consistently work across all files and should either all require manual prefixing with the base value or non of the links require the prefixing.

### Actual Behavior

The links work inconsistently. Sometimes they require manual prefixing with the base value, and other times they do not.

I believe this is not the intended behavior, hence I'm opening this issue. Any help or guidance would be greatly appreciated.

https://stackblitz.com/edit/github-rzkjm9?file=astro.config.mjs

### Participation

delucis:
#### Description

- Closes #1886
- Fixes an edge case how custom prev/next links in frontmatter are handled
- In general a custom prev/next link was applied directly to an existing link object, which produced the correct behaviour. How in the edge case where a link object wasn’t already available (e.g. pagination is disabled, or the page is first/last, so has no prev/next link), we were using a `makeLink()` utility designed for sidebar items that did some additional processing. This rarely caused issues, but if a user had `base` set, this led to inconsistencies where the base would prepended to links **only** in these rare edge cases.
- This PR fixes this by ensuring links created for pagination purposes are never processed and always reflect 1:1 with the authored link in frontmatter
- We had an old test that expected the processing, which converted `x` to `/x`, but I think that was a wrong expectation, so I updated that old test to reflect the new behaviour.

### DIFF
--- a/packages/starlight/__tests__/basics/pagination-with-base.test.ts
+++ b/packages/starlight/__tests__/basics/pagination-with-base.test.ts
@@ -1,0 +1,76 @@
+import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import type {
+	getPrevNextLinks as getPrevNextLinksType,
+	getSidebar as getSidebarType,
+} from '../../utils/navigation';
+
+vi.mock('astro:content', async () =>
+	(await import('../test-utils')).mockedAstroContent({
+		docs: [
+			['index.mdx', { title: 'Home Page' }],
+			['environmental-impact.md', { title: 'Eco-friendly docs' }],
+			['guides/authoring-content.md', { title: 'Authoring Markdown' }],
+			['guides/components.mdx', { title: 'Components' }],
+			['reference/frontmatter.md', { title: 'Frontmatter Reference' }],
+		],
+	})
+);
+
+describe('without base', async () => {
+	let getPrevNextLinks: typeof getPrevNextLinksType;
+	let getSidebar: typeof getSidebarType;
+	let sidebar: ReturnType<typeof getSidebar>;
+
+	beforeAll(async () => {
+		({ getPrevNextLinks, getSidebar } = await import('../../utils/navigation'));
+		sidebar = getSidebar('/reference/frontmatter/', undefined);
+	});
+
+	test('pagination links are formatted correctly with no frontmatter', () => {
+		const links = getPrevNextLinks(sidebar, true, {});
+		expect(links.prev?.href).toMatchInlineSnapshot(`"/guides/components/"`);
+		expect(links.next?.href).toMatchInlineSnapshot(`undefined`);
+	});
+
+	test('pagination links are formatted correctly with custom links in frontmatter', () => {
+		const links = getPrevNextLinks(sidebar, true, {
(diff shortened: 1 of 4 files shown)


packages/starlight/__tests__/basics/pagination-with-base.test.ts
```diff
@@ -0,0 +1,76 @@
...
+		sidebar = getSidebar('/reference/frontmatter/', undefined);
+	});
+
+	test('pagination links are formatted correctly with no frontmatter', () => {
+		const links = getPrevNextLinks(sidebar, true, {});
+		expect(links.prev?.href).toMatchInlineSnapshot(`"/guides/components/"`);
+		expect(links.next?.href).toMatchInlineSnapshot(`undefined`);
+	});
+
+	test('pagination links are formatted correctly with custom links in frontmatter', () => {
```
HiDeoo, line 35: Tiny thought: should the test name not mention "formatted correctly" but more something like "used as authored" to emphasize we're not formatting the link in this case?

HiDeoo (approved):
Having a new function not involving `formatPath()` definitely seems like the proper solution to fix this issue. The updated test also makes sense to me.

Thanks for fixing this :raised_hands:

packages/starlight/__tests__/basics/pagination-with-base.test.ts
delucis: Good point — updated and also switched away from snapshots to hardcoded values as I don’t think these should really change.

PR #2111, opened 2024-07-07
Update ar.json

sanabel-al-firdaws:


sanabel-al-firdaws:
google uses the word ملاحظة insted of ملحوظة the second word i don't see alot

delucis (approved):
Thanks!

PR #2122, opened 2024-07-10
Fix an Astro i18n → Starlight i18n config conversion issue

HiDeoo:
#### Description

This PR fixes the issue identified in this [comment](https://github.com/withastro/starlight/pull/1923#issuecomment-2213744449).

When using Astro’s `i18n` config for a multilingual site with a root locale (`prefixDefaultLocale` set to `false`), the `config.defaultLocale.locale` value was not being set correctly to `undefined` as expected in this case when converting the config to a Starlight i18n config.

This PR fixes the issue by checking for this specific case. No new tests are added as we already have tests covering this scenario but the expected Starlight i18n config output was incorrect so we just needed to update it.

Another potential, but more manual test, is to use the Astro’s `i18n` config from #1923 in the documentation and visiting a non-existent page.

### DIFF
--- a/.changeset/large-bananas-beg.md
+++ b/.changeset/large-bananas-beg.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an i18n configuration issue for multilingual sites when using Astro’s `i18n` config with `prefixDefaultLocale` set to `false`.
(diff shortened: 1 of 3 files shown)


delucis (approved):
Great work figuring that out @HiDeoo! Thank you 💖

PR #2126, opened 2024-07-12
Allow asides titles to contain markdown formatting

essential-randomness:
#### Description

- Closes #1915 and https://github.com/withastro/docs/issues/8711
- **What does this PR change?** 

Fixes a bug where `aside` titles would get cut out if any markdown formatting was added in them.  

Markdown of aside with formatting:
image
Before the fix
image
After the fix
image

I took the liberty to add `mdast-util-to-string` as a dependency. You already had `hast-util-to-string` so I assumed this would be fine and better than writing bespoke code to handle nested formatting. Let me know if you want me to remove it.


HiDeoo:
Thanks a lot for your contribution :raised_hands:

I did not yet dive in detail into the changes, altho they seems pretty straightforward at first glance (except maybe the lock file changes, looks like quite a few deps got updated).

One thing I'd like to point out and discuss before moving forward is the [`<Aside/>`](https://starlight.astro.build/guides/components/#asides) component. This would be a difference in behavior/features between the directive syntax and the component syntax.

I guess there are several possibilities to handle this, e.g. only support this for the directive syntax like in the current implementation, or also support it for the component syntax. I think I'm personally leaning towards the latter, but I'd like to hear people's opinions on this.

.changeset/mean-tigers-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Allow asides titles to contain markdown formatting
```
delucis, line 5: Small suggestion for the changeset base on our [style guide](https://contribute.docs.astro.build/docs-for-code-changes/pull-requests/#changesets)

```suggestion
Adds support for markdown formatting in aside titles
```

delucis (commented):
Thanks @essential-randomness! This looks great.

Regarding @HiDeoo’s question about what to do about `<Aside>`, that’s a great question!

My initial thought was “we can just require HTML there and use `set:html`” — i.e. usage like

```mdx
<Aside title="Custom <code>code</code> label">
```

But then we’ll still need to strip out the HTML for the `aria-label`. That may still be preferable to supporting full-on Markdown rendering there though?

Also curious whether people think it’s OK to merge this before fixing support in `<Aside>`?

HiDeoo:
> That may still be preferable to supporting full-on Markdown rendering there though?

I think so, this would align with the banner behavior for example and would be way more practical than let's say a slot.

> Also curious whether people think it’s OK to merge this before fixing support in `<Aside>`?

I personally think it's fine to merge this first, my comment was more about pointing out the potential difference and discuss it rather than having it block this PR.

delucis (approved):
Thanks again for the PR @essential-randomness! Happy with this and then we can think about how to tackle `<Aside>` support in a similar vein if and when it is needed.

PR #2137, opened 2024-07-22
Deleting unnecessary twitter meta tags, quick fixes

cevdetardaharan:
#### Description

1. Deleting "twitter:*" meta elements in favor of using "og:*" as fallback.
2. Deleting "initial-scale=1" because of it's default value.
3. Reducing code size.

These changes won't make any difference in visual perspective, but they will reduce file size in this project and every project that uses Starlight.

##### Resources
Twitter and OG meta elements:
1. https://developer.x.com/en/docs/twitter-for-websites/cards/guides/getting-started
2. https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup

`initial-scale=1`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag
> Controls the zoom level when the page is first loaded. Minimum: 0.1. Maximum: 10. Default: 1. Negative values: ignored.


packages/starlight/components/Head.astro
```diff
@@ -18,7 +18,7 @@ const headDefaults: z.input<ReturnType<typeof HeadConfigSchema>> = [
 	{ tag: 'meta', attrs: { charset: 'utf-8' } },
 	{
 		tag: 'meta',
-		attrs: { name: 'viewport', content: 'width=device-width, initial-scale=1' },
+		attrs: { name: 'viewport', content: 'width=device-width' },
```
delucis, line 21: While MDN might claim this is the default in practice it does not seem to be the case and without it I have seen sites break on mobile:

```suggestion
		attrs: { name: 'viewport', content: 'width=device-width, initial-scale=1' },
```

packages/starlight/components/Head.astro
```diff
@@ -36,20 +36,18 @@ const headDefaults: z.input<ReturnType<typeof HeadConfigSchema>> = [
 			type: config.favicon.type,
 		},
 	},
+	// Twitter Tags
+	{
+		tag: 'meta',
+		attrs: { name: 'twitter:card', content: 'summary_large_image' },
+	},
```
delucis, line 43: Was there a specific motivation for changing the order of this tag? I’d guess we can just leave it where it is to avoid introducing unnecessary churn?

delucis (changes requested):
Thank you for the PR @cevdetardaharan. Left a couple of notes.

packages/starlight/components/Head.astro
cevdetardaharan: I looked so many different resources and the only reason that it's needed was on Safari when you change orientation it's not changing the width but it's doing zoom-out, I don't have iPhone so I can't test it, but I'd like to know if it's still needed in 2024 (like a test on Safari) but still I'm okay with `initial-scale=1` to be in the codebase, so I can accept the suggestion. Thank you!

packages/starlight/components/Head.astro
cevdetardaharan: This is what X's (aka Twitter) documentation shows as an example, but I don't think it's strictly needed so if you want that tag to be there, I can accept this suggestion too. Thank you!

packages/starlight/components/Head.astro
```diff
@@ -36,20 +36,18 @@ const headDefaults: z.input<ReturnType<typeof HeadConfigSchema>> = [
...
 	// OpenGraph Tags
 	{ tag: 'meta', attrs: { property: 'og:title', content: data.title } },
 	{ tag: 'meta', attrs: { property: 'og:type', content: 'article' } },
 	{ tag: 'meta', attrs: { property: 'og:url', content: canonical?.href } },
 	{ tag: 'meta', attrs: { property: 'og:locale', content: lang } },
 	{ tag: 'meta', attrs: { property: 'og:description', content: description } },
 	{ tag: 'meta', attrs: { property: 'og:site_name', content: siteTitle } },
```
delucis, line 50:
```suggestion
	{ tag: 'meta', attrs: { property: 'og:site_name', content: siteTitle } },
	// Twitter Tags
	{
		tag: 'meta',
		attrs: { name: 'twitter:card', content: 'summary_large_image' },
	},
```

delucis (approved):

PR #2143, opened 2024-07-23
docs: add 'og:image:alt' metadata

julien-deramond:
#### Description

This PR adds a metadata in Starlight documentation to improve the accessibility of Twitter cards when this documentation is shared on this platform, or other cards on different social media platforms.

Feel free to drop this PR if this is a topic that has already been discussed, or if you think this is not necessary. In this case, don't spend too much time on my PR :)

> [!IMPORTANT]  
> After a discussion based on https://github.com/withastro/starlight/pull/2137, this PR will only implement `og:image:alt`, and Twitter will use this as a fallback

##### Twitter/X platform

Let's consider this page: https://astro.build/blog/astro-4120/.
It contains the following metadata:
```html
<meta name="twitter:image" content="https://astro.build/_astro/og_image_4_12.DvriEbIV.webp">
<meta name="twitter:image:alt" content="Astro 4.12: Server Islands">
```
When shared on Twitter (see https://x.com/astrodotbuild/status/1813994888427684008), if you inspect the Twitter card, you'll see that the image has an alternative text, which is good for a11y:
```html
<img alt="Astro 4.12: Server Islands" draggable="true" src="https://pbs.twimg.com/card_img/1813989406337904640/UEz_mzB7?format=jpg&amp;name=small" class="css-9pa8cd">
```

Let's consider our own Starlight documentation at https://starlight.astro.build/ja/getting-started/ (for instance).
It contains the following metadata:
```html
<meta property="twitter:image" content="https://starlight.astro.build/og.jpg?v=1">
```

The image is the following:

But there's no `twitter:image:alt`.
When shared on Twitter (see https://x.com/NewGyu/status/1814669986096685126), if you inspect the Twitter card, you'll see that the image doesn't have any alternative text:
```html
<img alt="" draggable="true" src="https://pbs.twimg.com/card_img/1814058933243318272/eD3h0-Rh?format=jpg&amp;name=small" class="css-9pa8cd">
```

It's probably debatable whether a Twitter card image is decorative or not, and should have an alt text or not. However, our Starlight documentation image contains text that, if important, should be accessible to everyone, so via a screen reader too, as it is an information.

This is the reason why I'm suggesting this enhancement.

##### Other social media platforms

For other social media platforms, they mostly use Open Graph, and it also supports the alternative text for images with `og:image:alt`. So this is basically the same explanation for the same concept

##### Translate or not translate?

I've forced the following English text: "Make your docs shine with Starlight".
The reason is that I'd say we should have the exact same information as in the thumbnail (which is not translated).
But that's debatable. Maybe we can translate it for screen readers users only, for once, they would have the most precise information :)

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -26,6 +26,7 @@ const NETLIFY_PREVIEW_SITE = process.env.CONTEXT !== 'production' && process.env
 
 const site = NETLIFY_PREVIEW_SITE || 'https://starlight.astro.build/';
 const ogUrl = new URL('og.jpg?v=1', site).href;
+const ogImageAlt = 'Make your docs shine with Starlight';
 
 export default defineConfig({
 	site,
@@ -57,6 +58,10 @@ export default defineConfig({
 				{
 					tag: 'meta',
 					attrs: { property: 'og:image', content: ogUrl },
+				},
+				{
+					tag: 'meta',
+					attrs: { property: 'og:image:alt', content: ogImageAlt },
 				},
 			],
 			customCss: process.env.NO_GRADIENTS ? [] : ['./src/assets/landing.css'],


delucis (commented):
Thanks @julien-deramond — I think this is a reasonable change!

As an aside: I’m honestly not sure what on earth Twitter is doing with any of this. Looking at the astro.build example, the markup looks something like this:

```html
<a
	href="..."
	aria-label="astro.build Astro 4.12: Server Islands | Astro"
	role="link"
>
	<div aria-label="Astro 4.12: Server Islands">
		<img alt="Astro 4.12: Server Islands" src="...">
	</div>
	<span>Astro 4.12: Server Islands | Astro</span>
</a>
```

Two `aria-label`s, an `alt`, and then actual text…

Nevertheless, this seems like a reasonable change to make.

One question: do you happen to know if both the `twitter:image:alt` and `og:image:alt` variants are required? Or would Twitter use the `og:` one if available?

julien-deramond:
> As an aside: I’m honestly not sure what on earth Twitter is doing with any of this. Looking at the astro.build example, the markup looks something like this:
> Two aria-labels, an unnecessary role, an alt, and then actual text…

Yeah, and they improved it, apparently. I stumbled upon a post where it was mentioned that the alt text was inside an HTML element with an `aria-hidden=true` before 🙃 

> One question: do you happen to know if both the twitter:image:alt and og:image:alt variants are required? Or would Twitter use the og: one if available?

There are apparently fallback rules (that I haven't tested) here: https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup

| Property | OpenGraph  | 
|---|---|
| `twitter:image` | `og:image` | 
| `twitter:image:alt` | `og:image:alt` |

We could rely on the `og:*` fallback as it seems to work. If we do that, for consistency, maybe we should drop `twitter:image`, `twitter:title`, etc. On one hand, it would reduce the size of the pages. 
But my own personal feeling would be to keep them both, as we would never know in the future if things change, if the fallback works differently, etc.

What do you think?

delucis:
#2137 coincidentally removes most of the existing `twitter:*` tags, so I think I’m OK with removing them. Apart from the `twitter:card` one, they all seem to map back to the more standard `og:*` ones, so I think I’d only use the `twitter:*` ones if for some reason I wanted more control for _different_ tags on Twitter from elsewhere.

julien-deramond:
Oh, I didn't notice https://github.com/withastro/starlight/pull/2137, funny coincidence. I can understand the logic behind getting rid of `twitter:*` (except `twitter:card` that is important). The Astro/Starlight community is big, so if there's an update, we can trust it to create some PRs to put it back if necessary.

delucis (approved):
Thanks for the quick update — looks great!

Also, didn’t address this earlier, but given we have the one static OG image, I think a non-translated alternative text makes sense too.

PR #2150, opened 2024-07-24
Sidebar state persistence

delucis:
#### Description

- Closes #1121 — this PR takes a different approach
- Implements state persistence in the site sidebar across page navigation
- There’s an older test deployment here that includes extra content in the sidebar to further test collapsed group handling: https://66be4aad9c468f0008849146--astro-starlight.netlify.app/

#### Scope

- Currently, this PR specifically targets navigations between two pages that share the same sidebar. Navigations between pages with _different_ sidebars are not impacted. Those will continue to behave as they do currently. Navigations across _sessions_ are also not impacted.
- Currently, restoring state does not happen on small viewports where the menu is collapsed behind a hamburger button. It felt weird to me to restore state in this context, but I’m open to hear feedback on this.

#### How does it work?

1. We generate a hash of the sidebar content (labels and link `href`)
2. This hash and the scroll position of the sidebar is stored in `sessionStorage` each page visit
3. Additionally, when a user expands/collapses a sidebar group, that modified state is also persisted
4. On a new page load, we check if the stored hash matches the sidebar hash for the new page, and if so, we restore collapsed state and scroll position.
5. That’s it.

#### More technical details
- The persistence script is included in `Page.astro` so it always runs, even in the absence of a sidebar. This makes sure visiting a page without a sidebar “clears” the stored state. Otherwise, for example, visiting the Starlight docs homepage and then returning to a docs page persists state, which feels wrong, I think. (Again, thinking about this as a UX improvement specific to navigating between pages with a shared sidebar.)
- The hashing takes place at build time and uses a fast and insecure hash algorithm rather than relying on one of the slower SHA algorithms.
- The hashing is done inside of the `<Sidebar>` component. In general, we try to keep logic out of these components to help make overriding simpler, and in an earlier commit the has was added as a `sidebarHash` property to the route data object. There are pros and cons to both approaches but I think this is the right way based on:
	- The hash is component specific — it is completely tied to our implementation’s behaviour, and should not be needed when a totally custom override is in use.
	- It is becoming a not uncommon pattern for people to filter sidebar entries before rendering the default sidebar component. By hashing in the component we are able to hash those correctly — otherwise in theory someone would need to filter AND generate a unique hash when reusing the default component in this way.
- The state-restore script is inlined in `Sidebar.astro` to ensure it runs as soon as the sidebar HTML is complete.
- This technique will not work automatically for entirely custom sidebars (not sure that’s possible really?) so those will need some extra help probably for people overriding. But open to feedback here too.

#### Todo

- [x] Remove nav from getting started page
- [x] Delete test pages
- [x] Delete extra sidebar config in `astro.config.mjs`


packages/starlight/components/SidebarPersistState.ts
```diff
@@ -0,0 +1,22 @@
...
+	const state = {
+		hash: target?.dataset.hash || '',
+		open: [...details].map((element) => element.open),
+		scroll: scroller?.scrollTop || 0,
+	};
+	sessionStorage.setItem('sl-sidebar-state', JSON.stringify(state));
+};
+
+// Store sidebar state before navigating. These will also store it on tab blur etc.,
+// but avoid using the `beforeload` event, which can cause issues with back/forward cache
```
delucis, line 17:
```suggestion
// but avoid using the `beforeunload` event, which can cause issues with back/forward cache
```

delucis:
Made an experimental update in 06dc927 to see if it might handle sidebars with collapsed groups better. Still needs some cleanup probably but the principle here is to only persist sidebar collapse state that has been directly modified by user interaction.

In other words, given a four group sidebar with each group marked with `collapsed: true` in the sidebar config:

- On first visit, the current page’s group is expanded, the other three are collapsed (as per Starlight’s current defaults)
- When navigating to a page in a different group (via a link in body, so no sidebar interaction yet), that new group will expand and the old group will collapse (again, as per Starlight’s current behaviour)
- If a user expands a group in the sidebar, that gets stored and persisted across navigations, while the other groups continue to collapse/expand depending on if they contain the current page

In theory, this improves the experience for sidebars with lots of collapsed groups. Nevertheless, it may also be somewhat confusing as it introduces two behaviours in parallel:

- Some groups are “sticky”, remembering user interaction (with no way to “unstick” them without closing and reopening the tab)
- Other groups continue to toggle back and forth depending on current page

Will need some user testing and feedback to decide whether something like this change makes sense!

matthewp (approved):

HiDeoo (approved):
Looks perfect to me, the best of all tested versions :star2: Amazing work and such a great improvement! :tada:

Small question: I think the Lunaria Status Overview is outdated/invalid as you removed all the test pages in https://github.com/withastro/starlight/pull/2150/commits/ae660352e7a6a9605aef2b1c28874d1b78ee6810 ?

delucis:
> Small question: I think the Lunaria Status Overview is outdated/invalid as you removed all the test pages in [ae66035](https://github.com/withastro/starlight/commit/ae660352e7a6a9605aef2b1c28874d1b78ee6810) ?

Yes! I asked @yanthomasdev about it — it’s a small bug where the Lunaria action bails early when it finds no content files, which means it couldn’t update that comment.

PR #2156, opened 2024-07-27
Fix bug for projects with spaces in their pathname

delucis:
#### Description

A user reported an issue [on Discord](https://discord.com/channels/830184174198718474/1266530451422449776) that seems to be caused by an improperly encoded pathname when they were building a project in a directory containing a space in the pathname. I initially assumed it to be a Windows bug, but it looks like it’s also reproducible on macOS and Linux.

To test this, I have renamed the directory containing one of our end-to-end fixtures to include a space. I could also add a separate fixture, but it felt kind of excessive to add an additional fixture just to test the pathname. Hopefully by testing this we’ll catch similar issues in the future.

This PR correctly resolves the import in `virtual:starlight/collection-config` to fix the bug.


ematipico (approved):

packages/starlight/__e2e__/collection-config.test.ts
```diff
@@ -1,6 +1,6 @@
 import { expect, testFactory } from './test-utils';
 
-const test = await testFactory('./fixtures/custom-src-dir/');
+const test = await testFactory('./fixtures/custom src-dir/');
```
HiDeoo, line 3: Small detail: should we add a comment mentioning why there is a space in the fixture name and that we use it to test this specific case, considering it may be the only fixture with a space in the name so we don't remove it by mistake?

HiDeoo (approved):
Looks perfect to me :thumbsup:

packages/starlight/__e2e__/collection-config.test.ts
delucis: Yes, good idea, thank you!

```suggestion
// This fixture contains a space in the directory so that we have a smoke test for building
// Starlight projects with pathnames like this, which are a common source of bugs.
const test = await testFactory('./fixtures/custom src-dir/');
```

PR #2166, opened 2024-08-05
Update dependencies

delucis:
#### Description

- Updates dependencies of the main Starlight package
- Main intention is to update Expressive Code, which fixed the `emitFile()` warning logs  in v0.35.5
- I initially tried updating more but ran into a couple of CI issues so decided to scale back the scope to get the EC change in

Note for the future from @HiDeoo: initial failures may have been because the E2E test fixtures hadn’t had their deps updated to match. May want to include those in the pnpm workspace (if possible for a nested directory)?

### DIFF
--- a/.changeset/slimy-kings-enjoy.md
+++ b/.changeset/slimy-kings-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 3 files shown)


HiDeoo (approved):
Looks perfect to me, and that warning gone is a great DX improvement even if everything was working as intended :thumbsup:

Issue #1961, opened 2024-06-05
Failed to get locale informations


torn4dom4n:
### What version of `starlight` are you using?

0.24.0

### What version of `astro` are you using?

4.9.3

### What package manager are you using?

bun

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

I get this bug after updating Starlight to v0.24.0

```shell
1:58:17 AM [vite] Error when evaluating SSR module /node_modules/@astrojs/starlight/utils/i18n.ts:
|- AstroUserError: Failed to get locale informations for the 'en' locale.
    at new AstroUserError (C:\Users\Windows\Documents\GitHub\AREA44\playastro\node_modules\astro\dist\core\errors\errors.js:96:9)
    at getLocaleInfo (:99:47)
    at <anonymous> (:5:48)
    at processTicksAndRejections (:12:39)

1:58:17 AM [vite] Error when evaluating SSR module /node_modules/@astrojs/starlight/integrations/shared/localeToLang.ts: failed to import "/node_modules/@astrojs/starlight/utils/i18n.ts"
|- AstroUserError: Failed to get locale informations for the 'en' locale.
    at new AstroUserError (C:\Users\Windows\Documents\GitHub\AREA44\playastro\node_modules\astro\dist\core\errors\errors.js:96:9)
    at getLocaleInfo (:99:47)
    at <anonymous> (:5:48)
    at processTicksAndRejections (:12:39)

1:58:17 AM [vite] Error when evaluating SSR module /node_modules/@astrojs/starlight/integrations/expressive-code/translations.ts: failed to import "/node_modules/@astrojs/starlight/integrations/shared/localeToLang.ts"
|- AstroUserError: Failed to get locale informations for the 'en' locale.
    at new AstroUserError (C:\Users\Windows\Documents\GitHub\AREA44\playastro\node_modules\astro\dist\core\errors\errors.js:96:9)
    at getLocaleInfo (:99:47)
    at <anonymous> (:5:48)
    at processTicksAndRejections (:12:39)

1:58:17 AM [vite] Error when evaluating SSR module /node_modules/@astrojs/starlight/integrations/expressive-code/index.ts: failed to import "/node_modules/@astrojs/starlight/integrations/expressive-code/translations.ts"
|- AstroUserError: Failed to get locale informations for the 'en' locale.
[...]

PR #2167, opened 2024-08-06
Define well-known RTL locales before calling `getLocaleInfo()` for default locale

delucis:
#### Description

- Fixes https://github.com/withastro/starlight/issues/1961#issuecomment-2270020341
- We were using `getLocaleInfo()` before defining our well-known RTL locales. This was only used in rare circumstances (environments that support neither [`Intl.Locale#getTextInfo()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo) nor the non-standard `Intl.Locale#textInfo` accessor), so hadn’t noticed the issue until now. (A StackBlitz or other web container running on Firefox is probably the only place you can really hit this.)

### DIFF
--- a/.changeset/odd-sheep-love.md
+++ b/.changeset/odd-sheep-love.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue detecting the built-in locale when running Starlight in a web container environment on Firefox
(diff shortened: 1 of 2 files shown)


HiDeoo (approved):
Well, that's definitely a super specific case, but glad it'll be fixed :tada:

PR #2168, opened 2024-08-06
Use Starlight `sidebar` user-config format for `<StarlightPage />` `sidebar` prop

HiDeoo:
#### Description

- Closes #2146 and #1683

This PR changes the data accepted by the `sidebar` prop of the `<StarlightPage />` component used by custom routes.

With this change, the `sidebar` prop can now accept a sidebar configuration identical to the one used by the `sidebar` Starlight configuration.

A few demo custom pages have been temporarily added to the docs to play around with the changes.

#### Remaining task

- [x] Remove the demo link in `docs/astro.config.mjs`
- [x] Remove `docs/src/pages/test-1.astro`
- [x] Remove `docs/src/pages/test-2.astro`
- [x] Remove `docs/src/pages/test-[index].astro`
- [x] Add a changeset (need to think about it, not sure we can provide an easy "migration" guide as the format can be quite different and involves a few different types of links if we wanted to cover everything — maybe a link to the sidebar guide?)

#### Idea

Could be worth maybe adding a new example to https://starlight-examples.netlify.app/ showing how to re-use a sidebar configuration between the Starlight configuration and a custom page where the custom page appends a few more items to the sidebar.


delucis (changes requested):
Oh boy, the code is so much simpler, not sure why we didn’t think to do it to begin with 😁 

Looks good to me — I think you can remove the test files. Also needs a changeset explaining how to migrate if you’re using `<StarlightPage>`.

HiDeoo:
> Oh boy, the code is so much simpler, not sure why we didn’t think to do it to begin with 😁

Yeah, I had the exact same thought when working on the changes, not sure why either 😅

I've removed the test pages and added a changeset (altho, not super happy with it, tried to use a `diff` code block but I thought it was not super easy to read, so I went with 2 code blocks instead, but this ends up being a massive changeset).

.changeset/chilled-kiwis-count.md
```diff
@@ -0,0 +1,56 @@
+---
+'@astrojs/starlight': minor
+---
+
+⚠️ **BREAKING CHANGE:** Updates the `<StarlightPage />` component `sidebar` prop to accept an array of [`SidebarItem`](https://starlight.astro.build/reference/configuration/#sidebaritem) like the Starlight global `sidebar` configuration.
```
delucis, line 5: I think we should probably pluralize “item” here to read a bit better, and also suggesting an alternative to just “global” to clarify things:

```suggestion
⚠️ **BREAKING CHANGE:** Updates the `<StarlightPage />` component `sidebar` prop to accept an array of [`SidebarItem`](https://starlight.astro.build/reference/configuration/#sidebaritem)s like the main Starlight `sidebar` configuration in `astro.config.mjs`.
```

.changeset/chilled-kiwis-count.md
```diff
@@ -0,0 +1,56 @@
...
+
+This change simplifies the definition of sidebar items in the `<StarlightPage />` component, allows for shared sidebar configuration between the global `sidebar` option and `<StarlightPage />` component, and also enables the usage of autogenerated sidebar groups with the `<StarlightPage />` component.
+If you are using the `<StarlightPage />` component with a custom `sidebar` configuration, you will need to update the `sidebar` prop to an array of [`SidebarItem`](https://starlight.astro.build/reference/configuration/#sidebaritem) objects.
+
+For example, the following custom page with a custom `sidebar` configuration defines a “Resources” group with a “New” badge, a link to the “Showcase” page which is part of the `docs` content collection, and a link to the Starlight website:
+
+```tsx
```
delucis, line 12: Shouldn’t these use Astro?

````suggestion
```astro
````

File: .changeset/chilled-kiwis-count.md [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+		},
+	]}
+>
+	<p>This is a custom page with a custom component.</p>
+</StarlightPage>
+```
+
+This configuration will now need to be updated to the following:
+
+```tsx
```
delucis, line 37:
````suggestion
```astro
````

delucis (commented):
I think the examples are good here! Yes, it’s big, but I agree this is clear (I also quickly tried a diff but it feels more confusing than helpful).

delucis (approved):

PR #2171, opened 2024-08-07
Build performance optimizations

delucis:
#### Description

- Aims to improve some build performance by avoiding some well-known slower operations
- Updates a few places dealing with trailing slashes to use our existing trailing slash utilities instead of using `.replace()` with a regular expression
- Updates the `stripExtension()` utility in sidebar generation to use lower-level string operations instead of a regular expression in `.replace()`. (In this case the logic is technically subtly different as the RegExp was matching `\w+` after a final period, but given the limited scope of usage for this utility on file paths ending in `.md`, `.mdx` etc., it should be fine.)
- Updates the `slugToParam()` utility to avoid an unnecessary `.replace()` in favour of a faster `.slice()`
- Refactors some slug processing in `linkFromInternalSidebarLinkItem()` to avoid the need to `.replace()` entirely

### DIFF
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -410,4 +410,7 @@ function applyPrevNextLinkConfig(
 }
 
 /** Remove the extension from a path. */
-const stripExtension = (path: string) => path.replace(/\.\w+$/, '');
+function stripExtension(path: string) {
+	const periodIndex = path.lastIndexOf('.');
+	return periodIndex < path.length - 1 ? path.slice(0, periodIndex) : path;
+}
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -412,5 +412,5 @@ function applyPrevNextLinkConfig(
 /** Remove the extension from a path. */
 function stripExtension(path: string) {
 	const periodIndex = path.lastIndexOf('.');
-	return periodIndex < path.length - 1 ? path.slice(0, periodIndex) : path;
+	return path.slice(0, periodIndex > -1 ? periodIndex : undefined);
 }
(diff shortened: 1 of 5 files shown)


packages/starlight/utils/navigation.ts
```diff
@@ -410,4 +410,7 @@ function applyPrevNextLinkConfig(
 }
 
 /** Remove the extension from a path. */
-const stripExtension = (path: string) => path.replace(/\.\w+$/, '');
+function stripExtension(path: string) {
+	const periodIndex = path.lastIndexOf('.');
+	return periodIndex < path.length - 1 ? path.slice(0, periodIndex) : path;
```
HiDeoo, line 415: Could comparing to [`-1`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf#return_value) a bit easier to understand?

HiDeoo (approved):
Lovely changes 🔥

packages/starlight/utils/navigation.ts
HiDeoo: Wait, just having second thought, is this correct? Isn't this always slicing?

delucis: Ohhhhhh, yes 😅

Probably not breaking tests because this always receives a path like `something.md`, but yeah, should fix the logic to not risk breaking if we ever pass something that doesn’t have an extension.

HiDeoo: Yeah, did not manage to figure out how to change my approval to request a change but now that you've seen my comment, it's fine ^^

delucis: Think this might be the best solution, especially given we know this always needs to slice currently:

```suggestion
	return path.slice(0, periodIndex > -1 ? periodIndex: undefined);
```

HiDeoo: Looks perfect to me and definitely easier to read/understand (and correct) 👍

PR #2179, opened 2024-08-08
fix: make link relative

micartey:
I used this template and it took me 35 minutes to understand that following the example is a dead end when trying to use the base option. Links need to be relative.

#### Description

- Closes no current issue but is related to #1407 
- What does this PR change? Makes a link in the base example relative to not mislead any other people
- Did you change something visual? Nope

### DIFF
--- a/examples/basics/src/content/docs/index.mdx
+++ b/examples/basics/src/content/docs/index.mdx
@@ -8,7 +8,7 @@ hero:
     file: ../../assets/houston.webp
   actions:
     - text: Example Guide
-      link: /guides/example/
+      link: guides/example/
       icon: right-arrow
       variant: primary
     - text: Read the Starlight docs
--- a/examples/tailwind/src/content/docs/index.mdx
+++ b/examples/tailwind/src/content/docs/index.mdx
@@ -18,7 +18,7 @@ hero:
     file: ../../assets/houston.webp
   actions:
     - text: Example Guide
-      link: /guides/example/
+      link: guides/example/
       icon: right-arrow
       variant: primary
     - text: Read the Starlight docs


delucis:
Thanks for the PR @micartey and sorry to hear this wasn’t a clear experience.

We intentionally make links absolute as in general it is much less confusing than ensuring relative links work well with all the different build output formats and hosting options out there. However, when using `base`, you do need to author links to also include `base`, so in your base the link would be something like `/base/guides/example`.

Maybe there’s somewhere we can make this clearer in the docs?

micartey:
Hey @delucis 

Maybe I did understand the documentation wrong.

I thought that using the `base` option doesn't require me to change all the links to manually set the base as a prefix?
This is also not very logical as the base option is being added as a prefix with `base/` before each link.

Having relative links is therefore only logical.

I do not understand what you mean with:
> intentionally make links absolute as in general it is much less confusing

Relative links are not confusion at all 🤔 
In my case, it was more confusing to understand that I somewhere had absolute links and therefore my website didn't work.

> ensuring relative links work well with all the different build output formats and hosting options out

This is an interesting statement as you use relative links in e.g. `astro.config.mjs`
Why is it there a good choice.

https://github.com/withastro/starlight/blob/976038306a3bc3ef7bde8867341b28859cb11ba2/examples/basics/astro.config.mjs#L15-L18

I would suggest sticking to one, but not using both.
And in case you want to use absolute links (which I highly advise against), to at least show a warning or error when using the `base` option. But the statement that I need to author links to include the `base` is clearly wrong - Not if you use relative links`

delucis:
Relative links depend on the URL format of the current page. For example let’s say you have the following link:

```html
<a href="guides">Guides</a>
```

If this link is on `https://example.com/product`, then this link points to `https://example.com/guides`.

If this link is on `https://example.com/product/`, then this link points to `https://example.com/product/guides`.

This can be a big source of confusion for people, especially if their server supports serving both `/product` AND `/product/` (not so uncommon). People often end up using trailing slashes inconsistently and don’t understand why their relative links only work correctly some of the time.

> you use relative links in e.g. astro.config.mjs

That is not a link, but the slug of a file in the `src/content/docs/` directory.

> I thought that using the base option doesn't require me to change all the links to manually set the base as a prefix?

`base` configures this prefix for some things:
- assets like JS, CSS, etc.
- links Starlight is responsible for like the sidebar

But unfortunately we can’t safely do it for other manually authored links because we can’t know if a user wants to link to something with `base` or without. Imagine you have a Starlight deployed at `example.com/docs` using `base: 'docs'`, but want to link to your landing page at `example.com` with `/`. If we transformed a link to `/` to be `/docs`, we would prevent anyone doing that.

So in general you’ll need to author links as you want them.

PR #2182, opened 2024-08-09
fix: using wrong field to deconstruct in 'CallToAction' component

hartmut-co-uk:
#### Description
Closes (no issue created)

What does this PR change?
- Fixes what a bug to pass html attribs to the **CallToAction** component.
- This allows to e.g. pass the class attribute `<CallToAction  class="px-1">...</CallToAction>`

Note: I'm actually not sure if the `Prop` interface definition is correct (`attrs?: Omit<HTMLAttributes<'a'>, 'href'> | undefined;`) or what the purpose for this is?!?

### DIFF
--- a/.changeset/wild-phones-pull.md
+++ b/.changeset/wild-phones-pull.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+fix: using wrong field to deconstruct in 'CallToAction' component
--- a/packages/starlight/components/CallToAction.astro
+++ b/packages/starlight/components/CallToAction.astro
@@ -11,7 +11,7 @@ interface Props {
 }
 
 const { link, variant, icon } = Astro.props;
-const { class: customClass, ...attrs } = Astro.props.attrs || {};
+const { class: customClass, ...attrs } = Astro.props || {};
 ---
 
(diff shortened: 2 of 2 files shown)


delucis:
Hi @hartmut-co-uk! Thanks for the PR.

I believe this code is working as expected. You can pass extra attributes using the `attrs` prop to actions (including `class` if you want) as shown in the reference for the [`hero` frontmatter property](https://starlight.astro.build/reference/frontmatter/#hero). Here’s an example with a custom `class` for each action:

```md
---
title: My Home Page
hero:
  title: 'My project with custom action appearance'
  actions:
    - text: Tell me more
      link: /getting-started/
      attrs:
        class: more-link
    - text: View on GitHub
      link: https://github.com/astronaut/my-project
      attrs:
        class: github-link
---
```

If you still believe there to be a bug please open an issue with a reproduction so we have a bit more context on what is not working.

hartmut-co-uk:
OK, thanks Understood @delucis!   
I'm re-using the component elsewhere and based on the example provided the following also works for me:
```html
<CallToAction attrs={{class: "minimal"}} link="/abc/">...</CallToAction>
```

delucis:
Ah cool, yes that’s not officially supported yet, but the experience will be improved once we merge #1784 — which removes `CallToAction` in favour of a proper `<LinkButton>` component.

PR #2183, opened 2024-08-09
i18n(it): Updates `guides/css-and-tailwind.mdx`

casungo:
#### Description

Updates `guides/css-and-tailwind.mdx`

The indentations are now updated to match the english ones too (not that it makes a difference in the page, its just more tidy to look at when doing a diff in VSCode)


HiDeoo (approved):
Thanks for updating the translation, this looks great :star2:

PR #2187, opened 2024-08-09
i18n(it): Updates `guides/pages.mdx`

casungo:
#### Description

Updates `guides/pages.mdx`


delucis (commented):
I’m guessing this update will need to wait for #2193 and #2194, which add the subheadings linked here. Otherwise looks good to me!

delucis (approved):
Links working as expected now! 🎉

PR #2193, opened 2024-08-09
i18n(it): Updates `reference/overrides.md`

casungo:
#### Description

Updates `reference/overrides.md`


docs/src/content/docs/it/reference/overrides.md
```diff
@@ -332,6 +345,12 @@ Componente contenente l'elemento `<h1>` per la pagina corrente.
 
 Le implementazioni dovrebbero garantire di impostare `id="_top"` sull'elemento `<h1>` come nell'implementazione predefinita.
 
+#### `DraftContentNotice`
+
+**Componente predefinito:** [`DraftContentNotice.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/DraftContentNotice.astro)
```
HiDeoo, line 350: It looks all other instances in this page are using "Componente standard" instead of "Componente predefinito". Not sure if one is better than the other, but I think they should be consistent. Do you think they should all be updated?

docs/src/content/docs/it/reference/overrides.md
```diff
@@ -50,6 +50,19 @@ Tag di lingua BCP-47 per le impostazioni internazionali di questa pagina, ad es.
 
 Il percorso di base in cui viene servita una lingua. `undefined` per gli slug della lingua di base.
 
+#### `siteTitle`
+
+**Tipo:** `string`
```
HiDeoo, line 55:
```suggestion
**tipo:** `string`
```

All the other instances in this page and also in the configuration references seems to be using "tipo" without capitalizing the first letter.

Note that if you think this should be changed, we should do that in a separate PR modifying all files with this pattern maybe.

docs/src/content/docs/it/reference/overrides.md
```diff
@@ -50,6 +50,19 @@ Tag di lingua BCP-47 per le impostazioni internazionali di questa pagina, ad es.
...
+
+Il titolo del sito per la lingua di questa pagina.
+
+#### `siteTitleHref`
+
+**Tipo:** `string`
```
HiDeoo, line 61:
```suggestion
**tipo:** `string`
```

Same as above

docs/src/content/docs/it/reference/overrides.md
HiDeoo, line 1: I think the [anchor change](https://github.com/withastro/starlight/pull/1620/files#diff-ad80506c0d18072cbd9a85e2090c2be41e5a5523fe9f2991a1c653e41d089591) from this [PR](https://github.com/withastro/starlight/pull/1620) is missing as the `<SiteTitle />` link in the Header override section now links to the `siteTitle` prop instead of the component. The anchor should be `#sitetitle-1` (cannot suggest a change for this).

HiDeoo (changes requested):

Left a few comments but except for those, the current changes look good to me.

docs/src/content/docs/it/reference/overrides.md
casungo: No, I think "Componente standard" should be used, will update it

delucis (approved):
Thank you for the updates @casungo and for the thorough review @HiDeoo!

PR #2213, opened 2024-08-14
Experimental minified inline scripts

delucis:
#### Description

Experiment using a Vite plugin to allow minifying scripts for use in a `<script is:inline>`


delucis:
The impact is not much to be honest. Comparing to other deploys of the Starlight docs, this approach is saving ~500B over all, ~200B gzipped, so probably not justified. I’d like to keep this branch around and test again once #2087 and #2150 land though as those both add more inline scripts.

delucis:
Updated this branch to use the minification strategy also for the new sidebar and tab persistence scripts. The size savings remain minimal: 700B reduction uncompressed, only about 100B reduction after the Brotli compression used by Netlify.

Over all, I’m not sure this makes much of a difference for performance. The only reason to perhaps consider keeping a system like this around would presumably be maintenance: scripts can be written in `.ts`, comments get stripped etc. Although I will add that I’m not 100% convinced of the benefits here either as it forces us to move inline scripts to a separate file, away from the template they are usually interacting with. Given this, I will close this as an interesting experiment for now.

PR #2216, opened 2024-08-16
Site Title Change Css Variable

sanabel-al-firdaws:
Its better if we use the value for h4 insted 

changed it from this
--sl-text-h4

to this 
--sl-text-xl

this has no visual diffrence for normal users but because i tried changing these values dynamiclly

--sl-text-h1
--sl-text-h2
..
--sl-text-h4

the title gets large weirdly...it shouldn't change

### DIFF
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -38,7 +38,7 @@ const { siteTitle, siteTitleHref } = Astro.props;
 	.site-title {
 		align-items: center;
 		gap: var(--sl-nav-gap);
-		font-size: var(--sl-text-h4);
+		font-size: var(--sl-text-xl);
 		font-weight: 600;
 		color: var(--sl-color-text-accent);
 		text-decoration: none;


delucis (commented):
Thanks for the PR @sanabel-al-firdaws!

The reason this uses the `h4` variable instead of `xl` is because `h4` changes depending on the screen size. On small screens it uses `xl`:

https://github.com/withastro/starlight/blob/8a861d16b586b019f61f30d93c61bdcd58e1503f/packages/starlight/style/props.css#L80

But on wider screens it uses the `2xl` variant:

https://github.com/withastro/starlight/blob/8a861d16b586b019f61f30d93c61bdcd58e1503f/packages/starlight/style/props.css#L175

We might need to think how best to handle this. I’m pretty sure there are many places we use the heading sizes for stuff that isn’t a heading, so could be a bit tricky to ensure that the override behaviour you want is consistent.

One way to solve in your own project would be to scope the `h4` size change to the Markdown content:


```css
/* Make H4s huge inside Markdown content */
.sl-markdown-content {
  --sl-text-h4: 5rem;
}
```

PR #2219, opened 2024-08-16
Fix a sidebar persistence issue when navigating between pages with different sidebars

HiDeoo:
#### Description

Fixes the sidebar persistence issue spotted by Chris.

The repro is:

- Go to https://starlight.astro.build/getting-started/
- Collapse some groups
- Navigate to the homepage
- Click “getting started”
- The groups are reset as expected
- Navigate to another page
- The stale collapse state is restored

If I'm not wrong, the `open` state is always loaded from session storage even when the stored hash and current hash don't match. This PR fixes this issue by only loading the `open` state from session storage when the stored hash and current hash match.

### DIFF
--- a/.changeset/lemon-ants-bow.md
+++ b/.changeset/lemon-ants-bow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a sidebar persistence issue when navigating between pages with different sidebar content.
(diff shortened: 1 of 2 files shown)


delucis (approved):
Yes, this fix makes sense to me! And can confirm it addresses the bug in the preview site. Thanks @HiDeoo 💖

PR #2222, opened 2024-08-16
i18n(fr): update `index`

HiDeoo:
#### Description

This PR updates the French version of the `index` page with the changes from #1784.


delucis (approved):
Ah, great point! I wonder if we should actually do a single `i18nIgnore` PR to update these in all languages?

But this PR LGTM 🙌

delucis:
(Also, missed the PR number: #2222 🥳)

HiDeoo:
> I wonder if we should actually do a single `i18nIgnore` PR to update these in all languages?

Good call, I'll make a PR after merging this one. I already merged `ko` earlier this morning and there is no other opened PRs for this file so might as well do it now 👍

PR #2227, opened 2024-08-17
i18n(ko-KR): update `pages.mdx`

jsparkdev:
#### Description

-  update `pages.mdx`
- #2168


docs/src/content/docs/ko/guides/pages.mdx
```diff
@@ -109,26 +109,25 @@ Markdown 페이지의 프론트매터와 유사하게 이 페이지에 대한 [
 
 ##### `sidebar`
 
-**타입:** `SidebarEntry[]`
+**타입:** [`SidebarItem[]`](/ko/reference/configuration/#sidebaritem)
```
HiDeoo, line 112:
```suggestion
**타입:** [`SidebarItem[]`](/ko/reference/configuration/#sidebaritem)  
```

This is missing the 2 trailing spaces so that `기본값` is on the next line.

Unrelated to this PR, but it also looks like these 2 spaces are missing for the `hasSidebar`, `headings`, `dir`, `lang`, and `isFallback` props following this part. Would you be up for fixing them at the same time?

HiDeoo (changes requested):
Thanks for updating the translation :raised_hands:

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view) and this looks great to me. I only left a suggestion regarding formatting.

docs/src/content/docs/ko/guides/pages.mdx
jsparkdev: Thank you! I fixed it.

HiDeoo (approved):
Amazing, thanks for taking the time to fix all of them, super appreciated :rocket:

PR #2242, opened 2024-08-20
Refactor sidebar persistence logic for better slow device performance

delucis:
#### Description

This PR refactors how the sidebar persistence logic works (yes, already 😅).

Before this PR, we restored state for sidebar groups in a single inline script at the bottom of the sidebar. This is mostly working very well, but on underpowered devices — or, perhaps more importantly, devices momentarily under some strain — this can mean a partial render of the sidebar can be shown before the sidebar is fully rendered, and state is restored only _after_ that. This can lead to a “flash of unrestored sidebar”. For example, here are two frames from a performance trace with 20x CPU slowdown, showing first a partial sidebar with groups expanded and then the restored state with groups collapsed:

performance trace screenshot

The refactor moves the restore logic into a small custom element. Every sidebar group (rendered by `<SidebarSublist>` renders this custom element, which triggers the state restore for that group. This means state can be restored incrementally as the sidebar renders instead of once after the full DOM is ready. The only exception is for the scroll position restoration. Setting `scrollTop` forces style recalc/layout work, so that still happens only once to minimise that. That means on a slow device, scroll restoration can still result in a visible jump in theory but it is overall still much more stable.

One detail I’m still looking at is how to generate an index for each sidebar group. Currently this is happening during rendering, but it might make more sense to move it to the sidebar generation logic directly.


packages/starlight/components/SidebarPersister.astro
```diff
@@ -0,0 +1,72 @@
...
+<sl-sidebar-state-persist data-hash={hash}>
+	<script is:inline>
+		(() => {
+			try {
+				if (!matchMedia('(min-width: 50em)').matches) return;
+				/** @type {HTMLElement | null} */
+				const target = document.querySelector('sl-sidebar-state-persist');
+				const state = JSON.parse(sessionStorage.getItem('sl-sidebar-state') || '0');
+				window._starlightScrollRestore = state.scroll;
+				if (!target || !state || target.dataset.hash !== state.hash) return;
```
HiDeoo, line 39:
```suggestion
				if (!target || !state || target.dataset.hash !== state.hash) return;
				window._starlightScrollRestore = state.scroll;
```

I think these 2 lines should be inverted otherwise we would restore the scroll position even if the hashes don't match. Tested it locally using a `docs/src/pages/test.astro` page using `<StarlightPage>` with a custom different sidebar and the scroll would be restored even with different sidebars.

HiDeoo (commented):
Really impressive refactor, tested it more and all the reproducible situations that were previously potentially showing a flash of unrestored sidebar no longer exhibit that behavior 👏 

Just left a tiny question/suggestion but otherwise this is looking great to me.

packages/starlight/components/SidebarPersister.astro
delucis: Oh yes — absolutely right! 😁

delucis:
Marked this as ready. I thought a bit more about the `SidebarRestorePoint` index generation and the current approach is pretty nicely contained and I think it should be reliable, so I think I’m happy with keeping it unless I hear other feedback!

packages/starlight/components/SidebarPersister.astro
```diff
@@ -0,0 +1,72 @@
...
+		/** Restored scroll position. Briefly stored on the `window` global to pass between inline scripts. */
+		_starlightScrollRestore?: number;
+	}
+}
+---
+
+<sl-sidebar-state-persist data-hash={hash}>
+	<script is:inline>
+		(() => {
+			try {
```
matthewp, line 33: Is the try/catch needed? I don't see anythink here that is likely to throw.

Also, you can do:

```js
{
  const target  = ...
}
```

instead of the function wrapper on line 32, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_let_const_class_or_function_declaration_in_strict_mode

packages/starlight/components/SidebarRestorePoint.astro
```diff
@@ -0,0 +1,12 @@
+---
+/** Unique symbol for storing a running index in `locals`. */
+const currentGroupIndexSymbol = Symbol.for('starlight-sidebar-group-index');
+const locals = Astro.locals as Record<typeof currentGroupIndexSymbol, number>;
```
matthewp, line 4: sneaky, i like it.

packages/starlight/components/SidebarPersister.astro
delucis: The `try`/`catch` is mainly to defend against the state loading in case `sessionStorage` is unavailable or for some reason it got some bad value stored to it so the `JSON.parse()` throws.

Good point about block scoping! Hadn’t considered that before for these inline scripts.

PR #2233, opened 2024-08-17
[i18nIgnore] docs: update `index.mdx` hero action buttons variants

HiDeoo:
#### Description

This PR updates all the homepage hero action buttons variants to match the changes from #1784.

PR #2245, opened 2024-08-21
Update index.mdx

ax-6:
I learned that the simplified Chinese version of index.mdx was outdated, so I updated it based on the English version of index.mdx.

### DIFF
--- a/docs/src/content/docs/zh-cn/index.mdx
+++ b/docs/src/content/docs/zh-cn/index.mdx
@@ -1,25 +1,30 @@
+以下是翻译内容：
+
 ---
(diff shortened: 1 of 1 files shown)


delucis:
Thank you @ax-6! In fact this was an error in our [i18n tracking system](https://i18n.starlight.astro.build/).

#1784 changed some frontmatter in the English homepage, marking the page as needing updating.

We then updated all translations in #2233 but used our `i18nIgnore` keyword, which means it did not mark the page as up-to-date.

Nevertheless, the page was already updated. Thanks all the same!

PR #2250, opened 2024-08-21
Stop publishing E2E tests

HiDeoo:
#### Description

This PR adds E2E tests and their configuration to the `.npmignore` file to avoid publishing them:

image

I tested the change before and after with `pnpm publish --dry-run --no-git-checks` in the `packages/starlight/` directory and confirmed that the E2E tests are no longer included after this change.

### DIFF
--- a/.changeset/slimy-beers-report.md
+++ b/.changeset/slimy-beers-report.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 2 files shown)


delucis (commented):
Ah, good catch — thanks @HiDeoo! This probably needs a changeset so we can release it?

delucis (approved):
Thanks again — end-to-end, but not end-to-end-to-npm 🚀

PR #2252, opened 2024-08-22
Build performance optimizations for projects with large sidebars

### LINKED ISSUE #2215: Large sidebars (i.e autogenerating all folders) causes build slowdowns due to repeated getSidebar calls

0.25.2

4.13.4

npm

Mac

Chrome

### Describe the Bug

On Starlight sites with a large (thousands of items) sidebar, which may be filtered with a Sidebar override (i.e to the current folder), individual pages will build slower as the sidebar size increases.

For example, with [`cloudflare/cloudflare-docs`](https://github.com/cloudflare/cloudflare-docs) as an example, there are ~4,000 pages and each top-level folder within `src/content/docs` is an autogenerate entry for the sidebar.

As part of [`generateRouteData`](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/route-data.ts#L41), [`getSidebar`](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/navigation.ts#L337) is called on each page with the current page's path (to mark a given page as current). With a very large sidebar, this can represent a large amount of the time spent rendering a page.

NodeJS Dev Tools profile:

image

Full build flamegraph:

image

I will try and follow-up with a proper MRE in StackBlitz!

https://github.com/cloudflare/cloudflare-docs

### Participation

HiDeoo:
#### Description

This PR is a draft experimenting with a new approach to fix #2215 by caching generated sidebars per locale to avoid regenerating them for each page during the build process.

A few details about the approach:

- It's using a basic `Map` at the moment.
  - I still want to try this approach against the SSR PR and maybe test with an LRU instead.
- The approach is used for the `docs` content collections pages and `<StarlightPage>` not specifying a custom sidebar (which means the sidebar generated based on the global `sidebar` config is used).
  - Using the same approach for `<StarlightPage>` with a custom sidebar would probably require a hash of the custom sidebar config so I decided to keep it simple for now and it's not required to get an idea of the potential performance improvements (there is also no guarantee that generating such a hash would not kill the performance improvements from the caching).

Here are some benchmarks I ran on my machine:

### Starlight Docs

- Images are cached from a previous run.

**Before:**

```
❯ hyperfine --runs 5 'pnpm build'
Benchmark 1: pnpm build
  Time (mean ± σ):     23.290 s ±  0.592 s    [User: 34.546 s, System: 3.046 s]
  Range (min … max):   22.431 s … 23.840 s    5 runs
```

**After:**

```
❯ hyperfine --runs 5 'pnpm build'
Benchmark 1: pnpm build
  Time (mean ± σ):     22.000 s ±  0.886 s    [User: 32.500 s, System: 2.769 s]
  Range (min … max):   21.124 s … 23.097 s    5 runs
```

### Cloudflare Docs (no patch)

- Images are cached from a previous run.
- Default Starlight with the changes from this PR applied without the [Starlight patch](https://github.com/cloudflare/cloudflare-docs/blob/2d1e3d754950cd15b844d9b0472d60c63cbd45e2/patches/%40astrojs%2Bstarlight%2B0.26.1.patch) that is applied to the Cloudflare docs.
- `astro check` has been removed from the `build` npm script as it's not really relevant for this benchmark.

**Before:**

```
❯ hyperfine --runs 5 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):     363.060 s ±  2.154 s    [User: 394.678 s, System: 24.588 s]
  Range (min … max):   360.170 s … 366.090 s    5 runs
```

**After:**

```
❯ hyperfine --runs 5 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):     230.533 s ±  2.276 s    [User: 264.999 s, System: 22.543 s]
  Range (min … max):   228.171 s … 233.879 s    5 runs
```

### Cloudflare Docs (with patch)

- Images are cached from a previous run.
- Starlight using the changes from this PR and an updated version of the [Starlight patch](https://github.com/cloudflare/cloudflare-docs/blob/2d1e3d754950cd15b844d9b0472d60c63cbd45e2/patches/%40astrojs%2Bstarlight%2B0.26.1.patch) that is applied to the Cloudflare docs (updated so that the cache of intermediate sidebars is keyed by locale + current section instead of just locale).
- `astro check` has been removed from the `build` npm script as it's not really relevant for this benchmark.

**Before:**

```
❯ hyperfine --runs 5 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):     210.619 s ±  2.512 s    [User: 240.743 s, System: 22.913 s]
  Range (min … max):   208.476 s … 214.525 s    5 runs
```

**After:**

```
❯ hyperfine --runs 5 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):     200.607 s ±  3.295 s    [User: 229.740 s, System: 21.582 s]
  Range (min … max):   196.067 s … 204.917 s    5 runs
```

A few interesting points:

- For the Starlight Docs, the improvement is not really significant but it's expected as it's mostly targeting sites with large sidebars.
- For the Cloudflare Docs with their current patch updated, the improvement is also not that huge but I think it's also expected as their current patch already trimmed down the number of entries generated for the sidebars a lot.
- For the Cloudflare Docs without the patch, the improvement is a bit more interesting, especially considering that most people will not have similar use cases and optimizations in a custom patch.


HiDeoo:
Undrafting the PR as I've been playing quite a bit with the changes and didn't see any downside to the current approach. I still want to do a little bit of play-testing with some SSR scenarios but I thought I might as well undraft the PR before I forget as usual and in case some users want to provide some feedback in the meantime.

packages/starlight/utils/navigation.ts
```diff
@@ -348,9 +329,25 @@ function sidebarFromDir(
...
+/**
+ * Intermediate sidebar represents sidebar entries generated from the user config for a specific
+ * locale and do not contain any information about the current page.
+ * These representations are cached per locale to avoid regenerating them for each page.
+ * When generating the final sidebar for a page, the intermediate sidebar is cloned and the current
+ * page is marked as such.
+ *
+ * @see getSidebarFromIntermediateSidebar
+ */
+const intermediateSidebars = new Map<string | undefined, SidebarEntry[]>();
```
delucis, line 341: Put a request out to @Fryuni [on Discord](https://discord.com/channels/830184174198718474/1070481941863878697/1303688158445309982) for a second opinion on this vs an LRU cache etc.

delucis (approved):
Went through this again now and seems really solid. I think I’m happy with this as-is 🚀

packages/starlight/utils/navigation.ts
delucis: Concluded it’s not an issue 🎉

packages/starlight/utils/navigation.ts
```diff
@@ -368,6 +363,37 @@ export function getSidebarFromConfig(
...
+				entry.isCurrent = true;
+				break;
+			}
+		} else if (entry.type === 'group') {
+			found = setIntermediateSidebarCurrentEntry(entry.entries, pathname);
+			if (found) break;
+		}
+	}
+	return found;
+}
```
kevinzunigacuellar, line 395: Since we are returning a boolean we can use it inside an if statement, this should clean up the code a bit.

I ran the tests locally and seems to be working, feel free to disregard.
```suggestion
function setIntermediateSidebarCurrentEntry(
	intermediateSidebar: SidebarEntry[],
	pathname: string
): boolean {
	for (const entry of intermediateSidebar) {
		if (entry.type === 'link' && pathsMatch(encodeURI(entry.href), pathname)) {
			entry.isCurrent = true;
			return true;
		}

		if (entry.type === 'group' && setIntermediateSidebarCurrentEntry(entry.entries, pathname)) {
			return true;
		}
	}
	return false;
}
```

PR #2253, opened 2024-08-22
Support the `class` attribute for hero action link buttons

HiDeoo:
This PR fixes an issue [reported](https://discord.com/channels/830184174198718474/1070481941863878697/1276132830547808337) on Discord where trying to use the `class` attribute in hero action link buttons was not working as expected.

The link button generated with the following frontmatter would not include the `TEST_CLASS_WARNING_REMOVE_ME` class:

```yaml
title: My Docs
template: splash
hero:
  actions:
    - text: Go
      icon: right-arrow
      link: /
      attrs:
        class: TEST_CLASS_WARNING_REMOVE_ME
```


packages/starlight/components/Hero.astro
```diff
@@ -50,12 +50,14 @@ if (image) {
...
 				<div class="sl-flex actions">
-					{actions.map(({ attrs, icon, link: href, text, variant }) => (
-						<LinkButton {href} {variant} icon={icon?.name} {...attrs}>
-							{text}
-							{icon?.html && <Fragment set:html={icon.html} />}
-						</LinkButton>
-					))}
+					{actions.map(
+						({ attrs: { class: className, ...attrs } = {}, icon, link: href, text, variant }) => (
+							<LinkButton {href} {variant} icon={icon?.name} class:list={[className]} {...attrs}>
```
delucis, line 55: Huh, interesting! Do you know why this extra work is needed? Because Astro already sets a `class` and uses that instead of the one spread in from `attrs`? Feels like maybe something we should report to the team.

delucis (approved):
Looks good @HiDeoo — I’m curious about one detail, but happy with the code.

packages/starlight/components/Hero.astro
HiDeoo: Good question, I always assumed this was the expected pattern to use with our `scopedStyleStrategy` when you want to accept a class that you would pass down to a child and the component also get a scoped class. Using `class` would result with one being overwritten by the other so you would have to use `class:list`. Thinking about it, I guess this may be handled automatically, never thought about it, could be a good question to ask 🤔

delucis: Asked [on Discord](https://discord.com/channels/830184174198718474/1070208455329198190/1278042063937474605), but I think we can still make this as a fix for now!

delucis (approved):
Thanks for the fix @HiDeoo!

PR #2256, opened 2024-08-25
feat(sitetitle): img component

### LINKED ISSUE #2161: SVG Logo during build doubles in dist/_astro folder

0.25.3

4.12.2

pnpm

Mac

Chrome

### Describe the Bug

I have logo.svg in /src/assets/logo.svg

Added link to logo in astro.config.mjs 
```json
export default defineConfig({
  integrations: [
	starlight({
		title: 'Документация Интеграм',
	        logo: {
		        src: './src/assets/logo.svg',
	        },
   })]
})
```

Added same logo in /src/content/docs/index.mdx 
as image for hero at frontmatter
```js
---
hero:
   image: 
       file: ../../../src/assets/logo.svg
---

```

After pnpm run build i'm getting 2 versions of same logo.svg in /dist/_astro/ folder with 2 type of hashes in filename as shown on screenshot

Screenshot 2024-07-30 at 22 20 00

Same stuff in reproduced stackblitz - adding logo to header through astro.config.mjs - doubles houston.webp in dist folder

**Probably it is unnecessary doubling of asset or am i doing smth wrong?**

https://stackblitz.com/edit/github-spzcpm?file=astro.config.mjs

### Participation

chappelo:
#### Description

- partially addresses #2161 by implementing Image for Logo, however, does not stop svg from being duplicated.

### DIFF
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -16,6 +16,7 @@ const { siteTitle, siteTitleHref } = Astro.props;
 					height={logos.dark.height}
 					width={logos.dark.width}
 					alt={config.logo.alt}
+					loading="eager"
 				/>
 				{/* Show light alternate if a user configure both light and dark logos. */}
 				{!('src' in config.logo) && logos.light && (
@@ -25,6 +26,7 @@ const { siteTitle, siteTitleHref } = Astro.props;
 						height={logos.light.height}
 						width={logos.light.width}
 						alt={config.logo.alt}
+						loading="eager"
 					/>
 				)}
 			</>
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -4,22 +4,40 @@ import config from 'virtual:starlight/user-config';
 import type { Props } from '../props';
 import { Image } from 'astro:assets';
 const { siteTitle, siteTitleHref } = Astro.props;
+
+const isDarkSvg = logos.dark?.format === 'svg';
+const isLightSvg = logos.light?.format === 'svg';
 ---
 
 <a href={siteTitleHref} class="site-title sl-flex">
 	{
 		config.logo && logos.dark && (
 			<>
-				<Image
-					class:list={{ 'light:sl-hidden': !('src' in config.logo) }}
-					src={logos.dark}
-					height={logos.dark.height}
-					width={logos.dark.width}
-					alt={config.logo.alt}
-					loading="eager"
-				/>
-				{/* Show light alternate if a user configure both light and dark logos. */}
-				{!('src' in config.logo) && logos.light && (
+				{!isDarkSvg && (
+					<Image
+						class:list={{ 'light:sl-hidden': !('src' in config.logo) }}
+						src={logos.dark}
+						height={logos.dark.height}
+						width={logos.dark.width}
+						alt={config.logo.alt}
+						loading="eager"
+					/>
+				)}
+				{isDarkSvg && (
+					<img
+						class:list={{ 'light:sl-hidden': !('src' in config.logo) }}
+						src={logos.dark.src}
+						alt={config.logo.alt}
+						loading="eager"
+					/>
+				)}
+			</>
+		)
+	}
+	{
+		config.logo && logos.light && (
+			<>
+				{!isLightSvg && (
 					<Image
 						class="dark:sl-hidden"
 						src={logos.light}
@@ -28,6 +46,9 @@ const { siteTitle, siteTitleHref } = Astro.props;
 						alt={config.logo.alt}
 						loading="eager"
 					/>
+				)}
+				{isLightSvg && (
+					<img class="dark:sl-hidden" src={logos.light.src} alt={config.logo.alt} loading="eager" />
 				)}
 			</>
 		)
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -2,6 +2,7 @@
 import { logos } from 'virtual:starlight/user-images';
 import config from 'virtual:starlight/user-config';
 import type { Props } from '../props';
+import { Image } from 'astro:assets';
 const { siteTitle, siteTitleHref } = Astro.props;
 ---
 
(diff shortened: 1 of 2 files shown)


packages/starlight/components/SiteTitle.astro
```diff
@@ -2,28 +2,29 @@
...
-				<img
+				<Image
 					class:list={{ 'light:sl-hidden': !('src' in config.logo) }}
-					alt={config.logo.alt}
-					src={logos.dark.src}
-					width={logos.dark.width}
+					src={logos.dark}
 					height={logos.dark.height}
+					width={logos.dark.width}
+					alt={config.logo.alt}
```
delucis, line 18: The `<Image>` component loads the image lazily by default, so we should explicitly override that behaviour here given this is high priority content:

```suggestion
					alt={config.logo.alt}
					loading="eager"
```

packages/starlight/components/SiteTitle.astro
```diff
@@ -2,28 +2,29 @@
...
 				{/* Show light alternate if a user configure both light and dark logos. */}
-				{!('src' in config.logo) && (
-					<img
+				{!('src' in config.logo) && logos.light && (
+					<Image
 						class="dark:sl-hidden"
+						src={logos.light}
+						height={logos.light.height}
+						width={logos.light.width}
 						alt={config.logo.alt}
```
delucis, line 27: And then same here:

```suggestion
						alt={config.logo.alt}
						loading="eager"
```

delucis (commented):
Thanks for tackling this @chappelo! I know you marked this as a draft, but I saw you were making good progress, so thought I’d leave some early feedback.

packages/starlight/components/SiteTitle.astro
chappelo: Understood, thanks.

chappelo:
@delucis thanks for the feedback. would be eternally grateful if you could help clarify a couple things regarding the expected outcomes 🙇 : 

`<Image>` component now used in `<SiteTitle>`:
The Image asset is being referened twice.
1. Once in index.mdx:
```
hero:
  tagline: Congrats on setting up a new Starlight project!
  image:
    file: ../../assets/houston.webp
```
    
2. A second time in astro config:
```
logo: {
  src: './src/assets/houston.webp',
},
```
Upon running npm build, three images are processed:
image

A 400x400 Hero `<Image>`(houston.CZZyCf7p_tstkr)
A 800x800 Logo `<Image>`(houston.CZZyCf7p_Z1FoQLb)
unused: houston.CZZyCf7p

I should be aiming to remove the unused asset  - which should have been automatically removed ?
https://github.com/withastro/starlight/issues/2161#issuecomment-2283933628 

Furthermore, if we want to reduce the output images down to one, I should further set the Logo's `<Image>` props too also be 400x400, so we don't build an 800x800 and 400x400 of the same image.

delucis:
> I should be aiming to remove the unused asset - which should have already been automatically removed ? [#2161 (comment)](https://github.com/withastro/starlight/issues/2161#issuecomment-2283933628)

That issue is specific to SVGs — for other image formats we definitely still want two different outputs because they’ll be different sizes as you noticed. That said, I don’t know if Astro is smart enough not to duplicate even SVGs, but that would be the test case rather than the `.webp` Houston image.

chappelo:
> > I should be aiming to remove the unused asset - which should have already been automatically removed ? [#2161 (comment)](https://github.com/withastro/starlight/issues/2161#issuecomment-2283933628)
> 
> That issue is specific to SVGs — for other image formats we definitely still want two different outputs because they’ll be different sizes as you noticed. That said, I don’t know if Astro is smart enough not to duplicate even SVGs, but that would be the test case rather than the `.webp` Houston image.

ahh woops I got carried with webp ..Will investigate with SVG! Thank you.

chappelo:
> > I should be aiming to remove the unused asset - which should have already been automatically removed ? [#2161 (comment)](https://github.com/withastro/starlight/issues/2161#issuecomment-2283933628)
> 
> That issue is specific to SVGs — for other image formats we definitely still want two different outputs because they’ll be different sizes as you noticed. That said, I don’t know if Astro is smart enough not to duplicate even SVGs, but that would be the test case rather than the `.webp` Houston image.

Good evening, sorry for the wait.

I've done some digging, and as you already stated, the Image component changes alone aren't enough - the SVGs are still being duplicated. I've tried playing w/ config and some methods from /assets but to no avail.

I assume this should be addressed in core astro build process, not starlight(?). I'd happily do it, if you think that is the appropriate step.

note: I'll take this off draft for now.

delucis:
Aha, good to know. My guess would be that because Astro hashes the options used with `<Image>` and `getImage()` when creating the image hash, those are taken into account even when the different `width`/`height` technically aren‘t impacting the output file, generating two hashes. Thanks for digging into it.

Bit fiddly to work around: either we manually switch to a regular `` when we detect an SVG was passed to Starlight or like you say, look into fixing this in Astro itself.

chappelo:
> Aha, good to know. My guess would be that because Astro hashes the options used with `<Image>` and `getImage()` when creating the image hash, those are taken into account even when the different `width`/`height` technically aren‘t impacting the output file, generating two hashes. Thanks for digging into it.
> 
> Bit fiddly to work around: either we manually switch to a regular `` when we detect an SVG was passed to Starlight or like you say, look into fixing this in Astro itself.

Cool thanks, will take another look this weekend!

chappelo:
> Aha, good to know. My guess would be that because Astro hashes the options used with `<Image>` and `getImage()` when creating the image hash, those are taken into account even when the different `width`/`height` technically aren‘t impacting the output file, generating two hashes. Thanks for digging into it.
> 
> Bit fiddly to work around: either we manually switch to a regular `` when we detect an SVG was passed to Starlight or like you say, look into fixing this in Astro itself.

Good evening, SVG appears to no longer double with the conditional svg checks!

So I've implemented the checks in the relevant components which use the Image component: Hero, SiteTitle.

Had to play around w/ SiteTitle render logic abit to handle cases like these:
image

Any issues let me know :)

delucis:
Sorry we haven’t had time to prioritize this. Given quite some time has passed and this isn’t an obvious fix, I think I’ll close this. Thanks again for trying to tackle this — I appreciate it!

PR #2269, opened 2024-08-31
Docs - add an example to the i18n custom labels section

BuckyBuck135:
#### Description

_- What does this PR change? Give us a brief description._
This PR adds some more detail to the "Extend translation schema" section of the i18n guide by adding an example of how to use the new custom label inside a component. indeed, as i was going through this process on my project, it took me a while to understand that I needed to use the `labels` props. I thought it could be a useful addition to the Starlight docs.

_- Did you change something visual? A before/after screenshot can be helpful._
Yes. I added a `<Steps>` components and a code block. See screenshots.

**Before**
before

**After**
after

I hope it's helpful.
Thank you,
Geoffrey

P.S. Sorry, I don't know why there are updates to `settings.json` and 'package.json`.

### DIFF
--- a/.astro/settings.json
+++ b/.astro/settings.json
@@ -1,0 +1,5 @@
+{
+	"_variables": {
+		"lastUpdateCheck": 1725113954442
+	}
+}
--- a/docs/src/content/docs/guides/i18n.mdx
+++ b/docs/src/content/docs/guides/i18n.mdx
@@ -253,26 +253,40 @@ You can provide translations for additional languages you support — or overrid
 
 ### Extend translation schema
 
-Add custom keys to your site’s translation dictionaries by setting `extend` in the `i18nSchema()` options.
-In the following example, a new, optional `custom.label` key is added to the default keys:
+You can add custom keys to your site’s translation dictionaries. In the following example, a new, optional `custom.label` key is added to the default keys:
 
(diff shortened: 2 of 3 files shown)


delucis:
Thanks for this @BuckyBuck135! Going to check with @HiDeoo but we are planning to change how this works in #1923, so it might not make sense to add this information at this point (although it is very useful!)

HiDeoo:
Indeed, as part of #1923, the documentation will now contain a brand new ["Accessing UI strings"](https://deploy-preview-1923--astro-starlight.netlify.app/guides/i18n/#accessing-ui-strings) section (still a work in progress) that will explain how to access UI strings using the new approach and all the helper functions that will be available.

image

Considering the current progress, I agree with Chris that we can potentially skip documenting and translating the old approach at this point.

I'll close this PR but thanks for the contribution nonetheless and also spotting an area that definitely needed improvement! :raised_hands:

I hope the new approach and its documentation will be more helpful when it's ready, and feel free to take a look at the link I shared above to let us know if you have any feedback or suggestions.

BuckyBuck135:
That new doc looks very promising and perfectly conveys the example I was trying to give.
 Thank you both for your technical and communication skills. Keep up the great work.
I'll find another way to get my first contributor alert :)

PR #2275, opened 2024-09-02
Fix Visual Studio Code workspace settings being ignored

HiDeoo:
I [learned](https://github.com/microsoft/vscode/issues/32693) yesterday that Visual Studio Code workspace settings (e.g. the `docs/.vscode/` folder) are not working if they are not in the root of the workspace.

Opening the Starlight monorepo in Visual Studio Code and checking the available launch tasks does indeed confirm this behavior as no tasks are available.

SCR-20240902-oaxc

I never noticed that issue as I don't rely on Visual Studio Code to launch the debugger.

This PR moves the `docs/.vscode/` folder to the root of the repository to fix the issue. Opening the monorepo in Visual Studio Code with this change applied now shows the available launch tasks.

SCR-20240902-obdv

I also updated the launch task to change the `cwd` to the `docs` folder so that setting breakpoints in the source code and starting the debugger works as expected.

SCR-20240902-obuu


delucis (approved):
Ah, I guess this makes sense — I never consciously configured these, so this is just the default config file that came from running `create astro` to scaffold the docs site. (It dates to the [very first commit in this repo](https://github.com/withastro/starlight/commit/4e18d9176ab2498bff8d3c4a3bd1aa8870c87754) 😁)

Happy to move these to help folks who may not have the Astro extension already and could use these.

PR #2279, opened 2024-09-03
Parse `<StarlightPage />` frontmatter asynchronously 

### LINKED ISSUE #2270: Passing a `frontmatter` containing type `reference()` causes an error in pages that use the `<StarlightPage />` component.

0.26.1

4.15.1

pnpm@9.9.0

Windows

Edge, Chrome, Firefox

### Describe the Bug

Passing a `frontmatter` containing type `reference()` causes an error in pages that use the `<StarlightPage />` component.

image

Clone the repository, run the `pnpm dev` command, and then go to <http://localhost:4321/page/> to see the error.

Reconstructive process: https://github.com/zhoushengdao/starlight-page-reference-error/commit/6fe29cb0b48cf1ff4df60533c0707103aea9802e

https://github.com/zhoushengdao/starlight-page-reference-error

### Participation

HiDeoo:
#### Description

- Closes #2270

This PR updates how the frontmatter data passed down to the `<StarlightPage />` component is parsed. Starting with Astro 4.14.0, schemas containing collection references contains an [async transform](https://github.com/withastro/astro/blame/a79a8b0230b06ed32ce1802f2a5f84a6cf92dbe7/packages/astro/src/content/runtime.ts#L585-L592) which would failed to be parsed due to our usage of Zod `safeParse()` instead of `safeParseAsync()`.

I decided to create a new `parseAsyncWithFriendlyErrors()` instead of always parsing schemas asynchrounously as `parseWithFriendlyErrors()` is used in the plugin `setup` hook `updateConfig` option and I did not want to change that behavior. The new function is now only used in this specific case.

Regarding testing, I decided to not add a test as we cannot really easily have a unit test with a custom schema using `extend` and having an E2E test for this would require using a different Astro version. Creating a unit test simply calling `parseAsyncWithFriendlyErrors()` with a schema containing an async refinement or transform doesn't provide much value imo. Let me know if you think otherwise.

Altho, I used the reproduction in #2270 and tested the fix using Astro `4.13.4` and the latest version and both are now working as expected.

### DIFF
--- a/.changeset/spicy-suns-marry.md
+++ b/.changeset/spicy-suns-marry.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue with frontmatter schemas containing collection references used with the `<StarlightPage />` component and an Astro version greater than `4.14.0`.
--- a/packages/starlight/utils/error-map.ts
+++ b/packages/starlight/utils/error-map.ts
@@ -25,11 +25,31 @@ export function parseWithFriendlyErrors<T extends z.Schema>(
 	input: z.input<T>,
 	message: string
 ): z.output<T> {
-	const parsedConfig = schema.safeParse(input, { errorMap });
-	if (!parsedConfig.success) {
-		throw new AstroError(message, parsedConfig.error.issues.map((i) => i.message).join('\n'));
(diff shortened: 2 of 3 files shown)


delucis (approved):
LGTM! Thanks for the investigations and quick fix @HiDeoo!

PR #2281, opened 2024-09-03
Ensure unhandled directives are restored without any extra whitespace

HiDeoo:
#### Description

This PR fixes an issue reported on Discord where the current process of restoring unhandled directives could change the expected output if the unhandled directive was followed by some characters due to extra whitespace being added.

Assuming the following input:

```markdown
## Environment variables (astro:env)
```

**Before:** (notice the space before the trailing `)`)

SCR-20240903-qkiu

**After:**

SCR-20240903-qklg

### DIFF
--- a/packages/starlight/integrations/asides.ts
+++ b/packages/starlight/integrations/asides.ts
@@ -66,7 +66,18 @@ function transformUnhandledDirective(
 ) {
 	const textNode = {
 		type: 'text',
-		value: toMarkdown(node, { extensions: [directiveToMarkdown()] }),
+		/**
+		 * `mdast-util-to-markdown` assumes that the tree represents a complete document (as it's an
+		 * AST and not a CST) and to follow the POSIX definition of a line (a sequence of zero or more
+		 * non- <newline> characters plus a terminating <newline> character), a newline is
+		 * automatically added at the end of the output so that the output is a valid file.
+		 * In this specific case, we can safely remove the newline character at the end of the output
+		 * before replacing the directive with its value.
+		 *
+		 * @see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206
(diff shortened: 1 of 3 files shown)


packages/starlight/integrations/asides.ts
```diff
@@ -66,7 +66,18 @@ function transformUnhandledDirective(
...
+		 * AST and not a CST) and to follow the POSIX definition of a line (a sequence of zero or more
+		 * non- <newline> characters plus a terminating <newline> character), a newline is
+		 * automatically added at the end of the output so that the output is a valid file.
+		 * In this specific case, we can safely remove the newline character at the end of the output
+		 * before replacing the directive with its value.
+		 *
+		 * @see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206
+		 * @see https://github.com/syntax-tree/mdast-util-to-markdown/blob/fd6a508cc619b862f75b762dcf876c6b8315d330/lib/index.js#L79-L85
+		 */
+		value: toMarkdown(node, { extensions: [directiveToMarkdown()] }).replace(/\n$/, ''),
```
delucis, line 80: Given our recent experiences with RegExp replaces in the sidebar, do you think it’s worth refactoring this for something like?
```js
	let markdown = toMarkdown(node, { extensions: [directiveToMarkdown()] });
	if (markdown.at(-1) === '\n') markdown = markdown.slice(0, -1);
```
Or do you think we hit this rarely enough that it doesn’t matter?

delucis (approved):
Makes sense — thanks for figuring that out! Left a small note re: performance, but seems solid to me.

packages/starlight/integrations/asides.ts
HiDeoo: Definitely, I should have spotted that. And that's also why I love reviews, always happy to make changes, specially when we spot a potential improvement, as small as it may be. Thanks for the review, I've updated the PR.

delucis (approved):
With all the time we save avoiding RegExp replaces, I have time to double approve!

PR #2282, opened 2024-09-04
Add WCAG AAA colour contrast option to theme editor

delucis:
#### Description

- Adds a toggle to the Starlight theme editor to allow people to opt in to shades that pass WCAG’S AAA requirements (7:1 contrast ratio for body text).
- Preview: https://deploy-preview-2282--astro-starlight.netlify.app/guides/css-and-tailwind/#color-theme-editor
- The `getPalettes()` function used to generate a colour palette now takes a minimum contrast to enforce and uses it for the shades which are currently marginal (pass AA requirements but not AAA).
- Contrast calculations are done with `culori`’s [`wcagContrast()`](https://culorijs.org/api/#wcagContrast) method. I used that to compute a contrast and then colours without sufficient contrast have their luminance incremented/decremented in 0.5% steps until the contrast requirement is met.

#### Browser testing

Tested on macOS with:

- Chrome 128.0.6613.113
- Firefox 129.0.2 and 130.0
- Safari 17.6

Tested on Android with:

- Firefox

### DIFF
--- a/docs/src/components/theme-designer/contrast-level.astro
+++ b/docs/src/components/theme-designer/contrast-level.astro
@@ -1,0 +1,86 @@
+---
+export interface Props {
+	labels: {
+		label: string;
+	};
+}
+const { labels = { label: 'WCAG Contrast Level' } } = Astro.props;
+---
+
+<contrast-level-toggle class="sl-flex">
+	<fieldset class="not-content">
+		<legend>{labels.label}</legend>
+		<div class="sl-flex">
+			<label class="sl-flex">
+				<input type="radio" name="contrast-level" value="4.5" checked />
+				AA
+			</label>
+			<label class="sl-flex">
+				<input type="radio" name="contrast-level" value="7" />
+				AAA
+			</label>
+		</div>
+	</fieldset>
+</contrast-level-toggle>
+
+<script>
+	import { minimumContrast } from './store';
+
+	class ContrastLevelToggle extends HTMLElement {
+		#fieldset = this.querySelector('fieldset')!;
+		constructor() {
+			super();
+			this.#fieldset.addEventListener('input', (e) => {
+				if (e.target instanceof HTMLInputElement) {
+					const contrast = parseFloat(e.target.value);
+					minimumContrast.set(contrast);
+				}
+			});
+		}
+	}
+
+	customElements.define('contrast-level-toggle', ContrastLevelToggle);
+</script>
+
+<style>
+	fieldset {
+		border: 0;
+		padding: 0;
+	}
+	fieldset > * {
+		float: left;
+		float: inline-start;
+		vertical-align: middle;
+	}
+	legend {
+		color: var(--sl-color-white);
+		margin-inline-end: 0.75rem;
+	}
+	label {
+		align-items: center;
+		padding: 0.25rem 0.75rem;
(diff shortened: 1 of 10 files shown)


docs/src/components/theme-designer/contrast-level.astro
```diff
@@ -0,0 +1,86 @@
...
+		padding: 0.25rem 0.75rem;
+		gap: 0.375rem;
+		background-color: var(--sl-color-gray-6);
+		font-size: var(--sl-text-xs);
+		cursor: pointer;
+	}
+	label:focus-within {
+		outline: 2px solid;
+		outline-offset: -3px;
+	}
```
HiDeoo, line 70: Should we apply some similar treatment to the `<preset-picker>` buttons while at it? Very difficult to notice the currently selected one atm?

docs/src/components/theme-designer/contrast-level.astro
```diff
@@ -0,0 +1,86 @@
...
+	labels: {
+		label: string;
+	};
+}
+const { labels = { label: 'WCAG Contrast Level' } } = Astro.props;
+---
+
+<contrast-level-toggle class="sl-flex">
+	<fieldset class="not-content">
+		<legend>{labels.label}</legend>
```
HiDeoo, line 12: This is pretty technical, should we link somehow to a reference or something describing what `AA` & `AAA` are?

HiDeoo (approved):
What a lovely PR, really amazing addition to the theme editor :clap:

docs/src/components/theme-designer/contrast-level.astro
delucis: Added in 5072db0 — thanks!

docs/src/components/theme-designer/contrast-level.astro
delucis: Didn’t find many amazing resources. Maybe the best link I dug up so far is the [MDN colour contrast guide](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast).

Not quite sure how best to integrate it though.

One option is a sentence in the intro text:

theme editor and preamble with a sentence reading "Use the Contrast Level option to specify which of the Web Content Accessibility Guideline colour contrast standards to meet"

Or less intrusive could be some kind of icon link in the group label itself, but also not super clear what the icon is in this case:

theme editor with an information icon next to the "WCAG Contrast Level" label

delucis: Alternatively could also label things with `4.5:1` / `7:1` although not sure that’s any _more_ helpful :Swe

delucis: Or as another option:

contrast level picker with two options: "Minimum (AA)" and "Enhanced (AAA)"

HiDeoo: Thanks for trying and sharing all the options 🙌 

As simple as it is, I think personally prefer the first option with a sentence and a link (and indeed, the MDN one seems like the best to link to).

delucis: Yeah, I think I do too. Added! Thanks for the helpful feedback.

PR #2284, opened 2024-09-05
i18n(ko-KR): update `css-and-tailwind.mdx`

jsparkdev:
#### Description

- update `css-and-tailwind.mdx`
- #2282


delucis (approved):
Thank you @jsparkdev! Let me know if you think the AA/AAA labels also needed translating. Because they refer to the WCAG levels, I left them as static, but I guess in theory some languages might want to use a different label.

jsparkdev:
@delucis If you mean the area that displays the image below, it would be better not to translate it. thank you
스크린샷 2024-09-05 오후 5 49 41

delucis:
Yes, exactly! Thanks for confirming 💖

PR #2285, opened 2024-09-05
Add support for translating sidebar badges

HiDeoo:
#### Description

- Closes #1721

This PR adds the ability to translate sidebar badges using the same approach as the site title. A badge text can be a string, or for multilingual sites, an object with values for each different locale.

```js
starlight({
  sidebar: [
    {
      slug: "test",
      badge: {
        en: "New",
        "pt-BR": "Novo",
      },
    },
  ],
});
```

This syntax is compatible with the string/object syntax to define badges. To support this feature, a new type of badge was added (`I18nBadge`) and can only be used in the sidebar configuration.


docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -520,6 +520,65 @@ Browsing the documentation in Brazilian Portuguese will generate the following s
...
+						text: {
+							en: 'Outdated',
+							'pt-BR': 'Descontinuado',
+						},
+						variant: 'caution',
+					},
+				},
+			],
+		},
+	],
```
delucis, line 556: Awesome stuff! Happy to add this.

I wonder if it would be simpler to only support localized strings using the `text` property though? I find it a bit weird to have two config options that are objects but that mean very different things depending on what the keys are, i.e. one object is “record of locales to labels” the other is “precisely typed config object with known keys”. (Without testing, I’d guess this also interferes with IDE autocomplete?)

Also, imagine this badge:
```js
badge: {
  en: 'New',
  fr: 'Nouveau',
}
```
If you decide you want to make it a `success` variant, this doesn’t work:
```diff
badge: {
  en: 'New',
  fr: 'Nouveau',
+  variant: 'success',
}
```
So overall, I think I’m in favour of a single syntax rather than two. What do you think?

HiDeoo: Ny initial implementation was what you described, then I somehow managed to convince myself that if you were using the shorthand notation, you would prefer to use that too with translated labels. I ended up not liking it personally but couldn't decide if I should remove it or not.

If you're having the same feeling too, I'll be more than happy to remove it (and it'll make the implementation simpler). I'll update the PR accordingly. Thanks again for the great input 🙌

delucis: Haha, yeah, going from `badge: 'New'` to `badge: { text: { en: 'New' } }` feels superficially like a lot, but I think in practice it’s OK and even it is slightly verbose, it definitely is clearer to explain/understand. Plus once you have a couple of locales the difference is kind of not so bad, e.g.
```js
badge: { text: { en: 'New', fr: 'Nouveau', de: 'Neu', it: 'Nuovo' } }
// vs
badge: { en: 'New', fr: 'Nouveau', de: 'Neu', it: 'Nuovo' }
```

kevinzunigacuellar: I second having only one config option with locales inside `text`. It's a key longer but a very small price to pay to have **localized** badges 💜.

HiDeoo: Thanks you both, super appreciated. I have updated the PR accordingly, definitely an improvement 👍

packages/starlight/utils/navigation.ts
```diff
@@ -446,3 +460,38 @@ function stripExtension(path: string) {
...
+		config.defaultLocale?.lang || config.defaultLocale?.locale || BuiltInDefaultLocale.lang;
+	const defaultText = text[defaultLang];
+
+	if (!defaultText) {
+		throw new AstroError(
+			`The badge text for "${itemLabel}" must have a key for the default language "${defaultLang}".`,
+			'Update the Starlight config to include a badge text for the default language.\n' +
+				'Learn more about sidebar badges internationalization at https://starlight.astro.build/guides/sidebar/#internationalization-with-badges'
+		);
+	}
```
delucis, line 494: Noting that one way to avoid needing this would be to use the same API we use for sidebar items themselves:

```js
{
  label: 'Andromeda',
  translations: { fr: 'Andromède' },
}
```

I don’t have a strong opinion — we use the `string | Record<string, string>` form for the site title, but the `label`/`translations` form here in the sidebar currently. I’m pretty sure I opted for the separate `translations` property when designing the sidebar API to enforce the presence of a default label at the schema level, but having `translations` keys proliferating everywhere also isn’t amazing.

Happy to keep this as is, but wanted to note this while approving!

delucis (approved):
LGTM! Thanks @HiDeoo — ready for the next minor 🫡

HiDeoo:
Updated the test coverage thresholds in https://github.com/withastro/starlight/pull/2285/commits/fd48b6b64ad3df725acd7a58e0df97774b727798 (no other changes since the approval)

PR #2288, opened 2024-09-06
Prevent Zod errors from crashing build

matthewp:
#### Description

- Errors are weakly typed and some `undefined`s were escaping due to a missing `typeof`, which this fixes.

### DIFF
--- a/.changeset/famous-glasses-clean.md
+++ b/.changeset/famous-glasses-clean.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):

delucis (approved):
Not quite sure what changed to cause us to start hitting this, but I believe you that the `error` object is inconsistent.

Might be the new use of async transforms perhaps (add by content layer) — could be a Zod issue of doing things subtly differently in `safeParse()` and `safeParseAsync()`.

Either way, approved!

PR #2289, opened 2024-09-06
Update astro dependency to 4.15.3 across monorepo

delucis:
#### Description

This PR updates the version of Astro we are using in the monorepo (for docs builds, tests, etc.) to the latest 4.15.4 release.

There were two additional changes needed:

1. Deduping Vite versions to ensure Astro and Vitest were both using the same version
2. Adding the `--noSync` flag to our type check command so that running type checks from the monorepo root doesn’t generate an `.astro/types.d.ts`

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -18,7 +18,7 @@
     "@astrojs/starlight": "workspace:*",
     "@lunariajs/core": "^0.1.1",
     "@types/culori": "^2.1.1",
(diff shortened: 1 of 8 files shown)


HiDeoo (approved):
Looks good to me :star2: 

Small note to remember that the new fixtures & example in #1255 & #2249 will need to be updated too.

PR #2296, opened 2024-09-07
i18n(fr): update `reference/configuration`

HiDeoo:
#### Description

This PR updates the French version of the `reference/configuration` page with the changes from #1255.


docs/src/content/docs/fr/reference/configuration.mdx
```diff
@@ -455,6 +455,16 @@ Définit si le système de recherche du site par défaut de Starlight, [Pagefind
...
 
+### `prerender`
+
+**Type :** `boolean`  
+**Par défaut :** `true`
+
+Définit si les pages générées par Starlight doivent être pré-rendues en HTML statique ou rendues à la demande par un [adaptateur SSR](https://docs.astro.build/fr/guides/server-side-rendering/).
+
+Les pages Starlight sont pré-rendues par défaut.
+Si vous utilisez un adaptateur SSR et que vous souhaitez générer les pages Starlight à la demande, définissez `prerender: false`.
```
delucis, line 466: My brain keeps stumbling of “et que vous” instead of “et vous” but I trust you obviously 😁

delucis (approved):
LGTM! Thanks @HiDeoo 🚀

docs/src/content/docs/fr/reference/configuration.mdx
delucis: Oh yeah and I see the exact same formulation in the Markdoc guide translation too. I’ll need to go learn what the purpose of the “que” is.

HiDeoo: Oh, it's not really mandatory in this case, it's used to avoid a repetition as the sentence technically should be "Si vous utilisez un adaptateur SSR et si vous souhaitez générer les pages Starlight à la demande, …" but to avoid the double "si", we use "que" (which is a "conjonction de subordination" here) for the second one.

[Reference](https://dictionnaire.lerobert.com/guide/conjonctions-de-subordination#G02100303)

Exemple from the reference:

> **_Si_** nous cassons quelque chose et **_qu_**'elle s'en aperçoit… (= **_Si_** nous cassons quelque chose et **_si_** elle s'en aperçoit…)

delucis: Oh, thank you! Wasn’t at all familiar with that usage of “que”.

PR #2301, opened 2024-09-07
Update `yummacss.com.png` thumbnail

rrenildopereiraa:
#### Description

- This change updates the `yummacss.com.png`

The old thumbnail was taken in a **beta version**, while the new one was taken in the **stable version**.

Comparison Image


delucis (approved):
Thanks for taking the time to update it!

PR #2308, opened 2024-09-09
i18n(fr): update `reference/configuration`

HiDeoo:
#### Description

Just realized that when updating the French version of the `reference/configuration` page in #2296, I missed one sentence 🙈


delucis (approved):
Oh yeah! I missed it too 😁

PR #2322, opened 2024-09-11
Cascade Layers

HiDeoo:
#### Description

- Closes #1048

This PR is a WIP that refactors all Starlight CSS to use [cascade layers](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers).

At the moment, Starlight will internally uses the following nested layers:

- `starlight.base`
- `starlight.reset`
- `starlight.core`
- `starlight.content`
- `starlight.components`
- `starlight.utils`

This slightly [reduces browser support](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+99%2C+Edge+%3E%3D+99%2C+Firefox+%3E%3D+98%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all) by 0.3% (`Chrome >= 88` → `Chrome >= 99`).

#### Testing

As I quickly realized that manually inspecting would not be practical for this PR to ensure there are no regressions, I created a new temporary package to help test the changes (`packages/local-prod-visual-diff`) that can be run using `pnpm test` from that directory.

For a list of routes, the tool will:

- Grab an entire page screenshot of the production version of the route (if it doesn't already exist locally)
- Start a local server and grab an entire page screenshot of the local version of the route
- Visually diff the two screenshots
- If there are significant differences, a diff image with differences in yellow/red will be generated and the test will fail

This has been very helpful to spot some tiny issues during the refactor, e.g. the following diff image that shows a tiny difference in the header `<Select>` component icons:

header

Another one showing an issue with the content of the last step in a set of `<Steps>`:

steps

Of course, this is only a bonus and this PR would still be a good Talking & Doc'ing candidate with many different eyes/browsers.

#### Remaining tasks

- [x] Tailwind v4 support
- [x] Figure out which layers we want to use (I mostly maximized the number of layers in the first implementation and we can reduce them during the review)
- [x] ~~Figure out if we should namespace the Tailwind layers, e.g. `tailwind.base` like we do at the moment for Starlight or if we should just use the default layers (not doing that)~~
- [x] Update the documentation theme editor Tailwind output
- [x] Address and remove the `TODO(HiDeoo)` comment in the `docs/astro.config.mjs` file
- [x] Remove the `packages/local-prod-visual-diff` package
- [ ] ~~See if we should remove the default scoped style strategy enforced by Starlight? (not doing that)~~ 
  - [ ] ~~We would need to refactor various components to handle `data-astro-cid-*` attributes. Not confident this PR is the right place to do that.~~
- [ ] Improve documentation
- [x] Add changesets for all impacted packages.


packages/starlight/style/layers.css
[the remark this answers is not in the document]
mayank99: What I've found is that order of imports is not a reliable way to ensure order of styles. I've run into many instances where Vite/Astro jumbles up the CSS order.

Should we move move this layer statement into a `<style>` tag in the `<head>`? Ideally this would be placed _after_ the user's own `<head>` content, but _before_ all the bundled stylesheets.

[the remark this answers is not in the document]
HiDeoo: This is something I'll need to investigate as this is not something I've encountered so far in this first iteration. The suggested approach could also decrease the DX for user customization. Adding that point to my todo list of things to investigate and I'll get back with my findings.

HiDeoo:
Small note that [Tailwind CSS v4.0 Beta](https://tailwindcss.com/docs/v4-beta) adds [native CSS cascade layers](https://tailwindcss.com/docs/v4-beta#native-css-cascade-layers) support.

mayank99:
Tailwind v4's layer implementation has good and bad parts.

The bad: they implicitly create very generic layer names (i.e. not namespaced).

The good: they allow control over importing their stylesheets(s) so you can put their layers inside your own layer or even change the names.

```css
@import "tailwindcss" layer(💨);

/* or */

@import "tailwindcss/base" layer(💨base);
@import "tailwindcss/theme" layer(💨theme);
@import "tailwindcss/utilities" layer(💨utilities);
```

HiDeoo:
Just updated the PR with the following changes:

- Updated the PR with all the changes in the codebase since the last update.
- Updated the list of routes used by the temporary visual diffing tool included in the PR (includes all the new component pages).
  - No new diffing issues were found with the updated content and new pages — which I assume is a good thing 😅 
- Refactored the Starlight Tailwind plugin to be pure CSS for Tailwind v4.
- Updated the Tailwind example to use Tailwind v4 and the new Starlight Tailwind CSS.
- Updated the Tailwind section of the documentation to reflect the changes (but still a ver rough draft).

Note that the documentation theme editor Tailwind output has not been updated yet and I'll do that in a follow-up commit.

.changeset/shaggy-games-retire.md
```diff
@@ -0,0 +1,43 @@
...
+export default defineConfig({
++ markdown: {
++   shikiConfig: {
++     theme: 'css-variables',
++   },
++ },
+});
+```
+
+Additionally, you can use [custom CSS](https://starlight.astro.build/guides/css-and-tailwind/#custom-css-styles) to control the appearance of the code blocks. Here is the previously used CSS variables for the fallback theme:
```
JusticeMatthew, line 27:
```suggestion
Additionally, you can use [custom CSS](https://starlight.astro.build/guides/css-and-tailwind/#custom-css-styles) to control the appearance of the code blocks. Here are the previously used CSS variables for the fallback theme:
```

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -48,16 +48,32 @@ Customize the styles applied to your Starlight site by providing additional CSS
...
+Tailwind CSS support in Astro projects is provided by the [Tailwind Vite plugin](https://tailwindcss.com/docs/installation/using-vite).
+Starlight provides complementary CSS to help configure Tailwind for compatibility with Starlight’s styles.
 
-The Starlight Tailwind plugin applies the following configuration:
+The Starlight Tailwind CSS applies the following configuration:
 
 - Configures Tailwind’s `dark:` variants to work with Starlight’s dark mode.
 - Uses Tailwind [theme colors and fonts](#styling-starlight-with-tailwind) in Starlight’s UI.
-- Disables Tailwind’s [Preflight](https://tailwindcss.com/docs/preflight) reset styles while selectively restoring essential parts of Preflight required for Tailwind’s border utility classes.
+- Restore essential parts of Preflight.
```
JusticeMatthew, line 76:
```suggestion
- Restores essential parts of Preflight.
```

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -93,7 +109,7 @@ If you already have a Starlight site and want to add Tailwind CSS, follow these
 
 <Steps>
 
-1.  Add Astro’s Tailwind integration:
+1.  Add Tailwind Vite plugin:
```
JusticeMatthew, line 112:
```suggestion
1.  Add the Tailwind Vite plugin:
```

JusticeMatthew (changes requested):
A few small issues I noticed

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -123,7 +139,7 @@ If you already have a Starlight site and want to add Tailwind CSS, follow these
 
     </Tabs>
 
-2.  Install the Starlight Tailwind plugin:
+2.  Install the Starlight Tailwind CSS:
```
anaxite, line 142:
```suggestion
2.  Install the Starlight Tailwind CSS plugin:
```

HiDeoo: I don't think `plugin` would be the proper word here as it could refer to what is now called "legacy plugins" in Tailwind CSS but what is being installed is indeed just CSS. Maybe "Install the Starlight Tailwind CSS declarations:" 🤔

anaxite: How about `Add the Starlight Tailwind CSS rules:`?

The main confusion for me is "install". I would use it for plugins or something that has an Install process, but not otherwise. If this has an installation process then it's fine.

HiDeoo: In that case, there is still an installation process through `npm`, e.g. `npm install` which installs a package that only ships CSS. It's similar to [`normalize.css`](https://www.npmjs.com/package/normalize.css) which can also be installed through `npm` but only ships CSS.

markgaze (commented):
Thanks for this, needed it for a project and it seems to be working pretty well for me! Just noticed a few things so thought I'd comment here 🙂

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -50,16 +50,41 @@ Customize the styles applied to your Starlight site by providing additional CSS
...
+```css /(my-(?:reset|layer))/
+@layer starlight.base,
+	starlight.reset,
+	my-reset,
+	starlight.core,
+	starlight.content,
+	starlight.components,
+	starlight.utils,
+	my-layer;
+```
```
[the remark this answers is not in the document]
delucis: Thanks for catching that! Do people have opinions on how helpful that would be? I guess you could still add your reset like this?
```css
@layer starlight.base,
	starlight.reset,
	starlight.my-reset,
	starlight.core,
	starlight.content,
	starlight.components,
	starlight.utils,
	my-layer;
```
Or are top-level layers preferable? Something like
```css
@layer starlight-base,
	starlight-reset,
	my-reset,
	starlight-core,
	starlight-content,
	starlight-components,
	starlight-utils,
	my-layer;
```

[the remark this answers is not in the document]
mayank99: I think I would do something like this:

```css
@layer my-reset, starlight, my-overrides;
```

Generally, it's best to have as few top-level layers as possible. Consumers should also not need to mess with Starlight's nested layers.

(I can share more feedback after I get the chance to play with it. Just trying to figure out how to do that)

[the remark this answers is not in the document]
delucis: > Just trying to figure out how to do that

Using https://gitpkg.vercel.app/ might be easiest?

1. Create a test project (or use Starlight site you already have set up
2. Update the Starlight dep:
	```sh
	npm install 'https://gitpkg.now.sh/HiDeoo/starlight/packages/starlight?hd-feat-cascade-layers'
	```
	
Alternatively you could just play around in the `docs` directory inside this monorepo using this branch.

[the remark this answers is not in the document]
markgaze: > Thanks for catching that! Do people have opinions on how helpful that would be? I guess you could still add your reset like this? Although it would require you naming your layer with a `starlight.` prefix which wouldn’t work for everyone. 
> ```css
> @layer starlight.base,
> 	starlight.reset,
> 	starlight.my-reset,
> 	starlight.core,
> 	starlight.content,
> 	starlight.components,
> 	starlight.utils,
> 	my-layer;
> ```
> Or are top-level layers preferable? Something like
> ```css
> @layer starlight-base,
> 	starlight-reset,
> 	my-reset,
> 	starlight-core,
> 	starlight-content,
> 	starlight-components,
> 	starlight-utils,
> 	my-layer;
> ```

Because we already have some pre-existing layers in our component library I was trying to slot them into the right order but I was overthinking it somewhat and ended up solving it with a `.not-content` class.

I agree with @mayank99 and the top-level ones are fine and you can use your first approach if something specific needs doing (or just add your CSS to one of the existing `starlight` layers!)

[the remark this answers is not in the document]
markgaze: > > Just trying to figure out how to do that
> 
> Using https://gitpkg.vercel.app/ might be easiest?
> 
> 1. Create a test project (or use a Starlight site you already have set up)
> 2. Update the Starlight dep:
> 	```sh
> 	npm install 'https://gitpkg.now.sh/HiDeoo/starlight/packages/starlight?hd-feat-cascade-layers'
> 	```
> 	
> Alternatively you could just play around in the `docs` directory inside this monorepo using this branch.

If you're using PNPM you can also do:

```json
"@astrojs/starlight": "github:markgaze/starlight#hd-feat-cascade-layers&path:/packages/starlight",
"@astrojs/starlight-tailwind": "github:markgaze/starlight#hd-feat-cascade-layers&path:/packages/tailwind",
```

in your `package.json` and it will work.

(I'm using my fork of the fork because I fixed the Tailwind package `files` as per my other review comment)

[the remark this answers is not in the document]
mayank99: The `github:` approach sounds amazing! I'm pleasantly surprised to see there is no build step in Starlight packages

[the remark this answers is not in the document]
markgaze: > The `github:` approach sounds amazing! I'm pleasantly surprised to see there is no build step in Starlight packages

I was too! If only I'd looked first and not wasted my time trying to `pnpm patch` them 🙈😂

The docs for that approach are [here](https://pnpm.io/cli/add#install-from-a-git-repository-combining-different-parameters), in case anyone needs them that sees this!

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -50,16 +50,28 @@ Customize the styles applied to your Starlight site by providing additional CSS
 
 You can see all the CSS custom properties used by Starlight that you can set to customize your site in the [`props.css` file on GitHub](https://github.com/withastro/starlight/blob/main/packages/starlight/style/props.css).
 
+### Cascade layers
+
+Starlight uses [cascade layers](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers) internally so that any custom unlayered CSS will override the default styles.
```
mayank99, line 55: Small suggestion: This part can be slightly expanded to mention the main purpose of using layers and to include the `starlight` layer name.

```suggestion
Starlight uses [cascade layers](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers) internally for managing the order of styles. This results in a more predictable CSS order and allows for easier overrides. Any custom unlayered CSS will override the default styles from the `starlight` layer.
```

mayank99 (approved):
Finally got the chance to try it out, and I'm happy to report it's working great! 🥳

When I used `customCss`, I found that the stylesheet was linked before Starlight's own CSS. I'm not sure if this behavior is intentional and will remain this way, but it works out great for the layer use-case.

For extra surety, I also tried overriding Starlight's `Head` component and it worked as expected for placing an inline `<style>` tag with my predefined layer order before all other `<head>` content.

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -294,8 +335,8 @@ import ThemeDesigner from '~/components/theme-designer.astro';
 		file](#custom-css-styles) to apply this theme to your site.
 	</Fragment>
 	<Fragment slot="tailwind-docs">
-		The example [Tailwind config file](#styling-starlight-with-tailwind) below
-		includes generated `accent` and `gray` color palettes to use in the
-		`theme.extend.colors` configuration object.
+		The example [Tailwind theme config](#styling-starlight-with-tailwind) below
```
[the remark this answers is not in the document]
delucis: I think “example” was intentional on my part here (although it’s a slightly clunky sentence) with the idea that the code shown was an _example_ of a Tailwind config with the generated theme palette etc. and that if you already had some config set, you probably wouldn’t just copy/paste the whole thing. (Unlike the plain CSS instructions above which just say “Add the following…”

Now that the Tailwind config is in CSS we _could_ potentially just show the variables for Tailwind too now?

````md

Add the following CSS variables to the `@theme` block in your Tailwind CSS file:

```css
/* Generated accent color palettes. */
--color-accent-200: ${dark['accent-high']};
--color-accent-600: ${light.accent};
/* etc. */
```
````

File: examples/tailwind/src/tailwind.css [hunk 8]
```diff
@@ -1,8 +1,17 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@layer base, starlight.base, starlight.reset, starlight.core, starlight.content, starlight.components, starlight.utils, theme, components, utilities;
```
delucis, line 1: Same again here:
```suggestion
@layer base, starlight, theme, components, utilities;
```

packages/starlight/components/Search.astro
```diff
@@ -163,321 +162,327 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
...
+			}
 		}
 	}
 </style>
 
 <style is:global>
-	[data-search-modal-open] {
-		overflow: hidden;
-	}
+	@import url('@pagefind/default-ui/css/ui.css') layer(starlight.core);
```
delucis, line 295: Ah interesting to see how to handle a case like this 👍

packages/starlight/components/Search.astro
```diff
@@ -163,321 +162,327 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
...
+			inset-block: 0;
+			inset-inline-start: var(--sl-search-tree-diagram-inline-start);
+			width: var(--sl-search-tree-diagram-size);
+			background: var(--sl-color-gray-4);
+			-webkit-mask: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='none' stroke='currentColor' stroke-linecap='round' viewBox='0 0 16 1000' preserveAspectRatio='xMinYMin slice'%3E%3Cpath d='M8 0v1000m6-988H8'/%3E%3C/svg%3E")
+				0% 0% / 100% no-repeat;
+			mask: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='none' stroke='currentColor' stroke-linecap='round' viewBox='0 0 16 1000' preserveAspectRatio='xMinYMin slice'%3E%3Cpath d='M8 0v1000m6-988H8'/%3E%3C/svg%3E")
+				0% 0% / 100% no-repeat;
+		}
+		#starlight__search .pagefind-ui__result-nested:last-child::before {
```
delucis, line 450: Do you remember why this change was needed? I remember we saw those styles being off in the search modal but don’t remember why only `:last-child` worked now.

packages/starlight/style/print.css
```diff
@@ -1,10 +1,5 @@
...
-	/**
-	This is increased in specificity so it works even when injected into pages before regular styles.
-	TODO: Just use `:root` once we move to @layer.
-	Same goes for the use of `!important` in this file.
-	*/
-	:root[data-theme]:is(:root) {
+	:root {
 		/* Colours — force light mode when printing */
 		--sl-color-white: hsl(224, 10%, 10%);
 		--sl-color-gray-1: hsl(224, 14%, 16%);
```
delucis, line 5: Very cool seeing all these styles cleaned up! 🎉 

Do you think we should add these in a `starlight.print` layer? Or is it intentional that these aren’t layered because we want them to override user styles?

packages/starlight/components/Select.astro
```diff
@@ -37,60 +37,64 @@ interface Props {
...
+		}
 
-	@media (min-width: 50rem) {
-		select {
-			font-size: var(--sl-text-sm);
+		.caret {
+			font-size: var(--sl-caret-size);
+			inset-inline-end: 0;
 		}
 	}
```
delucis, line 99: Ah, interesting seeing this (and also the couple of times where we couldn’t use our utilities in components).

Not sure what the “right” architectural approach is for these kinds of layer relationships where a core layer needs to override the later component layer. I guess we’ll maybe figure that out as we go though. (Maybe doesn’t matter too much, but should these e.g. go in an explicit `starlight.core-component-overrides` layer — that kind of question, which I don’t know the answer to right now.)

docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -155,22 +167,24 @@ If you already have a Starlight site and want to add Tailwind CSS, follow these
...
-3.  Create a CSS file for Tailwind’s base styles, for example at `src/tailwind.css`:
+3.  Create a CSS file for Tailwind’s base styles, for example at `src/tailwind.css`, including Starlight Tailwind CSS:
 
-    ```css
+    ```css {4}
     /* src/tailwind.css */
-    @tailwind base;
-    @tailwind components;
-    @tailwind utilities;
+    @layer base, starlight.base, starlight.reset, starlight.core, starlight.content, starlight.components, starlight.utils, theme, components, utilities;
```
delucis, line 174: And again:

```suggestion
    @layer base, starlight, theme, components, utilities;
```

delucis (commented):
I’m too sleepy to fully tackle the docs — will do those tomorrow morning — but getting in at least a review on all the code changes.

Amazing work as always @HiDeoo and I haven’t spotted anything major I don’t think!

> - [ ] See if we should remove the default scoped style strategy enforced by Starlight?

I was also wondering that the other day, but I agree with you that this PR already has enough going on 😁 

We could definitely reuse your visual diff script for that PR when it comes though!

.changeset/old-toes-flow.md
```diff
@@ -0,0 +1,165 @@
...
+           starlight({
+             title: "Docs with Tailwind",
+             customCss: ["./src/tailwind.css"],
+           }),
+         ],
+      +  vite: {
+      +    plugins: [tailwindcss()],
+      +  },
+       });
+      ```
```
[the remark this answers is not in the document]
HiDeoo: Good catch, I remember checking the latest supported version in `main` when writing the changesets but totally forgot to check this other PR :thumbsup:

.changeset/old-toes-flow.md
```diff
@@ -0,0 +1,165 @@
...
+  
+1. **Update Tailwind base styles.** Tailwind CSS base styles needs to be updated for Tailwind v4 and also to use Starlight Tailwind CSS.
+
+   ```diff
+   /* src/tailwind.css */
+   -@tailwind base;
+   -@tailwind components;
+   -@tailwind utilities;
+   +@layer base, starlight.base, starlight.reset, starlight.core, starlight.content, starlight.components, starlight.utils, theme, components, utilities;
```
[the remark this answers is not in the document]
HiDeoo: Definitely, everything was kept using the explicit form during the PR as quite a lot of changes went into the naming and number of layers, but if we're happy with the current approach, we can simplify it.

packages/local-prod-visual-diff/diff.test.ts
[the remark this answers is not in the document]
HiDeoo: Yeah, it was very helpful and actually caught a few things I would have missed otherwise. At least the code is in a git history now, and it's not that complex, so if we ever need it, we can dig it, re-use it, improve it, etc.

packages/starlight/components/Search.astro
HiDeoo: Great catch, this was changed in #2848 to fix an issue when using the [Pagefind metadata feature](https://pagefind.app/docs/metadata/) where the UI was broken. I totally made a mistake when updating the PR and fixing conflicts, good catch.

As is, if you use the feature, you get the old broken behavior:

SCR-20250415-jbhf

I updated the branch to include the fix which now renders properly:

SCR-20250415-jcnt

packages/starlight/style/print.css
HiDeoo: So far, for print styles, this is the approach I've been using with Cascade Layers so it easily override styles but I'll admit I may not have enough experience/feedback with print styles and layers. Definitely could be an area where with time, we get more feedback and switch to a different approach if needed.

packages/starlight/components/Select.astro
HiDeoo: Yeah, definitely an interesting scenario, I'm not sure yet what is the best practice here either, or even if there is one. An extra layer is definitely a solution but would increase the number of layers so not quite sure.

delucis: I guess the argument _for_ layering is to not worry about ordering as much, I would guess if two style sources are:

```css
/* Core style that wants to override a component */
@layer starlight.components {
  /* ... */
}
```

```css
/* Component styles to be overridden */
@layer starlight.components {
  /* ... */
}
```

Then if the specificity were the same, then the `starlight.components` layer would win even though the core style is trying to override it because component styles get pulled in after core styles.

So having an override layer would ensure an earlier core style can win. But I think we can tackle that as we go and work out what’s necessary, where we’re still seeing issues etc.

.changeset/old-toes-flow.md
```diff
@@ -0,0 +1,136 @@
...
+   	--color-gray-900: var(--color-slate-900);
+   	--color-gray-950: var(--color-slate-950);
+    /* Custom text font. */
+    --font-sans: 'Atkinson Hyperlegible';
+    /* Custom code font. */
+    --font-mono: 'IBM Plex Mono';
+   }
+   ```
+
+We recommend checking your site’s appearance when upgrading to make sure there are no style regressions caused by this change.
```
delucis, line 136: Might be a worth a link like this?

```suggestion
We recommend checking your site’s appearance when upgrading to make sure there are no style regressions caused by this change.

For full details about upgrading from Tailwind v3 to v4, see the [official upgrade guide](https://tailwindcss.com/docs/upgrade-guide).
```

.changeset/old-toes-flow.md
```diff
@@ -0,0 +1,136 @@
...
+   	--color-gray-700: var(--color-slate-700);
+   	--color-gray-800: var(--color-slate-800);
+   	--color-gray-900: var(--color-slate-900);
+   	--color-gray-950: var(--color-slate-950);
+    /* Custom text font. */
+    --font-sans: 'Atkinson Hyperlegible';
+    /* Custom code font. */
+    --font-mono: 'IBM Plex Mono';
+   }
+   ```
```
delucis, line 134: I found it a bit confusing that there are two lists of steps. I think this part is essentially a fourth step so maybe we can nest it like this?

````suggestion
1. **Update Tailwind customizations.** If you previously [customized your Tailwind theme configuration](https://starlight.astro.build/guides/css-and-tailwind/#styling-starlight-with-tailwind) in the `tailwind.config.mjs` file, such customizations must now be defined through CSS using the `@theme` block in your Tailwind base styles.

	1. Locate existing customizations in your `tailwind.config.mjs` file. The following example defines customizations for the accent colors, gray scale, and font families used by Starlight:
	
	   ```js
	   // tailwind.config.mjs
	   import starlightPlugin from '@astrojs/starlight-tailwind';
	   import colors from 'tailwindcss/colors';
	   
	   /** @type {import('tailwindcss').Config} */
	   export default {
	     content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
	     theme: {
	       extend: {
	         colors: {
	           // Custom accent colors.
	           accent: colors.fuchsia,
	           // Custom gray scale.
	           gray: colors.slate,
	         },
	         fontFamily: {
	           // Custom text font.
	           sans: ['"Atkinson Hyperlegible"'],
	           // Custom code font.
	           mono: ['"IBM Plex Mono"'],
	         },
	       },
	     },
	     plugins: [starlightPlugin()],
	   };
	   ```
	
	1. The above customizations can be migrated to the new `@theme` block in the file containing your Tailwind base styles as follows:
	
	   ```css
	   /* src/tailwind.css */
	   @layer base, starlight, theme, components, utilities;
	   
	   @import '@astrojs/starlight-tailwind';
	   @import 'tailwindcss/theme.css' layer(theme);
	   @import 'tailwindcss/utilities.css' layer(utilities);
	   
	   @theme {
	   	/* Custom accent colors. */
	   	--color-accent-50: var(--color-fuchsia-50);
	   	--color-accent-100: var(--color-fuchsia-100);
	   	--color-accent-200: var(--color-fuchsia-200);
	   	--color-accent-300: var(--color-fuchsia-300);
	   	--color-accent-400: var(--color-fuchsia-400);
	   	--color-accent-500: var(--color-fuchsia-500);
	   	--color-accent-600: var(--color-fuchsia-600);
	   	--color-accent-700: var(--color-fuchsia-700);
	   	--color-accent-800: var(--color-fuchsia-800);
	   	--color-accent-900: var(--color-fuchsia-900);
	   	--color-accent-950: var(--color-fuchsia-950);
	   	/* Custom gray scale. */
	   	--color-gray-50: var(--color-slate-50);
	   	--color-gray-100: var(--color-slate-100);
	   	--color-gray-200: var(--color-slate-200);
	   	--color-gray-300: var(--color-slate-300);
	   	--color-gray-400: var(--color-slate-400);
	   	--color-gray-500: var(--color-slate-500);
	   	--color-gray-600: var(--color-slate-600);
	   	--color-gray-700: var(--color-slate-700);
	   	--color-gray-800: var(--color-slate-800);
	   	--color-gray-900: var(--color-slate-900);
	   	--color-gray-950: var(--color-slate-950);
	    /* Custom text font. */
	    --font-sans: 'Atkinson Hyperlegible';
	    /* Custom code font. */
	    --font-mono: 'IBM Plex Mono';
	   }
	   ```
````

File: docs/astro.config.mjs [hunk 18]
```diff
@@ -31,6 +31,8 @@ const ogUrl = new URL('og.jpg?v=1', site).href;
 const ogImageAlt = 'Make your docs shine with Starlight';
 
 export default defineConfig({
+	// TODO(HiDeoo) Remove this, only used to avoid screenshoting the dev toolbar.
+	devToolbar: { enabled: false },
```
delucis, line 35: OK to start removing this and `local-prod-visual-diff` package I think!

delucis (commented):
Last few remarks on the changesets. Will looks at docs now.

delucis (approved):
I think this PR is in great shape to merge in our next minor! Many, many thanks to @HiDeoo for iterating over this and getting us to this point. In some ways this is a big architectural change but you’ve carried it out with such care that I think most people will hardly notice, while unlocking lots of improvements for those who need it.

PR #2327, opened 2024-09-13
Fix broken restoration of remark directives.

tritao:
#### Description

This fixes the logic in `remarkDirectivesRestoration` so it only restores remark directives that do not have node data.

If the node has a defined data, this means it was set explicitly and should be allowed to continue to be processed later on by `remark-rehype`.

### DIFF
--- a/.changeset/gold-coats-destroy.md
+++ b/.changeset/gold-coats-destroy.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fixes restoration of remark directives for nodes with custom data attached.
--- a/packages/starlight/__tests__/remark-rehype/asides.test.ts
+++ b/packages/starlight/__tests__/remark-rehype/asides.test.ts
@@ -270,3 +270,35 @@ test('lets remark plugin injected by Starlight plugins handle text and leaf dire
 		"<p>This is a:test of a sentence with a TEXT FROM REMARK PLUGIN directive handled by another remark plugin and some other text:name[content]{key="val"} directives not handled by any plugin.</p>"
 	`);
 });
+
+test('does not transform back directive nodes with data', async () => {
+	const processor = await createMarkdownProcessor({
+		remarkPlugins: [
+			...starlightAsides({
+				starlightConfig,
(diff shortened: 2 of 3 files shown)


HiDeoo:
Thanks for your contribution.

On a pure technical level, the changes make sense to me however I'm having a hard time thinking of a use case/scenario for this usage. Considering we would probably want to add a test case for this (e.g. something in the lines of this [existing one](https://github.com/withastro/starlight/blob/main/packages/starlight/__tests__/remark-rehype/asides.test.ts#L239-L272) but with a remark plugin setting some data and a rehype plugin), could you provide more details on how you're currently hitting this issue?

HiDeoo:
Oh, I see, I was definitely starting to imagine more complicated scenarios. This makes sense, thanks for the clarification 🙌 

I assume we could definitely have a test for this case too, I imagine it would looks something like this:

```ts
test('does not transform back directive nodes with data', async () => {
	const processor = await createMarkdownProcessor({
		remarkPlugins: [
			...starlightAsides({
				starlightConfig,
				astroConfig: {
					root: new URL(import.meta.url),
					srcDir: new URL('./_src/', import.meta.url),
				},
				useTranslations,
			}),
			// A custom remark plugin updating the node with data that should be consumed by rehype.
			function customRemarkPlugin() {
				return function transformer(tree: Root) {
					visit(tree, (node) => {
						if (node.type !== 'textDirective') return;
						node.data ??= {};
						node.data.hName = 'span';
						node.data.hProperties = { class: `api` };
					});
				};
			},
			remarkDirectivesRestoration,
		],
	});

	const res = await processor.render(`This method is available in the :api[thing] API.`);
	expect(res.code).toMatchInlineSnapshot(
		`"<p>This method is available in the <span class="api">thing</span> API.</p>"`
	);
});
```

I did not try it, just tweaked the existing test I linked earlier and I imagine this test would fail without your fix and pass after. Would you be up for trying to add such a test?

delucis (approved):
Thanks for fixing this @tritao and for helping debug and test @HiDeoo. Looks like this should be ready for our next release 🚀

PR #2329, opened 2024-09-16
Re-enable skipped navigation test

delucis:
#### Description

- @HiDeoo noticed this test file had been accidentally skipped when added in #1023
- This PR re-enables it, by adding `.test` to the filename, and fixes a test it contains to pass


delucis:
Ah, the snapshot change is actually unexpected — will need to work out why.

PR #2334, opened 2024-09-18
Fix Expressive Code UI labels

HiDeoo:
#### Description

This PR fixes an issue with Expressive Code UI labels not being displayed correctly in the UI. The issue was caused with the recent change to `i18next` `t` function which returns the key if a translation is not found. This PR fixes the issue by adding a check to see if the translation exists first and if not, it will return `undefined`.

Before:

image

After:

image


delucis (approved):
Looks perfect — thank you for the quick fix!

PR #2338, opened 2024-09-18
i18n(ko-KR): update `overrides.md`

jsparkdev:
#### Description

- update `overrides.md`
- #1923


HiDeoo (approved):
Thanks for updating this page :star2:

PR #2340, opened 2024-09-18
i18n(fr): update `reference/overrides`

HiDeoo:
#### Description

This PR updates the French version of the `reference/overrides` page with the changes from #1923.


delucis (approved):
Heh, you really didn’t need my approval for this one, but here it is anyway!

PR #2121, opened 2024-07-10
Components documentation improvements

HiDeoo:
#### Description

- Closes #1938 and #2010

This PR is a follow-up of #1938 to improve the components documentation and move away from the [single page](https://starlight.astro.build/guides/components/) currently used.

This pull request is a draft as it only includes a global component guide and the new references for the `<Tabs>` and `<Badge>` components. The idea is to first get feedback on the new format, discuss remaining questions and then continue with the rest of the components once the format is agreed upon.

The `<Badge>` page was chosen as it's relatively simple and small while the `<Tabs>` page was chosen as it's a more complex one which involves 2 components.

To give a brief overview of some of the changes per page:

##### Global Components Guide

- This is pretty much the [introduction](https://starlight.astro.build/guides/components/#using-a-component) of the current Components guide.
- It also includes a new section for `ComponentProps` to address #2010.

##### Component References

The global structure is as follows:

- Brief description of the component
- Preview of the component in a basic example with no code example
- Import statement
- A usage section with various use cases
  - Each usage includes a code example and the associated preview
- A props section with a list of all accepted props
  - A link to the component implementation (can be useful to see where/how extra props are spread for example)
  - This follow the pattern used in other references

Some other notes:

- Links exists between props mentioned in the usage section and the props section ~~and vice versa~~ (decided that it was not useful after discussing this).
- In the props section, the `required` tag is not included in the title (this allows for switching languages without losing position compared to the other reference pages).
- The component used to display code examples and/or previews is pretty much unstyled at this point of the PR:
  - The idea of the component is to make the examples more obvious with some kind of frame around it compared to the current examples.
  - We would need to decide if we want some controls on it or keep it as simple as possible:
    - I think one important thing to consider is that Starlight is not a component library/design system
    - Most examples researched include a "Show Code" / "Open in StackBlitz" button which I don't think are relevant for Starlight as we want to show both the code and the preview at the same time.
    - One of the controls I think could be relevant would be a toggle to switch the preview between light and dark mode although this one could be tricky to implement with Starlight's current code.
  - Do we want to keep the duplicated content in the markup (for the code and preview) like we currently have or do we want to have a more advanced setup, maybe relying on remark, to only have to specify the code once?
    - Not sure this is worth it considering the number of components and the complexity it could add

#### Remaining tasks

- [x] Once the format is agreed upon, convert the rest of the components to the new format
- [x] Use a custom UI string for the "Preview" word
  - [x] Wait for #1923 to be merged
- [x] Remove the existing components guide
[...]

PR #2343, opened 2024-09-18
i18n(fr): add `reference/icons`

HiDeoo:
#### Description

This PR adds the French version of the `reference/icons` page with the changes from #2121.

### DIFF
--- a/docs/src/content/docs/fr/reference/icons.mdx
+++ b/docs/src/content/docs/fr/reference/icons.mdx
@@ -1,0 +1,19 @@
+---
+title: Référence des icônes
+description: Une vue d'ensemble de tous les icônes disponibles dans Starlight.
+---
+
+Starlight fournit un ensemble d'icônes intégrées que vous pouvez afficher dans votre contenu en utilisant le composant `<Icon>`.
+
+## Utiliser les icônes
+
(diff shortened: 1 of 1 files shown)


delucis (approved):
Looks great! Noting that we’ve never added a `copiedLabel` prop to `<IconsList />` for translations so it always shows “Copied!”

Could be something to add in the future.

HiDeoo:
> Looks great! Noting that we’ve never added a `copiedLabel` prop to `<IconsList />` for translations so it always shows “Copied!”
> 
> Could be something to add in the future.

Technically, we could now also use `Astro.locals.t('expressiveCode.copyButtonCopied')` but it's not available in a lot of languages I think.

delucis:
> Technically, we could now also use `Astro.locals.t('expressiveCode.copyButtonCopied')` but it's not available in a lot of languages I think.

Yeah that’s true. Also for these one-off components I think I actually prefer the prop approach because then you are prompted to translate it in the same place as the translated content. Annoying for a component that gets repeated many times, but if only used once like this or the theme editor, then I think considering part of the “content” makes sense I think.

thomasbnt (approved):
Good!

PR #2344, opened 2024-09-18
docs: fix broken Markdown content related links

HiDeoo:
#### Description

Following the Astro Docs https://github.com/withastro/docs/pull/9337 & https://github.com/withastro/docs/pull/9344 pull requests, this PR fixes a few broken links related to Markdown content.

PR #2345, opened 2024-09-18
i18n(fr): add `components/using-components`

HiDeoo:
#### Description

This PR adds the French version of the `components/using-components` page with the changes from #2121.

### DIFF
--- a/docs/src/content/docs/fr/components/using-components.mdx
+++ b/docs/src/content/docs/fr/components/using-components.mdx
@@ -1,0 +1,96 @@
+---
+title: Utilisation des composants
+description: Utilisation des composants avec MDX et Markdoc avec Starlight.
+sidebar:
+  order: 1
+---
+
+Les composants vous permettent de réutiliser facilement un élément d'interface utilisateur ou de style de manière cohérente.
+Il peut s'agir par exemple d'une carte de liaison ou d'une intégration YouTube.
+Starlight prend en charge l'utilisation de composants dans les fichiers [MDX](https://mdxjs.com/) et [Markdoc](https://markdoc.dev/) et fournit des composants courants que vous pouvez utiliser.
+
+[Pour en savoir plus sur la création de composants, consultez la documentation d'Astro](https://docs.astro.build/fr/core-concepts/astro-components/).
+
+## Utilisation d'un composant avec MDX
+
+Vous pouvez utiliser un composant en l'important dans votre fichier MDX et ensuite l'afficher en tant que balise JSX.
+Ces balises ressemblent à des balises HTML, mais commencent par une lettre majuscule correspondant au nom utilisé dans votre instruction `import` :
+
+```mdx
+---
+# src/content/docs/exemple.mdx
+title: Bienvenue dans ma documentation
+---
+
+import { Icon } from '@astrojs/starlight/components';
+import CustomCard from '../../components/CustomCard.astro';
+
+<Icon name="open-book" />
+
+<CustomCard>
+	Les composants peuvent également contenir du **contenu imbriqué**.
+</CustomCard>
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/components/using-components.mdx
```diff
@@ -0,0 +1,96 @@
...
+---
+// src/components/Exemple.astro
+---
+
+<div class="not-content">
+	<p>Contenu non affecté par les styles par défaut de Starlight.</p>
+</div>
+```
+
+## Props des composants
```
HiDeoo, line 81: To keep consistency with https://starlight.astro.build/fr/reference/overrides/#props-des-composants and Astro Docs

docs/src/content/docs/fr/components/using-components.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+<Icon name="open-book" />
+
+<CustomCard>
+	Les composants peuvent également contenir du **contenu imbriqué**.
+</CustomCard>
+```
+
+Starlight étant basé sur Astro, vous pouvez ajouter la prise en charge de composants construits avec n'importe quel [framework d'interface utilisateur pris en charge (React, Preact, Svelte, Vue, Solid, and Alpine)](https://docs.astro.build/fr/guides/framework-components/) dans vos fichiers MDX.
+Pour en savoir plus sur [l'utilisation de composants avec MDX](https://docs.astro.build/fr/guides/integrations-guide/mdx/#utilisation-de-composants-dans-mdx), consultez la documentation Astro.
```
HiDeoo, line 36: Link updated with the change of https://github.com/withastro/starlight/pull/2344 which should be merged first.

docs/src/content/docs/fr/components/using-components.mdx
thomasbnt: “Props” is translated as ‘Propriétés’ in some files. Maybe change all the documentation _props_ -> _propriétés_ ?

HiDeoo: Oh, do you have specific links to this? We should definitely fix this. By quickly searching on the website, all "propriété(s)" are never used in the context of a component props.

props

--- 

This is indeed a tricky one. To give a bit more context on why we specifically picked "props" in this case, e.g. in the "Overrides Reference" page:

- This is a key concept in Astro and components.
- `props` is [heavily](https://docs.astro.build/fr/basics/astro-components/#props-de-composant) used in the French Astro documentation, same for `une props` or even `un prop`.
- If you're coming from the Astro documentation, you'll be familiar with this term for this feature and there is a high chance you'll be searching for it rather than "propriété".
- Other frameworks French documentation also tend to use `props` (e.g. [React](https://fr.react.dev/learn/passing-props-to-a-component)).

If we ever decide to change this, probably after a discussion with more people, I think we may want to do that on both the Starlight and Astro (which would be quite the effort) French documentation to keep consistency.

thomasbnt: > Oh, do you have specific links to this? We should definitely fix this. By quickly searching on the website, all "propriété(s)" are never used in the context of a component props.

This one for example _[La Propriété (props) page](https://docs.astro.build/fr/guides/routing/#la-propri%C3%A9t%C3%A9-props-page)_ or here _[transmettre des propriétés (Props)](https://docs.astro.build/fr/guides/framework-components/#transmettre-des-props-aux-composants-du-framework)_

HiDeoo: Oh, you mean in the Astro Docs, I thought you meant in the Starlight Docs.

Indeed, the Astro Docs is at the moment a mix of "props", "une props", "un prop", or "propriété" from what I can see. Could be worth discussing in the French Discord channel and see if it would be possible to agree on a fixed terminology and PR all instances. We did that at a few occasions in the Starlight repo, e.g. to rename all French example filenames, etc.

thomasbnt: We can leave props for the moment, and ask to French thread what is the best choice!

delucis (approved):
LGTM! And #2344 is merged 🫡

PR #2355, opened 2024-09-19
i18n(ko-KR): create `code.mdx`

jsparkdev:
#### Description

- create `code.mdx`

### DIFF
--- a/docs/src/content/docs/ko/components/code.mdx
+++ b/docs/src/content/docs/ko/components/code.mdx
@@ -16,7 +16,7 @@ import Preview from '~/components/component-preview.astro';
 	slot="preview"
 	code={`## 환영합니다
 
-안녕하세요 (**우주에서**)!`}
+**우주에서** 하는 인사!`}
 lang="md"
 title="example.md"
 ins={3}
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/ko/components/code.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+import Preview from '~/components/component-preview.astro';
+
+<Preview>
+
+<Code
+	slot="preview"
+	code={`## 환영합니다
+
+안녕하세요 (**우주에서**)!`}
```
HiDeoo, line 19: Just to double check, are the `(` and `)` around `**우주에서**` expected?

HiDeoo (commented):
Thanks for updating the translation :star2:

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). I left one tiny questions but other than that, everything looks good to me.

docs/src/content/docs/ko/components/code.mdx
jsparkdev: I've updated it. Thank you

jsparkdev:
@HiDeoo I've updated it. Thank you

HiDeoo (approved):
Wonderful work with all these translations! Thanks a lot for all your contributions :raised_hands: 

One last note in case you're interested, the sidebar group label could be translated [here](https://github.com/withastro/starlight/blob/main/docs/astro.config.mjs#L136-L138) (also noting that ["Resources"](https://github.com/withastro/starlight/blob/main/docs/astro.config.mjs#L162-L169) could be done in the same PR too).

PR #2366, opened 2024-09-19
i18n(ko-KR): update `astro.config.mjs`

jsparkdev:
#### Description

- update `astro.config.mjs` for translating sidebar labels.


HiDeoo (approved):
Thanks for this final update completing the translation of the sidebar too :raised_hands: Beautiful work!

PR #2371, opened 2024-09-19
Draft: Pagefind weight frontmatter

lorenzolewis:
#### Description

- Closes #2348
- This exposes a `pagefind.weight` property for page frontmatter to adjust the search index weight. 

Note: Currently we don't have any tests to verify the pagefind output. I've tested this by adding a weight of `7.5` to the `site-search.mdx` page. Before when searching for "pagefind" this page would be the 3rd on the list, now it is the 1st. This is purely for testing purposes.

### DIFF
--- a/docs/src/content/docs/guides/site-search.mdx
+++ b/docs/src/content/docs/guides/site-search.mdx
@@ -1,6 +1,8 @@
 ---
 title: Site Search
 description: Learn about Starlight’s built-in site search features and how to customize them.
+pagefind:
+  weight: 7.5
 ---
 
 import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
--- a/packages/starlight/components/Page.astro
+++ b/packages/starlight/components/Page.astro
@@ -38,7 +38,8 @@ const pagefindEnabled =
 	!Astro.props.entry.slug.endsWith('/404') &&
 	pagefindConfig !== false;
 
-const pagefindWeight = typeof pagefindConfig === 'object' && pagefindConfig.weight;
+const pagefindWeight =
+	pagefindEnabled && typeof pagefindConfig === 'object' && pagefindConfig.weight;
 ---
 
 <html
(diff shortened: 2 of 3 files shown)


docs/src/content/docs/guides/site-search.mdx
```diff
@@ -42,6 +42,44 @@ This text will be hidden from search.
 </div>
 ```
 
+## Adjust ranking in search results
+
+Pagefind's index ranking can be adjusted using the [`data-pagefind-weight`](https://pagefind.app/docs/weighting/#ranking-content-higher-or-lower) attribute. This can be used to [adjust a whole page](#adjust-whole-page-rank) or [individual sections of a page](#adjust-individual-section-rank).
+
+Learn more about about how the weight value impacts search results in the [Pagefind Weighting content guide](https://pagefind.app/docs/weighting/).
```
ematipico, line 49: It would be nice to explain to users when they should change the ranking, and why. The docs of pagefind only explain some technical stuff, but from the point of view of maintainer as myself, it would be nice to have some guidance on why I should use it.

lorenzolewis:
```suggestion
## Adjust ranking in search results

Pagefind's index ranking can be adjusted using the [`data-pagefind-weight`](https://pagefind.app/docs/weighting/#ranking-content-higher-or-lower) attribute. This can be used to [adjust a whole page](#adjust-whole-page-rank) or [individual sections of a page](#adjust-individual-section-rank).

For example, if there is a high-level guide that is a better entry point for readers than an error reference page with similar keywords, adjusting the weight of the guide page will cause it to be ranked higher in the Pagefind search results. 

Learn more about about how the weight value impacts search results in the [Pagefind Weighting content guide](https://pagefind.app/docs/weighting/).
```

@ematipico Thanks! Do you think this gives a good example of the usage here? Happy for any suggestions if anything else comes to mind 🥳

ematipico: We're getting somewhere I think. Now we should explain when to **increase** the page weight and when **decrease** the page weight.

PR #2377, opened 2024-09-19
Fix synced nested `<Tabs>` restoration issue

HiDeoo:
#### Description

This pull request fixes an issue reported on Discord regarding synced nested tabs.

Repro steps:

- Visit https://developers.cloudflare.com/d1/get-started/#1-create-a-worker
- Select the _Dashboard_ tab
- Reload the page
- The first tab panel is invisible

This issue is similar to #473 where the panels `hidden` property was not being set correctly due to the query selector matching all nested panels instead of just the direct children. The same issue can happen when restoring synced tabs as this part of the code was not including the same fix.

The restoration code now also uses the [`:scope` CSS pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope#direct_children) to only match the direct children.

I've included a test case to cover both the syncing of nested tabs (fixed in #473) and the restoration of synced nested tabs (fixed in this PR).

I also tested the fix locally on the Cloudflare docs and it now works as expected.

### DIFF
--- a/.changeset/beige-knives-visit.md
+++ b/.changeset/beige-knives-visit.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue with synced `<Tabs>` components containing nested `<Tabs>` causing tab panels to not render correctly.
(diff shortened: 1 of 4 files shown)


delucis (approved):
Thanks for fixing so quickly @HiDeoo! Will get this released right away for folks 🫡

PR #2380, opened 2024-09-20
Loosen i18n schema to pass through unknown keys by default

delucis:
#### Description

- This PR calls Zod’s [`.passthrough()`](https://github.com/colinhacks/zod?tab=readme-ov-file#passthrough) method on Starlight’s i18n schema by default.
- This keeps keys not known to the schema in the parsed output instead of stripping them out.
- This improves the DX of custom plugins: these already have types for their i18n strings generated via the `injectTranslations()` hook, but currently users would still need to explicitly extend the i18n schema to provide custom translations.
- With this change, users can skip extending the i18n schema, and just define custom translation data files directly.

### DIFF
--- a/packages/starlight/schemas/i18n.ts
+++ b/packages/starlight/schemas/i18n.ts
@@ -23,16 +23,26 @@ const defaultI18nSchema = () =>
 /** Type of Starlight’s default i18n schema, including extensions from Pagefind and Expressive Code. */
 type DefaultI18nSchema = ReturnType<typeof defaultI18nSchema>;
 
+/**
+ * Based on the the return type of Zod’s `merge()` method. Merges the type of two `z.object()` schemas.
+ * Also sets them as “passthrough” schemas as that’s how we use them. In practice whether or not the types
+ * are passthrough or not doesn’t matter too much.
+ */
+type MergeSchemas<A extends z.AnyZodObject, B extends z.AnyZodObject> = z.ZodObject<
+	z.objectUtil.extendShape<A['shape'], B['shape']>,
+	'passthrough',
+	B['_def']['catchall']
+>;
 /** Type that extends Starlight’s default i18n schema with an optional, user-defined schema. */
 type ExtendedSchema<T extends z.AnyZodObject> = T extends z.AnyZodObject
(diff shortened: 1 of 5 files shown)


HiDeoo (approved):
Looks perfect to me and a great improvement :rocket:

I wonder if linking in the `MergeSchemas` comment to [this return type](https://github.com/colinhacks/zod/blob/3032e240a0c227692bb96eedf240ed493c53f54c/src/types.ts#L2656-L2660) to have a reference could be useful or not, considering that if you hover the result of `schema1.merge(schema2).passthrough()`, e.g. Visual Studio Code, you will never see something close to that, especially for the `.catchall` type.

image

packages/starlight/schemas/i18n.ts
```diff
@@ -23,16 +23,26 @@ const defaultI18nSchema = () =>
 /** Type of Starlight’s default i18n schema, including extensions from Pagefind and Expressive Code. */
 type DefaultI18nSchema = ReturnType<typeof defaultI18nSchema>;
 
+/**
+ * Based on the the return type of Zod’s `merge()` method. Merges the type of two `z.object()` schemas.
+ * Also sets them as “passthrough” schemas as that’s how we use them. In practice whether or not the types
+ * are passthrough or not doesn’t matter too much.
```
delucis, line 29: You mean something like this @HiDeoo?

```suggestion
 * are passthrough or not doesn’t matter too much.
 * 
 * @see https://github.com/colinhacks/zod/blob/3032e240a0c227692bb96eedf240ed493c53f54c/src/types.ts#L2656-L2660
```

HiDeoo: Exactly, the previous `z.ZodIntersection` was very explicit and the new one can be kinda scary or magical if you don't exactly know where it comes from imo. Your call.

delucis: OK. Will add this. I also realised doing this that `ZodIntersection` was completely the wrong type technically — `merge()` doesn’t return an intersection but a composite object — but never mind, it worked anyway 😁

PR #2386, opened 2024-09-20
Clarify how to override default components that use multiple slots.

jwoyo:
The documentation lacked an explanation of how to override default components that use named slots. Passing all slots through is necessary. Omitting this step breaks the layout for example when user conditionally changes PageFrame.astro.

The commit adds an explanation and a code example.

```astro
---
// this could be: src/components/ConditionalPageFrame.astro
import type { Props } from '@astrojs/starlight/props';
import Default from '@astrojs/starlight/components/PageFrame.astro';
const isHomepage = Astro.props.slug === '';
---
{
	isHomepage ? (
		<section>My great landing page 🌟</section>
	) : (
		<Default {...Astro.props}>
			<slot />
			<Fragment slot="header"><slot name="header"/></Fragment>
			<Fragment slot="sidebar"><slot name="sidebar"/></Fragment>
		</Default>
	)
}
```


HiDeoo:

This is definitely something that we do not explicitly document and has come up a few times already. I'd like to get other people's opinion on this but my initial thought would be that this would better fit in the ["Reuse a built-in component"](https://starlight.astro.build/guides/overriding-components/#reuse-a-built-in-component) section rather than the "Use page data" one. I'm also not sure we would need an entire new code example for this as this would be mostly identical to other ones we already have but we could link to the [relevant Astro Docs section](https://docs.astro.build/en/basics/astro-components/#transferring-slots).

I think my idea would be to add a third bullet point to the ["Reuse a built-in component"](https://starlight.astro.build/guides/overriding-components/#reuse-a-built-in-component) unordered list that could look something like this:

> - If the built-in component you are reusing also contains [nameed slots](https://docs.astro.build/en/basics/astro-components/#named-slots), e.g. like the [`TwoColumnContent`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/TwoColumnContent.astro) component which uses a `right-sidebar` slot, you need to [transfer](https://docs.astro.build/en/basics/astro-components/#transferring-slots) theses as well using the `<slot name="right-sidebar" slot="right-sidebar" />` syntax.

Another possibility that comes to mind is to reference additional named slots in the ["Overrides Reference"](https://deploy-preview-2386--astro-starlight.netlify.app/reference/overrides/) page for the few components that have them. At the moment, only `TwoColumnContent` and `PageFrame` have named slots.

For example, below the "Default component: TwoColumnContent.astro" line, it could say:

> **Default component:** [`TwoColumnContent.astro`](https://deploy-preview-2386--astro-starlight.netlify.app/reference/overrides/#twocolumncontent:~:text=Default%20component%3A%20TwoColumnContent.astro)
> **Named slots:** `right-sidebar`

Like I said, I'd like to get your and other people's opinion on this before I make any changes. Let me know what you think.

jwoyo:
Thanks @HiDeoo for reviewing the proposed change. I agree that the information fits better as a third bullet point under [Reuse a built-in component](https://starlight.astro.build/guides/overriding-components/#reuse-a-built-in-component). Also referencing the named slots in the [Overrides Reference](https://deploy-preview-2386--astro-starlight.netlify.app/reference/overrides/#twocolumncontent:~:text=Default%20component%3A%20TwoColumnContent.astro) makes sense to me as most users probably won't check in the overridden file in depth and would appreciate a hint that they need to handle the slots.

I'll wait for others' opinions as well and proceed with your suggested change in this PR if I don't hear otherwise.

jwoyo:
Hey @HiDeoo. As I haven't heard any other opinions on that I've made the suggested changes:

- Removed the additional code example
- Added a third bullet point containing the code example with `TwoColumnContent`
- Added the **named slots:** line to the [Override References](https://deploy-preview-2386--astro-starlight.netlify.app/reference/overrides/)

docs/src/content/docs/guides/overriding-components.mdx
```diff
@@ -81,7 +81,7 @@ When rendering a built-in component inside a custom component:
 
 - Spread `Astro.props` into it. This makes sure that it receives all the data it needs to render.
 - Add a [`<slot />`](https://docs.astro.build/en/core-concepts/astro-components/#slots) inside the default component. This makes sure that if the component is passed any child elements, Astro knows where to render them.
-
+- If the built-in component you are reusing also contains [named slots](https://docs.astro.build/en/basics/astro-components/#named-slots), you need to transfer them as well. For example, the [`TwoColumnContent`](/reference/overrides/#twocolumncontent) component uses a `right-sidebar` slot. To keep it you need to transfer it using the `<slot name="right-sidebar" slot="right-sidebar" />` syntax (see [transferring slots](https://docs.astro.build/en/basics/astro-components/#transferring-slots)).
```
HiDeoo, line 84: The ideas of the changes that were suggested during the Talking & Doc'ing event were:

- Explicitly mention the only 2 components that have named slots
- Show the example you initially provided with an highlight on the line transferring a slot


````suggestion

If you are reusing the [`PageFrame`](/reference/overrides/#pageframe) or [`TwoColumnContent`](/reference/overrides/#twocolumncontent) components which contain [named slots](https://docs.astro.build/en/basics/astro-components/#named-slots), you also need to [transfer](https://docs.astro.build/en/basics/astro-components/#transferring-slots) these slots as well.

The example below shows a custom component that reuses the `TwoColumnContent` component which contains an additional `right-sidebar` named slot that needs to be transferred:

```astro {9}
---
// src/components/CustomContent.astro
import type { Props } from '@astrojs/starlight/props';
import Default from '@astrojs/starlight/components/TwoColumnContent.astro';
---

<Default {...Astro.props}>
	<slot />
	<slot name="right-sidebar" slot="right-sidebar" />
</Default>
```

````

jwoyo:
> The general consensus seems to be that both your initial suggestion with a example and some of the changes I suggested are good ideas so I left a new suggestion matching both of them. Let me know what you think about it.

Sounds perfect. Thank you for getting back! I've added the suggested change.

HiDeoo (approved):
Thanks for kicking this off and your patience. This looks like a great improvement to me with your latest changes :star2:

PR #2390, opened 2024-09-22
Move route data to `Astro.locals`

delucis:
#### Description

- Early draft PR exploring moving Starlight’s route data object to `Astro.locals` — currently this is passed down via `Astro.props` to all our templating components.

- There are some tricky nuances here for sure.

- For example, middleware runs for all routes on a site, which can include _non_ Starlight pages. For these we can’t generate route data. For now I’ve reflected this in the types for locals and asserted `Astro.locals.routeData!` in the components. There’s not really a sensible way to guard that without duplicating a null check in every component, but also it feels like route data _should_ be defined in all cases where Starlight’s `<Page>` is rendered.

- The `<StarlightPage>` component poses some challenges. Right now you pass some props and it generates route data. I’ve hacked this in by assigning that generated data to locals inside the component for now, but this does mean you can’t transform data for pages created this way with additional middleware. Not sure there’s any way to solve this?

	- One way might be to have a dedicated system of “route data middleware” implemented at the Starlight level instead of using generic Astro middleware, e.g.
	
	  ```js
	  starlight({
	    routeMiddleware: './some-file.ts',
	  })
	  ```
	  
	  We could bundle those in a virtual module and have both Starlight’s `locals.ts` and `<StarlightPage>` use them or something:
	  ```js
	  context.locals.routeData = await useRouteData(context);
	  for (const fn of routeMiddleware) {
	    context.locals.routeData = fn(context.locals.routeData);
	  }
	  ```

- The current branch rips out the prop drilling entirely. That means any overrides that rely on `Astro.props` will break. Could it be worth doing something to ease migration? Deprecate props but keep them around? Throw an error something like we did for `labels`? Something to think about.

- There’s probably a tidier way to do some of the code — just did the quick and easy thing for now. (For example, `Content` is currently typed as optional in the route data object to make it easy to throw it in where I needed it, without checking all the places that type is used. But could probably tidy that up to have a dedicated separate type.)

- Can also discuss naming here. So far it’s `Astro.locals.routeData`, but there’s probably an argument for something a bit more descriptive like `Astro.locals.starlightRoute` or even just `Astro.locals.starlight` potentially.

#### To-do

- [x] Remove demo middleware from docs


lorenzolewis:
Really exciting to see this!

> The current branch rips out the prop drilling entirely. That means any overrides that rely on `Astro.props` will break. Could it be worth doing something to ease migration? Deprecate props but keep them around? Throw an error something like we did for `labels`? Something to think about.

With the i18n update didn't we have a somewhat breaking change (possibly that's the `labels` update you're talking about)? I feel like if it's at all possible to keep current behavior around and mark it as depreciated then it would be much more helpful. This change will probably impact A LOT of users who have spent time crafting component overrides. 

> Can also discuss naming here. So far it’s `Astro.locals.routeData`, but there’s probably an argument for something a bit more descriptive like `Astro.locals.starlightRoute` or even just `Astro.locals.starlight` potentially.

Is there any norm here in Astro-land for 3rd party integrations? This has a potential for name collisions if a user has implemented middleware, correct? If that's the case then I think it would be good to at the very least prefix any locals with `starlight`.

HiDeoo:
Sharing a few quick thoughts before I forget them as I've been playing with the idea locally too and we can discuss more in depth later.

> For now I’ve reflected this in the types for locals and asserted `Astro.locals.routeData!` in the components.

As this impacts components and not user components, I've been experimenting with a getter function instead that throws if the route data is not available, which would mean it was called in a non-Starlight route.

> The `<StarlightPage>` component poses some challenges.

Definitely a tricky one indeed, been trying a few things but haven't found something that I'm happy with yet too.

> Deprecate props but keep them around? Throw an error something like we did for `labels`?

As we have seen from the `0.28.0` reports, it's relatively easy to craft explicit errors for `Astro.props.*` usage in user code altho it gets confusing for users when it's coming from a plugin.

If we compare to labels, in a panel of 15 plugins, only 3 where using `Astro.props.labels` so the impact was pretty low. This would definitely not be the case for all other props. Still need to think about this one but if we go with the same approach as `labels`, we should make sure to mention it in the changeset/breaking change that users should make sure their plugins are up to date (which is something I didn't think about for `labels`).

delucis:
Spent a bit of time revisiting this PR and feeling quite positive about the idea. Made the following adjustments, which helps it feel better:

1. I pulled 404 route data up into our route data process as well. This means that we can guarantee the `locals` will always contain Starlight data, even if it is only the 404 route for pages Starlight doesn’t know about. Avoids the need for a getter that throws an error (which I also played with for a bit) and means we really centralise all the data operations in the route data layer.

2. I renamed `locals.routeData` to `locals.starlightRoute`. Feels better to namespace it a bit and also feels better when using things like `entry.data` where the repeat “data” in the path — `locals.routeData.entry.data` — felt kind of repetitive and vague. `starlightRoute` expresses things nicely I think.

There’s a type error I’m not 100% sure about — I guess we’re pulling some of our virtual modules into view for type checking that we weren’t previously? Not sure, but haven’t investigate too much. Next up I’d like to tackle the idea of a route data “middleware”/“pipeline” so that users and plugins have a place to plug in and modify stuff.

delucis:
Context from today’s work: fixed type checks by moving around some code.

Basically the issue was that this PR exposes the `StarlightRouteData` type ambiently in `App.Locals`, but that type was colocated with some stuff that imported some of our non-public `virtual:*` modules. The refactor creates a dedicated `utils/routing/types.ts` file which avoids imports from our modules using virtual modules. Naming may not be perfect yet. For example, this also moves a few sidebar-related types up to that type file, but it does at least ensure we aren’t forced to expose more public type for the `virtual:*` modules.

trueberryless:
I'm also very excited about this PR!

Just to keep track of what has been achieved already and what needs to be done after reading the latest changes:

- [x] We* have a nice and clear name with some namespacing (Personally, I really agree with @lorenzolewis and @delucis)
- [x] `Astro.locals.starlightRoute` is guarenteed not null because @delucis moved 404 to route data as well. This means we don't need a getter which throws an error like @HiDeoo suggested.
- [ ] `<StarlightPage>` needs to support middleware (tricky because currently we* just set `Astro.locals` to the generated values)
- [ ] Discuss proper naming for the type file because it now includes some sidebar-related stuff

Please correct me if any bullet point is wrong or misleading in any way, I'll correct and keep the list up to date if you want.

\*We: I don't see myself included, props to @delucis,@lorenzolewis and @HiDeoo

trueberryless:
Regaring the `<StarlightPage>`:

I'm not that familiar with how Astro works as you guys are, so maybe - the more I think about it, the more I lean towards probably - this idea is **dumb**, but:

Would it be possible to manually call the [render](https://github.com/withastro/astro/blob/98f9e833a5610cbd6f966a827451318435d31483/packages/astro/src/core/app/index.ts#L241) function from Astro in the `<StarlightPage>` component which would allow to run middleware by the user afterwards?

Please just ignore this idea if it's stupid... 🤣

delucis:
As this PR is going to impact plugin authors as well as regular Starlight users I wanted to give @HiDeoo, @martrapp, @Fevol, @trueberryless, and @lorenzolewis and extra heads up in case you have feedback.

You can read about how route data is now handled in the new [“Route Data” guide](https://deploy-preview-2390--astro-starlight.netlify.app/guides/route-data/) in the preview deploy of our docs.

Basically anywhere you currently use `Astro.props` in component overrides will need updating to use `Astro.locals.starlightRoute` instead. Something that is new is the concept of [“route middleware”](https://deploy-preview-2390--astro-starlight.netlify.app/guides/route-data/#how-to-customize-route-data), which will allow plugins to inject a function that can modify the data for the current route. So any plugin that does something like `Astro.props.siteTitle = 'Some other title'`, filters the sidebar etc. will now be able to do that using a middleware function without necessarily injecting an override.

If any of you have thoughts on this change and how it impacts you, please let me know!

trueberryless:
I have really been looking forward to this PR for a some months, writing a lot with HiDeoo about how **powerful** these changes actually are. In my opinion, @HiDeoo describes it very well here (summed up from Discord):

> Basically all the mutations these overrides do would be moved to a middleware, meaning the override would be left rendering something really small, e.g. a dropdown or something else, which can be exported to a component that the user can compose very easily in their own overrides. And for cases more complex, plugins will be able to also add things to route data, so another plugin could check for example `Astro.locals.isBlogRoute === true` or something like that and have some special handling for these very complex cases.

But at the same time, I think that the powerfulness of the `Astro.locals` is not demonstrated enough in the docs. In my humble opinion, adding a more complex example, like an override from Starlight Blog, to the docs would be awesome, alto I get the idea of documenting a very basic idea, like the `!` added to all `siteTitle`s...

Maybe, I'll just look at how HiDeoo will adapt the blog plugin and take it as a premium reference in order to write a short blog post on my own website about how overrides in Starlight are done the right and perfect way. I think that some documentation of that kind will help other plugin creators as well, but please let me know what you (Chris and plugin authors) think about my thoughts...

trueberryless:
Also, [I have a small suggestion](https://github.com/withastro/starlight/pull/2778) which I think should be discussed in the mentioned PR, not here...

martrapp:
Cool approach! 

This might take a bit of time for people to migrate but doesn't seem too big a change and it is well documented. 

Any ETA yet? 

Shouldn't the examples include a call to next() at the end to enable possible chaining?

trueberryless:
> Shouldn't the examples include a call to next() at the end to enable possible chaining?

I think as Chris mentioned in the description, that the for loop handles all middleware, so no next is needed:

```js
for (const fn of routeMiddleware) {
  context.locals.routeData = fn(context.locals.routeData);
}
```

HiDeoo:
> But at the same time, I think that the powerfulness of the `Astro.locals` is not demonstrated enough in the docs. In my humble opinion, adding a more complex example, like an override from Starlight Blog, to the docs would be awesome, alto I get the idea of documenting a very basic idea, like the `!` added to all `siteTitle`s...

Personally I think the Starlight feature is only `context.locals.starlightRoute` or `Astro.locals.starlightRoute`. The fact that it is attach to `context.locals` and `Astro.locals` is just a way to make it accessible to the user. The documentation links to the Astro documentation for the Astro render context which is the Astro feature used under the hood and I think the example you're suggesting is more of a plugin author taking advantage of the Astro feature rather than the new Starlight change in this case.

I guess a similar example could be a Starlight plugin adding an Astro integration to update the Astro config to add a remark plugin. That's something possible but the Starlight documentation does not have an example of that because it's mostly Astro features under the hood.

That's only my opinion of course but definitely open to discussing it further.

trueberryless:
> How breaking must a change be to justify a major version change? 😬

I'm not entirely sure how the Starlight maintainers see this, but I think right now Starlight is not yet ready for v1 release... That's I think the reason why only minor releases happen. Once most of the issues are fixed and most users are satisfied, Starlight v1 will probably be released, but generally speaking that's a good question...

trueberryless:
> I think the example you're suggesting is more of a plugin author taking advantage of the Astro feature rather than the new Starlight change in this case.

That's a very good point and 100% true. I meant exactly this advantage. This is the reason, why I came to the conclusion that writing my own blog to remind myself how I can use the features of Astro and Starlight in an exemplary way is probably the best option, leaving the current docs as they are. Thanks for the quick response!

HiDeoo:
> I'm not entirely sure how the Starlight maintainers see this, but I think right now Starlight is not yet ready for v1 release...

There are definitely some planned features and fixes before a v1 release. Technically speaking about [semver](https://semver.org/), "major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable." but I think we've been pretty good about only releasing breaking changes in minor versions with detailled changelog and migration guides.

delucis:
Yeah, in general, we treat each Starlight minor like a small “major”. So no (intentional 😅) breaking changes in patches, but there can be breaking changes between minors as we iron out an API that we’ll keep for v1. That also aligns with how npm handles things, for example a dependency range of `^0.30.0` can install any 0.30.x patches, but won’t also include 0.31, unlike for later versions where `^1.30.0` would theoretically include all v1 releases, including e.g. 1.99.9.

Re: ETA, depends a bit — I have some details still to work on. But maybe end of next week is a plausible timeline.

martrapp:
Thanks, everyone, for the refreshers! I can explain it! ;-)

Most of the time I’ve worked with software, versioning usually started at 1.0. I'm still too used to thinking of 0.x versions as immature precursors that might never fully materialize. But that definitely doesn't apply to Starlight ;-) That's probably where my unease comes from. Apparently, I’m not the only one who finds years of 0.x versions odd. See Anthony Fu’s article on this for more: https://antfu.me/posts/epoch-semver.
Cheers!

trueberryless:
Quick comment: I love the documentation addition from https://github.com/withastro/starlight/commit/a7eebada417e93368d568690a64bb4141e1da91b

Really nice example 👍

martrapp:
> Really nice example 👍

First thing I'll do when it comes out, maybe a bit truncated ;-)

trueberryless:
@delucis Can we please ask Sarah if we can leave the middleware? 🥺 

It looks awesome!!!

image

Or do we not even have to ask Sarah in the Starlight Docs, only in Astro i guess...

CHRIS CAN WE LEAVE THE MIDDLEWARE IN PLS??? 🥺 🥺 🥺 🥺 🤣

delucis:
Just updated the PR with proper APIs for plugins to use and a new middleware runner that adds support for `next()` callbacks to allow middleware to delay code execution until after other middleware runs.

Summary:

- Plugins can no longer directly update `routeMiddleware` in the Starlight config.
- Instead, they use a new `addRouteMiddleware()` callback. This has the same signature as the Astro `addMiddleware()` requiring an `entrypoint` and an option `order` property.
- When we run plugins, we collect all middleware and then apply them based on the `order` values.
- Middleware handlers can `await next()` to delay code execution until later middleware has completed. Calling `next()` is optional.
- Docs have been updated to reflect these changes.

Marked this as ready for review as I think now we have all the pieces in place?

I’m still a bit nervous about just _how_ breaking these changes are. I imagine a lot of plugins and projects will be impacted.

One possibility:

We _could_ walk back the changes to drop `Astro.props` and do that later potentially.

In this scenario:

1. We still generate things on locals and run middleware
2. We then pass the route data as `Astro.props` down through components as currently

That would mean existing overrides would continue to work while people migrate. However, it would not be possible to avoid a future breaking change because at some point we’d still remove the prop drilling. So it may make sense to do it now and make sure we release this when we have time to help with updating as many plugins as possible.

---

Things still on my radar:

- There are some broken links that need fixing
- I still haven’t been able to revisit some feedback on the “Route Data” docs page @sarah11918 shared, but that felt lie it could follow later if necessary.

docs/src/content/docs/reference/plugins.md
```diff
@@ -23,11 +23,17 @@ interface StarlightPlugin {
...
+        order?: 'pre' | 'post';
+      }) => void;
       astroConfig: AstroConfig;
       command: 'dev' | 'build' | 'preview';
       isRestart: boolean;
       logger: AstroIntegrationLogger;
-      injectTranslations: (Record<string, Record<string, string>>) => void;
+      injectTranslations: (dict: {
+        [lang: string]: Record<string, string>;
+      }) => void;
```
HiDeoo, line 36: Curiosity: is this just a readability change to make it more obvious the keys in the top-level object are langs?

docs/src/content/docs/guides/route-data.mdx
HiDeoo, line 1: Considering middlewares may run for non-Starlight pages, and in this case `starlightRoute` will contain the data of a 404, I think we may need to document a little bit more this behavior and add documentation regarding 404s.

Some questions coming to mind are:

- Should we document that `id`s for 404 route will always be `404`?
- Would a user need to discriminate a real 404 vs a 404 entry data for a custom page?
- Is having `starlightRoute` always defined to avoid some nullish coalescing operator or a type-guard the best approach? 
- Would a discriminated union be helpful in this case, e.g. `{ isStarlightRoute: false, starlightRoute: undefined }` and `{ isStarlightRoute: true, starlightRoute: StarlightRouteData }`?

HiDeoo (commented):
Played quite a bit this week-end with the new changes we discussed. This is working pretty well. Left a few questions and high-level comments/concerns but overall, no real issues with the new approach.

Amazing work 👏

.changeset/chilled-bees-pump.md
```diff
@@ -0,0 +1,36 @@
...
+
+Moves route data to `Astro.locals` instead of passing it down via component props
+
+⚠️ **Breaking change:**
+Previously, all of Starlight’s templating components, including user or plugin overrides, had access to a data object for the current route via `Astro.props`.
+This data is now available as `Astro.locals.starlightRoute` instead.
+
+To update, refactor any component overrides you have:
+
+- Remove imports of `@astrojs/starlight/props`, which are no longer required.
```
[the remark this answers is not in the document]
trueberryless: ", which ..., which" would sound bad, maybe something like:

> Remove deprecated imports of `@astrojs/starlight/props`, which are no longer required.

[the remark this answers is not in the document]
HiDeoo: Oh, personally I was suggesting to replace the existing end of sentence, but open to anything as wording is not my strong suit 😅

docs/src/content/docs/guides/route-data.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+Customizing route data can be useful when you want to modify how Starlight processes your data in a way not possible with existing configuration options.
+
+For example, you may want to filter sidebar items or customize titles for specific pages.
+Changes like this do not require modifying Starlight’s default components, only the data passed to those components.
+
+### How to customize route data
+
+You can customize route data using a special form of “middleware”.
+This is a function that is called every time Starlight renders a page and can modify values in the route data object.
```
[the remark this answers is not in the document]
delucis: Thinking about this a bit more. I really don’t like how this leaks everywhere.

I moved stuff to an Astro middleware, which is clean and allowed me to type `Astro.locals.starlightRoute` as always being defined, which is nice for component code because you don’t need guards everywhere to check if stuff is available. But the fact it means we may generate lots of route data unnecessarily (e.g. on non-Starlight pages), or run twice (e.g. on custom pages) feels BAD 😓 The image endpoint could be worked around, but it still feels like a code smell.

Given that, we probably need to handle it a bit differently and align it more with these docs which say “every time Starlight renders a page”. That is when this data is actually needed (unlike the `Astro.locals.t()` which makes sense to support in other pages).

To do that, I think we probably need to do something like this:
- Move the route data process back to our `Page` component (similar to how the `StarlightPage` component does it currently).
- Type `Astro.locals.starlightRoute` as potentially `undefined`.
- Try to improve DX for components that depend on the route data. Couple of ideas:
  - Provide a utility like `assertRouteData()` that throws if the data isn’t available (i.e. component used outside of a route), e.g.
    ```astro
    ---
    import { assertRouteData } from '@astrojs/starlight/route-data';
    
    assertRouteData();
    const {} = Astro.locals.starlightRoute;
                         // ^ guaranteed to be defined
    ---
    ```
  - Use a getter API instead that throws on non Starlight routes? That might feel more natural to people and would avoid extra imports:
    ```astro
    ---
    const {} = Astro.locals.starlightRoute;
                         // ^ under the hood, this is a getter with an
                         // initial implementation that throws, then we replace
                         // the implementation when generating route data
    ```

What do you think of that @HiDeoo?

docs/src/content/docs/reference/plugins.md
delucis: Ah this was technically a fix: the current docs don’t include the parameter name (`dict: `), just the type which isn’t valid TypeScript. A bit unfortunate that it wraps to be honest.

docs/src/content/docs/reference/plugins.md
```diff
@@ -23,11 +23,17 @@ interface StarlightPlugin {
       config: StarlightUserConfig;
       updateConfig: (newConfig: StarlightUserConfig) => void;
       addIntegration: (integration: AstroIntegration) => void;
+      addRouteMiddleware: (config: {
+        entrypoint: string;
+        order?: 'pre' | 'post';
```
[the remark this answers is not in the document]
delucis: Yeah, I wasn’t sure. I guess probably we should as their may be some programmatic cases where someone wants to toggle between all three potentially 👍

docs/src/content/docs/guides/route-data.mdx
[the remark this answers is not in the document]
HiDeoo: I think this is closer to the initial approach I described in my first comment to the PR when I was playing with the idea. I _don't_ think this would change any of the assumptions I made so far in my experiments but I would definitely have to confirm that once doable.

Regarding `assertRouteData()`:

- If used internally (our own components), that would mean someone would not be able to use such component outside of Starlight. This was not something officially supported but it was technically possible by passing the expected data as a prop, it would have been possible I guess with the previous approach too but would no longer be possible with this one (depending on what the check actually is).
- I think we would need a version that does not throw, e.g. `isRouteData()` or something too?
- May need to double check on that but I don't think we can have such utility asserting the type of something without passing such thing, so at least `assertRouteData(Astro)`.

Regarding a getter approach, this is exactly what I used in some early tests, seemed to work well altho if it throws by default, it prevents the easy creation of let's say a component doing something slightly different if used in a Starlight page vs in a non-Starlight page. A try/catch would be needed in that case. Not a big deal but just a thought I wanted to mention.

I think aligning on the approach described in the documentation (and probably the one people would expect?) makes more sense and was probably my biggest concern during the last review.

docs/src/content/docs/reference/plugins.md
HiDeoo: Wow, excellent catch, I totally missed that multiple times 😅

.changeset/chilled-bees-pump.md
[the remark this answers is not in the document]
trueberryless: Ah alright! I didn't understand that right then sry, that makes more sense! 👍

docs/src/content/docs/guides/route-data.mdx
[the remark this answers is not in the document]
trueberryless: > I think aligning on the approach described in the documentation (**and probably the one people would expect?**) makes more sense and was probably my biggest concern during the last review.

Just to give a quick feedback of a guy who doesn't understand 80% of the technical stuff that's being discussed here: Yes, I can confirm that the approach currently* described in the docs logically makes a lot of sense (although maybe hard to achieve behind the scenes) and is what I would expect! As I said, just my thoughts and honest feedback...

*Current docs ref: 

> You can customize route data using a special form of “middleware”. 
 This is a function that is called every time Starlight renders a page and can modify values in the route data object.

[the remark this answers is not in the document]
delucis: > someone would not be able to use such component outside of Starlight

I _think_ this is a reasonable position for us to take? Our components are designed to be used by us, not just anywhere, so I’m not sure it’s a goal to support that. If it IS, then we have to drop the locals idea entirely and stick to props. That’s also an option if we decide it’s important.

In a way there are two separate concerns:

1. Have a way to process data for each route (route middleware)
2. Simplify components and make the current route more widely available by moving route data to a global context (locals) 

We don’t _have_ to do both 1 and 2 necessarily. We could have a process that runs the route middleware, but then still maintain the prop drilling approach we use currently. That would make this feature non-breaking.

docs/src/content/docs/reference/plugins.md
delucis: Me too 😁

docs/src/content/docs/guides/route-data.mdx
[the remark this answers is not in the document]
HiDeoo: > > someone would not be able to use such component outside of Starlight
> 
> I _think_ this is a reasonable position for us to take?

Definitely, maybe I wasn't clear enough but I'm all for it, I was just mentioning it while thinking about everything.

It's already impossible to properly support that CSS-wise and most of the time doesn't make sense to use it outside of Starlight. The maintenance cost of supporting it would be too high too to justify it. The stance of saying to people "copy the code and cherry-pick what you need" like we're currently doing is the best way to go imo.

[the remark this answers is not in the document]
delucis: OK, gotcha! I personally do think that there are arguments that global context is not ideal if it can be avoided as a general rule, but I guess we’re deciding that the advantage of having an easily accessible data object for all components is powerful enough to justify this.

In that case, I’ll play with the getter option and we can call out explicitly in the changelog that you can’t use Starlight components elsewhere. (Well, you still could technically I guess. You would just need to set `Astro.locals.starlightRoute = myPropsObject` instead of passing `myPropsObject` into a component.)

For the future it would be good to understand the use cases people have for reusing things and see if we can address those with a proper approach. For example, if it is “I need the nav bar but nothing else” or “I just want search”, maybe there are targeted ways to support _just_ those use cases. (For example with a dedicated simpler template some people have asked for. Or turning something like `<Search>` into a user component.)

packages/starlight/route-data.ts
```diff
@@ -0,0 +1,8 @@
+import type { APIContext } from 'astro';
+export type { StarlightRouteData } from './utils/routing/types';
+
+type RouteMiddlewareHandler = (context: APIContext) => void | Promise<void>;
```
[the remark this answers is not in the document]
delucis: Fixed in 0de3e367

docs/src/content/docs/guides/route-data.mdx
delucis: I think this comment is resolved by moving to a getter that throws an error and deferring the middleware to only run in actual routes?

docs/src/content/docs/reference/plugins.md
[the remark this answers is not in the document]
delucis: Fixed in ca58aed3

docs/src/content/docs/guides/route-data.mdx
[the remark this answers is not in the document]
delucis: Made the changes to switch to a getter and generate route data inside the routes themselves rather than in the middleware — feels good I think. LMK if it seems like a reasonable approach.

packages/starlight/locals.ts
```diff
@@ -1,8 +1,38 @@
...
+ * avoid generating route data in this middleware which also runs for non-Starlight route.
+ */
+export function initializeStarlightRoute(context: APIContext) {
+	const state: { routeData: StarlightRouteData | undefined } = { routeData: undefined };
+	Object.defineProperty(context.locals, 'starlightRoute', {
+		get() {
+			if (!state.routeData) {
+				throw new AstroError(
+					'`locals.starlightRoute` is not defined',
+					'This usually means a Starlight component is being rendered outside of a Starlight page, which is not supported.'
```
HiDeoo, line 29: Should this also mention that it could be also that something is accessing `local.starlightRoute` outside of Starlight? I mean it's not only that case of rendering a Starlight component, it could be user-code accessing it, plugin code, etc. ?

packages/starlight/locals.ts
```diff
@@ -1,8 +1,38 @@
...
+ *
+ * This ensures Starlight components can easily access `starlightRoute` without needing type guards,
+ * we can throw a helpful message if `starlightRoute` is accessed on non-Starlight pages, and we
+ * avoid generating route data in this middleware which also runs for non-Starlight route.
+ */
+export function initializeStarlightRoute(context: APIContext) {
+	const state: { routeData: StarlightRouteData | undefined } = { routeData: undefined };
+	Object.defineProperty(context.locals, 'starlightRoute', {
+		get() {
+			if (!state.routeData) {
```
HiDeoo, line 26: Is a `try/catch` like you had in [`13d2038` (#2390)](https://github.com/withastro/starlight/pull/2390/commits/13d203846592f23b23120d6236f19fa9b389bd83) the recommended approach to check if it's defined? We don't provide any helper or boolean to avoid a `try/catch`?

I'm fine with that, just wondering.

I assume we should maybe document this behavior too in `docs/src/content/docs/reference/route-data.mdx`?

packages/starlight/locals.d.ts
```diff
@@ -13,5 +13,6 @@ declare namespace StarlightApp {
 declare namespace App {
 	interface Locals {
 		t: import('./utils/createTranslationSystem').I18nT;
+		starlightRoute: import('./utils/routing/types').StarlightRouteData;
```
HiDeoo, line 16: Maybe we should add a small JSDoc comment here for `starlightRoute` and also one for `t`.

When writing code, natural auto-completion for `Astro.locals` feels undocumented:

SCR-20250211-jokq

Compared to the one we provide for something in `Astro.locals.starlightRoute`:

SCR-20250211-jonr

HiDeoo (commented):
Done a new pass at reviewing the latest changes and getter approach. No breaking changes on my end and really happy with this latest version :clap:

Left a few comments/questions but really minor things. Amazing work!

packages/starlight/locals.ts
delucis: Hmm, good question! What are the scenarios/use cases? Components that should behave in a certain way inside Starlight routes but also work in other places?

packages/starlight/locals.d.ts
delucis: Great idea — tried doing this in [`883d4a5`](https://github.com/withastro/starlight/pull/2390/commits/883d4a54935a5ce9478f62bd478a26bee38b5022)

packages/starlight/locals.ts
delucis: When you say “outside of Starlight” where do you mean exactly? Somewhere other than “outside of a Starlight page”?

packages/starlight/locals.ts
HiDeoo: I guess this is now be way less common/important in this new implementation vs the previous one, considering now a route middleware will only run for Starlight routes which was not the case before.

Considering route data is now no longer in `Astro.props` but in `Astro.locals`, it allows to basically access the route data in lot of new places, so yeah, could be in a user-component and based on the presence of route data, you could decide to do something different. I could even imagine scenarios where you could check for it in another regular middleware.

I'm fine with the current approach if you need to check for it, just think it should be noted in the documentation that this can potentially throw an error.

packages/starlight/locals.ts
delucis: Ahh I see, yeah I see what you mean. OK, maybe something like this helps?

```suggestion
					'This usually means a component that accesses `locals.starlightRoute` is being rendered outside of a Starlight page, which is not supported.\n\n' +
						'If this is a component you authored, you can do one of the following:\n\n' +
						'1. Avoid using that component in non-Starlight pages.\n' +
						'2. Wrap the code that reads `locals.starlightRoute` in a  `try/catch` block and handle the cases where `starlightRoute` is not available.\n\n' +
						'If this is a Starlight built-in or third-party component, you may need to report a bug or avoid this use of the component.'
```

Renders like this and even shows the offending component:

Astro error pane showing the above error nicely formatted and the component where the error occurred

packages/starlight/locals.ts
delucis: I think I’m OK with requiring a `try/catch` for now? In the future we could always consider a `hasStarlightRoute` boolean or something? Although I don’t love if that would _maybe_ be confusing to other middleware? It would only indicate `starlightRoute` has been attached, _not_ whether it _might_ be attached later I guess.

packages/starlight/locals.ts
HiDeoo: Amazing, covers everything 🤩

delucis:
OK, I think this should be ready!

Noting I had to make some small link fixes to Hindi and Indonesian docs pages, so this should be merged with a Lunaria tracking directive in the commit message:

```
@lunaria-ignore:docs/src/content/docs/hi/**.mdx;docs/src/content/docs/id/**.mdx
```

Paging @yanthomasdev to confirm that will do what I want!

HiDeoo (approved):
Amazing work and such a great set of changes. Lots of future improvements to come for everyone thanks to this PR :star_struck:

trueberryless:
OMG, you guys have actually done it! I can't put it into words how excited I am! 🎉🍾🎂🎈🎁

PR #2404, opened 2024-09-24
i18n(zh-cn): Update `overrides.md`

Nin3lee:
#### Description

- Update `overrides.md`
- #1923


HiDeoo (approved):
Thanks for updating this translation :raised_hands:

PR #2417, opened 2024-09-26
[docs] add starlight overrides map link

sarah11918:
#### Description

 This is an unsolicited PR to add @HiDeoo 's [interactive Starlight Overrides Map](https://starlight-overrides-map.netlify.app/) to the Overriding Components guide.

I always expect to find this link here, or in the reference section, and then when I don't, I have to track down via searching Hideoo's repos. (Because I don't know how to bookmark?)

Unless Hideoo does not want this site publicized, I think this would be a good place for the link!


apatel369 (approved):
LGTM

delucis (approved):
Triple approve! ✅ ✅ ✅

PR #2421, opened 2024-09-27
WIP(zh-cn): add `using-components.mdx`

Nin3lee:
#### Description

- add `using-components.mdx`
- #2344 
- #2121 
- NOTE: the `CI/check` link will fail, because some referenced link are not exist, I will update in this PR step by step.

### DIFF
--- a/docs/src/content/docs/zh-cn/components/using-components.mdx
+++ b/docs/src/content/docs/zh-cn/components/using-components.mdx
@@ -1,0 +1,94 @@
+---
+title: 使用组件
+description: 通过 Starlight 使用 MDX 和 Markdoc 中的组件
+sidebar:
+  order: 1
+---
+
+组件能让你更轻松、更连贯地复用部分 UI 或者样式。
+示例可能包括链接卡片或嵌入 YouTube。
+Starlight 支持使用 [MDX](https://mdxjs.com/) 和 [Markdoc](https://markdoc.dev/) 文件中的组件，并提供了一些常用组件供你使用。
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/zh-cn/components/using-components.mdx
```diff
@@ -0,0 +1,94 @@
...
+与 MDX 不同的是，Markdoc 中的组件不需要导入。
+以下示例为，在 Markdoc 文件中渲染 Starlight 的 [卡片组件](/zh-cn/components/cards/)：
+
+```markdoc
+---
+# src/content/docs/example.mdoc
+title: 欢迎来到我的文档
+---
+
+{% card title="Stars" icon="star" %}
```
HiDeoo, line 50:
```suggestion
{% card title="星星" icon="star" %}
```

Took the translation from the card grids page, but not sure it's 100% correct. Feel free to tweak it if needed.

docs/src/content/i18n/zh-cn.json
HiDeoo, line 1: The file should be named `zh-CN.json` like the [built-in one](https://github.com/withastro/starlight/blob/main/packages/starlight/translations/zh-CN.json).

Note that just like in the current version, the custom translation won't work due to an unrelated issue with the i18n system. I'm planning to open a PR later to fix it and having this custom translation will be an additional check to visually ensure that the fix works as expected.

HiDeoo (approved):
Thanks for the update, looks perfect to me now :raised_hands:

PR #2443, opened 2024-10-07
fix bottom padding in search dialog

kevinzunigacuellar:
#### Description

- Fixes CSS issue where bottom padding is not applied in the search dialog. 

| Before                                                                                                               | After                                                                                                                |
|----------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------|
| image | image |

### DIFF
--- a/.changeset/thin-guests-listen.md
+++ b/.changeset/thin-guests-listen.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks @kevinzunigacuellar! Tested this in Safari, Chrome, and Firefox and can confirm it works as expected 🌟

PR #2444, opened 2024-10-07
i18n fixes and improvements

HiDeoo:
#### Description

This PR includes the following changes:

### Fix translations for languages with a region subtag

Languages with a region subtag (e.g. `zh-CN`) are not properly translated in the current implementation. This is visible in the https://starlight.astro.build/zh-cn/components/cards/ page:

- The built-in UI strings are not applied, e.g. "Search" in the search bar.
- The custom UI strings are not applied, e.g. "Preview" in the component preview.

image

This issue is due to the fact that `Astro.currentLocale` returns a BCP 47 language tag (usually called a `lang` in Starlight) and that the translation system creation is expecting a language path from the URL (usually called a `locale` in Starlight) that is later converted to a BCP 47 language tag. Passing down a `lang` instead of a `locale` fails that conversion making the translation of languages with a region subtag not work.

This PR changes the translation system creation to always use a BCP 47 language tag.

image

### Use `Astro.locals.t()` when possible

In various places, we were still creating more translation systems than necessary instead of using the already configured `Astro.locals.t()`. This PR replaces those instances with `Astro.locals.t()`.

### Docs

The documentation ["Accessing the current locale" example](https://deploy-preview-2444--astro-starlight.netlify.app/guides/i18n/#accessing-the-current-locale) has been updated to use [`getRelativeLocaleUrl()`](https://docs.astro.build/en/reference/api-reference/#getrelativelocaleurl) so that the generated link is always correct.


HiDeoo:
Thanks for the feedback 🙌 

> I noticed that the file `locals.ts` is getting the wrong locale

This PR refactors `useTranslations()` to now [accepts a `lang` instead of a `locale`](https://github.com/withastro/starlight/pull/2444/files#diff-bbce9364b05eb1346b991999ece70b963b2ae9316488366d2d0eac50c707ea7c). In Starlight, a locale usually represents the `path` used in i18n, e.g. `zh-cn` while `lang` is a BCP 47 language tag, e.g. `zh-CN` so this may be what you're seeing, which is a change from the previous behavior?

daniel-locatelli:
What I meant is that the UI translation problem is still happening when I use this current pull request.

I forked this repo and used as a local package to test on my own Astro project.
The problem seems to be that the `locals.ts` is incorrectly getting the locale "en" in place of "pt-br" for example.

However, that did not happen when I tried directly on the "docs" in the monorepo. When you log `locals.ts` in the monorepo it outputs the correct locale. In my case the output was "pt-BR", which is still not in the right format, but can be fixed just lowering the case.

I made a video so you can better understand. At the bottom left is this current pull request. At the right is my own project, you can see it is not logging the wright locale:

https://github.com/user-attachments/assets/c1f89f54-9522-4d70-9bf7-9d77e527f0c5

HiDeoo:
I see, I think there are 2 points here:

> The problem seems to be that the locals.ts is incorrectly getting the locale "en" in place of "pt-br" for example.

Always getting `en` is definitely an issue. Altho, it should be at a different level, likely when we [transform](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/i18n.ts#L73-L98) the Starligh i18n config to an Astro i18n config. This part of the code should not be impacted by the changes in this PR.

I tried reproducing the issue locally but I couldn't using the following config I assumed from your video:

```js
defaultLocale: "en",
locales: {
  en: { label: "English", lang: "en" },
  "pt-br": { label: "Português do Brasil", lang: "pt-BR" },
},
```

 Would you be able to share your Starlight config so I can investigate further?

> In my case the output was "pt-BR", which is still not in the right format, but can be fixed just lowering the case.

This is expected with the current changes as `Astro.currentLocale` returns what is called a `lang` in Starlight (or a [`code`](https://docs.astro.build/en/guides/internationalization/#custom-locale-paths) in Astro i18n) and not a `locale` as I described in this [comment](https://github.com/withastro/starlight/pull/2444#issuecomment-2422677456).

daniel-locatelli:
I already tried different configs too. Using Astro's native i18n or with starlight. No luck so far.
I am about to start a new project and see if I manage to reproduce the error.

My current config:
```
// https://astro.build/config
export default defineConfig({
  site: getSite(),
  base: BASE_PATH,
  output: "server",
  adapter: cloudflare({
    platformProxy: {
      enabled: true,
    },
  }),
  experimental: {
    serverIslands: true,
  },
  integrations: [
    sitemap(),
    react(),
    icon(),
    tailwind(),
    starlight({
      title: "ArchCompute",
      defaultLocale: "en",
      locales: {
        en: { label: "English", lang: "en" },
        "pt-br": {
          label: "Português Brasileiro",
          lang: "pt-BR",
        },
      },
      disable404Route: true,
    }),
    mdx(),
  ],
});

```

HiDeoo:
Oh, that just hit me, which version of Astro are you using?

I had a similar issue a few days ago that was an issue in Astro which was fixed in [Astro 4.16.4](https://github.com/withastro/astro/releases/tag/astro%404.16.4). Would you be able to test using the latest Astro version?

daniel-locatelli:
Oh man... that was it! Thanks a lot.
Shame on me for not think about updating Astro 😅 It was at 4.15.11

Also thanks for this PR!

HiDeoo:
Amazing to hear, thanks for the follow-up and checking it out :raised_hands: And no problem, I had the same problem and it took me a while to figure it out, it's a bit tricky to spot an upstream i18n issue while working in on another i18n issue :sweat_smile:

packages/starlight/utils/createTranslationSystem.ts
```diff
@@ -56,7 +56,7 @@ export function createTranslationSystem<T extends i18nSchemaOutput>(
 	 * - `exists()` method for checking if a key exists in the dictionary.
 	 * - `dir()` method for getting the text direction of the locale.
 	 *
-	 * @param {string | undefined} [locale]
+	 * @param {string | undefined} [lang]
```
delucis, line 59: The comment description on line 52 also needs updating (thanks GitHub for not letting me suggest the change directly).

delucis (approved):
Looks great! I noticed one tiny detail in a comment, but no other notes — will fix that and conflicts now so we can get a patch out with these important fixes.

PR #2452, opened 2024-10-08
i18n(de): translate badges

trueberryless:
#### Description

Translation of `badges.mdx` into German language.


docs/src/content/docs/de/components/badges.mdx
```diff
@@ -0,0 +1,150 @@
...
+### Abzeichen anpassen
+
+Passe Abzeichen an, indem du andere `<span>`-Attribute wie `class` oder `style` mit benutzerdefiniertem CSS verwendest.
+
+<Preview>
+
+```mdx "style={{ fontStyle: 'italic' }}"
+import { Badge } from '@astrojs/starlight/components';
+
+<Badge text="Individuell" variant="success" style={{ fontStyle: 'italic' }} />
```
HiDeoo, line 106: Just to double check, and this is probably my translation tool that is not good enough, is "Individuell" the expected word for "Custom"?

docs/src/content/docs/de/components/badges.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+## `<Badge>` Eigenschaften
+
+**Implementation:** [`Badge.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/user-components/Badge.astro)
+
+Die `<Badge>`-Komponente akzeptiert die folgenden Eigenschaften und auch alle [anderen `<span>`-Attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes):
+
+### `text`
+
+**erforderlich**
```
HiDeoo, line 134:
```suggestion
**Erforderlich**  
```

Should this be capitalized for consistency with "Typ" and "Voreinstellung"? Let me know what you think.

HiDeoo (commented):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). Left a few very minor questions but other than that, this looks good to me.

docs/src/content/docs/de/components/badges.mdx
trueberryless: There are probably many different words possible for the german translation of "custom", "individuell" is probably NOT one of them xD. Nonetheless, I think it fits the situation perfectly, but I would love to hear thoughts from other Native Speakers, f.e. @delucis

docs/src/content/docs/de/components/badges.mdx
trueberryless: imho it should not be capitalized...

why?

required (and erforderlich as well) are adjectives 

type (and Typ and Voreinstellung) are nouns

in German, the rules for capitalization are complex but generally speaking only nouns are capitalized.

someone could argue that all three must be written capitalized not only because of consistency but also because it's the first letter of a new "sentence" / "paragraph". but I think it's not a sentence and therefore it can be written small...

same as before: would love to hear some comments from other native speakers...

Maxframe: Hmm.. Personally, I would prefer to capitalize all of them just for the sake of consistency. Even though 'erforderlich' is an adjective, I believe that capitalizing it along with 'Typ' and 'Voreinstellung' looks cleaner and more uniform. From a design perspective, this approach seems more visually balanced.

docs/src/content/docs/de/components/badges.mdx
```diff
@@ -0,0 +1,150 @@
...
+<Badge
+	slot="preview"
+	text="Individuell"
+	variant="success"
+	style={{ fontStyle: 'italic' }}
+/>
+
+</Preview>
+
+## `<Badge>` Eigenschaften
```
Maxframe, line 126:
```suggestion
## `<Badge>`-Eigenschaften
```

Little missing Bindestrisch. Apart from that LGTM. Great work :)

HiDeoo (approved):
Thanks for updating the PR and also for the extra review @Maxframe , really appreciated :star2:

PR #2467, opened 2024-10-11
i18n(de): update translation getting-started

trueberryless:
#### Description

Translation of `getting-started.mdx` into German language.


docs/src/content/docs/de/getting-started.mdx
```diff
@@ -5,33 +5,34 @@ description: Lerne wie du deine nächste Dokumentationsseite mit Starlight und A
 
 import { Tabs, TabItem } from '@astrojs/starlight/components';
 
-## Erstelle ein neues Projekt
+Starlight ist ein voll funktionsfähiges Dokumentationsthema, welches auf dem [Astro](https://astro.build) Framework aufbaut.
```
trueberryless, line 8: Need help with the translation of "documentation theme". Currently chose "Dokumentationsthema" because online translators suggest it but it feels wrong. Not entirely sure what it means in the English version, so I cant find the right German word...

docs/src/content/docs/de/getting-started.mdx
```diff
@@ -89,28 +102,25 @@ Da es sich bei Starlight um eine Beta-Software handelt, wird es regelmäßig Upd
 
 Starlight ist eine Astro-Integration und wird wie jede `@astrojs/*`-Integration aktualisiert:
```
HiDeoo, line 103: This sentence is missing the changes from https://github.com/withastro/starlight/pull/1241/files

docs/src/content/docs/de/getting-started.mdx
```diff
@@ -126,4 +136,4 @@ In der wachsenden Liste der Anleitungen in der Seitenleiste findest du Hilfe bei
 
 Wenn du keine Antwort in dieser Dokumentation finden kannst, besuche bitte die [Astro Docs](https://docs.astro.build/de/) für die vollständige Astro-Dokumentation. Deine Frage kann vielleicht beantwortet werden, wenn du verstehst, wie Astro im Allgemeinen funktioniert.
 
-Du kannst auch nach bekannten [Starlight-Problemen auf GitHub](https://github.com/withastro/starlight/issues) suchen und im [Astro Discord](https://astro.build/chat/) Hilfe von unserer aktiven, freundlichen Community erhalten! Du kannst Fragen in unserem `#Support`-Forum stellen oder unseren speziellen `#starlight`-Channel besuchen, um aktuelle Entwicklungen und mehr zu diskutieren.
+Du kannst auch nach bekannten [Starlight-Problemen auf GitHub](https://github.com/withastro/starlight/issues) suchen und im [Astro Discord](https://astro.build/chat/) Hilfe von unserer aktiven, freundlichen Community erhalten! Du kannst Fragen in unserem `#support`-Forum stellen oder unseren speziellen `#starlight`-Channel besuchen, um aktuelle Entwicklungen und mehr zu diskutieren.
```
HiDeoo, line 139: I think the entire "Fehlerbehebung" section may be missing the changes from https://github.com/withastro/starlight/pull/1026/files#diff-130a89ed8dd4068e1a366e0a6a31adb124cc60387ba42faadf457e83f9583948

It also looks like the title is not a match with the English version which mention Starlight which is not the case here.

HiDeoo (changes requested):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). I left a few notes regarding some potential missing update but otherwise, the translation looks good to me.

docs/src/content/docs/de/getting-started.mdx
trueberryless: thank you very much for spotting my mistake! I did miss this completely, my bad!

docs/src/content/docs/de/getting-started.mdx
trueberryless: I rewrote the entire section, thank you for spotting the missing translation!

However, this paragraph (especially the endings of both sentences) sounds repetitive and weird in German, respectively. Therefore I rewrote it a little bit different than a 1:1 translation. I hope this is okay...

> If your answer cannot be found in these docs, please visit the [full Astro Docs](https://docs.astro.build) for complete Astro documentation.
Your question may be answered by understanding how Astro works in general, underneath this Starlight theme.

HiDeoo: It definitely does not have to be a 1:1 translation if you think it's better while still keeping the main points, and you definitely know better than me what is best. And if ever needed, we can always come back to this later and discuss it further.

PR #2470, opened 2024-10-11
i18n(de): update translation reference/frontmatter

trueberryless:
#### Description

Translation of `reference/frontmatter.mdx` into German language.


docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -338,16 +380,85 @@ sidebar:
...
+};
+```
+
+Mehr über Schemata für Inhaltssammlungen erfährst du in [„Definieren eines Sammelschemas“](https://docs.astro.build/de/guides/content-collections/#definieren-eines-sammelschemas) in den Astro-Dokumenten.
+
+`docsSchema()` nimmt die folgenden Optionen an:
+
+### `extend`
+
+**Typ:** Zod-Schema oder Funktion, die ein Zod-Schema zurückgibt
```
HiDeoo, line 418:
```suggestion
**Typ:** Zod-Schema oder Funktion, die ein Zod-Schema zurückgibt  
```

To get the default value on a new line.

docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -18,30 +19,37 @@ Willkommen auf der Info-Seite!
...
-**type:** `string`
+**Typ:** `string`
 
 Die Seitenbeschreibung wird für die Metadaten der Seite verwendet und wird von Suchmaschinen und in der Vorschau von sozialen Medien angezeigt.
 
+### `slug`
+
+**Typ**: `string`
+
+Setzt den Slug der Seite außer Kraft. Siehe [„Benutzerdefinierte Slugs definieren“](https://docs.astro.build/de/guides/content-collections/#benutzerdefinierte-slugs-definieren) in der Astro-Dokumentation für weitere Details.
```
HiDeoo, line 36:
```suggestion
Setzt den Slug der Seite außer Kraft. Siehe [„Benutzerdefinierte Slugs definieren“](https://docs.astro.build/de/guides/content-collections/#defining-custom-slugs) in der Astro-Dokumentation für weitere Details.
```

This page is not translated in the Astro Docs so we might as well link to the correct anchor until then.

docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -221,34 +241,51 @@ prev:
...
 ---
+# src/content/docs/example.md
 # Diese Seite aus dem Suchindex ausblenden
 pagefind: false
 ---
 ```
 
+### `draft`
+
+**Typ:** `boolean`
```
HiDeoo, line 273:
```suggestion
**Typ:** `boolean`  
```

To get the default value on a new line.

docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -311,14 +351,15 @@ sidebar:
 
 #### `badge`
 
-**type:** <code>string | <a href="/de/reference/configuration/#badgeconfig">BadgeConfig</a></code>
+**Typ:** <code>string | <a href="/de/reference/configuration/#badgeconfig">BadgeConfig</a></code>
 
 Füge der Seite in der Seitenleiste ein Abzeichen hinzu, wenn es in einer automatisch generierten Gruppe von Links angezeigt wird.
 Bei Verwendung einer Zeichenkette wird das Abzeichen mit einer Standard-Akzentfarbe angezeigt.
 Optional kann ein [`BadgeConfig` Objekt](/de/reference/configuration/#badgeconfig) mit den Feldern `text` und `variant` übergeben werden, um das Abzeichen anzupassen.
```
HiDeoo, line 358: This sentence seems to be missing the changes from this [PR](https://github.com/withastro/starlight/pull/1530/files#diff-6d478912950527786985d16b5fef998fc9a98ffc282f20ebe557b822d0e717bd).

docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -154,13 +168,14 @@ interface HeroConfig {
     link: string;
     variant: 'primary' | 'secondary' | 'minimal';
     icon: string;
```
HiDeoo, line 170:
```suggestion
    variant?: 'primary' | 'secondary' | 'minimal';
    icon?: string;
```

From https://github.com/withastro/starlight/pull/1784/files#diff-6d478912950527786985d16b5fef998fc9a98ffc282f20ebe557b822d0e717bd

docs/src/content/docs/de/reference/frontmatter.md
```diff
@@ -1,12 +1,13 @@
 ---
-title: Frontmatter Referenz
+title: 'Referenz: Frontmatter'
```
HiDeoo, line 2: This seems to be a "style" change from the other references. Is this expected? Do you think such change should be applied to other references?

As a non-native speaker, I unfortunately cannot really comment on this change so let me know.

HiDeoo (changes requested):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view). Left a few suggestions and questions but otherwise looks good to me.

docs/src/content/docs/de/reference/frontmatter.md
trueberryless: Oh, okay, I thought I can just link to the future... I did the same mistake on many other commits then...

trueberryless: Do you (@HiDeoo ) think its a good idea if I just change this common error in all my PRs and make a new issue on the Starlight repo with all the links that needs to be fixed in the future, cause I surely cannot remember them all...?

docs/src/content/docs/de/reference/frontmatter.md
trueberryless: OMG!!! I just discovered that it is possible to set a different label in the global sidebar in Starlight.

Hence I think I will remove all the duplicate mentions of "Referenz" in the sidebar, f.e. the sidebar label of the Frontmatter would be "Frontmatter" only because the folder already says "Referenzen". 

But on the actual side, the title could either be consistent with "Referenz: Frontmatter" or "Frontmatter Refernz". What would you like more @HiDeoo ?

docs/src/content/docs/de/reference/frontmatter.md
HiDeoo: I think it's better to actually fix them in PRs if they're spotted like this one and if later some broken anchors are found that may have been missed, then we can fix them at that time.

docs/src/content/docs/de/reference/frontmatter.md
HiDeoo: I think this may be one case where another native speaker may be required. If we simply take a look at the English version, "Reference" is repeated but maybe that does not convert well to some other languages and I definitely cannot be the judge of that.

image

trueberryless: Totally fair. I would say, it's also possible in German, grammatically and so, to include "Referenz" in every page title somehow. 

But what makes it kinda ugly is the fact, that in German words can theoretically be chained together indefinitely, e.g. "Konfigurationsreferenz" while in English the words are still separated. This leads me to believe that the English version is much easier to read, which could be countersteered, by not including the reference part in the German version.

That's the reason I would like to make this rather huge style change in the German version. But let's wait for some other native speakers to jump right into this discussion! 😊

HiDeoo: Discussed the proposed change with Chris and Hippo and they're both on board with the suggested change so we're going to go ahead with it. Thanks a lot for the patience until we figured this out and also suggesting the improvement.

HiDeoo (approved):
Thanks for updating the PR, everything looks ready to go now :rocket:

PR #2473, opened 2024-10-14
Update customization.mdx

waleedsadek-panx:
Social link isn't being displayed on the website. Added an extra line.

#### Description

On the https://starlight.astro.build/guides/customization/#social-links social inks h2 header isn't being displayed. Just added an extra line to display it. 
No

### DIFF
--- a/docs/src/content/docs/guides/customization.mdx
+++ b/docs/src/content/docs/guides/customization.mdx
@@ -196,6 +196,7 @@ defineConfig({
   </TabItem>
 </Tabs>
 
+
 ## Social links
 
 Starlight has built-in support for adding links to your social media accounts to the site header via the [`social`](/reference/configuration/#social) option in the Starlight integration.


HiDeoo:
Thanks for your contribution.

I may be missing something but I'm not entirely sure to see what is not being displayed and fixed with the proposed change.

| Before   | After    |
| -------- | -------- |
| image | image |

Would you mind providing more details on the issue?

waleedsadek-panx:
I just realized it's only an issue on `Firefox` as I checked `Chrome` and it's displayed but on `firefox` the `Social Links` H2 header are missing like so. 
Screenshot_20241015_081633

HiDeoo:
Interesting, would you be able to check with all extensions disabled? I tested in Firefox and cannot repro either. Maybe an adblocker or something similar picking "Social links" as something that they should hide?

waleedsadek-panx:
You're absolutely right! It's the ad blocker as soon as I disabled it, it worked!

PR #2482, opened 2024-10-14
i18n(de): update translation overriding-components

trueberryless:
#### Description

Translation of `guides/overriding-components.mdx` into German language.


HiDeoo (approved):
Thanks for the update, looks perfect to me :star2:

PR #2488, opened 2024-10-15
i18n(de): translate resources/community-content

trueberryless:
#### Description

Translation of `resources/community-content.mdx` into German language.


trueberryless:
I did not translate `🌟 SUB 1 MINUTE RUN` cause it sounds so cool in German and the translation would be scuffed. If I should translate it nevertheless, let me know!

HiDeoo (approved):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://hackmd.io/1KYnrpUQR0SAGTzHIUoUDg?view) and everything looks good to me 👍

PR #2505, opened 2024-10-23
docs: i18n improvements

HiDeoo:
#### Description

This PR implements tiny i18n related improvements for the documentation:

- Removes the `slugToLocaleData()` internal dependency in the languages list component in favor of using `Astro.currentLocale`.
- Adds a translatable "copied" label to the icons list as discussed in https://github.com/withastro/starlight/pull/2343#pullrequestreview-2313253701

### DIFF
--- a/docs/src/components/icons-list.astro
+++ b/docs/src/components/icons-list.astro
@@ -2,10 +2,18 @@
 import { Icon } from '@astrojs/starlight/components';
 import { Icons } from '../../../packages/starlight/components/Icons';
 
+interface Props {
+	labels?: {
+		copied?: string;
+	};
+}
+
+const { copied = 'Copied!' } = Astro.props.labels ?? {};
+
 const icons = Object.keys(Icons) as (keyof typeof Icons)[];
 ---
 
-<div class="icons-grid">
+<div class="icons-grid" data-label-copied={copied}>
 	{
 		icons.map((icon) => {
 			return (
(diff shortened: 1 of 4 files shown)


docs/src/components/icons-list.astro
```diff
@@ -2,10 +2,18 @@
 import { Icon } from '@astrojs/starlight/components';
 import { Icons } from '../../../packages/starlight/components/Icons';
 
+interface Props {
+	labels?: {
```
HiDeoo, line 6: This uses the same data structure used in the theme editor for labels which would be nice in case we ever need more labels in the future altho I guess an argument could be made to use a flat string property instead of an object in this case.

docs/src/content/docs/reference/icons.mdx
```diff
@@ -16,4 +16,4 @@ A list of all available icons is shown below with their associated names. Click
 
 import IconsList from '~/components/icons-list.astro';
 
-<IconsList />
+<IconsList labels={{ copied: 'Copied!' }} />
```
HiDeoo, line 19: Duplicated with the component default label but marks translations as outdated.

docs/src/components/icons-list.astro
```diff
@@ -2,10 +2,18 @@
...
+	};
+}
+
+const { copied = 'Copied!' } = Astro.props.labels ?? {};
+
 const icons = Object.keys(Icons) as (keyof typeof Icons)[];
 ---
 
-<div class="icons-grid">
+<div class="icons-grid" data-label-copied={copied}>
```
HiDeoo, line 16: The "copied" label is added only once to the DOM and not for every icons.

delucis: Perfect. We could even delegate the click handling here too I guess to avoid assigning a listener for every button? Doesn’t have to be in this PR though!

delucis (approved):
Looks great! Tested a few pages including ones with fallback content and everything seems to be working as expected. Happy we can get this content translated properly too 🙌

PR #2507, opened 2024-10-23
Fixes ToC highlighting after window resize

### LINKED ISSUE #2506: Right Sidebar Locks and Stops Syncing with Content After Window Resize

0.27.1

4.15.4

npm

win11

edge

### Describe the Bug

Normally, the current heading in the right sidebar TOC (Table of Contents) updates dynamically as you scroll through the main content. However, if you resize the browser window to a smaller width, causing the right sidebar to hide, and then resize it back to a larger width where the sidebar becomes visible again, the TOC stops updating. It becomes "frozen" at the position it was in before the window was resized and no longer syncs with the current scroll position in the main content.

Reproduce: This bug can be reproduced on the official Starlight website.

### Participation

HiDeoo:
#### Description

- Closes #2506

This PR fixes a table of contents highlighting issue after the window is resized.

When resizing the window, the current code disconnect the `IntersectionObserver` and re-runs the `observe` function after a small amount of time. Altho, this function starts by checking if the variable containing the `IntersectionObserver` is defined, which is still the case after a disconnect, and if it is, it doesn't re-run the logic to observe the elements.

This PR fixes the issue by setting the observer variable to `undefined` after disconnecting it so that the observer is re-created with the new `rootMargin` after the window is resized.

### DIFF
--- a/.changeset/silent-fireants-retire.md
+++ b/.changeset/silent-fireants-retire.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a table of contents highlighting issue after resizing the window.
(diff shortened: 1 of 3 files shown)


delucis (approved):
Ah perfect! Thank you @HiDeoo. I also wouldn’t be surprised if this is the same bug as one @TheOtterlord once reported on mobile, maybe with the viewport resizing because of browser chrome hiding/showing.

PR #2515, opened 2024-10-27
fix: apply `scrollbar-gutter: stable` to prevent layout shift with modals

### LINKED ISSUE #2514: Slight layout shift when opening search modal due to scroll bar disappearing

0.28.3 (whatever is on https://starlight.astro.build/ right now)

whatever is used on https://starlight.astro.build/ right now

whatever is used on https://starlight.astro.build/ right now

Mac

Chrome Beta 131

### Describe the Bug

When opening the page search modal, there's a small layout shift of the background due to the scroll bar disappearing in modal view.

It's difficult to avoid the layout shift, maybe one could reserve some width for the scrollbar in modal view as well.

To reproduce:
1. Click on the search modal
2. Observe layout shift

https://stackblitz.com/github/withastro/starlight/tree/main/examples/basics?file=README.md

### Participation

corneliusroemer:
#### Description

Closes #2514

Adds `scrollbar-gutter: stable` to html to prevent layout shift when scrollbar appears/disappears, e.g. when opening the search modal.

Before

After

### DIFF
--- a/packages/starlight/style/reset.css
+++ b/packages/starlight/style/reset.css
@@ -11,6 +11,7 @@
 html {
 	color-scheme: dark;
 	accent-color: var(--sl-color-accent);
+	scrollbar-gutter: stable;
 }
 
 html[data-theme='light'] {


delucis (commented):
Thanks for the issue and PR @corneliusroemer!

One issue with this approach is that on short pages without a scrollbar, you can end up with this slightly odd gap at the end of the nav bar:

Starlight top navigation bar. The background color stops short of the right-hand edge leaving an incongruous gap.

I’m wondering if maybe there’s a better way to handle this to avoid that?

corneliusroemer:
@delucis Thanks for having a look! There probably are ways of overflowing the navbar into the scrollbar gutter but I'm not a CSS expert so this is out of my depth. Stackoverflow has some suggestions: https://stackoverflow.com/a/70277203/7483211, e.g. combining scrollbar-gutter with `overflow-y: auto;`. The answer also has some nice fallbacks for browsers that don't (yet) support scrollbar-gutter (old Safari and Safari mobile)

delucis:
Hmm, thanks for the link. Those options seem to have other problems unfortunately. We’ll need to dig in and see if there’s a good way to solve this. I think given the current two options of small layout shift in the background on modal open and displaying a visible gap in the title bar, I’d choose the layout shift for now. But would love to solve it better if we can.

HiDeoo:
> Those options seem to have other problems unfortunately.

This has also been my experience so far with `scrollbar-gutter: stable` in this kind of scenarios, specially with fixed position elements.

To my knowledge, there are definitely some "hacky" ways to achieve this, but getting them right on all platforms, especially mobile, is super tricky.

This is a scenario where the obvious fix (`overscroll-behavior` on the `<dialog>`) does not work due to the [spec](https://drafts.csswg.org/css-overscroll-1/#overscroll-behavior-properties):

> An element that is not scroll container must accept but ignore the values of this property.

The WHATWG is [aware](https://github.com/whatwg/html/issues/7732) of the feedback, there are also on top of this some browser bugs and differences of behavior between Chrome, Firefox and Safari (bugs linked in the issue).

I'm personally not sure there is at the moment a perfect or even good solution for this sadly.

PR #2525, opened 2024-10-29
i18n(de): update `overriding-components.mdx`

trueberryless:
#### Description

- update `overriding-components.mdx`
- #2417


delucis (approved):
Super! Thanks for the update @trueberryless 🙌

PR #2539, opened 2024-10-29
Add catppuccin-starlight theme

TheOtterlord:
#### Description

Add catppuccin-starlight theme.

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -110,6 +110,11 @@ A theme is a Starlight plugin that changes the visual appearance of a site with 
 		title="starlight-theme-obsidian"
 		description="Starlight theme inspired by the style of Obsidian Publish sites."
 	/>
+	<LinkCard
+		href="https://github.com/TheOtterlord/catppuccin-starlight"
+		title="catppuccin-starlight"
+		description="📝 Soothing pastel theme for Starlight"
+	/>
 </CardGrid>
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/resources/plugins.mdx
```diff
@@ -110,6 +110,11 @@ A theme is a Starlight plugin that changes the visual appearance of a site with
 		title="starlight-theme-obsidian"
 		description="Starlight theme inspired by the style of Obsidian Publish sites."
 	/>
+	<LinkCard
+		href="https://github.com/TheOtterlord/catppuccin-starlight"
+		title="catppuccin-starlight"
+		description="📝 Soothing pastel theme for Starlight"
```
delucis, line 116: I know this is the repo description but given nothing else uses an emoji here, I'd suggest we leave it out. For fear of starting an emoji arms race between plugins 😄 

```suggestion
		description="Soothing pastel theme for Starlight"
```

delucis (approved):
Thanks @TheOtterlord! Just left one small comment, but otherwise LGTM! 😻

docs/src/content/docs/resources/plugins.mdx
TheOtterlord: Haha, fair enough. It's the standard tagline template for catppuccin, so I just copy-pasted.

trueberryless:
Nice to see some new themes coming to life 🧬! 👍

PR #2546, opened 2024-10-31
Stop silencing i18n content collection related errors

HiDeoo:
#### Description

This PR removes the existing behavior of silencing all errors related to the `i18n` content collection when we try to load it.

Initially, the error silencing was introduced in https://github.com/withastro/starlight/pull/1458 to prevent a runtime error when the `i18n` content collection did not exist or was empty.

Since then, this error was [removed](https://github.com/withastro/astro/pull/8382) in Astro and [an empty array is now returned](https://github.com/withastro/astro/pull/10356) when the `i18n` content collection does not exist or is empty.

Both changes are available since Astro v4.4.14 (which is higher than the minimum required version for Starlight (v4.14.0)) and this means that we can now entirely remove the try/catch block entirely.

Note that the warning silencing is still in place.

Screenshot of a possible error that is now visible:

SCR-20241031-obyr

### DIFF
--- a/.changeset/big-ducks-rhyme.md
+++ b/.changeset/big-ducks-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue where i18n content collection related errors, e.g. malformed JSON or YAML, would not be reported.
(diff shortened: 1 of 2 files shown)


delucis (approved):
Wonderful. Thank you for the PR and thorough testing. This will be a nice one for helping catch things like syntax errors.

PR #2548, opened 2024-10-31
Fix invalid locale issue

HiDeoo:
#### Description

This PR fixes a build issue visible in https://github.com/withastro/docs/pull/9890

For project without a root locale like the docs, for pages not matching the default locale slug, we were not falling back to the default locale.

I also tested the fix locally on the failing Astro Docs branch and it now builds properly.

### DIFF
--- a/.changeset/great-toys-train.md
+++ b/.changeset/great-toys-train.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Ensures that pages not matching the default locale slug in projects without a root locale fallback to the default locale.
+Fixes a URL localization edge case. In projects without a root locale configured, slugs without a locale prefix did not fall back to the default locale as expected.
--- a/.changeset/great-toys-train.md
+++ b/.changeset/great-toys-train.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Ensures that pages not matching the default locale slug in projects without a root locale fallback to the default locale.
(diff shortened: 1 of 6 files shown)


.changeset/great-toys-train.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Ensures that pages not matching the default locale slug in projects without a root locale fallback to the default locale.
```
HiDeoo, line 5: Feels like this could maybe be simplified somehow, gives a bit of a headache to read 🤯

delucis: I tried!
```suggestion
Fixes a URL localization edge case. In projects without a root locale configured, slugs without a locale prefix did not fall back to the default locale as expected.
```

delucis (approved):

HiDeoo:
@delucis Just realized the fix is wrong, please hold on merging

delucis (changes requested):
@HiDeoo said the fix is wrong!

HiDeoo:
To explain with a bit more details, I just realized we actually had 2 implementations of `slugToLocale()`, 1 tested and the other one not tested, and this PR only updated 1 of them.

I just pushed new changes removing the duplicated code and both code paths now use the same underlying implementation which ends up being tested.

I think there could be some potential reorganization of the code needed, e.g. `packages/starlight/integrations/shared/slugToLocale.ts` is not only used by integrations. We could do that later in a separate PR or now if preferred but I don't think I now have a clear idea on the structure/naming we could use.

delucis (approved):
Thanks for catching the duplication!

delucis:
> I think there could be some potential reorganization of the code needed, e.g. `packages/starlight/integrations/shared/slugToLocale.ts` is not only used by integrations.

I guess it will depend what else is using these? But for sure I think we can release this fix and refactor later.

PR #2565, opened 2024-11-03
Add YAML support to the FS translation system

HiDeoo:
#### Description

This PR fixes an issue where the FS translation system (only used with some internal integrations or remark plugins at the moment) was not handling YAML files even though Astro Content Collections support YAML files.

### DIFF
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -177,6 +177,7 @@
   "devDependencies": {
     "@astrojs/markdown-remark": "^5.1.0",
     "@playwright/test": "^1.45.0",
+    "@types/js-yaml": "^4.0.9",
     "@types/node": "^18.16.19",
     "@vitest/coverage-v8": "^1.6.0",
     "astro": "^4.15.3",
@@ -196,6 +197,7 @@
     "hast-util-to-string": "^3.0.0",
     "hastscript": "^9.0.0",
     "i18next": "^23.11.5",
+    "js-yaml": "^4.1.0",
     "mdast-util-directive": "^3.0.0",
     "mdast-util-to-markdown": "^2.1.0",
     "mdast-util-to-string": "^4.0.0",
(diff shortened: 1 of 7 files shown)


packages/starlight/package.json
```diff
@@ -196,6 +197,7 @@
     "hast-util-to-string": "^3.0.0",
     "hastscript": "^9.0.0",
     "i18next": "^23.11.5",
+    "js-yaml": "^4.1.0",
```
HiDeoo, line 200: Using the same parser used in Astro.

packages/starlight/utils/translations-fs.ts
```diff
@@ -23,15 +28,18 @@ export function createTranslationSystemFromFs<T extends i18nSchemaOutput>(
...
+		for (const file of files) {
+			const filePath = path.parse(file);
+			if (!contentCollectionFileExtensions.includes(filePath.ext)) continue;
+			const id = filePath.name;
+			const url = new URL(filePath.base, i18nDir);
+			const content = fs.readFileSync(new URL(file, i18nDir), 'utf-8');
+			const data =
+				filePath.ext === '.json'
+					? JSON.parse(content)
+					: yaml.load(content, { filename: fileURLToPath(url) });
```
HiDeoo, line 40: As YAML is a superset of JSON, we could technically have only used the YAML parser, although I decided against it for 2 reasons:

- We have a test erroring when parsing an empty JSON file, which would no longer have been an error if we used the YAML parser.
- For malformed JSON files, an YAMLException would be thrown, which I think could be confusing for users to mention YAML when they are using JSON files.

delucis (approved):
The logic and tests all look perfect — thanks @HiDeoo!

Hitting this does make me a tiny bit nervous about the content loader future of all formats being available, but at least for now we can align with Astro’s built-in support 🎉

PR #2571, opened 2024-11-04
Prettier ignore a malformed YAML test file

HiDeoo:
This PR prevents Prettier to try formatting malformed YAML files used during tests.


delucis (approved):
Thanks for the quick fix!

PR #2573, opened 2024-11-04
Throw an error if a showcase image does not have the required dimensions

delucis:
#### Description

- This PR adds a validation to our showcase card component to ensure passed images have the expected dimensions.
- We often get images submitted that do not have the required dimensions and contributors don’t get any feedback about this until we review the PR. The aim here is that they will get local dev errors if previewing the site, or failing that, CI failures when they open the PR. It will also help PR reviewers spot the issue more easily.
- There is currently one image that has the wrong dimensions. Will resize it in a subsequent commit so we can see this fail once in this PR.


HiDeoo (approved):
Beautiful 🤩 

image

PR #2578, opened 2024-11-05
Expose FS translation system to plugins

HiDeoo:
#### Description

This plugin exposes the file system translation system to Starlight plugins throught the `setup` hook using a new `useTranslations()` callback function.

This change requires the addition of a new plugin hook (currently named `init`, I picked the first thing coming to mind to get started) which now exposes the `injectTranslations()` callback function (which was previously available in the `setup` hook).

The documentation so far is only a rough draft to get an idea of everything changed.

#### Remaining tasks

- [x] Restore the `docs/astro.config.mjs` file
- [x] Restore the `docs/package.json` file
- [x] Restore the `docs/src/content/i18n/zh-CN.json` file
- [x] Remove the entire `packages/demo-plugin/` directory
- [x] Run `pnpm i` to remove mention of the `starlight-demo-plugin` package in the lockfile
- [x] Remove all remaining `// TODO(HiDeoo)` comments (if any)
- [ ] Order the changeset in the release PR so that `.changeset/polite-fishes-remain.md` comes first (``Deprecates the Starlight plugin `setup` hook in favor of the new `config:setup` hook which provides the same functionality.``)


packages/starlight/integrations/expressive-code/index.ts
```diff
@@ -152,7 +151,7 @@ export function getStarlightEcConfigPreprocessor({
 				},
 				...otherStyleOverrides,
 			},
-			getBlockLocale: ({ file }) => pathToLocale(file.path, { starlightConfig, astroConfig }),
+			getBlockLocale: ({ file }) => pathToLang(file.path, { astroConfig, starlightConfig }),
```
HiDeoo, line 154: Expressive Code technically expects a Starlight language here so I changed it but this is not a bug fix (no changeset) as Expressive Code lower-cases the language identifier anyway.

docs/src/content/docs/reference/plugins.md
```diff
@@ -163,70 +245,75 @@ The example above will log a message that includes the provided info message:
...
-  // Define the additional plugin translations in the `I18n` interface.
-  interface I18n {
-    'myPlugin.doThing': string;
-  }
-}
-```
+**type:** `(path: string) => string`
 
-You can also infer the types for the `StarlightApp.I18n` interface from a source file if you have an object containing your translations.
+A callback function to get the language for a given full file path.
```
delucis, line 278: Aligning this with the previous suggestion, which I also like because it makes clear the arguments to pass as well as the purpose.

```suggestion
Call `pathToLang()` with a file path to get the language for that file.
```

One question here though: in general these docs always say “full” file path. What happens if I do `pathToLang('src/content/docs/en/foo.md')`? Or even `pathToLang('en/foo.md')`? Is it dependent on being an absolute path?

docs/src/content/docs/reference/plugins.md
```diff
@@ -163,70 +245,75 @@ The example above will log a message that includes the provided info message:
...
 
-For example, given the following source file:
+This can be particularly useful in [remark or rehype plugins](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) that a Starlight plugin may use to process Markdown or MDX files.
+The [virtual file format](https://github.com/vfile/vfile) used by such plugins includes the [full path](https://github.com/vfile/vfile#filepath) of the file being processed, which can be used with `pathToLang` to determine the language of the file.
+The returned language can be used with the [`useTranslations`](#usetranslations) helper to get UI strings for that language.
```
delucis, line 282: Suggesting some tiny tweaks here:
- removed the “Starlight plugin” part as it helps simplify the sentence and hopefully it’s obvious on this page we’re referring to plugins
- added `()` to the mention of the functions like we usually do

```suggestion
This can be particularly useful when adding [remark or rehype plugins](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) to process Markdown or MDX files.
The [virtual file format](https://github.com/vfile/vfile) used by these plugins includes the [full path](https://github.com/vfile/vfile#filepath) of the file being processed, which can be used with `pathToLang()` to determine the language of the file.
The returned language can be used with the [`useTranslations()`](#usetranslations) helper to get UI strings for that language.
```

delucis (commented):
This looks great @HiDeoo! Left some suggestions/questions/discussions, but they’re honestly all around the “outside” of the core functionality or on the docs. I think overall this seems very solid 🚀

.changeset/large-balloons-compete.md
```diff
@@ -0,0 +1,29 @@
...
+
+This addition changes how Starlight plugins add or update translation strings used in Starlight’s localization APIs.
+Plugins previously using the [`injectTranslations()`](https://starlight.astro.build/reference/plugins/#injecttranslations) callback function from the plugin [`setup`](https://starlight.astro.build/reference/plugins/#hookssetup) hook should now use the same function available in the [`init`](https://starlight.astro.build/reference/plugins/#hooksinit) hook.
+
+```diff
+export default {
+  name: 'plugin-with-translations',
+  hooks: {
+-   setup({ injectTranslations }) {
++   init({ injectTranslations }) {
```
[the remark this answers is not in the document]
HiDeoo: > Regarding the hook naming, I was wondering if we should follow the colon-separated pattern used by Astro integrations?

I personally like that idea.

> but maybe a bit misleading if you don’t set up i18n config here, only translations.

I guess, altho if we compare with our configuration file, we use the `locales` key to configure i18n, not `i18n` so I think it's fine.

> I guess we might have a config:done hook like Astro does that gives access to the final resolved Starlight config?

Maybe, I don't really see the need for it right now and I'm not even sure what you would do with it.

> which your current docs kind of suggests: “initialization function called before Starlight is initialized (during the astro:config:setup integration hook)”

Well, to be fair, like I said, I didn't really put any thought at all about the naming of the hook, and I documented it mostly to give an overview of all the changes and mimick what we had before.

I think if we ever need something else to be done before `config:setup` that is not `i18n`, we could still add an `xyz:setup` and call it before `config:setup` like we do with `i18n:setup` and it would be fine.

> If we do rename setup() it doesn’t have to be in this PR and can be done in a backwards compatible way where we still call setup() if defined but mark it as deprecated.

I think we maybe do want to do that at the same time, would be easier to communicate and organize docs? And we can still soft deprecate the `setup` method in the code so it's still run.

packages/starlight/__tests__/plugins/vitest.config.ts
```diff
@@ -63,6 +59,34 @@ export default defineVitestConfig({
 						},
 					});
 				},
+				async setup({ config, updateConfig, useTranslations, pathToLang }) {
+					await Promise.resolve();
```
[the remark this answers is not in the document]
HiDeoo: I'll add a comment, this is mostly to ensure type-wise it supports async `setup()` while avoiding to add a type test for it as type tests are still a bit broken within a Vitest workspace setup like we have.

packages/starlight/integrations/expressive-code/index.ts
```diff
@@ -5,9 +5,10 @@ import {
 } from 'astro-expressive-code';
 import { addClassName } from 'astro-expressive-code/hast';
 import type { AstroIntegration } from 'astro';
-import type { StarlightConfig } from '../../types';
-import type { createTranslationSystemFromFs } from '../../utils/translations-fs';
-import { pathToLocale } from '../shared/pathToLocale';
+import type { StarlightConfig, StarlightPlugin } from '../../types';
+import { pathToLang } from '../shared/pathToLang';
```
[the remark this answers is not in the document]
HiDeoo: Maybe this deserves a comment somewhere, but it's for the `<Code>` component basically which runs in a different context.

HiDeoo:

Holding off on the suggested changes so I can tackle them all at once when we have a clearer idea of what we want to do regarding the name.

docs/src/content/docs/reference/plugins.md
HiDeoo: I can add a few test cases for these, but they all fallback to the default locale.

.changeset/large-balloons-compete.md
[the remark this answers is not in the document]
delucis: OK, that sounds good! So if you think `i18n` is not problematic, I think it’s nice to have that concise hook name. And +1 to do the backwards compatible deprecation of `setup` in favour of `config:setup`.

Regarding what you might do with a `config:done`: most often I see the `astro:config:done` hook to share the final config with some other code via the integration closure, e.g.

```js
function integration() {
  let lazy = {};
  return {
    name: 'example',
    hooks: {
      'astro:config:setup'() {
        // add something referencing `lazy` e.g. a Vite plugin
        // that will run _after_ the done hook but wants the
        // final config shape
      },
      'astro:config:done'({config}) {
        lazy.config = config;
      },
    },
  };
}
```

But not suggesting we add that now. No point until some more clear needs emerge.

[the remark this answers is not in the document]
HiDeoo: Perfect, I'll get started asap on the new changes, thanks again.

HiDeoo:
Updated the branch with the changes discussed earlier, hope I did not forget anything:

- Adds the `config:setup` hook and deprecate the `setup` one
  - I added a few tests to ensure we make sure to totally remove `setup` before v1
- Renames the new hook to `i18n:setup`
- Cleans up and renames the new path utility to `absolutePathToLang`
  - It's fairly easy to add some checks to ensure the paths are the ones we expect for a document in a content collection altho, this would fail as soon as something like `import { Content } from '../../something.md'` is used (this would even fail with our `EmptyMarkdown` file).
- Adds a test to ensure we migrate `@astrojs/starlight-docsearch` to the `config:setup` hook before v1
- Adds a new `HookParameters` utility type and its documentation
  - A lot of plugin are already doing something like `Parameters<StarlightPlugin['hooks']['setup']>['0']['something']` but now this would be even longer with the addition of a nullable so the new type should help with that like in Astro.
- Updates `starlight-links-validator` to a new version where I refactored some very old legacy types to avoid type errors
  - The link validation is currently failing, I've opened #2598 to fix this
- Updates the `useTranslations()` documentation & example to mention it's a BCP-47 tag and to use a region subtag in the example

I have a question: this PR adds multiple changesets but I think `.changeset/polite-fishes-remain.md` should come first. Altho, I don't know if there is a way to enforce that from this PR, right?

docs/src/content/docs/reference/plugins.md
```diff
@@ -41,12 +47,100 @@ A plugin must provide a unique name that describes it. The name is used when [lo
 
 ## `hooks`
 
-Hooks are functions which Starlight calls to run plugin code at specific times. Currently, Starlight supports a single `setup` hook.
+Hooks are functions which Starlight calls to run plugin code at specific times.
+
+To get the type of a hook's arguments, use the `HookParameters` utility type and pass in the hook name.
+In the following example, the `configSetup` function `options` parameter is typed to match the parameters of the `config:setup` hook:
```
delucis, line 53: Suggesting a small simplification here to help make the sentence a little bit clearer:

```suggestion
In the following example, the `options` parameter is typed to match the arguments passed to the `config:setup` hook:
```

docs/src/content/docs/reference/plugins.md
```diff
@@ -41,12 +47,100 @@ A plugin must provide a unique name that describes it. The name is used when [lo
...
-Plugin setup function called when Starlight is initialized (during the [`astro:config:setup`](https://docs.astro.build/en/reference/integrations-reference/#astroconfigsetup) integration hook).
-The `setup` hook can be used to update the Starlight configuration or add Astro integrations.
+function configSetup(options: HookParameters['config:setup']) {
+  options.useTranslations('en');
+}
+```
+
+### `i18n:setup`
+
+Plugin internationalization setup function called when Starlight is initialized (during the [`astro:config:setup`](https://docs.astro.build/en/reference/integrations-reference/#astroconfigsetup) integration hook).
```
delucis, line 65: What do you think about omitting the Astro hook detail? Part of me feels like it might be superfluous information that distracts from the focus of this section, but I also get that it could be useful information.

```suggestion
Plugin internationalization setup function called when Starlight is initialized.
```

docs/src/content/docs/reference/plugins.md
```diff
@@ -41,12 +47,100 @@ A plugin must provide a unique name that describes it. The name is used when [lo
...
+The `i18n:setup` hook can be used to inject translations strings for the plugin in various locales.
```
delucis, line 66: Small tweak suggestion:

```suggestion
The `i18n:setup` hook can be used to inject translation strings so a plugin can support different locales.
```

docs/src/content/docs/reference/plugins.md
```diff
@@ -41,12 +47,100 @@ A plugin must provide a unique name that describes it. The name is used when [lo
...
+These translations will be available in the [`config:setup`](#configsetup) hook and plugin UI.
```
delucis, line 67: Suggesting a small tweak here. “plugin UI” sounds a bit like some kind of UI for managing a plugin somehow to me, so wanted to try to emphasise “components” to help with understanding that this is in your website’s UI and how to access them.

```suggestion
The `i18n:setup` hook can be used to inject translation strings so a plugin can support different locales.
These translations will be available via [`useTranslations()`](#usetranslations) in the `config:setup` hook and in UI components via [`Astro.locals.t()`](/guides/i18n/#using-ui-translations).
```

docs/src/content/docs/reference/plugins.md
```diff
@@ -41,12 +47,100 @@ A plugin must provide a unique name that describes it. The name is used when [lo
...
+
+This hook is called with the following options:
```
delucis, line 69: I think it might be worth repeating the hook name here rather than just referring to it with “this“:

```suggestion
The `i18n:setup` hook is called with the following options:
```

packages/starlight/__tests__/plugins/config.test.ts
```diff
@@ -120,6 +122,80 @@ describe('validation', () => {
...
+			const docSearchPlugin = starlightDocSearch({
+				appId: 'YOUR_APP_ID',
+				apiKey: 'YOUR_SEARCH_API_KEY',
+				indexName: 'YOUR_INDEX_NAME',
+			});
+			expect(docSearchPlugin.hooks.setup).not.toBeDefined();
+			expect(docSearchPlugin.hooks['config:setup']).toBeDefined();
+		}
+	);
+});
```
delucis, line 197: FWIW, I would be happy to just change the DocSearch plugin right away, bumping the minimum supported Starlight version, and release a new minor. Can also happen in a follow-up PR though as we can release a new plugin version after this Starlight minor.

packages/starlight/index.ts
```diff
@@ -66,15 +65,11 @@ export default function StarlightIntegration(
 				);
 
 				const integrations = pluginResult.integrations;
+				const useTranslations = pluginResult.useTranslations;
+				const absolutePathToLang = pluginResult.absolutePathToLang;
```
delucis, line 69: Should we refactor this to this?
```suggestion
				const { integrations, useTranslations, absolutePathToLang } = pluginResult;
```

delucis (commented):
I think this looks good! Left a few small notes, but nothing significant jumped out to me — just cleaning up the demo plugin and removing any other TODOs.

> I have a question: this PR adds multiple changesets but I think `.changeset/polite-fishes-remain.md` should come first. Altho, I don't know if there is a way to enforce that from this PR, right?

Good question! I’m not 100% sure how Changesets orders them beyond the major/minor/patch subheadings. Not sure if it’s chronological by commit time or something else. I’d guess we should just tweak the CHANGELOG in the release PR when the time comes.

packages/starlight/__tests__/plugins/config.test.ts
HiDeoo: Done, one more changeset it is 😅

HiDeoo:
> I’d guess we should just tweak the CHANGELOG in the release PR when the time comes.

Makes sense, that was also my guess, thanks. I added a new todo to the PR body to remember about that.

packages/docsearch/index.ts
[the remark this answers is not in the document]
HiDeoo: Ah, excellent catch, totally missed that, will do that now.

delucis (approved):
Perfect! OK, should be ready to go I think! I’m thinking of trying to get `Astro.locals` ready too, so that these two changes breaking plugins can go out together. I think we’re almost there.

HiDeoo:
> I’m thinking of trying to get `Astro.locals` ready too, so that these two changes breaking plugins can go out together. I think we’re almost there.

Makes sense 👍 I already gathered some feedback/concerns regarding the route data PR to publish when it's ready for review, really excited about this one 🤩

trueberryless:
Wait, this PR also got merged today?!? Holy moly, how incredible is that! 💝

PR #2580, opened 2024-11-06
Feat: Add links to "Open a PR" on resources pages (`plugins.mdx` & `community-content.mdx`)

trueberryless:
#### Description

- add 2 new sections to the [CONTRUBUTING.md](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md) file:
  - [Plugins and Integrations](https://github.com/trueberryless/starlight/blob/feat-contribution-enhancement/CONTRIBUTING.md#plugins-and-integrations)
  - [Community Content](https://github.com/trueberryless/starlight/blob/feat-contribution-enhancement/CONTRIBUTING.md#community-content)
    - [Video Content](https://github.com/trueberryless/starlight/blob/feat-contribution-enhancement/CONTRIBUTING.md#video-content) (new example MUST be reviewed...)
- link to these sections from the Starlight website
  - `plugins.mdx` "Open a PR" links to [Plugins and Integrations](https://github.com/trueberryless/starlight/blob/feat-contribution-enhancement/CONTRIBUTING.md#plugins-and-integrations) (to original repo and main branch logically...)
  - `community-content.mdx` "Open a PR" links to [Community Content](https://github.com/trueberryless/starlight/blob/feat-contribution-enhancement/CONTRIBUTING.md#community-content) (to original repo and main branch logically...)

I'm not sure if the maintainers want these changes to exist, because it kinda feels intended that the "Call to action" sections have no links on both sites but on the `showcase.mdx` site it does.
Probably the reason is, that we want people to add their Showcases on their own, but Plugins and other Content is maybe to complex for the people, so they should just provide a link and we change the files on our own. But I don't know and therefore I through this PR in here...


trueberryless:
Also I'll move the misplaced import statement to another PR cause it doesn't have to do anything with this one...

CONTRIBUTING.md
```diff
@@ -288,6 +288,108 @@ To add a language, you will need its BCP-47 tag and a label. See [“Adding a ne
...
+
+It's great to see plugins and other integrations enhance or change features of Starlight.
+We have a collection of plugins and tools on our [plugins](https://starlight.astro.build/resources/plugins/) page.
+If you’ve built a Starlight plugin, theme or some other tool, adding it to the plugin page is just a pull request away!
+
+1. Set up a development environment by following the [“Setting up a development environment”](#setting-up-a-development-environment) instructions.
+2. Add a new entry for your plugin or tool in the right section in `docs/src/content/docs/resources/plugins.mdx`.
+
+   - The new entry must be appended at the end of an existing list.
+   - The `href` attribute must be the URL of your Starlight site. If your documentation is hosted on a subdomain or subdirectory, include that in the URL.
```
delucis, line 301: This line might need tweaking a bit compared to the showcase version. Although now I’m thinking about it, I also wonder if we need to repeat these sections entirely for each content type given how similar they all are 🤔

Maybe there’s a nice way to have generic instructions for all of these?

trueberryless: Good point. Now that I looked through the existing plugins I found that there is no consistency whether to link to the GitHub repo or the website. I think I'll change the text to Github cause 88% (22/25) currently link to GitHub. Maybe we change the other 3 URLs to their GitHub (I checked all 3, they are public available) so it is consistent...

trueberryless: Making it generic?

I'm tending toward "no" because it feels nice to have dedicated section for each of them because the user can look at them without any confusion / scrolling around much. However, the general idea is not bad, if we can find a fitting **heading** for all 3 sections, I'll change it (because just calling the entire group "Resources" feels wrong...

HiDeoo: > Although now I’m thinking about it, I also wonder if we need to repeat these sections entirely for each content type given how similar they all are 🤔

I'm also personally leaning towards the idea of having a single section:

- These pages don't usually get a lot of friction when content is added to them compared for example to the showcase page which requires a screenshot at a specific size, etc.
- The existing content is usually enough to get a good idea of how it's structured and what's needed to add more content (e.g. not like the showcase page where you need to know the exact size of the screenshot (which is not mentioned in the page itself) or the location of the files as the path is abstracted away).

I think a single section mentioning which files to edit depending on the content type you want to add and a sentence mentioning to either append a `<LinkCard>` or to an array might be enough and maybe more digestible for the reader. Only my opinion though, definitely open to other ideas!

trueberryless: I'm gonna be honest: One section doesn't taste as good as 3, but that's just me...
Let's see what you guys think! I'm excited... 🎉

trueberryless: > @trueberryless: Maybe we change the other 3 URLs to their GitHub (I checked all 3, they are public available) so it is consistent...

Actually I looked closer into all 3 plugins (starlight-ghostcms, star-wrap, FeelBack) and realised that each of them has a reason not to link to GitHub: the Starlight Plugin is a kinda "subset" of what the product offers, so we'll definitely not change their links to their GitHubs... My bad...

trueberryless:
Looking back at what I did here two years ago, I think such a detailed documentation of all resources pages is not actually that necessary. The `Sites` and `Themes` section already cover the most difficult additions I would say, and I guess we noticed that not many people encounter problems with adding new plugins or other community content.

Besides: Why did I include a Rick Roll as an example? Seems like a had a funny day back then 😅 

I'm gonna close this for now, but keep the branch, if we ever decide that we need more detailed documentation for these sections, we can take this as a reference. 👍

PR #2581, opened 2024-11-06
Move import statement in `resources/plugins.mdx` after Aside (because it's already used before it's imported...)

trueberryless:
#### Description

- I don't know if this is a bug, but I'm not sure why it actually works, but I think that:
- import statement is misplaced and therefore moved to the top of the page (after the Aside)

Reason: The components imported are already used before being imported, or am I blind?

(already mentioned shortly in #2580)

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -10,6 +10,8 @@ Have you built a plugin or a tool for Starlight?
 Open a PR adding a link to this page!
 :::
 
+import { CardGrid, LinkCard } from '@astrojs/starlight/components';
+
 ## Plugins
 
 [Plugins](/reference/plugins/) can customize Starlight configuration, UI, and behavior, while also being easy to share and reuse.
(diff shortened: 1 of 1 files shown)


delucis:
Ah, the position of import statements is actually unimportant in MDX (and even in regular JS as it happens), which admittedly might seem a bit surprising. The way that JS module loading works is that first all `import` statements are collected from the file and evaluated and then other code is run. For example, this is totally valid:

```js
const result = callFunction('x');

import { callFunction } from './somewhere';
```

For legibility reasons you wouldn’t usually do that of course, but it’s not wrong as such.

trueberryless:
> For example, this is totally valid:
> 
> ```js
> const result = callFunction('x');
> 
> import { callFunction } from './somewhere';
> ```

LOL, that reminds me of some weirdness I heard about `var`. I don't know if it still has this "functionality" but you could use a variable before declaring it. Comming from `C#` this just seems like a horror movie 😂

delucis:
Yeah `var` also gets hoisted IIRC (at least the name declaration does, not the value). Although weird behaviour of `var` is also why everyone prefers `let` and `const` these days.

Anyway, given this I think we can close this PR. Thanks again! 🙌

PR #2551, opened 2024-11-01
Update `astro-expressive-code` to v0.38

hippotastic:
#### Description

Updates the `astro-expressive-code` dependency to the latest version (0.38).

The new version allows using `ec.config.mjs` to selectively override individual Expressive Code styles and settings provided by Starlight themes and plugins, speeds up Shiki language loading, and adds the config option `expressiveCode.shiki.injectLangsIntoNestedCodeBlocks`. See the [Expressive Code release notes](https://expressive-code.com/releases/#0380) for full details.

PR #2582, opened 2024-11-06
Update dev dependencies

delucis:
#### Description

- Updates development versions of Astro and other dependencies
- Should finally fix Astro’s Ecosystem CI which has been failing for a while now
- Two non-dependency changes:
  - Due to a type (non)issue in the docs Astro config, I cast the type of the Markdoc grammar to any for now. We can solve upstream in Expressive Code later.
  - Looks like the upgraded Tailwind version changed order of some of its generated code, but looks like a correct bug fix to me of ensuring user CSS is generated after Tailwind boilerplate instead of before.


docs/astro.config.mjs
```diff
@@ -179,7 +179,12 @@ export default defineConfig({
 					autogenerate: { directory: 'resources' },
 				},
 			],
-			expressiveCode: { shiki: { langs: [markdocGrammar] } },
+			expressiveCode: {
+				shiki: {
+					// TODO: Remove `any` once Expressive Code supports grammars with comments.
```
HiDeoo, line 184: For reference, should be fixed with [`f594927` (#2551)](https://github.com/withastro/starlight/pull/2551/commits/f594927614879a3e290c48a197e5aed7dabb3cd0)

HiDeoo (approved):
Everything looks perfect to me, even locally :thumbsup:

PR #2579, opened 2024-11-06
Added social icon for Nostr

RafidMuhymin:
#### Description

- Added social icon for Nostr.

**Icon Reference**: https://github.com/mbarulli/nostr-logo/blob/main/SVG/nostr-icon-purple-on-white.svg. I optimized the icon and made changes to align with the other social icons.

PR #2584, opened 2024-11-07
add flickr to social icons

sarah11918:
#### Description

This PR adds a new social icon option `flickr` and a corresponding svg path for the Flickr logo.

I don't know how to test this, or if I've missed anything, but this maybe comes close? :smile:


delucis (commented):
A couple of small details missed, but this a great start! Here’s another recent social icon PR which is a good reference: https://github.com/withastro/starlight/pull/2579

We’ll also need to:
- Update the error message expected in `packages/starlight/__tests__/basics/config-errors.test.ts`
- Add a changeset

Ah and we’ll need to double check the icon itself is working. Right now looks like this for me:

missing flickr icon

Looks like the path you added is much larger than the 32x32 size we expect.

---

Also taking a look at these icons and wondering if they’ll really be understood as indicating Flickr?

a grid of different flickr icons most of which are just two dots of the same grey side by side

We haven’t done this often but for the `pkl` icon we use `opacity` to create a lower contrast on the central part of the icon and maybe should do the same here for one of the circles 🤔

sarah11918:
Yeah, admittedly it's not an ideal logo they've chosen. In other situations, I've used a generic camera icon personally. And, if we shouldn't use monochrome anyway, then I'm happy to just close this, as the coloured versions wouldn't make sense. It's not my fault they've chosen poorly with their branding 😅

sarah11918:
Alright, happy to close the two dots branding and shame Flickr about their branding! Just thought it was worth a try!

PR #2589, opened 2024-11-08
Support passing more options to the DocSearch component

delucis:
#### Description

- This PR exposes a few more of the DocSearch component’s options in the Starlight DocSearch Plugin’s config.
- We’re still limited to serializable options here, so it’s a subset of DocSearch’s full set of options (which includes stuff like callback methods for transforming results and providing custom UI components).
- This PR also adds a README to the package which was missing and updates our GitHub labeler action config to apply a `🌟 docsearch` label to PRs like this one.
- I also omitted the `initialQuery` option for now because it feels like that would perhaps be a mistake to control from a global location like the plugin config. If anything, perhaps it should go through i18n APIs or something else. Seemed safest to omit for now and we can revisit if someone has a specific use case and requests it.
- We don’t have tests set up for the DocSearch plugin given it’s pretty simple. I tested this by adding the plugin to a local example and testing out the different config options, which worked successfully.

### DIFF
--- a/packages/docsearch/index.ts
+++ b/packages/docsearch/index.ts
@@ -1,17 +1,37 @@
 import type { StarlightPlugin } from '@astrojs/starlight/types';
+import type docsearch from '@docsearch/js';
 import type { AstroUserConfig, ViteUserConfig } from 'astro';
 import { z } from 'astro/zod';
 
-/** Config options users must provide for DocSearch to work. */
-const DocSearchConfigSchema = z.object({
-	appId: z.string(),
-	apiKey: z.string(),
-	indexName: z.string(),
-});
-export type DocSearchConfig = z.input<typeof DocSearchConfigSchema>;
+type SearchOptions = Parameters<typeof docsearch>[0]['searchParameters'];
+
+/** DocSearch configuration options. */
+const DocSearchConfigSchema = z
+	.object({
+		// Required config without which DocSearch won’t work.
+		/** Your Algolia application ID. */
+		appId: z.string(),
+		/** Your Algolia Search API key. */
+		apiKey: z.string(),
+		/** Your Algolia index name. */
+		indexName: z.string(),
+		// Optional DocSearch component config (only the serializable properties can be included here)
+		/** The maximum number of results to display per search group. Default is `5`. */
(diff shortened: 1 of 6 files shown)


packages/docsearch/index.ts
```diff
@@ -1,17 +1,37 @@
...
+		/** Your Algolia index name. */
+		indexName: z.string(),
+		// Optional DocSearch component config (only the serializable properties can be included here)
+		/** The maximum number of results to display per search group. Default is `5`. */
+		maxResultsPerGroup: z.number().optional(),
+		/** Disable saving recent searches and favorites to the local storage. Default is `false`. */
+		disableUserPersonalization: z.boolean().optional(),
+		/** Whether to enable the Algolia Insights plugin and send search events to your DocSearch index. Default is `false`. */
+		insights: z.boolean().optional(),
+		/** The Algolia Search Parameters. See https://www.algolia.com/doc/api-reference/search-api-parameters/ */
```
HiDeoo, line 25:
```suggestion
		/**
		 * The maximum number of results to display per search group.
		 * @default 5
		 */
		maxResultsPerGroup: z.number().optional(),
		/**
		 * Disable saving recent searches and favorites to the local storage.
		 * @default false
		 */
		disableUserPersonalization: z.boolean().optional(),
		/**
		 * Whether to enable the Algolia Insights plugin and send search events to your DocSearch index.
		 * @default false
		 */
		insights: z.boolean().optional(),
		/**
		 * The Algolia Search Parameters.
		 * @see https://www.algolia.com/doc/api-reference/search-api-parameters/
		 */
```

Small nit to use TSDoc tags, I personally think the highlighting of the types is more obvious as it matches the one from your code theme for that type:

| Before | After |
| --- | --- |
| SCR-20241109-brkh | SCR-20241109-brup |

HiDeoo (approved):
Left a small nit related suggestion but other than that, looks good to me 🌟

packages/docsearch/index.ts
delucis: Thanks! Yeah, I’ll admit I just copy-pasted from the Algolia docs, but agree it makes sense to use the proper tags.

PR #2595, opened 2024-11-09
Add my two new plugins to Starlight `resources/plugins.mdx` page

trueberryless:
#### Description

- Add these two plugins:
  - [Starlight Cooler Credit](https://starlight-cooler-credit.trueberryless.org/)
  - [Starlight Contributor List](https://starlight-contributor-list.trueberryless.org/)

I'm NOT adding Cooler Credit at the end of the list for the following reason: 
I think it looks really nice if [Starlight Sidebar Topics](https://starlight-sidebar-topics.netlify.app/) and [Starlight Sidebar Topics Dropdown](https://starlight-sidebar-topics-dropdown.trueberryless.org/) are next to each other (side-by-side) because the Dropdown version is actually a copy of the original...

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -94,6 +94,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		description="Add an interactive site graph inside your page's sidebar."
 	/>
 	<LinkCard
+		href="https://github.com/trueberryless-org/starlight-cooler-credit"
+		title="starlight-cooler-credit"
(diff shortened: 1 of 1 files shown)


HiDeoo (changes requested):
Thanks for the new additions :star2:

Let's actually move `starlight-cooler-credit` to the bottom of the list with the other new one and keep a strict append-only approach. Allowing user-defined order once can potentially set a precedent for future requests where a user could just ask "I'd like my plugin to be at the top of the list because its name starts with an 'A'" or any other reason, valid or not. Definitely easier to just keep it simple and potentially avoid any unwanted discussions.

trueberryless:
Okay, should be good to go now!

PR #2601, opened 2024-11-11
Bump Node versions to 20 in workflows

trueberryless:
#### Description

- bump all node version to 20 in workflow files
- I don't know if it's intended that some are still at v18, so I just created this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  NODE_VERSION: 18
+  NODE_VERSION: 20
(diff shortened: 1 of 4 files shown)


HiDeoo:
Thanks for the pull request :raised_hands:

Starlight follows Astro [prerequisites](https://docs.astro.build/en/install-and-setup/#prerequisites) and regarding Node.js, `v18.17.1` or `v20.3.0` or higher (`v19` is not supported) is required. When it comes to testing (`ci.yml`), we definitely want to keep testing against the minimum supported version, which is `v18.17.1`.

We could technically update the Node.js versions in the other workflows but I personally think that having them all to the same version is a great reminder to update them all at once when Astro no longer supports Node.js 18.

PR #2606, opened 2024-11-14
Add `grid-template-columns` to`<CardGrid>`

delucis:
#### Description

- This PR adds a base `grid-template-columns` rule to the `<CardGrid>` component to apply on smaller screens as one `100%` wide column.

- This helps avoid overflow when there is complex content inside the grid children. We already [apply this rule as an override in Astro Docs](https://github.com/withastro/docs/blob/a9065e9f5bf616abd8b769b5250ed2535f8762d7/src/styles/custom.css#L3-L11) to avoid overflow, but I only just remembered to upstream this.

- **Why can this be an issue?** Without an explicit column width, the grid uses an implicit width based on context. For example, without this rule, cards on the Astro Docs homepage would overflow due to the code block in the “Start a new project” card:

  narrow viewport screenshot of Astro Docs without this change. A card featuring a code example overflows the viewport

- Adding an explicit column width avoids that and ensures children are constrained as expected:

  A similar screenshot of Astro Docs but now the cards all fit inside the viewport instead of overflowing

### DIFF
--- a/.changeset/few-dogs-grab.md
+++ b/.changeset/few-dogs-grab.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Makes `<CardGrid>` more resilient to complex child content on smaller viewports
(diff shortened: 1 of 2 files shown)


HiDeoo (approved):
Looks like a great improvement to upstream :star2: :rocket:

PR #2611, opened 2024-11-18
Fix i18n type issue with multiple data content collections

HiDeoo:
#### Description

This PR fixes an issue initially spotted in https://github.com/withastro/docs/pull/9240.

To reproduce the issue:

1. Create a new data content collection by creating an `docs/src/content/nav/en.ts` file with the following content:

   ```ts
   export default {}
   ```
1. Run `pnpm dev` to generate types for the new data content collection.

At this point, all `Astro.locals.t()` calls would indicate a type error and you would get no autocomplete suggestions for it.

On top of extracting the `CollectionEntry<'i18n'>` type outside of the `UserI18nSchema` conditional, I also had to cherry pick a fix I made in #2578 as #2380 made the `i18nSchemaOutput` fallback type too loose.

### DIFF
--- a/.changeset/blue-walls-kiss.md
+++ b/.changeset/blue-walls-kiss.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a UI string type issue in project with multiple data content collections.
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thank you TypeScript for being the gift that keeps giving and thank you @HiDeoo for your persistence in the face of adversity! 😅

PR #2612, opened 2024-11-18
Astro 5 + collection loaders + legacy collections support

HiDeoo:
#### Description

This PR is a draft adding Astro 5 support to Starlight, adding collection loaders for our content collections and support for legacy collections.

The first goal was to get everything working with Astro 5. There is room for improvements and potential refactoring, but the goal was to get everything in place first to get a better overview of what needs to be done and later iterate on it.

~~The MDX dependency case is temporarily workaround by making it a peer dependency of Starlight and installing it manually everywhere it's needed until we have a better solution (in discussion).~~

Note that the collection locations are not yet user-configurable.

#### Remaining tasks

- [x] Ensure all `// TODO(HiDeoo)` comments are addressed and removed
- [x] Figure out if legacy collections support should be documented more (it's only a sentence + link in the manual setup page now)
- [x] Peer dependency bump for `@astrojs/starlight` in some packages (see changesets for the one that needs an update)
- [x] Add changesets
- [ ] Merge with Lunaria directives to ignore the changes to all `code.mdx` files
	- [ ] `@lunaria-ignore:src/content/docs/**/code.mdx;src/content/docs/code.mdx;`


docs/src/content/docs/components/code.mdx
```diff
@@ -85,14 +85,14 @@ You can then pass this imported string to the `<Code>` component to include it o
 # src/content/docs/example.mdx
 
 import { Code } from '@astrojs/starlight/components';
-import importedCode from '/src/env.d.ts?raw';
+import importedCode from '/tsconfig.json?raw';
```
HiDeoo, line 88: With the [changes to the TypeScript configuration](https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/#changed-typescript-configuration), I removed all `env.d.ts` files so this example is no longer working (and relevant?). I switched to showing instead the default (and new) `tsconfig.json` file for Astro projects (we have a verbatim copy in the basics example).

patches/starlight-links-validator@0.13.2.patch
HiDeoo, line 1: I made a small patch to fix a type issue with the link validator that I'll fix in the package once Astro 5 is released.

packages/starlight/vitest.config.ts
```diff
@@ -18,12 +18,14 @@ export default defineConfig({
...
+				// Since Vitest 2.1.2, coverage is collected for `*.astro` files.
+				'**/*.astro',
 			],
 			thresholds: {
-				lines: 88,
+				lines: 87,
 				functions: 90,
 				branches: 90,
-				statements: 88,
+				statements: 87,
```
HiDeoo, line 28: Since [Vitest 2.0.0](https://github.com/vitest-dev/vitest/releases/tag/v2.0.0), empty lines are now ignored in coverage by default which can cause significant differences in coverage results. I reduced the thresholds only to adjust for this change.

packages/starlight/vitest.config.ts
```diff
@@ -18,12 +18,14 @@ export default defineConfig({
 				'props.ts',
 				// Main integration entrypoint — don’t think we’re able to test this directly currently.
 				'index.ts',
+				// Since Vitest 2.1.2, coverage is collected for `*.astro` files.
+				'**/*.astro',
```
HiDeoo, line 22: Since [Vitest 2.1.2](https://github.com/vitest-dev/vitest/releases/tag/v2.1.2), `*.astro` files are now included in the coverage report by default.

packages/starlight/vitest.workspace.ts
```diff
@@ -1 +1 @@
-export default ['__tests__/*'];
+export default ['__tests__/*/vitest.config.ts'];
```
HiDeoo, line 1: Since [Vitest 2.1.0](https://github.com/vitest-dev/vitest/releases/tag/v2.1.0), `*` matches anything, folders like before but also files which ends up being treated as a Vitest config so we explicitly matches against our explicit config files to ignore files at the root of the `__tests__/` folder.

HiDeoo:
The E2E tests are currently failing due to what seems to be an Astro issue related to the new `content.config.ts` file. I shared a small repro to Matt with my initial understanding of the issue.

I'll update the PR once I get confirmation and the bug is fixed.

HiDeoo:
Updated the branch to use Astro beta 12 which includes Vite 6 stable and the latest Vitest version adding support for Vite 6.

A few test assertions were updated to be explicitly awaited as Vitest auto-await them at the moment, which is a behavior that will be removed in Vitest 3 and starting with Vitest [`2.1.5`](https://github.com/vitest-dev/vitest/releases/tag/v2.1.5), a warning is shown when this happens.

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -33,7 +33,7 @@ The page description is used for page metadata and will be picked up by search e
 
 **type**: `string`
 
-Override the slug of the page. See [“Defining custom slugs”](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs) in the Astro docs for more details.
+Override the slug of the page. See [“Defining custom IDs”](https://docs.astro.build/en/guides/content-collections/#defining-custom-ids) in the Astro docs for more details.
```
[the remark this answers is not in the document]
HiDeoo: PR merged and the title did not change.

packages/markdoc/package.json
```diff
@@ -17,12 +17,12 @@
     "./components": "./components.ts"
   },
   "devDependencies": {
-    "@astrojs/markdoc": "^0.11.5",
+    "@astrojs/markdoc": "^0.12.1",
     "@astrojs/starlight": "workspace:*",
-    "vitest": "^1.6.0"
+    "vitest": "2.1.6"
```
HiDeoo, line 22: Decided to pin Vitest version to [`2.1.6`](https://github.com/vitest-dev/vitest/releases/tag/v2.1.6) which introduced support for Vite 6. It was reverted in [`2.1.7`](https://github.com/vitest-dev/vitest/releases/tag/v2.1.7) and I was hoping this would be a quick enough fix but it looks like it may take a bit longer. As we're not affected by the issue (mostly `resolve.conditions` related), and considering support was added and removed in patch versions, I think it's better to pin to that version for now. I'll keep an eye on it, update and unpin when it's resolved.

HiDeoo:
Not upgrading yet to [Astro `v5.0.3`](https://github.com/withastro/astro/releases/tag/astro%405.0.3) yet as it looks like this version may introduce a regression making some of our E2E tests fail.

Opened <https://github.com/withastro/astro/issues/12652> to track the issue and will upgrade once we have a confirmation and a fix.

.changeset/polite-snails-sip.md
```diff
@@ -0,0 +1,13 @@
+---
+'@astrojs/starlight-docsearch': minor
+'@astrojs/starlight-tailwind': minor
```
delucis, line 3: I messed up at one point, so we’re accidentally already in major version territory for the Tailwind package 😄 (it’s currently at v2.0.3)

```suggestion
'@astrojs/starlight-tailwind': major
```

delucis (commented):
Didn’t get much into the main code yet — will tackle that this afternoon — but left a first couple of small notes/questions from the first 73 files I looked at 😁

Thanks again for the great work on this @HiDeoo. Shame the flu stopped me being able to review it sooner!

.changeset/polite-snails-sip.md
HiDeoo: Good catch, I totally missed that :thumbsup:

docs/src/content/docs/de/components/code.mdx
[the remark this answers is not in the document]
HiDeoo: Yes, it's tracked in the last todo of the PR body:

> Merge with Lunaria directives to ignore the changes to all code.mdx files (if we decide to keep the changes)

packages/starlight/loader.ts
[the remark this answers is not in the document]
HiDeoo: I hesitated between both as we exposes our schemas in `@astrojs/starlight/schema` so decided to keep our current approach but happy to change it :thumbsup:

docs/src/content/docs/reference/overrides.md
```diff
@@ -73,7 +73,7 @@ The slug for this page generated from the content filename.
 
 **Type:** `string`
 
-The unique ID for this page based on the content filename.
+The slug for this page or the unique ID for this page based on the content filename if using the [`legacy.collections`](https://docs.astro.build/en/reference/legacy-flags/#collections) flag.
```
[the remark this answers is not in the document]
HiDeoo: It's available in `Astro.props.entry` in this case.

package.json
```diff
@@ -50,6 +50,9 @@
         "playwright",
         "search-insights"
       ]
+    },
+    "patchedDependencies": {
+      "starlight-links-validator@0.13.2": "patches/starlight-links-validator@0.13.2.patch"
```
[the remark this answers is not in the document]
HiDeoo: Exactly, I'll make a PR to remove the patch once a new version of the links validator is released.

packages/starlight/loader.ts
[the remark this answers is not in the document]
delucis: Yeah, I suspected that was the reason. I don’t _think_ I made that naming decision particularly thoughtfully 😅 I’m guessing that was where `docsSchema` came from and then later I added `i18nSchema` without thinking that `schema` now included schemas plural.

For loaders I can imagine it growing even more if we ever offer loaders tailored to specific situations.

docs/src/content/docs/reference/overrides.md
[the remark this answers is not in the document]
delucis: Ah right. I guess it’s also not really documented in the Astro docs (except in the [docs for loader authors](https://docs.astro.build/en/reference/content-loader-reference/#filepath)), so we may need to think about whether we need to document it at all (bit tricky given it’s loader-dependent).

Bigger question: do we still need `slug` as well as `id` now they are the same? Or do we deprecate `slug` (and remove it from docs)?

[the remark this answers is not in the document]
HiDeoo: They're the same if not using `legacy.collections` so I don't think we can deprecate it now.

packages/starlight/loader.ts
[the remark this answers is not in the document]
HiDeoo: Just updated the branch with this refactor.

docs/src/content/docs/reference/overrides.md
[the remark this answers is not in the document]
delucis: I guess it depends how we think about `legacy.collections` support. Presumably that option will go away with Astro v6. My guess would be that Starlight v1 happens before v6, so I’d argue we should probably drop `legacy.collections` support for v1 at the latest. In that scenario, deprecating it now so it still works for legacy collection users but is not recommended might still make sense to me.

packages/starlight/utils/navigation.ts
```diff
@@ -228,6 +228,9 @@ function treeify(routes: Route[], baseDir: string): Dir {
 		// Build the tree
 		.forEach((doc) => {
 			const parts = getBreadcrumbs(doc.id, baseDir);
+			// If the breadcrumb is empty, the document is at the root level so we can fallback to the
+			// document ID.
+			if (parts.length === 0) parts.push(doc.id);
```
delucis, line 233: Just to make sure I understand: is this because previously `index.md` could be an ID and this would no longer generate any breadcrumbs because it has a slug of `"/"` or `""` or something? I’d like to understand why this handling is different/if it should be part of `getBreadcrumbs()`.

packages/starlight/utils/route-data.ts
```diff
@@ -85,7 +83,7 @@ function getLastUpdated({ entry }: PageProps): Date | undefined {
 		try {
 			return frontmatterLastUpdated instanceof Date
 				? frontmatterLastUpdated
-				: getNewestCommitDate(entry.id);
+				: getNewestCommitDate(entry.filePath);
```
delucis, line 86: I guess for now we’re still only supporting on-disk `src/content/` documents, so no need to handle entries which don’t include `filePath`? (i.e. remote content loaders)

packages/starlight/utils/route-data.ts
```diff
@@ -105,10 +103,8 @@ function getEditUrl({ entry, id, isFallback }: PageProps): URL | undefined {
 		// If a URL was provided in frontmatter, use that.
 		url = editUrl;
 	} else if (config.editLink.baseUrl) {
-		const srcPath = project.srcDir.replace(project.root, '');
-		const filePath = isFallback ? localizedId(id, config.defaultLocale.locale) : id;
 		// If a base URL was added in Starlight config, synthesize the edit URL from it.
-		url = ensureTrailingSlash(config.editLink.baseUrl) + srcPath + 'content/docs/' + filePath;
+		url = ensureTrailingSlash(config.editLink.baseUrl) + entry.filePath;
```
delucis, line 107: What happens here with legacy collections? Is `filePath` still available?

**Edit:** Ah, I see now in `normalizeCollectionEntry()` you ensure it is present for those entries too 👍

packages/starlight/utils/git.ts
```diff
@@ -105,7 +105,9 @@ export function getAllNewestCommitDate(directory: string): [string, number][] {
 
 	return Array.from(latestDates.entries()).map(([file, date]) => {
 		const fileFullPath = resolve(repoRoot, file);
-		const fileInDirectory = relative(directory, fileFullPath);
+		let fileInDirectory = relative(rootPath, fileFullPath);
+		// Format path to unix style path.
+		fileInDirectory = fileInDirectory?.replace(/\\/g, '/');
```
delucis, line 110: Is this an Astro v5-related change? Or a bug fix for an existing issue?

packages/starlight/style/shiki.css
delucis, line 1: Oh… I think this file predates Expressive Code support? 😅 Not something to tackle in this PR, but I’ve no idea if we still need this.

packages/starlight/loaders.ts
```diff
@@ -0,0 +1,84 @@
...
+) {
+	return (
+		(typeof srcDir === 'string' ? srcDir : srcDir.pathname).replace(
+			typeof root === 'string' ? root : root.pathname,
+			''
+		) +
+		'content/' +
+		collection
+	);
+}
```
delucis, line 84: Should these maybe be exported from a secondary utility module? As this is the top-level publicly exported module we advertise for direct use, it would probably be cleaner not to only export loader functions from it, not also utilities that aren’t part of the public API (I think?)

packages/starlight/integrations/virtual-user-config.ts
```diff
@@ -42,7 +46,19 @@ export function vitePluginStarlightUserConfig(
...
+	const rootPath = fileURLToPath(root);
+	const docsPath = resolveCollectionPath('docs', srcDir);
+
+	let collectionConfigImportPath = resolve(
+		fileURLToPath(srcDir),
+		legacy.collections ? './content/config.ts' : './content.config.ts'
+	);
+	// If not using legacy collections and the config doesn't exist, fallback to the legacy location.
+	// We need to test this ahead of time as we cannot `try/catch` a failing import in the virtual
+	// as this would fail at build time when Rollup tries to resolve a non-existenting path.
```
delucis, line 58: Small typo:

```suggestion
	// as this would fail at build time when Rollup tries to resolve a non-existent path.
```

delucis (commented):
Don’t think I spotted any showstoppers here! Just a bunch of smaller suggestions and questions.

docs/src/content/docs/reference/overrides.md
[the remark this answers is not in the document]
HiDeoo: Makes sense, I think I initially read that yesterday as "removing it". I've pushed the changes in the latest update but may be worth checking my wording.

packages/starlight/loader.ts
```diff
@@ -0,0 +1,84 @@
...
+) {
+	return (
+		(typeof srcDir === 'string' ? srcDir : srcDir.pathname).replace(
+			typeof root === 'string' ? root : root.pathname,
+			''
+		) +
+		'content/' +
+		collection
+	);
+}
```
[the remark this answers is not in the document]
HiDeoo: Definitely, these helpers came up at a very late stage of the PR when I was 100% sure we could not support user-configurable collection paths and I decided to implement those helpers to remove all hardcoded collection paths from the codebase and make it easier in the future to support user-configurable collection paths but I indeed totally forgot to move them somewhere else, thanks, great catch 🙌

packages/starlight/utils/route-data.ts
HiDeoo: Yeah, the typechecking will fail when we make `filePath` optional in our `StarlightDocsEntry` type and we would need to take care of that when doing so.

packages/starlight/utils/git.ts
HiDeoo: This is an Astro v5-related change related to the fact that we now resolve from `rootPath` and without that change our tests started failing. If you want a detailed breakdown on exactly why, it'll have to wait when I get access to a Windows machine and examine the differences.

packages/starlight/utils/route-data.ts
delucis: In general I guess this will also fall into the category of “stuff we need to know about loaders”, so we can toggle features (e.g. Git dates) based on that.

delucis (commented):
Amazing! Thanks for those quick adjustments.

Re: the remaining to-dos in the PR body:

> * [ ] Figure out if legacy collections support should be documented more (it's only a sentence + link in the manual setup page now)

As briefly discussed earlier in the review, I guess legacy support is more of a stop-gap, so it may be OK to have this mainly in the changeset for people migrating like you do now.

> * [ ]  Peer dependency bump for `@astrojs/starlight` in some packages (see changesets for the one that needs an update)

I think this was done maybe? Or are there still some more that need bumping?

> * [ ]  Merge with Lunaria directives to ignore the changes to all `code.mdx` files (if we decide to keep the changes)

Happy to keep the changes! We just need to work out the Lunaria magic 🪄

HiDeoo:
> > * [ ]   Peer dependency bump for `@astrojs/starlight` in some packages (see changesets for the one that needs an update)
> 
> I think this was done maybe? Or are there still some more that need bumping?

For `@astrojs/starlight-docsearch`, `@astrojs/starlight-tailwind`, and `@astrojs/starlight-markdoc`, we still need to manually bump the `peerDependencies` of `@astrojs/starlight` to `">=0.30.0"`, right? This won't be done automatically?

> > * [ ]   Merge with Lunaria directives to ignore the changes to all `code.mdx` files (if we decide to keep the changes)
> 
> Happy to keep the changes! We just need to work out the Lunaria magic 🪄

My guess is `@lunaria-ignore:src/content/docs/**/code.mdx`? Does that look right?

> Maybe one more thought: we don’t really have an official “Upgrade Lunaria” _guide_ in the docs

I'm confused, do you mean Starlight?

HiDeoo:
> > For `@astrojs/starlight-docsearch`, `@astrojs/starlight-tailwind`, and `@astrojs/starlight-markdoc`, we still need to manually bump the `peerDependencies` of `@astrojs/starlight` to `">=0.30.0"`, right? This won't be done automatically?
> 
> Ah yes, I think you’re right 👍

I'll do that right now.

> > > Maybe one more thought: we don’t really have an official “Upgrade Lunaria” _guide_ in the docs
> > 
> > 
> > I'm confused, do you mean Starlight?
> 
> Er… yes, 😂 Sorry, left my brain elsewhere I guess.

Great idea, altho as you proposed, maybe you doing it after the release so we have the direct link and to avoid my poor wording, let's do that after.

.changeset/green-suns-drive.md
```diff
@@ -0,0 +1,53 @@
+---
+"@astrojs/starlight": minor
+---
+
+Adds support for Astro v5, drops support for Astro v4.
+
+⚠️ **BREAKING CHANGE:** Astro v4 is no longer supported. Make sure you [update Astro](https://docs.astro.build/en/guides/upgrade-to/v5/) and any other official integrations at the same time as updating Starlight:
```
sarah11918, line 7:
```suggestion
## Upgrade Astro and dependencies

⚠️ **BREAKING CHANGE:** Astro v4 is no longer supported. Make sure you [update Astro](https://docs.astro.build/en/guides/upgrade-to/v5/) and any other official integrations at the same time as updating Starlight:
```
Not sure whether the headers are strictly speaking necessary, but it feels like this could be broken into 2 sections to make each part of the process obvious.

.changeset/green-suns-drive.md
```diff
@@ -0,0 +1,53 @@
...
+
+```sh
+npx @astrojs/upgrade
+```
+
+_Community Starlight plugins and Astro integrations may also need to be manually updated to work with Astro v5. If you encounter any issues, please reach out to the plugin or integration author to see if it is a known issue or if an updated version is being worked on._
+
+Astro 5.0 introduces a new version of [content collections](https://docs.astro.build/en/guides/content-collections/) which Starlight is built on top of. We recommend converting the `docs` (and optionally the `i18n`) collection(s) used by Starlight to the new Content Layer API which brings several performance improvements and added capabilities.
```
sarah11918, line 15:
```suggestion
## Update your collections

⚠️ **BREAKING CHANGE:** Starlight's internal [content collections](https://docs.astro.build/en/guides/content-collections/), which organize, validate, and render your content, have been updated to use Astro's new Content Layer API and require configuration changes in your project.
```
Same thing here, just thinking of explicitly calling this out as another step, and focusing on what happened and to do about it.

Noting this removes the notion of "benefits" which could still be included!

.changeset/green-suns-drive.md
```diff
@@ -0,0 +1,53 @@
...
+
+1. **Move the content config file.** This file no longer lives within the `src/content/config.ts` folder and should now exist at `src/content.config.ts`.
+
+   The previous location is still supported, and is required if the [`legacy.collections`](https://docs.astro.build/en/reference/legacy-flags/) flag is enabled.
```
sarah11918, line 19:
```suggestion
```
Am super tempted to not mention this, as we didn't in the upgrade guide during general upgrade advice, only for people NOT performing any upgrade steps. Avoids the risk of them not upgrading later, and they ARE making other upgrades here now. This should happen at the same time.

.changeset/green-suns-drive.md
```diff
@@ -0,0 +1,53 @@
...
+   -  docs: defineCollection({ schema: docsSchema() }),
+   +  docs: defineCollection({ loader: docsLoader(), schema: docsSchema() }),
+   -  i18n: defineCollection({ type: 'data', schema: i18nSchema() }),
+   +  i18n: defineCollection({ loader: i18nLoader(), schema: i18nSchema() }),
+    };
+   ```
+
+1. **Update other collections.** To update any other collections you may have, follow the [“Updating existing collections”](https://docs.astro.build/en/guides/upgrade-to/v5/#updating-existing-collections) section in the Astro 5 upgrade guide.
+
+If you are unable to make any changes to your collections at this time, you can enable the [`legacy.collections` flag](https://docs.astro.build/en/reference/legacy-flags/) which is supported by Starlight and will allow you to keep your collections in their current state until the legacy flag is no longer supported.
```
sarah11918, line 53:
```suggestion
If you are unable to make any changes to your collections at this time, including Starlight's default `docs` and `i18n` collections, you can enable the [`legacy.collections` flag](https://docs.astro.build/en/reference/legacy-flags/) to upgrade to v5 without updating your collections. This legacy flag exists to provide temporary backwards compatibility, and will allow you to keep your collections in their current state until the legacy flag is no longer supported.
```
Wanted to make sure the "your collections" was understood to also include the ones they might not think of as *their* collections!

sarah11918 (commented):
It's happening! Some thoughts for your consideration

.changeset/green-suns-drive.md
delucis: Ah, I think headings might be tricky because once this ends up inside the CHANGELOG there’s an expected heading hierarchy (and this entire changeset will be inside a bullet list). We’d need to double check what level these have to be if we think a heading is important.

delucis (approved):
Here goes nothing!

5k GitHub Stars, 5k “used-by” projects on GitHub, we must be ready for Astro 5k 🚀 

Massive thanks to @HiDeoo for guiding this work and start our new adventures with ✨ _Content Layer_ ✨

PR #2613, opened 2024-11-18
Fix `sidebar` frontmatter use in internal links

delucis:
#### Description

- Fixes use of the `sidebar` frontmatter options to control appearance of a sidebar link when included using the “internal” link syntax, i.e. `{ slug: 'page/name' }` or the shorthand of `'page/name'`
- Currently `sidebar.label`, `sidebar.badge`, and `sidebar.attrs` in frontmatter are ignored with only `title` from frontmatter being used when an explicit label is not set.
- This PR fixes that and updates a sidebar test to match

### DIFF
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -13,7 +13,7 @@ import { createPathFormatter } from './createPathFormatter';
 import { formatPath } from './format-path';
 import { BuiltInDefaultLocale, pickLang } from './i18n';
 import { ensureLeadingSlash, ensureTrailingSlash, stripLeadingAndTrailingSlashes } from './path';
-import { getLocaleRoutes, routes, type Route } from './routing';
+import { getLocaleRoutes, routes, type Route, type StarlightDocsEntry } from './routing';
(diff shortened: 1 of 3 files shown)


HiDeoo (approved):
Oh, interesting we did not hit this in the Starlight docs, nice catch and it now works as expected when testing locally with the fix :tada:

packages/starlight/utils/navigation.ts
```diff
@@ -13,7 +13,7 @@ import { createPathFormatter } from './createPathFormatter';
 import { formatPath } from './format-path';
 import { BuiltInDefaultLocale, pickLang } from './i18n';
 import { ensureLeadingSlash, ensureTrailingSlash, stripLeadingAndTrailingSlashes } from './path';
-import { getLocaleRoutes, routes, type Route } from './routing';
+import { getLocaleRoutes, routes, type Route, type StarlightDocsEntry } from './routing';
```
HiDeoo, line 16:
```suggestion
import { getLocaleRoutes, routes, type Route } from './routing';
```

PR #2616, opened 2024-11-19
Fix language switcher `trailingSlash` edge case

### LINKED ISSUE #2609: `trailingSlash` is not respected for the multilingual sites selector

0.29.0

4.16.13

npm

Linux

Firefox

### Describe the Bug

when in the Astro config I have `trailingSlash: 'never'`, It seems that links for the language switch are not set to respect the same attribute as they are always ending with `/`

ex: `/fr/` instead of `/fr`

Repository: https://github.com/tcgdex/documentation

https://github.com/tcgdex/documentation

### Participation

delucis:
#### Description

- Closes #2609
- Handles an edge case where we were still generating URLs with trailing slashes in the language switcher (and some `<head>` metadata) even when Astro’s `trailingSlash: 'never'` was set.
- In general, we get away without formatting these URLs because we’re just changing the local prefix for `Astro.url` which is already in the correct format.
- The edge case is that a `URL` object can never have a `pathname` of `""`. Setting `pathname = ""` still results in a `pathname` of `"/"`. This means that when prefixing a root locale homepage, we can’t rely only on `pathname` to know if a trailing slash should be present or not.
- Instead I’m importing our context object to retrieve the user’s configured `trailingSlash` option.
- Wasn’t 100% sure how best to fix this. In the end I made the `trailingSlash` config a parameter to our `localizedUrl()` utility as this makes testing simpler, rather than importing the user config into the utility directly which would mean having to set up two separate test environments. But open to feedback if people think it would be better to avoid the extra parameter.

### DIFF
--- a/.changeset/thick-cooks-call.md
+++ b/.changeset/thick-cooks-call.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an edge case to correctly avoid a trailing slash when navigating from a root locale homepage to another language via Starlight’s language switcher when `trailingSlash: 'never'` is set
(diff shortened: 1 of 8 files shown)


HiDeoo (approved):
Tested locally too and the edge case seems to be fixed :raised_hands:

I personally think it's fine to add the extra argument, we don't use that function in many places and it makes it easy to test without duplicating fixtures with multiple configs.

PR #2623, opened 2024-11-22
remove forced paint

FredKSchott:
#### Description

- Fixes a forced paint currently affecting Astro docs, others.
- I did this via the GH UI and haven't tested it myself, you've been warned!

### DIFF
--- a/packages/starlight/components/SidebarPersister.astro
+++ b/packages/starlight/components/SidebarPersister.astro
@@ -56,12 +56,12 @@ declare global {
 	<slot />
 
 	<script is:inline>
-		(() => {
+		requestAnimationFrame(() => {
 			const scroller = document.getElementById('starlight__sidebar');
 			if (!window._starlightScrollRestore || !scroller) return;
 			scroller.scrollTop = window._starlightScrollRestore;
 			delete window._starlightScrollRestore;
-		})();
+		});
 	</script>
 </sl-sidebar-state-persist>
 


trueberryless:
There is currently a [draft PR](#2569) which moves the inline script into `Page.astro` because there are other restoration issues when overriding the sidebar.

Not sure how the maintainers will handle this, but I think we either wait for #2569 to merge or merge this one first and update the other PR...

.changeset/gorgeous-pugs-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix an issue where the scrollbar was causing a forced paint halfway through page render
```
trueberryless, line 5:
```suggestion
Fixes an issue where the scrollbar was causing a forced paint halfway through page render
```

Starlight and Astro prefer writing their changesets in a present tense verb.

delucis:
I remember testing this in the past but seems I never left any feedback.

Just tested again for sanity and confirmed that this does not seem to impact paint timing at all. Inspecting performance traces in Chrome shows identical frame-by-frame rendering between Starlight’s production site and the deploy preview.

It may be that something else causes paint and therefore the `requestAnimationFrame()` added here still fires at the same time as it does currently? Either way this PR is not making any difference, so I’m going to close it to tidy up PRs for now. Would still be a good thing to investigate whether we can improve this on page loads though.

PR #2549, opened 2024-10-31
Add new plugin to Starlight (`resources/plugin.mdx`)

trueberryless:
#### Description

- add new plugin called `starlight-sidebar-plugin-dropdown`
- Starlight plugin to split your docs page into multiple subpages and switch between them with a dropdown menu in the sidebar.

**1000 thanks to @HiDeoo the actual idea for such a plugin was their idea! (I just changed the visuals a bit)**

Try out the plugin [here](https://starlight-sidebar-topics-dropdown.trueberryless.org/).

PR #2624, opened 2024-11-23
i18n(de): update `resources/plugins.mdx`

trueberryless:
#### Description

- update `resources/plugins.mdx`
- #2595


trueberryless:
I think Lunaria is showing that #2549 is not yet translated, but I think the content is actually translated cause the plugins `starlight-sidebar-topics` and `starlight-sidebar-topics-dropdown` already exist in the German translation...

jsparkdev (approved):
As a non-native speaker, I've reviewed it using a translator and all the updates you've made appear to be correct.

LGTM!

PR #2625, opened 2024-11-23
i18n(de): update `reference/overrides.md`

trueberryless:
#### Description

- update `reference/overrides.md`
- #2386


docs/src/content/docs/de/reference/overrides.md
```diff
@@ -211,7 +212,8 @@ Diese Komponente wird innerhalb von [`<PageFrame>`](#pageframe) gerendert und is
 
 #### `TwoColumnContent`
 
-**Standardkomponente:** [`TwoColumnContent.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/TwoColumnContent.astro)
+**Standardkomponente:** [`TwoColumnContent.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/TwoColumnContent.astro)  
+**Benannte Slots:** `right-sidebar`
```
jsparkdev, line 216:
```suggestion
**Benannte Slot:** `right-sidebar`
```

This is a minor but simple fix.

trueberryless: Good catch, thank you!🎉👍

PR #2627, opened 2024-11-23
i18n(fr): update `guides/overriding-components` & `reference/overrides`

HiDeoo:
#### Description

This PR updates the French version of the `guides/overriding-components` & `reference/overrides` pages with the changes from #2386.


jsparkdev (approved):
As a non-native speaker, I reviewed it using a translator and all the corrections seem to be correct!

LGTM!

PR #2632, opened 2024-11-23
i18n(ko): fix Markdown newline in `reference/overrides.md`

trueberryless:
#### Description

- there is no newline in the Korean translation from #2629 
- other translations and original have two spaces before the newline to enforce the newline (#2627, #2386)


jsparkdev (approved):
@trueberryless  Good catch, thanks.

PR #2633, opened 2024-11-23
Fix Safari VoiceOver issue

HiDeoo:
#### Description

The change may look a bit weird at first but it looks like the number of Safari/VoiceOver issue with `display: contents;` is growing. A new "fun" one I just discovered seems to be related to a web component having `display: contents;` and a `<script>` element as a direct child which seems to make VoiceOver read the content of the `<script>` element.

This reproduces at least on macOS Sequoia 15.1 with Safari 18.1.

Here is a video before the fix…

https://github.com/user-attachments/assets/0d6beffb-5051-42b2-9882-9236bf687770

Here is a video after the fix:

https://github.com/user-attachments/assets/c0208683-2376-43c8-8f90-05cbb50b2976

If this is accepted, maybe we could add a comment in the file frontmatter explaining why we have this attribute on a `script` element with a link to this PR and the videos?

### DIFF
--- a/.changeset/strange-icons-explode.md
+++ b/.changeset/strange-icons-explode.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a VoiceOver issue with Safari where the content of a `<script>` element could be read before the sidebar content.
--- a/packages/starlight/components/SidebarPersister.astro
+++ b/packages/starlight/components/SidebarPersister.astro
@@ -13,10 +13,10 @@
   - The state is parsed from session storage and restored.
   - This is a progressive enhancement, so any errors are swallowed silently.
 
-	The `aria-hidden` attribute on script tags is used to prevent a Safari/VoiceOver bug where the
-	script is read out loud due to them being inside a container with `display: contents`.
-	@see https://bugs.webkit.org/show_bug.cgi?id=283645
-	@see https://github.com/withastro/starlight/pull/2633
+  The `aria-hidden` attribute on script tags is used to prevent a Safari/VoiceOver bug where the
+  script is read out loud due to them being inside a container with `display: contents`.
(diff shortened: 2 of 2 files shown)


KonnorRogers:
I tested the current Starlight docs home page on MacOS 14.6 and found similar results:

VO + Safari 17.6 reads the script tag contents (unexpected)
VO + Chrome 131 / Firefox 132 correctly skip the script tag contents (expected)

delucis:
Thanks for catching this @HiDeoo! Linking a WebKit bug I just reported based on this: https://bugs.webkit.org/show_bug.cgi?id=283645

Given the bug is CSS related I'm curious whether setting `display: none` on the script would also work to fix this. But I don't really mind what approach we use as long as we're not announcing code to people 😅 

I [shared this issue on Mastodon](https://m.webtoo.ls/@swithinbank/113534442704057404) and got various responses including some [recommending avoiding `display: contents` entirely](https://toot.cafe/@aardrian/113534956445587342). I understand that sentiment given the difficulty testing the full matrix of OS x browser x screen reader combinations and the apparent flakiness of support for `display: contents`, but in a handful of cases it's really really useful, so I think we can just proceed with caution and take extra care to test whenever we use it (and use it as a last resort).

HiDeoo:
> Given the bug is CSS related I'm curious whether setting `display: none` on the script would also work to fix this. But I don't really mind what approach we use as long as we're not announcing code to people 😅

The script already get that style from the user agent style sheet and I guess it's not enough in this case.

image

> but in a handful of cases it's really really useful, so I think we can just proceed with caution and take extra care to test whenever we use it (and use it as a last resort).

That's also my opinion, especially in this one I think it's super useful, the bug is definitely not a common case and we don't overuse this value anyway.

delucis (approved):
Thanks for the fix! I also tested on Safari 18.1 on macOS 15.1 and confirmed both that the bug exists and that this PR fixes it.

PR #2635, opened 2024-11-25
Fix language selector invalid value

HiDeoo:
This issue fixes an issue where the language selector could have an invalid value when the user navigates back to a page in a different language.

Repro:

1. Visit [`http://localhost:4321/`](http://localhost:4321/)
1. Change the language to French
1. Press the back button

At expected at this point, the page is in English however the language selector is still set to French.

In Chrome & Safari, a refresh of the page would fix the issue. However, in Firefox, a refresh of the page is not enough to fix the issue, you have to manually focus the address bar and press enter again.

This is not something visible with the theme selector as we have some JS code to update the picker when the page is loaded.

Setting the to `autocomplete` attribute to `off` on the `select` element prvent the browser to automatically select a cached value for this field based on the last user input. Our `<Select>` component is only used for the theme and theme selector so I applied the change to the component itself.


delucis (commented):
Oh, nice fix! Would have assumed we needed a bit of extra JS for this.

Are we sure it works? I just tried in Chrome (v131 on macOS 15) and still saw the old behaviour:

https://github.com/user-attachments/assets/e62ad408-ee82-45e0-af5e-3d23f2c5aa53

HiDeoo:
Oh, interesting, I did not checked the deploy preview yet but indeed, this works in dev but not in the production version. I wonder why the difference 🤔 I'll investigate a bit more as it would be nice to avoid JS for this if possible somehow.

trueberryless:
I think that the production preview works properly, because it works on my machine :)

I'm using Arc v1.29.0 on Windows 11 and here you can see the comparison between current Starlight (left) and preview of this PR (right).

https://github.com/user-attachments/assets/7d1a6afb-1498-431f-b2fc-7da05d1101dc

PS: I'm sorry that my screen recorder didn't capture the dropdown menu, what are you guys using for recording?

HiDeoo:
After investigating a bit more and finding so many cases/bug reports regarding how the `autocomplete` attribute is behaving, and considering the explicit value we provide can be ignored, domain-dependent, markup dependant, change based on user's settings or browser, I think it's safe to say that we sadly can't rely on it even tho, per spec, it should work…

I'll have to update this PR to use a JS based approach instead when I get the time.

> PS: I'm sorry that my screen recorder didn't capture the dropdown menu, what are you guys using for recording?

Depending on the machine I use, I'll either use ScreenFlow or the built-in macOS screen recording tool.

HiDeoo:
Turns out after more investigation that in this specific case, `autocomplete="off"` works in all browsers but what's making the difference in production is the [back/forward cache (or bfcache)](https://web.dev/articles/bfcache/) (that's also why `autocomplete="off"` is still present in the changes and works in dev where the bfcache does not work).

I updated the PR with some code to hook to the [`pageshow`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event) event which has a [`persisted`](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted) property, which is true if the page was restored from bfcache. If this is the case, we ensure that the selected value is the correct one.

I made an E2E test for it while adding the code but I did not submit it:

- By default, the bfcache is disabled in playwright.
- We can change that but due to a Chrome limitation, this does not work in headless mode.
- The [`new` headless implementation](https://github.com/microsoft/playwright/issues/33566) available in Playwright v1.49.0 removes that limitation but it felt a bit too new to switch to it as there are still some issues with it.

packages/starlight/components/LanguageSelect.astro
```diff
@@ -46,4 +46,18 @@ function localizedPathname(locale: string | undefined): string {
 		}
 	}
 	customElements.define('starlight-lang-select', StarlightLanguageSelect);
+	window.addEventListener('pageshow', (event) => {
```
HiDeoo, line 49: A better event would be `resume` from the Document API but it's [not available in Safari](https://caniuse.com/mdn-api_document_resume_event) at the moment.

packages/starlight/components/LanguageSelect.astro
```diff
@@ -46,4 +46,18 @@ function localizedPathname(locale: string | undefined): string {
...
+		if (!event.persisted) return;
+		document
+			.querySelectorAll<HTMLSelectElement>('starlight-lang-select select')
+			.forEach((select) => {
+				// If the page was loaded from a cache, the language select selected index might not be in
+				// sync with the current page.
+				const markupSelectedIndex =
+					select.querySelector<HTMLOptionElement>('option[selected]')?.index;
+				if (markupSelectedIndex !== select.selectedIndex) {
+					select.selectedIndex = markupSelectedIndex ?? 0;
```
trueberryless, line 59:
```suggestion
					select.selectedIndex = markupSelectedIndex ?? select.selectedIndex;
```

Please correct me if I'm wrong, but wouldn't it be better to leave the dropdown as it is if markupSelectedIndex is undefined?

Or formulated differently: Why did you fallback to the default locale in this case, HiDeoo? 

This is just another option to consider, you are probably right to default to 0...

HiDeoo: If the selected index from the markup is not defined, we're in the case of the default locale with no `selected` attribute in the markup. For example, if you check the markup on https://starlight.astro.build/, you can see that there is no `selected` attribute for "English" which makes the first option automatically selected.

image

Falling back to `0` ensures that the first option is selected in this case even if `select.selectedIndex` has a different value due to a cache restoration.

Basically, when a cache is restored, the markup is always the source of truth and we never fallback to `select.selectedIndex` which may not be reliable.

packages/starlight/components/LanguageSelect.astro
```diff
@@ -46,4 +46,18 @@ function localizedPathname(locale: string | undefined): string {
 		}
 	}
 	customElements.define('starlight-lang-select', StarlightLanguageSelect);
+	window.addEventListener('pageshow', (event) => {
+		if (!event.persisted) return;
+		document
+			.querySelectorAll<HTMLSelectElement>('starlight-lang-select select')
+			.forEach((select) => {
```
delucis, line 53: Rather than this query selector, would it maybe make sense to move the event listener inside the custom element where we already get the `<select>`?

It means duplicating the `pageshow` listener, but we never have these more than twice on a page, so it feels like it would be “cleaner” to encapsulate it there. Unless there are other lifecycle reasons not to do it like that.

packages/starlight/components/Select.astro
```diff
@@ -17,7 +17,7 @@ interface Props {
 <label style={`--sl-select-width: ${Astro.props.width}`}>
 	<span class="sr-only">{Astro.props.label}</span>
 	<Icon name={Astro.props.icon} class="icon label-icon" />
-	<select value={Astro.props.value}>
+	<select value={Astro.props.value} autocomplete="off">
```
delucis, line 20: The intention is still to add the `autocomplete` attribute here right? Just for the browsers/conditions where it does work?

May be worth a comment inline to remind ourselves why we added it.

delucis (commented):
Thanks @HiDeoo! I’ll admit there’s part of me that is unsure if this is even a “bug” to be fixed or just “how form elements work and are expected to work”, but the thing that we should definitely fix is that switching language should always work (which it doesn’t currently in this scenario because _re-selecting_ a language doesn’t change anything). So I guess this is one approach to fixing that.

I left a couple of queries and also looks like this script takes us 10 bytes over our JS side limit, so that budget would need increasing for CI to pass.

delucis (approved):
Ready for launch! Thanks again @HiDeoo 🙌

PR #2636, opened 2024-11-26
feat(link): add target="_blank" for external social links

TinsFox:
### Description

This PR enhances the user experience of social media links by making them open in new tabs, allowing users to maintain their current browsing session while accessing external social platforms.

#### Changes Made
- Added `target="_blank"` attribute to social media anchor tags
- Maintains the main site navigation context for users
- Follows web accessibility best practices

#### Before
Social media links would navigate away from the current page, disrupting the user's browsing experience.

#### After
Social media links now open in new tabs, preserving the user's current session on the main site.

#### Technical Details
Modified the social media link component:
html

<a href={url} rel="me" class="sl-flex">

<a href={url} rel="me" target="blank" class="sl-flex">

### DIFF
--- a/packages/starlight/components/SocialIcons.astro
+++ b/packages/starlight/components/SocialIcons.astro
@@ -12,7 +12,7 @@ const links = Object.entries(config.social || {}) as [Platform, SocialConfig][];
 	links.length > 0 && (
 		<>
 			{links.map(([platform, { label, url }]) => (
-				<a href={url} rel="me" class="sl-flex">
+				<a href={url} rel="me" class="sl-flex" target="_blank">
 					<span class="sr-only">{label}</span>
 					<Icon name={platform} />
 				</a>


HiDeoo:

The current behavior is actually an intentional design decision. We avoid opening links in new tabs by default as shown in various UX researches like this [one](https://www.nngroup.com/articles/new-browser-windows-and-tabs/) from the Nielsen Norman Group. In general, it's best to avoid forcing a user to open something in a new tab and let them decide how they want to open a link, e.g. a basic click = same window, cmd/ctrl + click or middle click = new window/tab.

Thanks to Starlight [component overrides](https://starlight.astro.build/guides/overriding-components/), users can still explicitly opt-in to open links in new tabs even though we do not recommend it as the default behavior.

Considering the above, I'm going to close this PR. If you have any questions or concerns, feel free to reach out on [Discord](https://astro.build/chat/) in the `#starlight` channel or open a new discussion.

PR #2640, opened 2024-11-27
feat: add Simple Layout without default styles and components

totto2727:
## Overview

- Addition of `SimpleStarlightPage.astro` file with non-user CSS styling and removal of all but the minimum required elements
- Extracted part of Page.astro into a separate component

## Details

A `StarlightPage` component already exists that provides layouts for custom pages.
However, if you want to display content across the full screen, I don't need the header and sidebars included in `StarlightPage`.

ex. Reveal.js slides, etc.

However, it is troublesome to set up the `head` from scratch just for that custom page, and there is more code to manage.
If possible, it is preferable to have the same structure as the page using the `StarlightPage` component.

So I added a component that provides only the minimum layout required.

If you already have a solution to this problem, I would appreciate it if you could let me know.

### DIFF
--- a/packages/starlight/components/Page.astro
+++ b/packages/starlight/components/Page.astro
@@ -13,7 +13,6 @@ import ContentPanel from 'virtual:starlight/components/ContentPanel';
 import FallbackContentNotice from 'virtual:starlight/components/FallbackContentNotice';
 import DraftContentNotice from 'virtual:starlight/components/DraftContentNotice';
 import Footer from 'virtual:starlight/components/Footer';
-import Head from 'virtual:starlight/components/Head';
 import Header from 'virtual:starlight/components/Header';
 import Hero from 'virtual:starlight/components/Hero';
 import MarkdownContent from 'virtual:starlight/components/MarkdownContent';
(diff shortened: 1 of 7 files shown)


packages/starlight/components/StarlightHeadlessPage.astro
Nadwey, line 1: This component is named `StarlightHeadlessPage`, but it is referred to as `SimpleStarlightPage` elsewhere.

packages/starlight/components/Page.astro
```diff
@@ -22,101 +21,85 @@ import PageSidebar from 'virtual:starlight/components/PageSidebar';
...
+		/* Add scroll padding to ensure anchor headings aren't obscured by nav */
+		html {
+			/* Additional padding is needed to account for the mobile TOC */
+			scroll-padding-top: calc(1.5rem + var(--sl-nav-height) + var(--sl-mobile-toc-height));
+		}
+		main {
+			padding-bottom: 3vh;
+		}
+		@media (min-width: 50em) {
+			[data-has-sidebar] {
```
Nadwey, line 53:
```suggestion
			:global(:root[data-has-sidebar]) {
```

PR #2641, opened 2024-11-28
Add tests and functionality for handling double quotes in file names in FileTree

iskandersierra:
#### Description

- Closes #2608
- What does this PR change?
  Allows to use double quotes in `FileTree` component to indicate the full file name in cases where the file name contains white space. Like: `- "my file name" this is the comment`
  Yes, but I don't know how to test the visuals from this project. I just added some tests for this case and change the code parsing the content of the list.
  I would still need to update the docs

### DIFF
--- a/docs/src/content/docs/components/file-tree.mdx
+++ b/docs/src/content/docs/components/file-tree.mdx
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
 
 </Preview>
 
+### Escape white-space in file names
+
+To include white-space in a file or directory name, escape it with double quotes.
+
+<Preview>
+
+```mdx {5,6}
+import { FileTree } from '@astrojs/starlight/components';
+
+<FileTree>
+
+- "main folder/"
+  - "The Header.astro"
+  - …
+
+</FileTree>
+```
+
+<Fragment slot="markdoc">
+
+```markdoc {2,3}
+{% filetree %}
+- "main folder/"
+  - "The Header.astro"
+  - …
+{% /filetree %}
+```
+
+</Fragment>
+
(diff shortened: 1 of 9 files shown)


iskandersierra:
I might need to consider detecting the special double-quotes `“` and `”` because the markdown is turning the regular double quotes into those.

image

iskandersierra:
It is working when the comments contain some nodes, but not when the comments are plain text 
image

iskandersierra:
I think there is no need to implement this feature as proposed. There is already  away to create a file name with spaces, if we use a tag like `<span>` to enclose the file name.

Maybe the only required change is in the docs to indicate that fact, but there is no need to complicate things with using quotes because they introduce too many cases when combined with markup tags.

delucis:
Hi @iskandersierra — sorry for the delayed response, this PR dropped off my radar over the winter break.

> Maybe the only required change is in the docs to indicate that fact, but there is no need to complicate things with using quotes because they introduce too many cases when combined with markup tags.

Oh that’s interesting! I hadn’t considered that could work. And yes, ideally the `"file name"` syntax would have been a nicer shorthand than `<span>file name</span>`, but I hadn’t thought about how that would interact with the default Smartypants plugin Astro adds.

Maybe we could do a couple things:

1. Document that if we think it’s useful
2. Add a test for that behaviour to make sure we don’t accidentally break it.

delucis:
Going to close this PR for now — thanks again for the exploration @iskandersierra! 💖 

And if you or anyone else wants to follow up on the [suggestions above](https://github.com/withastro/starlight/pull/2641#issuecomment-2586726075), I’d be happy to take a look.

PR #2645, opened 2024-11-30
Fix favicon support for query and fragment in URLs

### LINKED ISSUE #2639: favicon expects a relative url but treats it as a filename

0.29.2

4.16.10

npm

linux

chrome

### Describe the Bug

The [favicon](https://github.com/withastro/starlight/blob/dc481aea7990e06dc6fa980d15ee02db2c67b0a0/packages/starlight/schemas/favicon.ts#L16) configuration parameter expects a relative URL pointing to the favicon.  However, the configuration validation treats the value as a [file path](https://github.com/withastro/starlight/blob/dc481aea7990e06dc6fa980d15ee02db2c67b0a0/packages/starlight/schemas/favicon.ts#L18) in order to detect the file type.

The current approach causes issues when the relative url ends with anything beyond the file extension - for example, when a primitive cache busting approach is applied:

```ts
  integrations: [
    starlight({
      favicon: '/favicon.svg?v=1732649058923',
```

```bash
❯ npm run dev

> @example/starlight-basics@0.0.1 dev
> astro dev

[AstroUserError] Invalid config passed to starlight integration
  Hint:
    favicon must be a .ico, .gif, .jpg, .png, or .svg file
  Location:
    /home/projects/withastro-starlight-qjpiye/node_modules/astro/dist/core/errors/errors.js:120:5
  Stack trace:
    at new AstroUserError (file:///home/projects/withastro-starlight-qjpiye/node_modules/astro/dist/core/errors/errors.js:120:5)
    at Module.parseWithFriendlyErrors (/home/projects/withastro-starlight-qjpiye/node_modules/@astrojs/starlight/utils/error-map.ts:6:10)
    at astro:config:setup (/home/projects/withastro-starlight-qjpiye/node_modules/@astrojs/starlight/index.ts:45:59)
    at async Module.createContainer (file:///home/projects/withastro-starlight-qjpiye/node_modules/astro/dist/core/dev/container.js:39:14)
    at async Module.dev (file:///home/projects/withastro-starlight-qjpiye/node_modules/astro/dist/core/dev/dev.js:40:19)
```

The value for favicon should be parsed as a relative url and the type extracted based on the path and not the entire configuration value.

https://stackblitz.com/edit/withastro-starlight-qjpiye?file=astro.config.mjs

### Participation

techfg:
#### Description

- Closes #2639 
- What does this PR change? Supports query and fragment on favicon url when detecting image type


packages/starlight/utils/path.ts
```diff
@@ -50,3 +50,14 @@ export function ensureHtmlExtension(path: string) {
 	}
 	return ensureLeadingSlash(path);
 }
+
+export function splitQueryAndFragment(url: string) {
```
delucis, line 54: We should document this function probably like the other functions in this file:

```suggestion
/** Parse a URL into pathname, search params, and hash components. */
export function splitQueryAndFragment(url: string) {
```

packages/starlight/utils/path.ts
```diff
@@ -50,3 +50,14 @@ export function ensureHtmlExtension(path: string) {
...
+	const [urlWithQuery, hash] = url.split('#');
+	const [urlBase, queryString] = (urlWithQuery || '').split('?');
+
+	return {
+		base: urlBase || '',
+		query: queryString ? `?${queryString}` : '',
+		fragment: hash ? `#${hash}` : '',
+	};
```
delucis, line 62: We could potentially lean on native URL parsing here rather than rolling our own. What do you think? In this example, I’d suggest returning the properties exactly as they are named in the web standard API: `pathname` rather than `base`, `search` rather than `query` (`query` in the `URL` object excludes the `?`), and `hash` rather than `fragment`:

```suggestion
        const { pathname, search, hash } = new URL(url, 'https://example.com');
	return { pathname, search, hash };
```

If the `example.com` looks a bit funny here, the way this works is that it resolves the first argument relative to the second, so for relative arguments (e.g. `/foo`) we need a base to resolve against otherwise it can’t construct a full URL. We’re only using the origin at all, so `example.com` is ignored in output here.

delucis, line 62: But a second question: the usage in `favicon.ts` only needs the pathname right? We’re not using the search and hash?

In that case, a `stripSearchAndHash()` function might make more sense that just returns the pathname directly?

.changeset/neat-deers-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes favicon url parsing of query and fragment
```
delucis, line 5: Suggesting a small tweak to this changeset:

```suggestion
Fixes support for favicon URLs that contain a search query and/or hash
```

delucis (commented):
Thank you @techfg! Left a few suggestions.

packages/starlight/utils/path.ts
techfg: @delucis - Thanks for all the input!  Regarding above, the reason I didn't use `URL` for the parsing is due to #2647 as the favicon can be absolute or relative.  I considered detecting if it is absolute or internal and then using `URL` but that seemed rather fragile in terms of approach.  Additionally, to detect absolute vs relative, my options were to add a dependency to do the detection or roll my own and didn't feel that either was ideal.  Since we really need to only grab `base` regardless of whether its absolute or relative, I landed on the approach I took.  Happy to adjust, just need to make sure that we handle absolute and relative in this case - let me know your thoughts.

techfg: @delucis - Agreed, in this case we only need `base`, keeping in mind it may be an absolute or relative url so I chose `base` vs. `pathname` since it could be more than just the "path" returned.  I included `query` and `fragment` in case it was ever needed in future since the parsing is required either way so seemed to make sense to return.  I chose `query` and `fragment` instead of using `search` and `hash` since this function, at least as written, separates a `url` in to its `parts` and `query` and `fragment` are the names given to the parts of a url (as opposed to the properties of URL itself which separates a url even further in to distinct properties).  All that is to say is why I chose what I did but happy to adjust however you think is best, just let me know what you think.

packages/starlight/utils/path.ts
techfg: Will adjust once other feedback is resolved and final names determined.

.changeset/neat-deers-pretend.md
techfg: Will adjust once other feedback is resolved and final names determined.

packages/starlight/utils/path.ts
delucis: I think that’s OK because the code doesn’t use the data this function returns for anything other than checking the extension.

techfg: Sorry, just wanted to clarify what you mean by "OK" - do you mean leave as I have it or change to detect absolute vs relative URL and then use URL accordingly?

delucis: - Use `URL` directly
- No need to detect absolute vs relative because we’re not using this function to get the final URL at all, only to extract the extension.

techfg: TIL -  `URL` constructor will ignore `base` when `url` is absolute so `new URL('https://starlight.build.dev/favicon.svg', 'https://www.example.com') will succeed.

I had assumed (I should have actually tested it ☹️ ) that it will fail, hence all comments above, sorry!

I'll adjust to use URL and update PR, thanks!

.changeset/neat-deers-pretend.md
techfg: Updated PR.

packages/starlight/utils/path.ts
techfg: Eliminated `splitQueryAndFragment` util function as its no longer necessary given use of `URL` constructor.

packages/starlight/utils/path.ts
techfg: Eliminated `splitQueryAndFragment` util function as it's no longer necessary given use of `URL` constructor.

techfg: Eliminated `splitQueryAndFragment` util function as it's no longer necessary given use of `URL` constructor.

delucis (approved):
Lovely! And it’s much simpler now too. Great work 🙌

PR #2646, opened 2024-11-30
chore: remove rel `shortcut` from favicon

### LINKED ISSUE #2112: favicon icons don't need `rel="shortcut icon"` - can just be `rel="icon"`

0.25.0

4.11.5

npm

Windows

Edge

### Describe the Bug

image

No need for `rel="shortcut icon"`, should be just `rel="icon"`.

`shortcut` is not, and never has been, a valid link relation. Read this [amazing article](https://mathiasbynens.be/notes/rel-shortcut-icon) by Mathias Bynens from ten years ago that explains why we never needed shortcuts and why rel="icon" is just fine.

(from: https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs)

n/a

### Participation

techfg:
#### Description

- Closes #2112
- What does this PR change? Removes rel `shortcut` from favicon.  As discussed in #2112, 'shortcut' only existed for IE <=8 which Starlight does not support.  While agents should properly parse the rel value even with shortcut, it is non-conforming and should not be used (see [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#icon)).
- Did you change something visual? No.

### DIFF
--- a/packages/starlight/__tests__/basics/head.test.ts
+++ b/packages/starlight/__tests__/basics/head.test.ts
@@ -91,9 +91,10 @@ describe('createHead', () => {
 		const defaultFavicon = {
 			tag: 'link',
 			attrs: {
-				rel: 'shortcut icon',
+				rel: 'icon',
 				href: '/favicon.svg',
 				type: 'image/svg+xml',
+				'data-favicon': 'default',
 			},
 		} as const;
 
--- a/.changeset/lazy-rats-decide.md
+++ b/.changeset/lazy-rats-decide.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Remove 'shortcut' from favicon rel attribute
--- a/packages/starlight/components/Head.astro
+++ b/packages/starlight/components/Head.astro
@@ -32,9 +32,10 @@ const headDefaults: z.input<ReturnType<typeof HeadConfigSchema>> = [
 	{
 		tag: 'link',
 		attrs: {
-			rel: 'shortcut icon',
+			rel: 'icon',
 			href: fileWithBase(config.favicon.href),
 			type: config.favicon.type,
+			'data-favicon': 'default',
 		},
 	},
 	// OpenGraph Tags
--- a/packages/starlight/utils/head.ts
+++ b/packages/starlight/utils/head.ts
@@ -89,7 +89,12 @@ function getImportance(entry: HeadConfig[number]) {
 		// The default favicon should be below any extra icons that the user may have set
 		// because if several icons are equally appropriate, the last one is used and we
 		// want to use the SVG icon when supported.
-		if (entry.tag === 'link' && 'rel' in entry.attrs && entry.attrs.rel === 'shortcut icon') {
+		if (
+			entry.tag === 'link' &&
+			'rel' in entry.attrs &&
+			entry.attrs.rel === 'icon' &&
+			entry.attrs['data-favicon'] === 'default'
+		) {
 			return 70;
 		}
 		return 80;


packages/starlight/__tests__/basics/head.test.ts
```diff
@@ -91,9 +91,10 @@ describe('createHead', () => {
 		const defaultFavicon = {
 			tag: 'link',
 			attrs: {
-				rel: 'shortcut icon',
+				rel: 'icon',
 				href: '/favicon.svg',
 				type: 'image/svg+xml',
+				'data-favicon': 'default',
```
delucis, line 97: Thank you for the PR @techfg!

Could you explain why the `data-favicon` attribute was added in this PR? It’s not mentioned in the PR description and I don’t believe it’s a web standard.

SnowDingo (commented):
Thank you for your PR @techfg !!

I looked over your code and it looks wonderful. Just one quick thing -(same as @delucis's comment but) I'm curious about what the `data-favicon` is doing.

Best Regards,
SnowDingo

techfg:
> Thank you for your PR @techfg !!
> 
> I looked over your code and it looks wonderful. Just one quick thing -(same as @delucis's comment but) I'm curious about what the `data-favicon` is doing.
> 
> Best Regards, SnowDingo

@SnowDingo - Please see the reply I left to @delucis's question for an answer to this.

delucis:
> Please see the reply I left to @delucis's question for an answer to this.

I don’t see any reply, but maybe I’m looking in the wrong place? I see this:
GitHub issue comment by delucis.

techfg:
@delucis - I think I figured it out, [see this](https://github.com/orgs/community/discussions/10369), didn't realize that you had to "submit" comments, don't recall having to do that before.  Anyway, I've submitted so hopefully you see them now.  Will submit comments on #2645 as well.

packages/starlight/__tests__/basics/head.test.ts
delucis: Ahhh, understood. OK.

Reading the comment in the head processing code, it says:

> The default favicon should be below any extra icons that the user may have set because if several icons are equally appropriate, the last one is used and we want to use the SVG icon when supported.

The current heuristic of relying on `shortcut icon` honestly seems not a great idea (I’m looking at you past self 👀). I wonder if we can just short based on the extension? So instead of needing the extra data attribute, we make sure an SVG favicon is listed last?

techfg: Agreed on the prior approach not being ideal (e.g., there could be multiple with `shortcut icon` to begin with), was going to mention but wasn't really relevant 😃.

The concern I'd have with detecting SVG directly is that there may not even be an SVG and even if there is, there may be multiple (e.g., config contains custom `head` attributes for different sizes).

I guess the question comes down to what the comment you referenced was really intending to do - I can think of two ways to interpret:

1. It was attempting to ensure that the `favicon` defined in the Starlight configuration property was always last (regardless of its type - e.g., svg, png, ico, jpg, etc.)
2. It was attempting to ensure that any SVG (regardless of whether it was the Starlight configuration property favicon or any other `head` configuration `link`) was last

It appears that [head config](https://github.com/withastro/starlight/commit/c6c1b6727140a76c42c661f406000cc6e9b175de) was added to Starlight prior to [favicon](https://github.com/withastro/starlight/commit/31b8a5aed2bca363c1b05c683b020e596b70bf4a) which provides some context to the comment, but it still is a little ambiguous in terms of what it was intending to do unless I'm just overthinking the comment 😄.

In doing some research on how user agents/browsers determine which browser to use, [Wikipedia](https://en.wikipedia.org/wiki/Favicon#:~:text=Firefox%20and%20Safari%20will%20use,%C3%9716%2C%20otherwise%20the%20ICO.) seemed to be provide some information for anything definitive.  Unfortunately, every browser is a little different assuming the information there is accurate and at the end of the day, user agents/browsers are free to change how they choose a favicon in the future.  

[This article](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs) seems to be the de facto guide on favicons but there is no way at build/runtime for Starlight to know what was added to `head` configuration regarding favicon reliably (or at least without doing a lot of inspection of `rel` & `type` attributes) and it could really be anything with any number of meta tags.

All the above said, do you recall what you were ultimately intending to do with the `favicon` config property and sort weight?  Depending on that, I can think of a few options:

1. Remove the check entirely and just leave at `80`
2. Detect any `link` that is for an `svg` and weight at `70`
3. Detect only the `favicon` config property if it's an `svg` and weight at `70`
4. Detect only the `favicon` config property regardless of type and weight at `70` (this is what the prior code kind of did and what the code in the PR does currently)

One other note is that there is no way to "disable" starlight from generating the favicon link (unless `Head` is overridden) since it seems [merge](https://github.com/withastro/starlight/blob/main/packages/starlight/utils/head.ts#L63) would not detect them as the same (see [stackblitz](https://stackblitz.com/edit/withastro-starlight-kugfxe?file=astro.config.mjs) repro) so there is no way for a user to completely control sort order short of overriding `Head`.  Possibly `favicon` config should support a value of `false` to disable it but that's a separate feature request and may not be necessary.  

In meantime, since the `head` config doesn't provide a way to control sort order of meta tags I think minimizing what is sorted may be best so possibly option 1 should be followed.  If the goal is to control sorting of favicon at some level, I think option 4 may be best since it's basically saying Starlights `favicon` config property will always come last.

Sorry for the long winded comment, this one isn't as straightforward as it may first appear, I think it all comes down to seeing if you can recall why you conditioned the sort weight originally and then deciding from there how to best adjust/handle 😄 

Lemme know your thoughts, thanks!

delucis:
Thanks again for this PR @techfg! As it introduces a few issues we haven’t had time to resolve and the presence of `shortcut` isn’t actually harmful, I think we’ll close this and are revisiting some aspects of how `<head>` management works in #2927 anyway, so a likely future solution will look a bit different.

PR #2650, opened 2024-12-02
Move `@types/js-yaml` to normal dependencies

raviqqe:
#### Description

This PR moves the `@types/js-yaml` package into a normal dependency in the `@astrojs/starlight` package.

Because we export TypeScript files directly in the `exports` field in a `package.json` file of the `@astrojs/starlight` package, we need to include all dependencies' `@types/*` packages in normal dependencies. Otherwise, we get following errors on type check of, for example, `astro.config.ts` even with `skipLibCheck: true` in `tsconfig.json`:

```
node_modules/@astrojs/starlight/utils/translations-fs.ts:4:18 - error TS7016: Could not find a declaration file for module 'js-yaml'. '/Users/raviqqe/src/github.com/raviqqe/stak/doc/node_modules/js-yaml/dist/js-yaml.mjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/js-yaml` if it exists or add a new declaration (.d.ts) file containing `declare module 'js-yaml';`

4 import yaml from 'js-yaml';
```

### DIFF
--- a/.changeset/serious-badgers-build.md
+++ b/.changeset/serious-badgers-build.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Include `@types/js-yaml` package in non-dev dependencies
+Moves `@types/js-yaml` package to non-dev dependencies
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thank you @raviqqe!

PR #2657, opened 2024-12-06
docs: use single listener in icons reference

HiDeoo:
#### Description

As discussed [here](https://github.com/withastro/starlight/pull/2505#discussion_r1816760391), this PR refactors the icons list in the icons reference to use a single listener for all icons instead of one per button.


delucis (approved):
Thanks for remembering to clean this up! 🧼

PR #2661, opened 2024-12-07
Enlarge header icons

illright:
#### Description

Related to https://github.com/withastro/starlight/discussions/1701, but not explicitly proposed there. The original proposal suggests to increase the space between the social icons, but I believe the icons themselves should be larger too.

This PR changes the appearance of social icons and the icons in language/theme selects to be larger. It also increases the touch area of the social icons to align it with the touch area of selects (~44.5px on desktop, ~48px on mobile). Finally, it marginally increases the space between the icon and the label, just something that's been bugging me visually a little bit.

<details><summary>Screenshots</summary>

**Desktop (light)**

<table><tr><th>Before</th><th>After</th></tr><tr><td>
Scherm­afbeelding 2024-12-07 om 19 22 40

</td><td>
Scherm­afbeelding 2024-12-07 om 19 23 17

</td></tr></table>

**Desktop (dark)**

<table><tr><th>Before</th><th>After</th></tr><tr><td>
Scherm­afbeelding 2024-12-07 om 19 22 46

</td><td>
Scherm­afbeelding 2024-12-07 om 19 23 23

</td></tr></table>

**Mobile (light)**

<table><tr><th>Before</th><th>After</th></tr><tr><td>
Scherm­afbeelding 2024-12-07 om 19 21 17
</td><td>
Scherm­afbeelding 2024-12-07 om 19 16 59
</td></tr></table>

**Mobile (dark)**

<table><tr><th>Before</th><th>After</th></tr><tr><td>
Scherm­afbeelding 2024-12-07 om 19 21 22
</td><td>
Scherm­afbeelding 2024-12-07 om 19 17 12
</td></tr></table>

</details>

### DIFF
--- a/.changeset/rich-toys-drive.md
+++ b/.changeset/rich-toys-drive.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Enlarge header icons
--- a/packages/starlight/components/MobileMenuFooter.astro
+++ b/packages/starlight/components/MobileMenuFooter.astro
@@ -18,7 +18,6 @@ import type { Props } from '../props';
 		margin-inline-end: auto;
 		gap: 1rem;
 		align-items: center;
-		padding-block: 1rem;
 	}
 	.social-icons:empty {
 		display: none;
--- a/packages/starlight/components/Select.astro
+++ b/packages/starlight/components/Select.astro
@@ -29,7 +29,7 @@ interface Props {
 
 <style>
 	label {
-		--sl-label-icon-size: 0.875rem;
+		--sl-label-icon-size: 1.2rem;
 		--sl-caret-size: 1.25rem;
 		--sl-inline-padding: 0.5rem;
 		position: relative;
@@ -63,7 +63,7 @@ interface Props {
 	select {
 		border: 0;
 		padding-block: 0.625rem;
-		padding-inline: calc(var(--sl-label-icon-size) + var(--sl-inline-padding) + 0.25rem)
+		padding-inline: calc(var(--sl-label-icon-size) + var(--sl-inline-padding) + 0.5rem)
 			calc(var(--sl-caret-size) + var(--sl-inline-padding) + 0.25rem);
 		margin-inline: calc(var(--sl-inline-padding) * -1);
 		width: calc(var(--sl-select-width) + var(--sl-inline-padding) * 2);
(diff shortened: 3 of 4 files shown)


SnowDingo (approved):
Thank you @illright for your contribution!

I took a look at the changes you've made and it seems like everything is working properly and as it should. I think your change made the Starlight docs more accessible.
So I will approve the PR.

Snowdingo

delucis (commented):
Thanks again for the PR @illright! Sharing some quick thoughts:

This has some definite positive sides (mainly of interest to me is making the clickable size a bit larger for accessibility), but I’m not so confident on the increased size and spacing overall.

A couple of points:

- Because the icon size increase also requires the text to increase in the theme and language pickers, it has reduced available space inside those. This is probably not an obvious detail, but one issue we have with these menus is how they accommodate different strings. For example, in English the theme picker uses a set of relatively similar length strings (light/dark/auto) but in other languages this can vary quite a bit, e.g. Spanish which ranges from the short “Claro” to the long “Automático”.

  Currently, these menus are of a fixed width which is also why they contain in some cases (e.g. English) extra whitespace between the selected value and the caret on the right. The fixed width has the advantage of visual stability when selected values change (UI to the left of the menu does not need to shift horizontally to accommodate longer values) as well as predictability as you don’t need to click through all values to see if your layout works for all the different possibilities. But it has the drawback that we need to select a single “compromise” value for the menu width that will work for as many cases as possible without becoming too stretched for shorter values.
  
  This is not a new constraint in this PR, just noting that if scaling up the icons and text, we would _also_ need to scale up the full field including whitespace to match.

- I think the larger social links more or less work in the case of Starlight’s own docs (although personally I find the spacing a little wide), but I suspect the size + spacing increase could look less nice for sites using more links. It’s fairly common to have 4 or 5 of these, which starts to take up a lot of space at this size.

Here’s an example that maybe demonstrates the two issues above:

| | |
|---|---|
| Before | image |
| After | image |

Given this, I’m not sure what would be best here. Perhaps scoping things just to the clickable size of the icons with minimal visual changes would be an option? Curious to hear if this extra information triggers any new thoughts for you!

delucis:
Thanks again for the PR @illright — as we haven’t resolved the issues mentioned above I’ll close this for now. If necessary, we can discuss further here or in #1701.

illright:
Yeah, that's fair :D Sorry for asking for feedback and then going MIA like that — I've been experiencing a drop in motivation and kept putting off the response. I'll get back to it once I have some more energy

PR #2664, opened 2024-12-12
Publish provenance for public packages

HiDeoo:
#### Description

While working in a workflow that involves strict security requirements, I noticed that Starlight public packages do not include any [provenance](https://github.blog/security/supply-chain-security/introducing-npm-package-provenance/). This PR fixes that by adding a `publishConfig` block to public packages.

- [Astro PR](https://github.com/withastro/astro/pull/8737) doing the same 
- [Astro PR](https://github.com/withastro/astro/pull/8749) fixing permissions forgotten in the first PR

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ jobs:
     name: Release
     if: ${{ github.repository_owner == 'withastro' }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -10,6 +10,9 @@ jobs:
     name: Release
     if: ${{ github.repository_owner == 'withastro' }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
```
delucis, line 15: Hmm, looks like this still isn’t the right permissions combo in our case. The workflow failed after merging: https://github.com/withastro/starlight/actions/runs/12316634406/job/34377330918

Might need write access to create the release PR?

delucis: Ah there was https://github.com/withastro/astro/pull/8752 which updated to `contents: write`. Not much additional context in there though 😁

delucis: Made a fix PR: https://github.com/withastro/starlight/pull/2665

PR #2672, opened 2024-12-14
i18n(ko-KR): update `using-components.mdx`

jsparkdev:
#### Description

- update `using-components.mdx`
- #2612


HiDeoo (approved):

PR #2676, opened 2024-12-14
fix: update links in `i18n.mdx`

jsparkdev:
#### Description

- update wrong links in `i18n.mdx`


HiDeoo (approved):
Great catch, thanks for fixing these links :raised_hands:

PR #2684, opened 2024-12-14
i18n(ko-KR): update `overrides.md`

jsparkdev:
#### Description

- update `overrides.md`
- #2612


delucis (approved):
Thanks for the update @jsparkdev! 🌟

PR #2685, opened 2024-12-14
docs: deprecated `Astro.props.slug` → `Astro.props.id`

julien-deramond:
#### Description

While migrating some projects to Astro v5 (+ Starlight), building the website showed warning messages that `Astro.props.slug` is deprecated (still works right now).

If I'm not wrong, `Astro.props.slug` can be safely replaced by `Astro.props.entry.slug`.

This PR applies this change in Starlight documentation in the `docs/src/content/docs/guides/overriding-components.mdx` override example.

**Note**: I wasn't sure whether or not I needed to modify also the translated files. Please tell me if I need also to modify the other files.

This PR doesn't contain any changesets as it's a documentation modification only.


docs/src/content/docs/guides/overriding-components.mdx
```diff
@@ -141,7 +141,7 @@ In the following example, a component overriding Starlight's [`Footer`](/referen
 import type { Props } from '@astrojs/starlight/props';
 import Default from '@astrojs/starlight/components/Footer.astro';
 
-const isHomepage = Astro.props.slug === '';
+const isHomepage = Astro.props.entry.slug === '';
```
HiDeoo, line 144:
```suggestion
const isHomepage = Astro.props.id === '';
```

Good catch, totally missed this reference in the docs.

> If I'm not wrong, `Astro.props.slug` can be safely replaced by `Astro.props.entry.slug`.

The deprecation path is actually to [update collections](https://github.com/withastro/starlight/releases/tag/@astrojs/starlight@0.30.0) to use Astro's new Content Layer API with a loader and in this case, you would use [`Astro.props.id`](https://starlight.astro.build/reference/overrides/#id). As [described](https://docs.astro.build/en/guides/upgrade-to/v5/#updating-existing-collections) in the Astro v5 upgrade guide, Content layer collections do not have a reserved `slug` field. Instead, all updated collections will have an `id`. Starlight currently still provides a `slug` field for backwards compatibility, but it'll be removed in Starlight v1 where the [`legacy.collections`](https://docs.astro.build/en/reference/legacy-flags/) flag will no longer be supported and only the `id` field will be available.

HiDeoo (changes requested):
Thanks for the PR :raised_hands:

I left a suggestion to the code change but great catch.

 > I wasn't sure whether or not I needed to modify also the translated files. Please tell me if I need also to modify the other files.

 This is perfect, we usually avoid when possible to modify the translated files as updating them could mark them as updated even though they're not.

docs/src/content/docs/guides/overriding-components.mdx
julien-deramond: Initially, the solution didn't work, likely due to a compilation error in my project. After carefully retesting, the method indeed functions correctly. Thanks HiDeoo for the detailed explanation and references, let's proceed with committing this suggestion.

HiDeoo (approved):
Amazing, thanks for the quick update :star2:

PR #2687, opened 2024-12-14
feat(header): enhance frosted glass effect on landing page #2686

markteekman:
Add gradient mask to header backdrop for a more realistic frosted glass effect that fades towards the bottom.

#### Description

- Closes #2686 (discussion)
- Enhances the frosted glass effect on the landing page header by:
-- Adding a gradient mask to the backdrop
-- Making the effect fade out towards the bottom
-- Keeping the effect scoped to landing page only
-- Based on and credits to: https://www.joshwcomeau.com/css/backdrop-filter/

This creates a more natural and modern look while maintaining good performance.

Before:
Scherm­afbeelding 2024-12-14 om 10 19 32

After:
Scherm­afbeelding 2024-12-14 om 10 19 38

Before:
Scherm­afbeelding 2024-12-14 om 10 20 42

After:
Scherm­afbeelding 2024-12-14 om 10 20 58

### DIFF
--- a/docs/src/assets/landing.css
+++ b/docs/src/assets/landing.css
@@ -18,8 +18,14 @@
 [data-has-hero] header {
 	border-bottom: 1px solid transparent;
 	background-color: transparent;
-	-webkit-backdrop-filter: blur(16px);
+}
+
+[data-has-hero] .header-backdrop {
 	backdrop-filter: blur(16px);
+	-webkit-backdrop-filter: blur(16px);
+	mask-image: linear-gradient(to bottom, black 0% 50%, transparent 50% 100%);
+	-webkit-mask-image: linear-gradient(to bottom, black 0% 50%, transparent 50% 100%);
+	z-index: -1;
 }
 
 [data-has-hero] .hero > img {
--- a/packages/starlight/components/Header.astro
+++ b/packages/starlight/components/Header.astro
@@ -29,6 +29,7 @@ const shouldRenderSearch =
 		<ThemeSelect {...Astro.props} />
 		<LanguageSelect {...Astro.props} />
 	</div>
+	<div class="header-backdrop"></div>
 </div>
 
 <style>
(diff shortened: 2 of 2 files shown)


delucis (commented):
Thanks for the PR and the cool idea @markteekman! I think this might be slightly unfortunate timing because although I love the idea in principle, I find this clashes quite a bit with the landing page banner we added yesterday — it feels weird to me to have that permanent blur from the banner as the initial state. (I’ll also note that in the current implementation the extra backdrop element prevents elements behind it from being clickable, so breaks the banner link.)

If we didn’t have banners and this effect kicked in on scroll, I think it would work nicely, but as-is, I’m not quite convinced.

markteekman:
@delucis ah jep, totally forgot to throw the `overflow-y: clip;`on there, my bad 😅 But yeah I agree, I wasn't sold on what it does to the banner as well. It's a nice touch for the rest of the website when you scroll but the banner kinda settles it for this one.

Considering adding the effect on scroll, that would take a small snippet to make that work. I've recorded an example of how it looks:

https://github.com/user-attachments/assets/2bc3b9c5-6ad1-4fea-a9bf-6f15028b4744

For this to work I would have to add this snippet to `Header.astro`:

```js
<script>
	const root = document.documentElement;
	const scrollThreshold = 10;

	window.addEventListener('scroll', () => {
		if (window.scrollY > scrollThreshold) {
			root.setAttribute('data-scrolled', '');
		} else {
			root.removeAttribute('data-scrolled');
		}
	});
</script>
```

Although I'm not a fan of adding JS (however small) to add only a small visual enhancement. It looks cool now though, I like how the effect kicks in when you scroll. So I'll leave it up to you what to do with it 🙂

delucis:
Thanks — yeah, I agree JS is probably overkill for this 😀 

Thanks again for the exploration — I’d noticed this “issue” with blurred overlays before but hadn’t thought of this solution, so I learned something new even if we can’t use it in our case.

PR #2688, opened 2024-12-14
Fix autogenerated sidebar issue

HiDeoo:
#### Description

As [reported](https://discord.com/channels/830184174198718474/1317414583492808747) on Discord, this PR fixes an issue with autogenerated sidebar when using Starlight with Astro's new Content Layer API where nested group names would be sluggified.

I made a mistake two times when porting the `treeify()` and `getBreadcrumbs()` functions to Astro v5, the first time when I implemented it and the second time when Chris asked me clarifications about the changes. I convinced myself that these functions were dealing with slugs instead of paths (legacy ID), which is not the case and makes no sense in retrospect.

This PR reverts the changes I made to these functions and now, when using a loader, before sorting and treeifying the routes, it computes a localized file path relative to the collection's root. This mimics the behavior of the legacy ID. I also added a test with an autogenerated sidebar for a folder containing uppercase letters and spaces to make sure it works as expected as we didn't have one before.


delucis (approved):
Thanks for the quick fix! Not sure I even realised we supported this, but glad it’s working for people and you could fix it so quickly 🙌

bomberstudios:
@HiDeoo the real MVP. Thanks a lot for the quick fix! 🧡

PR #2702, opened 2024-12-16
Fix issue with user-defined autogenerated sidebar groups

HiDeoo:
#### Description

This PR is very similar to #2688 but applies the same fix to user-defined autogenerated sidebar group whereas the previous PR only fixed the issue for the default autogenerated sidebar group.

For user-defined autogenerated sidebar groups, before calling `treeify()`, we filter routes but that filtering was always based on IDs. Similar to #2688, this PR computes an equivalent of legacy IDs before filtering routes when using a loader.


delucis (approved):
Thanks for the fix @HiDeoo!

PR #2705, opened 2024-12-16
i18n(de): update German translation

trueberryless:
#### Description

- #2676
- #2612
- #2685 
- #2671

PR #2707, opened 2024-12-16
Fix missing import in i18n collection docs

delucis:
#### Description

- While reviewing #2705 I noticed we accidentally removed this import in #2612


sarah11918 (approved):
Great catch! Thank you, translators! 💜

Issue #2693, opened 2024-12-14
Accessibility audit with several findings (serious and best practise)


markteekman:
### What version of `starlight` are you using?

Latest

### What version of `astro` are you using?

Latest

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome and Firefox

### Describe the Bug

# Accessibility audit

As promised a while ago here: https://github.com/withastro/starlight/issues/1950#issuecomment-2156391347 I spent a couple of hours to do a thorough audit of the Astro Starlight theme accessibility wise. The theme is really solid already,  so this list consist mainly of findings to further improve the accessibility:

## ⚠️ Serious

- Elements must meet minimum color contrast ratio thresholds.
  - Found on: https://starlight.astro.build/guides/authoring-content/ (in dark mode).
  - Comments in examples in dark mode just mis a tiny bit to meet the minimal requirement of 4.5:1. Although I would suggest to aim a little higher, 6:1 for example.
  - Screenshot:
![Scherm­afbeelding 2024-12-14 om 12 03 04](https://github.com/user-attachments/assets/333512ea-3512-4724-95f8-5e9d2aa2045e)
- Heading structures should follow the numerical flow (h1 > h2 > h3 etc.)
  - Found on: https://starlight.astro.build/getting-started/
  - There is an `h2` before the `h1` on the page ("On this page" in the sidebar)
  - Screenshot:
![Scherm­afbeelding 2024-12-14 om 12 35 43](https://github.com/user-attachments/assets/6e4829e3-5217-4b8f-b26a-474391020b13)
- Interactive elements should have at least one or more visual indicators (besides color).
  - Found on: https://starlight.astro.build/.
  - The buttons and link in the hero have no hover/focus effect, consider for example:
    - Changing the underline on hover/focus.
    - Changing te background color on hover/focus.
    - Animating the icons on hover/focus.
    - Screenshot:
![Scherm­afbeelding 2024-12-14 om 12 47 32](https://github.com/user-attachments/assets/870f90ad-3c3e-43bd-9d62-11b52c83aa58)
- Links opening in a new tab should announce it to the screen reader.
  - Found on: https://starlight.astro.build/.
[...]

PR #2708, opened 2024-12-16
Fixes code example colour contrast

delucis:
#### Description

- Addresses a colour contrast issue reported in #2693
- Expressive Code performs colour contrast correction automatically, but we had misconfigured a background colour setting, so Starlight’s default theme was not being properly taken into account in these calculations.
- This PR fixes that and ensures code blocks have sufficient text contrast.
- Shout out to @hippotastic for helping with the solution.


hippotastic:
Just a heads up: The code block styles on this page seem to be broken for me on iOS Safari in the deploy preview, not sure why that is the case:

https://deploy-preview-2708--astro-starlight.netlify.app/environmental-impact/#caching

The live site looks good. I can't imagine this one-line change causing this, so there might be something else going on.

delucis:
> Just a heads up: The code block styles on this page seem to be broken for me on iOS Safari in the deploy preview, not sure why that is the case:
> 
> [deploy-preview-2708--astro-starlight.netlify.app/environmental-impact#caching](https://deploy-preview-2708--astro-starlight.netlify.app/environmental-impact/#caching)

Oh that’s very odd… and I can reproduce the exact same behaviour in the dev server too. It appears that the Expressive Code styles are not loaded at all on that page.

delucis:
Figured out this was a caching issue due to how the Astro 5 content collection implementation caches Markdown across builds. Rebuilding on Netlify with a clear cache fixed it.

This does make this change a bit riskier than it appears. Users might need to `astro sync --force` or manually remove the cache (e.g. via UI in deployment environments) to avoid issues when upgrading.

hippotastic (approved):
LGTM! I checked the contrast in various pages, including tricky areas like marked lines including comments that had issues in the live site, and now the contrast levels are accessible everywhere.

PR #2713, opened 2024-12-17
i18n(ko-KR): update `i18n.mdx`

jsparkdev:
#### Description
- update `i18n.mdx`
- #2707


HiDeoo (approved):
Thanks for the quick update :raised_hands:

PR #2717, opened 2024-12-17
Fix list item spacing bug on Firefox

delucis:
#### Description

- Firefox renders margins applied to `<br>` unlike other browsers.

- We had a style which unintentionally applied `margin-bottom` to `<br>` inside list items, which would cause odd spacing in lists only in Firefox. For example, in the [Astro docs](https://docs.astro.build/en/reference/error-reference/):

   unordered list with irregular spacing of items

- This PR fixes this by excluding `<br>` from that style, and also adds `<code>` to avoid the same issue impacting list items with a last child that is an inline code element.

### DIFF
--- a/.changeset/proud-apricots-warn.md
+++ b/.changeset/proud-apricots-warn.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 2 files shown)


HiDeoo (approved):
Ah, it's been a while since that issue came out 😅 Tested in Firefox and this is indeed fixed.

PR #2724, opened 2024-12-18
Add farcaster to social links

dionysuzx:
#### Description

This PR should allow `'farcaster'` to be used as an option in the social icons schema. This is so that starlight users can add a social link to their farcaster site (eg. warpcast) like so:

```javascript
social: {
    ...
    farcaster: 'https://warpcast.com/@someone`
}
```

Do I need to add a changeset..?

### DIFF
--- a/.changeset/real-pugs-fold.md
+++ b/.changeset/real-pugs-fold.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Adds social link support for Farcaster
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thanks @dionysuzx! I believe this was the intention when someone first added the icon, but I guess we forgot to add the `social` schema part.

We’ll need a changeset so this is reflected in our changelog, but I can add one now before merging.

PR #2726, opened 2024-12-18
Adds icon for phone

techfg:
#### Description

- Closes #2725 
- What does this PR change? Adds social link icon for phone
    - Icon Source: https://iconscout.com/unicons/free-line-icons/phone
    - Processed through SVGOMG

image


techfg:
@delucis - 

Thanks for the review and makes sense regarding social.  I hadn't planned on including it in social since it's really not a "social" thing but noticed `email` was in social so felt `phone` was similar and made sense.  PR updated to remove :)

delucis (approved):
Thanks! 🙌 

And yeah, `email` suffers from the same issue to be honest — I don’t think we’d fully thought through that policy at the time, so it snuck in. “E-mail” is a bit more common in various languages, but it still isn’t ideal for many. Guess that ship has sailed though 😅

PR #2729, opened 2024-12-18
docs: update `starlight-links-validator`

HiDeoo:
#### Description

This PR updates the `starlight-links-validator` plugin to the version [`0.14.0`](https://github.com/HiDeoo/starlight-links-validator/releases/tag/starlight-links-validator%400.14.0) which adds support for Astro v5 and Starlight v0.30.0.

- The temporary patch introduced in #2612 is removed as it is no longer needed.
- The `--force` flag is removed from the `linkcheck` script as the plugin now automatically does this.


delucis (approved):
Thanks for following up on this and getting the plugin updated so quickly @HiDeoo! 🚀

PR #2736, opened 2024-12-20
Update minimum supported Astro version to 5.1

delucis:
#### Description

- Updates our development versions of Astro to 5.1.1
- Updates the minimum supported of Astro to 5.1.0
- This bump is needed for both #2708 and #157


HiDeoo:
Small note that the dev version car not be updated until https://github.com/withastro/astro/issues/12652 is fixed.

delucis:
> Small note that the dev version can not be updated until [withastro/astro#12652](https://github.com/withastro/astro/issues/12652) is fixed.

Ohhhh, you even said that already in https://github.com/withastro/starlight/pull/2612#issuecomment-2520667064

Damn, that makes this really tricky. We can’t really have a dev version lower than our minimum supported version I don’t think?

lilnasy:
> Small note that the dev version can not be updated until [withastro/astro#12652](https://github.com/withastro/astro/issues/12652) is fixed.

It can if you manually set process.env.NODE_ENV to development before running the failing test. It corrects astro's wrong guess that the environment is production, and it correctly looks for the development data store instead of the production one.

lilnasy:
> This is what we do in the Astro tests:

The default of data store retrieval is to assume production, so wouldn't implementing the linked workaround this arrive at the same failure?

I imagine the variable would need to be definitely set to development, not just deleted, which is what worked for me.

delucis:
> I imagine the variable would need to be definitely set to development, not just deleted, which is what worked for me.

Not sure exactly why, but it does look like [deleting it](https://github.com/withastro/starlight/pull/2736/commits/f8fc1d92b3137c43638c37946b601002b51ad773) fixes our E2E tests in this context.

packages/starlight/__e2e__/test-utils.ts
```diff
@@ -25,6 +25,8 @@ export function testFactory(fixturePath: string) {
 		} = {}
 	): Promise<Server> {
 		const { mode } = options;
+		// Reset NODE_ENV so it can be re-set by `dev()` or `build()`.
+		delete process.env.NODE_ENV;
```
HiDeoo, line 29: Maybe this should also contains a todo comment to remove this once fixed? Would hate to miss an issue in the long run because we're doing this in an E2E test but that's not something users do.

ascorbic:
Install `astro@experimental--data-store-location`

delucis:
Thanks @ascorbic! The experimental release does seem to fix things. You can see it running successfully a branch off this PR here: #2776

packages/starlight/__e2e__/test-utils.ts
HiDeoo: Considering that Matt created an amazing PR fixing it, I guess this comment is irrelevant now 😄

delucis: Astro moves fast!

PR #2738, opened 2024-12-24
feat: add details animation styles

kevinzunigacuellar:
#### Description

- This PR adds details element expand and retract animations.
- Caution: `interpolate-size` is an experimental feature.
- `allow-discrete` is recent baseline since August, 2024 [Can I use?](https://caniuse.com/?search=allow-discrete)

### DIFF
--- a/packages/starlight/components/Page.astro
+++ b/packages/starlight/components/Page.astro
@@ -5,6 +5,7 @@ import type { Props } from '../props';
 import '../style/props.css';
 import '../style/reset.css';
 import '../style/shiki.css';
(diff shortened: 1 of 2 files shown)


delucis (commented):
These look so smooth @kevinzunigacuellar — thanks for sharing! I think the lower availability `allow-discrete` might be OK given this is a progressive enhancement and we could place these styles behind an `@supports` query. However, I don’t think we can depend on `interpolate-size` just yet until it is standard.

Love the exploration though and appreciate you sharing this code 💖

PR #2749, opened 2025-01-03
Expose navigation functions via the internal module.

brian-montgomery:
#### Description

    Currently, if a developer wants to utilize the `sidebar` config structure to do something custom on a page utilizing the `<StarlightPage>` component, there is no way to access that generated config.

    The `starlight/utils/navigation` module exposes two functions that the `<Sidebar>` component uses: `getSidebar` and `getSidebarFromConfig`. However, those functions are not exposed publicly to developers. This PR simply exposes those functions via the existing `@astrojs/starlight/internal` module.

    No, this is just a developer enhancement for niche scenarios that want to do something custom on a

### DIFF
--- a/packages/starlight/internal.ts
+++ b/packages/starlight/internal.ts
@@ -1,6 +1,7 @@
 /**
- * @file This file contains utility functions imported by the `<Code>` component.
+ * @file This file contains utility functions imported by the `<Code>` and `<Sidebar>` components.
  */
 
 export { useTranslations } from './utils/translations';
 export { getStarlightEcConfigPreprocessor } from './integrations/expressive-code';
+export { getSidebar, getSidebarFromConfig } from './utils/navigation';


HiDeoo:
Thanks for your contribution 🙌 

The `@astrojs/starlight/internal` only exposes some internal helpers that are used by Starlight itself or some of its direct dependencies. Exporting new functions that do not match this criteria would break this contract as they are not meant to be used by the end user. That's also why this module is not documented and if we need to expose new functions for the end user, this is definitely something that should be documented.

At the moment, customization in Starlight is done using [component overrides](https://starlight.astro.build/guides/overriding-components/) and soon through [a new API](https://github.com/withastro/starlight/pull/2390) based on middlewares. Your description of the issue you're facing is a bit vague so I think the best approach would be to first start with a [new discussion](https://github.com/withastro/starlight/discussions/new/choose) to make sure what you're trying to achieve is not already possible, and if it's not, we can discuss and think how to make it possible in the future, and if it means exposing new functions, these would definitely be documented and not exposed through that module.

Considering all of this, I'm going to close this PR for now. Feel free to open a new discussion so we can discuss further and I hope you understand the reasons behind this decision. Thanks again for your contribution!

PR #2751, opened 2025-01-05
docs: fix invalid import in `ComponentProps` example

HiDeoo:
#### Description

As [reported](https://discord.com/channels/830184174198718474/1070481941863878697/1325375830402273323) on Discord, the `ComponentProps` example contains an invalid import statement.

This PR fixes the import statement.

PR #2756, opened 2025-01-07
i18n(ko-KR): update `using-components.mdx`

jsparkdev:
#### Description

- update `using-components.mdx`
- #2751


HiDeoo (approved):
Thanks for the quick update 🙌

PR #2757, opened 2025-01-07
Fix translation issue for languages with a region subtag

HiDeoo:
#### Description

- Closes https://github.com/withastro/docs/issues/10630

This PR fixes an issue for languages with a region subtag (e.g. `zh-CN`) where custom UI strings are not properly translated. On top of the linked issue, this issue is also visible in https://starlight.astro.build/zh-cn/components/asides/ where the `component.preview` label shows "Preview" instead of "预览" even tho [a translation exists](https://github.com/withastro/starlight/blob/main/docs/src/content/i18n/zh-CN.json) ([same page in the deploy preview link](https://deploy-preview-2757--astro-starlight.netlify.app/zh-cn/components/asides/) to compare and see the fix).

The issue is related to the fact that the `loadTranslations()` function loading user translations was not properly updated for Astro v5 changes regarding the `id` property, which means our `userCollections` dictionary would contain an `zh-cn` key instead of `zh-CN`.

This PR fixes the issue by using the new `filePath` property instead of the old `id` property when not using legacy collections.

We had tests supposed to cover this, unfortunately, like I missed this spot during the Astro v5 migration, I also missed updating our mocking helper for dictionaries (`mockDict()`) like I did for documents (`mockDoc()`) so everything was passing even tho the issue was present.

### DIFF
--- a/packages/starlight/virtual.d.ts
+++ b/packages/starlight/virtual.d.ts
@@ -8,6 +8,8 @@ declare module 'virtual:starlight/plugin-translations' {
 	export default PluginTranslations;
 }
 
+// TODO: Move back to `virtual-internal.d.ts` when possible. For example, when dropping support for
+// legacy collections, `utils/translations.ts` would no longer need to import project context.
 declare module 'virtual:starlight/project-context' {
 	const ProjectContext: {
 		root: string;
--- a/packages/starlight/virtual.d.ts
+++ b/packages/starlight/virtual.d.ts
@@ -7,3 +7,16 @@ declare module 'virtual:starlight/plugin-translations' {
 	const PluginTranslations: import('./utils/plugins').PluginTranslations;
 	export default PluginTranslations;
 }
+
+declare module 'virtual:starlight/project-context' {
+	const ProjectContext: {
+		root: string;
+		srcDir: string;
+		trailingSlash: import('astro').AstroConfig['trailingSlash'];
+		build: {
+			format: import('astro').AstroConfig['build']['format'];
+		};
+		legacyCollections: boolean;
+	};
+	export default ProjectContext;
+}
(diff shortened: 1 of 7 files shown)


packages/starlight/virtual.d.ts
```diff
@@ -7,3 +7,16 @@ declare module 'virtual:starlight/plugin-translations' {
...
+		root: string;
+		srcDir: string;
+		trailingSlash: import('astro').AstroConfig['trailingSlash'];
+		build: {
+			format: import('astro').AstroConfig['build']['format'];
+		};
+		legacyCollections: boolean;
+	};
+	export default ProjectContext;
+}
```
HiDeoo, line 22: This move from `virtual-internal.d.ts` → `virtual.d.ts` is required due to the new import of `virtual:starlight/project-context` in `packages/starlight/utils/translations.ts` which would now fail for users [running `tsc --noEmit`](https://github.com/withastro/starlight/blob/main/docs/package.json#L12) on their project.

delucis (approved):
LGTM! Thanks for the quick fix. Hopefully we can drop legacy collection support in the near-ish future to avoid the extra virtual module exposure.

packages/starlight/virtual.d.ts
```diff
@@ -7,3 +7,16 @@ declare module 'virtual:starlight/plugin-translations' {
 	const PluginTranslations: import('./utils/plugins').PluginTranslations;
 	export default PluginTranslations;
 }
+
+declare module 'virtual:starlight/project-context' {
```
delucis, line 11: Maybe we even add a note to that effect?

```suggestion
// TODO: Move back to `virtual-internal.d.ts` when possible. For example, when dropping support for
// legacy collections, `utils/translations.ts` would no longer need to import project context.
declare module 'virtual:starlight/project-context' {
```

HiDeoo: Good call, I guess this would also highly change when we get back to bundling but it's an easy comment making sure we don't forget 👍

PR #2759, opened 2025-01-07
i18n(de): update `using-components.mdx`

trueberryless:
#### Description

- #2751


HiDeoo (approved):
Thanks for the quick update 🙌

PR #2762, opened 2025-01-07
Prevent title from being translated

HiDeoo:
#### Description

This PR adds the [`translate` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate) to the header site title to indicates that the element must not be translated.

> Although not all browsers recognize this attribute, it is respected by automatic translation systems such as Google Translate, and may also be respected by tools used by human translators. As such it's important that web authors use this attribute to mark content that should not be translated.

For example, commenting the `logo` configuration option in the Astro config of the Starlight Docs and asking Google Translate to translate the English version of the website to French will result in the following:

SCR-20250107-ovfr

After this PR, the header site title will no longer be translated:

SCR-20250107-ovia

### DIFF
--- a/.changeset/smart-moons-repair.md
+++ b/.changeset/smart-moons-repair.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Prevents the header title from being translated by automatic translation systems.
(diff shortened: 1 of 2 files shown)


delucis (approved):
Interesting! I hadn’t encountered this attribute yet.

At first I wondered if this might be undesirable in some cases if you had a title that was not a brand name — if it was a descriptive phrase like “Pancake Recipes” or something, maybe someone would like to have that auto-translated to “Recettes de crêpes”. That said, I looked through our site showcase and thought a bit about Starlight users, and I think this would only very rarely cause issues most likely, so I think it could be a reasonable change.

PR #2778, opened 2025-01-10
Step up and Badge on

trueberryless:
#### Description

Personally - having some experience with translating the docs - I think that creating `.mdx` files immediately, instead of renaming later (because _often_ md files end up having to be mdx files because of components) has the benefit for translators, that the changes are better traceable. That's the first reason for this PR.

The second reason is more relatable for the maintainers probably: I think that the new route-data guide should use the Step component instead of having just the numbered list...

I also found a reasonable place in the route-data reference for a component: the `slug` could have a badge label somewhere. Here I think the change to a `.mdx` file is not that necessary because the Badge doesn't need to be there, further discussion maybe?

### DIFF
--- a/docs/src/content/docs/guides/route-data.mdx
+++ b/docs/src/content/docs/guides/route-data.mdx
@@ -3,6 +3,8 @@ title: Route Data
 description: Learn how Starlight’s page data model is used to render your pages and how you can customize it.
 ---
 
+import { Steps } from '@astrojs/starlight/components';
+
 When Starlight renders a page in your documentation, it first creates a route data object to represent what is on that page.
 This guide explains how route data is generated, how it is used, and how you can customize it to modify Starlight’s default behavior.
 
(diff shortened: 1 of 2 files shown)


delucis (commented):
Thanks @trueberryless! Great idea with the steps component.

I think I will remove the badge for now (but OK with keeping the file MDX). I don’t love how it looks (by inheriting the bold font weight from the heading) and how it calls attention to the property, but I appreciate the idea!

trueberryless:
> I think I will remove the badge for now (but OK with keeping the file MDX). I don’t love how it looks (by inheriting the bold font weight from the heading) and how it calls attention to the property, but I appreciate the idea!

Yeah, I also wasn't sure where to put it and if it even makes sense. I'm glad you removed it and I really appreciate leaving it as MDX tho, this may help a lot ❤️

PR #2779, opened 2025-01-10
i18n(de): Update German translation with #2390

trueberryless:
#### Description

- Update German translation
- #2390


trueberryless:
Currently, I just want to see if the link checks pass... And I just realised, I could have done that locally...

PR #2782, opened 2025-01-11
Fix outdated links to Expressive Code documentation

delucis:
#### Description

- In several places in our documentation we were still linking to Expressive Code README documents on GitHub that predated the full https://expressive-code.com docs site (and mostly no longer contain the same content as it moved to the website).
- This PR fixes those links to point to relevant parts of the EC documentation site instead.

### DIFF
--- a/.changeset/tender-tools-kick.md
+++ b/.changeset/tender-tools-kick.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a documentation link in the JSDoc comment for the `StarlightExpressiveCodeOptions` type
--- a/docs/src/content/docs/guides/authoring-content.mdx
+++ b/docs/src/content/docs/guides/authoring-content.mdx
@@ -220,13 +220,13 @@ var fun = function lang(l) {
 
 ### Expressive Code features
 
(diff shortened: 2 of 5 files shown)


HiDeoo (approved):
Excellent idea. Just went through all of them and they all make sense to me :raised_hands:

Should we also update https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/expressive-code/index.ts#L23 to point to https://expressive-code.com/guides/themes/ while at it?

trueberryless (approved):
LGTM 🎉 

The `<Code>` component implementation is also left untouched to link to GitHub, perfect!
Every link to a specific heading also makes sense, I really like the differenciation between the [styling configuration](https://expressive-code.com/key-features/text-markers/#configuration) and the [configuration reference](https://expressive-code.com/reference/configuration/).

delucis:
> Should we also update [`main`/packages/starlight/integrations/expressive-code/index.ts#L23](https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/expressive-code/index.ts?rgh-link-date=2025-01-12T10%3A29%3A27Z#L23) to point to [expressive-code.com/guides/themes](https://expressive-code.com/guides/themes/) while at it?

Great call — I’d only searched in the docs, so missed this one. Updated in 3955cfe.

PR #2800, opened 2025-01-13
Added starlight-to-pdf tool to Plugins and Integrations page

Linkerin:
#### Description

The PR adds a link to the [starlight-to-pdf](https://github.com/Linkerin/starlight-to-pdf) cli tool for converting Starlight into PDF into the [Plugins and Integrations](https://starlight.astro.build/resources/plugins/) page, `Community tools and integrations` section.

A solution to the following discussion: [Export docs to PDF #964](https://github.com/withastro/starlight/discussions/964).

PR #2802, opened 2025-01-13
i18n(de): update `plugins.mdx`

trueberryless:
#### Description

- update `plugins.mdx`
- #2800

### DIFF
--- a/docs/src/content/docs/de/resources/plugins.mdx
+++ b/docs/src/content/docs/de/resources/plugins.mdx
@@ -187,6 +187,6 @@ Diese Community-Tools und Integrationen können verwendet werden, um deiner Star
 	<LinkCard
 		href="https://github.com/Linkerin/starlight-to-pdf"
 		title="starlight-to-pdf"
-		description="Ein CLI-Werkzeug zur Konvertierung von Starlight-Websites in PDF-Dateien."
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/de/resources/plugins.mdx
```diff
@@ -184,4 +184,9 @@ Diese Community-Tools und Integrationen können verwendet werden, um deiner Star
 		title="@hideoo/starlight-plugin"
 		description="Ein Generator, um schnell Starlight-Plugins zu erstellen."
 	/>
+	<LinkCard
+		href="https://github.com/Linkerin/starlight-to-pdf"
+		title="starlight-to-pdf"
+		description="Ein CLI-Werkzeug zur Konvertierung von Starlight-Websites in PDF-Dateien."
```
delucis, line 190: In the Astro docs glossary for German (https://github.com/withastro/docs/blob/main/i18n-guides/deutsch.md), “CLI” is translated as “Kommandozeile”, so I guess the consistent translation here would be:

```suggestion
		description="Ein Kommandozeilenwerkzeug zur Konvertierung von Starlight-Websites in PDF-Dateien."
```

What do you think?

delucis (commented):
Thanks @trueberryless! Just had one query.

delucis (approved):
Thanks again @trueberryless 🌟

PR #2814, opened 2025-01-18
i18n(fr): update `reference/icons`, `components/icons` & `components/using-components`

HiDeoo:
#### Description

This PR updates the French version of the `reference/icons`, `components/icons`, and `components/using-components` pages with the changes from #2805.

Decided to group these changes together as they all relate to the same PR and are fairly small changes.


jsparkdev:
As a non-native speaker, I have verified the content using a translator, and all translations appear to be consistent.

LGTM!

PR #2822, opened 2025-01-22
Add clientOptionsModule option to DocSearch plugin

KianNH:
#### Description

- Closes https://github.com/withastro/starlight/discussions/2806
- Adds a `clientOptionsModule` option to the `starlightDocSearch` plugin.
  - This takes the shape of a path to a JavaScript or TypeScript file which will be imported (if present) and merged with existing options in the `starlight/docsearch-config` virtual module.
  - If an option exists in both, i.e `appId`, the option in the `clientOptionsModule` file will be used.

#### Considerations

- Updating documentation

https://github.com/withastro/starlight/blob/66a4131c63ec8ff425c6f04c1c496abbdfddc945/docs/src/content/docs/guides/site-search.mdx?plain=1#L111-L118

- Tests

Whilst this is probably not a large enough package to warrant it, I did setup try out setting up Vitest for this package where `vitePluginDocSearch` is passed in `vitest.config.ts` but since it would need a lot of work to setup a harness to test `starlightDocSearch` I opted against including it.

### DIFF
--- a/packages/docsearch/index.ts
+++ b/packages/docsearch/index.ts
@@ -49,29 +49,34 @@ const DocSearchConfigSchema = z
 		z
 			.object({
 				/**
-				 * A JavaScript module containing a default export of options
-				 * which are passed to the DocSearch library.
+				 * The path to a JavaScript or TypeScript file containing a default export of options to
+				 * pass to the DocSearch client.
 				 *
-				 * It can be used to configure options that are not serializable,
-				 * such as `transformSearchClient` or `resultsFooterComponent`.
+				 * The value can be a path to a local JS/TS file relative to the root of your project,
+				 * e.g. `'/src/docsearch.js'`, or an npm module specifier for a package you installed,
+				 * e.g. `'@company/docsearch-config'`.
 				 *
-				 * Supports local JS files relative to the root of your project,
-				 * e.g. `'/src/docsearch.js'`, and JS you installed as an npm
-				 * module, e.g. `'@company/docsearch-config'`.
+				 * Use `clientOptionsModule` when you need to configure options that are not serializable,
+				 * such as `transformSearchClient()` or `resultsFooterComponent()`.
+				 *
+				 * When `clientOptionsModule` is set, all options must be set via the module file. Other
+				 * inline options passed to the plugin in `astro.config.mjs` will be ignored.
 				 *
 				 * @see https://docsearch.algolia.com/docs/api
+				 *
 				 * @example
 				 * // astro.config.mjs
-				 * starlightDocSearch({
-				 *   // ...
-				 *   clientOptionsModule: "./src/config/docsearch.ts",
-				 * })
+				 * // ...
+				 * starlightDocSearch({ clientOptionsModule: './src/config/docsearch.ts' }),
+				 * // ...
 				 *
 				 * // src/config/docsearch.ts
 				 * import type { DocSearchClientOptions } from '@astrojs/starlight-docsearch';
 				 *
 				 * export default {
-				 *   // ...
+				 *   appId: '...',
+				 *   apiKey: '...',
+				 *   indexName: '...',
 				 *   getMissingResultsUrl({ query }) {
 				 *     return `https://github.com/algolia/docsearch/issues/new?title=${query}`;
 				 *   },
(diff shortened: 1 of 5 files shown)


packages/docsearch/index.ts
```diff
@@ -36,6 +38,38 @@ const DocSearchConfigSchema = z
...
 		 */
 		searchParameters: z.custom<SearchOptions>(),
+		/**
+		 * A JavaScript module containing a default export. These options
+		 * are merged with other configuration options and provided to
+		 * the DocSearch library.
+		 *
+		 * In the event that a configuration option exists in both,
+		 * for example in the plugin options and the module, the
+		 * value in the module will be used.
```
HiDeoo, line 48: I wonder if this behavior could be confusing? Would we need to support some kind of deep-merging for some values?

This makes me wonder if making only one of the approach usable at the same time could be a good alternative, e.g. either `starlightDocSearch({ appID: 'XYZ' })` or `starlightDocSearch('./docsearch.config.ts')` 🤔 

Or maybe it should be the options that we currently have, or an object with only the new option (currently `clientOptionsModule`) but not a mix of both and we error if it's the case?

packages/docsearch/index.ts
```diff
@@ -36,6 +38,38 @@ const DocSearchConfigSchema = z
...
+		 *
+		 * It can be used to configure options that are not serializable,
+		 * such as `transformSearchClient` or `resultsFooterComponent`.
+		 *
+		 * Supports local JS files relative to the root of your project,
+		 * e.g. `'/src/docsearch.js'`, and JS you installed as an npm
```
HiDeoo, line 54: Did you mean?

```suggestion
		 * e.g. `'./src/docsearch.js'`, and JS you installed as an npm
```

packages/docsearch/index.ts
```diff
@@ -36,6 +38,38 @@ const DocSearchConfigSchema = z
 		 * @see https://www.algolia.com/doc/api-reference/search-api-parameters/
 		 */
 		searchParameters: z.custom<SearchOptions>(),
+		/**
+		 * A JavaScript module containing a default export. These options
```
HiDeoo, line 42: I think this should specify what the default export should be.

HiDeoo (commented):
Not a complete review or request for some changes, but I left a few comments mostly regarding the JSDoc to start some discussion around the approach.

One other question coming to mind is if we should expose a `defineDocSearchConfig` helper just for the typings or something similar?

packages/docsearch/index.ts
delucis: Ah, I hadn’t considered this. It would also avoid needing to merge stuff — the virtual module would simply re-export the module if present or the inline config if not. Might be cleaner I guess and easier to document even?

> If you need to use options X, Y, Z, move the configuration to a separate file…

Are there any circumstances where it might make sense to define some stuff in `astro.config.mjs` even if you have a dedicated config module?

Not a strong opinion but probably a slightly preference for something like `{ clientOptionsModule: "./path" }` over a string argument just so we have room if there are ever _non_ client options that make sense with/without a client options module.

delucis (commented):
Looking good @KianNH! Seconding @HiDeoo’s idea to make the option module all-or-nothing, so if you use that, you move _all_ config there, unless there’s a reason not to do that anyone can see.

I also agree we should export something to help with typing the options module. Given this module will end up in client-side code, I might suggest just a type rather than a `defineClientOptions()` utility just to avoid the extra no-op function in the bundle. (Would also allow us to export it directly from `@astrojs/starlight-docsearch` rather than needing a dedicated file just for that.) That would help us assert that e.g. you shouldn’t set the `container` property which Starlight takes care of.

So something like:

```ts
import type docsearch from '@docsearch/js';

export type DocSearchClientOptions = Omit<Parameters<typeof docsearch>[0], 'container'>;
```

And then users can write config as:

```ts
import type { DocSearchClientOptions } from '@astrojs/starlight-docsearch';

export default { 
  // ...
} satisfies DocSearchClientOptions;
```

delucis (approved):
Thanks again for this @KianNH! I took the liberty of adding some docs and I’ll get it released 🙌

PR #2827, opened 2025-01-24
docs: add more community plugins resources

adrian-ub:
add [starlight-theme-black](https://github.com/adrian-ub/starlight-theme-black)

PR #2830, opened 2025-01-28
i18n(fr): Update `resources/plugins.mdx` from #2827

thomasbnt:
#### Description
Update `resources/plugins.mdx` from #2827


ArmandPhilippot (approved):
LGTM!

PR #2836, opened 2025-01-29
i18n(fr): update `guides/site-search`

HiDeoo:
#### Description

This PR updates the French version of the `guides/site-search` page with the changes from #2822.

### DIFF
--- a/docs/src/content/docs/fr/guides/site-search.mdx
+++ b/docs/src/content/docs/fr/guides/site-search.mdx
@@ -1,6 +1,8 @@
 ---
 title: Recherche
 description: Découvrez les fonctionnalités de recherche intégrées à Starlight et comment les personnaliser.
+tableOfContents:
+  maxHeadingLevel: 4
 ---
 
 import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/site-search.mdx
```diff
@@ -110,13 +112,62 @@ Avec cette configuration mise à jour, la barre de recherche de votre site ouvri
 
 #### Configuration de DocSearch
 
-Le module d’extension DocSearch de Starlight supporte également la personnalisation du composant DocSearch avec les options supplémentaires suivantes :
+Le module d’extension DocSearch de Starlight supporte la personnalisation du composant DocSearch avec les options supplémentaires suivantes spécifiées dans la configuration du module d’extension :
```
HiDeoo, line 115: We don't really have a real translation for "inline" and "en ligne" feels weird: https://french.stackexchange.com/questions/54639/what-is-the-translation-of-inline-in-french

delucis (approved):
LGTM!

And I think your approach to “inline” makes sense — it’s equivalent to the « dans le corps du … » examples you linked and seems clear to me.

PR #2841, opened 2025-01-29
i18n(de): update `guides/site-search.mdx`

trueberryless:
#### Description

- update `guides/site-search.mdx`
- #2822

### DIFF
--- a/docs/src/content/docs/de/guides/site-search.mdx
+++ b/docs/src/content/docs/de/guides/site-search.mdx
@@ -1,6 +1,8 @@
 ---
 title: Suche
 description: Finde heraus, welche Suchfunktionen in Starlight integriert sind und wie du sie anpassen kannst.
+tableOfContents:
+  maxHeadingLevel: 4
 ---
 
 import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/de/guides/site-search.mdx
```diff
@@ -94,9 +96,9 @@ Wenn du Zugang zum [Algolia DocSearch-Programm](https://docsearch.algolia.com/)
    			title: 'Website mit DocSearch',
    			plugins: [
    				starlightDocSearch({
-   					appId: 'YOUR_APP_ID',
-   					apiKey: 'YOUR_SEARCH_API_KEY',
-   					indexName: 'YOUR_INDEX_NAME',
+   					appId: 'DEINE_APP_ID',
```
trueberryless, line 99: I translated the var names here because I intuitively translated them below too...
Pls let me know if it's better to keep them in English or translate them to German in both scenarios!

trueberryless: (6 out of 15 languages currently translate them into the specific language, the others leave them in English...)
Maybe we should globally specify how to handle it because I think that currently I only know that it should be translated from the specific [GERMAN.md in Astro](https://github.com/withastro/docs/blob/56990e2206a68c42da808f5c1d771f69e13548fb/i18n-guides/deutsch.md?plain=1#L65)

delucis: Ah yeah, different languages seem to do this differently and I tend to leave it up to translator discretion. In general I would encourage translating example values like this, but I guess it may vary a bit language-to-language what is best.

Anyway this looks great!

delucis (approved):
Sieht super aus — danke @trueberryless! 🥳

PR #2843, opened 2025-01-29
Removed broken links from showcase

alvinometric:
These were 404s for me


kevinzunigacuellar (commented):
Thanks for the PR! I also get 404 from these links. If we are removing them from the showcase, we should also remove their images. Can you include them?

docs/src/components/showcase-sites.astro
```diff
@@ -5,11 +5,7 @@ import FluidGrid from './fluid-grid.astro';
 
 <FluidGrid>
 	<Card title="Athena OS" href="https://athenaos.org" thumbnail="www.athenaos.org.png" />
-	<Card
-		title="PubIndexAPI Docs"
-		href="https://docs.pubindexapi.com/"
-		thumbnail="docs.pubindexapi.com.png"
-	/>
+
```
trueberryless, line 8:
```suggestion
```

Like e.g. here...

alvinometric:
Good call @kevinzunigacuellar ! I didn't even see the `.png` at the end, I thought these were URLs

HiDeoo (approved):
Thanks for the contribution, great call and thanks everyone for the reviews :raised_hands:

Looks perfect to me :rocket:

PR #2848, opened 2025-01-31
Fix some Pagefind UI issues

### LINKED ISSUE #2840: CSS variable `--sl-color-accent-light` in Search component is not defined

0.31.1

5.1.10

pnpm

Windows

Chrome

### Describe the Bug

Go to https://starlight.astro.build/ , press F12, and then find the `#starlight__search > div > form` element, scroll down the Styles, and see the bug.

Image

> Maybe this is a typo in https://github.com/withastro/starlight/blob/86afe959885349c9de710338ee8b05c7032f09dd/packages/starlight/components/Search.astro#L259

https://starlight.astro.build/

### Participation

HiDeoo:
#### Description

- Closes #2840

This PR fixes a few UI issues related to the Pagefind UI.

Note that I'm not super happy with the changesets, I think I didn't manage to get the right balance between what is changed and the level of detail in the changeset, maybe even 1 changeset could be enough? Happy to iterate on this based on feedback.

##### Unknown CSS variable

This issue was the first one reported in #2840. This PR removes the unknown CSS variable (`--pagefind-ui-primary`) and now uses the desired one (`--sl-color-text`) which was already the color used when the browser was falling back to the default color when the variable was not defined.

The `--pagefind-ui-primary` variable is used in 2 places in Pagefind, the first one being the "Load more results" button:

|        | Dark Theme | Light Theme |
| ------ | ---------- | ----------- |
| Before | SCR-20250131-jxqm   | SCR-20250131-jxti  |
| After  | SCR-20250131-nlol   | SCR-20250131-nlpp  |

The second place is the checkbox background color when using [filters](https://pagefind.app/docs/filtering/) where some extra CSS was added to fix the tick color:

|        | Dark Theme | Light Theme |
| ------ | ---------- | ----------- |
| Before | SCR-20250131-nqvc   | SCR-20250131-nqwc   |
| After  | SCR-20250131-nrhn   | SCR-20250131-nrip   |

_Ultimately, some changes could be made upstream to avoid this issue and rely on accent colors for the checkbox and should be beneficial for all Pagefind users as it's also an issue with a default Pagefind setup._

##### Metadata

When using the Pagefind [metadata feature](https://pagefind.app/docs/metadata/), the UI was broken. I tried to use the same UI as inline code at the moment for the tags.

|        | Dark Theme | Light Theme |
| ------ | ---------- | ----------- |
| Before | SCR-20250131-ohaq   | SCR-20250131-ohbk   |
| After  | SCR-20250131-oogb   | SCR-20250131-oohr   |

This PR improves the UI but as usual when it comes to UI, that's not my strong suit so happy to iterate on this based on feedback/suggestions.


delucis (commented):
I think this looks great @HiDeoo! I don’t have major feedback on the design — I think it looks pretty good (definitely better than before 😅) — and given this is a less used Pagefind feature, I’m happy to have it working nicely like this.

Re: the changesets, yeah I _guess_ these could be condensed into a single one? Maybe even as simple as this?

```md
Fixes styling of filter and metadata elements in Pagefind search UI
```

The “load more” button hasn’t actually changed, so probably doesn’t need mentioning IMO.

But no notes from me on the code changes!

kevinzunigacuellar (approved):
Looks great! LGTM ✅ 🚀

PR #2858, opened 2025-02-04
Support multisite search on pagefind

XREvo:
#### Description

- Closes https://github.com/withastro/starlight/discussions/2856
- It add more configuration option to pagefind, by allowing [mergeIndex](https://pagefind.app/docs/multisite/)
- No visual changes had been made with this update.

This had been tested locally with 2 instances :
- one on port 3000 and the other on port 3001
- with this configuration  
``` ts
export default defineConfig({
	... ,
	integrations: [
		starlight({
			...,
			pagefind: {
				indexWeight: 2.0,
				mergeIndex: [{
					bundlePath: 'http://localhost:3001/pagefind/',
					indexWeight: 0.5,
				}]
			},
		}),
	],
});
```
- and network panel show that both sites where queried for search  
image

#### What's next

- [x] Completing the doc to let users known that they can use it (i'll do it in the the next days)


docs/src/content/docs/reference/configuration.mdx
```diff
@@ -435,7 +435,7 @@ When setting `themes`, you must provide at least one dark and one light theme fo
 
 #### `useStarlightUiThemeColors`
 
-**type:** `boolean`  
+**type:** <code>boolean | <a href="#pagefindoptions">PagefindOptions</a></code>
```
delucis, line 438: This appears to have been changed in error?
```suggestion
**type:** `boolean`
```

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -447,7 +447,7 @@ When using custom themes and setting this to `true`, you must provide at least o
 
 ### `pagefind`
 
-**type:** <code>boolean | <a href="#pagefindoptions">PagefindOptions</a></code>  
+**type:** <code>boolean | <a href="#pagefindoptions">PagefindOptions</a></code>
```
delucis, line 450: The double space at the end of this line is meaningful and is interpreted as a line break in Markdown. You can see this in the docs preview where now the two lines collapse together:

type and default properties running together onto one line

We should add back the spaces:

```suggestion
**type:** <code>boolean | <a href="#pagefindoptions">PagefindOptions</a></code>  
```

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -457,19 +457,37 @@ This will also hide the default search UI if in use.
 
 Pagefind cannot be enabled when the [`prerender`](#prerender) option is set to `false`.
 
-Set `pagefind` to an object to configure the Pagefind search client.
-See [“Customize Pagefind's result ranking”](https://pagefind.app/docs/ranking/) in the Pagefind documentation for more details about using the `pagefind.ranking` option to control how search result ranking is calculated.
+Set `pagefind` to an object to configure the Pagefind search client :
```
delucis, line 460: Ah, there’s no space before colons in English 😄 

```suggestion
Set `pagefind` to an object to configure the Pagefind search client:
```

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -457,19 +457,37 @@ This will also hide the default search UI if in use.
...
 	};
+	mergeIndex?: [{
+		bundlePath: string;
+		indexWeigth?: number;
+		basePath?: string;
+		baseUrl?: string;
+		excerptLength?: number;
+		mergeFilter?: any;
+		highlightParam?: string;
+		language?: string;
```
delucis, line 483: I’m not 100% sure about exposing `excerptLength` and `highlightParam` here given we don’t expose them for the current site.

It feels like `excerptLenght` should match the defaults, otherwise you could have discrepancies between e.g. the excerpt length of a result from another site and the excerpt length of a result from the current site.

`highlightParam` _potentially_ makes more sense if you had this customized differently across sites using Pagefind, but in the context of Starlight it’s not even possible currently to configure highlights, so it seems odd to add the option.

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,118 @@
...
+    /**
+     * Multiply all rankings for this index by the given weight.
+     */
+	indexWeight: indexWeightSchema,
+    /**
+     * Merge this filter object into all search queries in this index.
+     *
+     * Only applies in multisite setups.
+     */
+	mergeFilter: z.any().nullish(),
```
delucis, line 63: Could we provide a better type/schema for this? What shape does `mergeFilter` expect?

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,118 @@
...
+		 *
+		 * @see https://pagefind.app/docs/ranking/#configuring-term-similarity
+		 */
+		termSimilarity: z.number().min(0).default(9),
+	});
+const pagefindIndexOptionsSchema = z.object({
+    /** 
+	 * Overrides the URL path that Pagefind uses to load its search bundle
+	 * */
+	basePath: z.string().nullish(),
```
delucis, line 43: General comment on this and the other schema entries here: elsewhere we used Zod’s `.optional()` method rather than `.nullish()`, so I think we should stick to the same here — there’s no real need to support setting `null` explicitly.

delucis (commented):
Thank you for the PR @XREvo! I’ve left a few comments.

One additional note: could you remove the changes to the non-English docs? In general we only change the English docs in implementation PRs and then allow our translators to update their individual languages to ensure they can fully update things correctly for their linguistic context.

packages/starlight/schemas/pagefind.ts
XREvo: The best I can do is to set it as `z.object().optional()` because the content of the object is dynamic.  
The property's name are dynamic and the content is either a string or an array of strings, which led to smething like that :   
```json
{
  my_property_1: "value",
  my_property_2: [ "value 1", "value 2" ]
}
```

But I don't find a way to configure the part where "property names are dynamic" in zod.

Does `z.object().optional()` instead of `z.any().optional()` would be enough for you?

XREvo: Well, my proposal doesn't work. Zod wait for an object schema to be specified when you use `z.object(...)`. It cannot manage object with an unknown schema, except for `any`.

So what do you prefer? Should I stay with `z.any()` but  add a detail about expected formalism in comment? Or simply delete the property?

docs/src/content/docs/reference/configuration.mdx
XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/dac389fe127b29c644a73fa0cc88ebe1ac273acc

docs/src/content/docs/reference/configuration.mdx
XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/6b4fa0ff41a11967cca25c110270ef3880c24025

docs/src/content/docs/reference/configuration.mdx
XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/080be3cd09b3a29ff7e4129e4443947baa019013 for the code and https://github.com/withastro/starlight/pull/2858/commits/c2b802f5629a440592bbfd0a3ad44f18d043f4dc for the doc

packages/starlight/schemas/pagefind.ts
XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/4c780934b0eea3cf4d5f909d76eccaa008c36426

XREvo:
I've reverted my modification of translations with https://github.com/withastro/starlight/pull/2858/commits/10d99675bc5ad762c7e1d0b480fe7688f3bd75d6

packages/starlight/schemas/pagefind.ts
HiDeoo: You may be looking for [records](https://github.com/colinhacks/zod#records) in this case, e.g. `z.record(z.string(), z.union([z.string(), z.string().array()]))` which would be an object where keys are string and values are either a string or an array of strings.

XREvo:
I let you resolve the conversations after review and i'm waiting for your feedback about the mergeFilter property.  
And thanks for the review 🙏

packages/starlight/schemas/pagefind.ts
XREvo: That's exactly what I was looking for 🤩  
Thanks @HiDeoo!

XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/7dd2fd76418ea911844b464dd122997da65b5793

packages/starlight/schemas/pagefind.ts
XREvo: Fixed with https://github.com/withastro/starlight/pull/2858/commits/4c780934b0eea3cf4d5f909d76eccaa008c36426

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+	});
+const pagefindIndexOptionsSchema = z.object({
+    /** 
+	 * Overrides the URL path that Pagefind uses to load its search bundle
+	 * */
+	basePath: z.string().optional(),
+    /**
+	 * Appends the given baseURL to all search results. May be a path, or a full domain
+	 * */
+	baseUrl: z.string().optional(),
```
delucis, line 47: Should probably clean up the closing lines of these comments:

```suggestion
    /** 
	 * Overrides the URL path that Pagefind uses to load its search bundle
	 */
	basePath: z.string().optional(),
    /**
	 * Appends the given baseURL to all search results. May be a path, or a full domain
	 */
	baseUrl: z.string().optional(),
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+    /**
+     * Multiply all rankings for this index by the given weight.
```
delucis, line 49: As in the above comments for the ranking weights, I think adding a link to the Pagefind docs could be handy where possible:

```suggestion
     * Multiply all rankings for this index by the given weight.
     *
     * @see https://pagefind.app/docs/multisite/#changing-the-weighting-of-individual-indexes
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+     */
+	indexWeight: indexWeightSchema,
+    /**
+     * Merge this filter object into all search queries in this index.
+     *
+     * Only applies in multisite setups.
+	 * 
+	 * Expected formalism is an object :
+	 * - with keys as the filter names 
+	 * - and values as the filter values, represented as a string or an array of strings.
```
delucis, line 59: Similarly here, I think rather than us going into detail documenting this, it’s OK to link directly to Pagefind’s docs:

```suggestion
	 * @see https://pagefind.app/docs/multisite/#filtering-results-by-index
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+     */
+	mergeFilter: z.record(
+		z.string(),
+		z.string().or(
+			z.array(z.string()).nonempty()
+		)
+	).optional(),
+    /**
+     * Language of this index.
```
delucis, line 68: Same here:

```suggestion
     * Language of this index.
     *
     * @see https://pagefind.app/docs/multisite/#merging-a-specific-language-index
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
-		.default({}),
+			pagefindIndexOptionsSchema.extend({
+				/**
+				 * Set Pagefind’s `bundlePath` mergeIndex option.
+				 * 
+				 * This option is not part of the `PagefindIndexOptions` type on pagefind, 
+				 * but it's presence is checked before using it on the pagefind_ui module.
+				 *
+				 * @see https://pagefind.app/docs/multisite/#searching-additional-sites-from-pagefind-ui
+				 * @see https://github.com/CloudCannon/pagefind/blob/v1.3.0/pagefind_ui/modular/modular-core.js#L192
```
delucis, line 109: And similarly this link feels not necessarily helpful to a user, I think I’d suggest removing it and leaving only the docs link:
```suggestion
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+
 const pagefindSchema = z.object({
+	/**
+	 * Set Pagefind’s `pageLength` ranking option.
+	 * 
+	 * Configure how search result in the current website are weighted by Pagefind
+	 * compared to other sit'e's merge indexes.
+	 *
+	 * @see https://pagefind.app/docs/multisite/#changing-the-weighting-of-individual-indexes
+	 */
```
delucis, line 85: I think the first line here was left in in error? And also suggesting a tweak to the rest of this comment for clarity:

```suggestion
	/**
	 * Configure how search results from the current website are weighted by Pagefind
	 * compared to results from other sites when using the `mergeIndex` option.
	 *
	 * @see https://pagefind.app/docs/multisite/#changing-the-weighting-of-individual-indexes
	 */
```

packages/starlight/schemas/pagefind.ts
```diff
@@ -1,43 +1,117 @@
...
+    /**
+	 * Appends the given baseURL to all search results. May be a path, or a full domain
+	 * */
+	baseUrl: z.string().optional(),
+    /**
+     * Multiply all rankings for this index by the given weight.
+     */
+	indexWeight: indexWeightSchema,
+    /**
+     * Merge this filter object into all search queries in this index.
```
delucis, line 53: Reading the Pagefind docs again I think this wording might be more accurate:

```suggestion
     * Apply this filter configuration to all search results from this index.
```

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -457,19 +457,35 @@ This will also hide the default search UI if in use.
...
 		termFrequency?: number;
 		termSaturation?: number;
 		termSimilarity?: number;
 	};
+	mergeIndex?: [{
+		bundlePath: string;
+		indexWeigth?: number;
+		basePath?: string;
+		baseUrl?: string;
+		mergeFilter?: any;
```
delucis, line 480: I guess we can update this type too, now that we landed on a valid schema:

```suggestion
		mergeFilter?: Record<string, string | string[]>;
```

delucis (commented):
Thanks for your patience on this one @XREvo! I think it looks good, and the `mergeFilter` schema seems correct to me.

I left a bunch of comments, but please don’t feel overwhelmed — they’re all tiny details to improve the documentation/comments 😁 

I’m also happy to handle applying these if you prefer.

XREvo:
Thanks @delucis for this careful review  
I've added all your changes, they were legit to my eyes.

delucis (approved):
Thanks again for the PR @XREvo! Will get this published as part of our next minor release 🙌

PR #2865, opened 2025-02-10
i18n(zh-cn): Update `authoring-content.mdx`

Nin3lee:
#### Description

- Update `authoring-content.mdx`
- #2782

### DIFF
--- a/docs/src/content/docs/zh-cn/guides/authoring-content.mdx
+++ b/docs/src/content/docs/zh-cn/guides/authoring-content.mdx
@@ -223,13 +223,13 @@ var fun = function lang(l) {
 
 ### Expressive Code 功能
 
-Starlight 使用 [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code) 来扩展代码块的格式化功能。
+Starlight 使用 [Expressive Code](https://expressive-code.com/) 来扩展代码块的格式化功能。
 Expressive Code 的文本标记和窗口外框插件是默认启用的。
 可以使用 Starlight 的 [`expressiveCode` 配置选项](/zh-cn/reference/configuration/#expressivecode) 来配置代码块的渲染。
 
 #### 文本标记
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/zh-cn/guides/authoring-content.mdx
```diff
@@ -223,24 +223,24 @@ var fun = function lang(l) {
 
 ### Expressive Code 功能
 
-Starlight 使用 [Expressive Code](https://github.com/expressive-code/expressive-code/tree/main/packages/astro-expressive-code) 来扩展代码块的格式化功能。
+Starlight 使用 [Expressive Code](https://expressive-code.com/) 来扩展代码块的格式化功能。
```
HiDeoo, line 226: Cannot comment this as it's out of review range, but right above this change, there is a code block with the following content:

> 长单行代码块不应换行。如果它们太长，它们应该水平滚动。这一行应该足够长长长长长长长长长长长长来证明这一点。

This does not exist in the English version, would you be up for removing it?

HiDeoo (changes requested):
Thanks for the update 🙌 

All updated links looks good to me, altho I noticed some unrelated content that I think should be removed from this translation file. Let me know if you're interested in removing it (could also be done in a follow-up PR).

docs/src/content/docs/zh-cn/guides/authoring-content.mdx
Nin3lee: Thanks for your observant review, HiDeoo! It's true that this sentence is not exist in the English version.

> 长单行代码块不应换行。如果它们太长，它们应该水平滚动。这一行应该足够长长长长长长长长长长长长来证明这一点。

It means that "the code block which is too long in single line should not be wrapped. If they are too long, you should let it scroll horizonally . And this line of code is loooooooooong enough to proof the conclusion."

This likely stems from a moment of inspiration the translator experienced during the translation process. In terms of the purpose of this line of text, it more closely resembles a coding convention used during programming. While removing this humorous code line might leave the translator with some regret in the moment, from the perspective of document structure, it's more appropriate to remove it to maintain consistency.

HiDeoo (approved):
Thanks for the explanation and update, looks perfect to me 👍

PR #2881, opened 2025-02-15
Adding all samples for badges in sidebar as example

mseibert:
#### Description

- With this PR all types of badges will be shown directly in the docs which is important to help people chose the right one for their use case.
- The current status quo is less verbose. But it puts the work on the reader to try out every badge type to see and test it. That is cumbersome, takes time and unnecessary.
- The visual appeal of the badges should be visible directly.

Deeplink: https://deploy-preview-2881--astro-starlight.netlify.app/guides/sidebar/#badge-variants-and-custom-styling

image


trueberryless:
Hello and welcome to the Starlight project!

Thank you for taking the time to create such a detailed Pull Request!

I'm not one of the maintainers, but I generally think that they want to keep the documentation as simple as possible. Looking at the differences between the current documentation and this request, I think that we can agree that your request takes up more space then the current docs:

| Current documentation      | Your request |
| ----------- | ----------- |
| image | image |

In my opinion this is not necessary because if the reader wants to look at all the different Badges, they can just go to the Badges page and under the Usage section, they can see all the Badges available: https://deploy-preview-2881--astro-starlight.netlify.app/components/badges/#usage

image

So personally, I think that it is not necessary to list all the Badge variantes on the Sidebar Navigation page, because there is already an exclusive page for all the props and things regarding Badges. Moreover, I really think that the documentation should not be too bloated and more subtle and discreet. I really like the section as it is (just my thoughts).

One last thing to notice:

> I think if you make a docs change in the Starlight repository, you don't need to include a changeset because the website is deployed by Netlify. As far as I know, you just need to include a changeset if a change somewhere inside the package folder is happening.

mseibert:
> In my opinion this is not necessary because if the reader wants to look at all the different Badges, they can just go to the Badges page and under the Usage section, they can see all the Badges available: https://deploy-preview-2881--astro-starlight.netlify.app/components/badges/#usage

Touché, I did not see this other page. Maybe we could agree that a link to this other section could be useful?

> See how to use all the different variants, different sizes, and custumizations in the ["Usage" section of the Badge component](https://starlight.astro.build/components/badges/#usage).

That would have helped me not come up with this suggestion here. Or even leaner:

> [Learn more about using and customizing badges](https://starlight.astro.build/components/badges/#usage).

What do you think?

trueberryless:
I agree that a link could not hurt anyone, I like both examples, maybe a combination of the things I like from both:

> Learn more about [using and customizing badges](https://starlight.astro.build/components/badges/#usage).

(Not the entire sentence is a link + its lean as you said)

mseibert:
> In the meantime you could adjust the changes, remove the changeset and edit the title and description if you want... 

Will do.

mseibert:
Back to start and added link at right point: 

image

mseibert:
> I think if you make a docs change in the Starlight repository, you don't need to include a changeset because the website is deployed by Netlify. As far as I know, you just need to include a changeset if a change somewhere inside the package folder is happening.

Erased changeset.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -390,6 +390,8 @@ The configuration above generates the following sidebar:
 	]}
 />
 
+Learn more about [using and customizing badges](https://starlight.astro.build/components/badges/#usage).
```
trueberryless, line 393: As far as I know, I think it should be possible to link to other pages without the domain name, e.g.:

```suggestion
Learn more about [using and customizing badges](/components/badges/#usage).
```

SnowDingo (approved):
Thank you @mseibert for your contribution!

I agree with your latest changes, and I think the addition of the link makes the documentation more concise and also user-friendly.
Because I am not a Starlight maintainer, I can't merge this but I would approve this PR.

Thank you!!
SnowDingo

trueberryless:
> @trueberryless Good catch. Sorry for missing that.

No problem, looks good to me now! 🎉

trueberryless (approved):

delucis (approved):
Thank you for the contribution @mseibert and thanks everyone for reviewing this — I think it ended up in a great spot 🙌

PR #2882, opened 2025-02-15
i18n(de): update German translation with #2578

trueberryless:
#### Description

- update all German doc files which were changed in #2578


trueberryless:
I think we need to merge so that the file `reference/plugins.md` is ignored (because this file also contains changes in the other PR merged today)

This should be contained in the commit message:

```
@lunaria-ignore:docs/src/content/docs/de/reference/plugins.md;src/content/docs/de/reference/plugins.md
```

(not necessary if this PR is merged as second -> probably some merge conflicts to solve then...)

delucis (approved):
LGTM! And no need for the Lunaria directive thanks to merging other translations first.

PR #2884, opened 2025-02-15
i18n(fr): add `guides/route-data` and `reference/route-data`

HiDeoo:
#### Description

This PR adds the French version of the pages `guides/route-data` and `reference/route-data`.


trueberryless (approved):
LGTM 🎉 

I don't know why my translation tool from French to English says that the French term "données de route" to "**road** data" and not "**route** data, but that's probably just the translation tool.

I reviewed this as non-native speaker!

docs/src/content/docs/fr/reference/route-data.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+**Type :** `{ minHeadingLevel: number; maxHeadingLevel: number; items: TocItem[] } | undefined`
+
+Table des matières de la page courante si celle-ci est activée.
+
+### `headings`
+
+**Type :** `{ depth: number; slug: string; text: string }[]`
+
+Un tableau de toutes les en-têtes Markdown extraites de la page courante.
```
ArmandPhilippot, line 138: I don't think `en-tête` has the same meaning as `titre`... IMO, [`en-tête`](https://www.larousse.fr/dictionnaires/francais/en-t%C3%AAte/29917) would correspond more to the frontmatter of a Markdown file so it could be confusing. But maybe it's just me. 😄 
```suggestion
Un tableau de toutes les titres Markdown extraits de la page courante.
```

ArmandPhilippot (approved):
Only a small nit, otherwise LGTM!

docs/src/content/docs/fr/reference/route-data.mdx
HiDeoo: Interesting, for me it's a bit of the opposite and why it's called heading and not title.

I know both are used in French, even in our own docs <https://starlight.astro.build/fr/guides/authoring-content/#en-t%C3%AAtes> but that's not a reference considering I may have been the one to write that part ^^

I think some references I found at the time were https://docs.github.com/fr/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax and https://docusaurus.io/fr/docs/next/markdown-features/toc both using "en-tête".

Altho, I'm pretty sure we can find as many references using "titre".

For consistency with the existing French docs, I'll ignore the suggestion right now, altho I'm totally open to discussing it further and if we decide to change it, make a pass on the entire French docs to make sure it's consistent everywhere in a dedicated PR.

PR #2886, opened 2025-02-16
i18n(fr): update `reference/plugins` & `guides/i18n`

HiDeoo:
#### Description

This PR updates the French version of the `reference/plugins` & `guides/i18n` pages with the changes from #2390 and #2578.


docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -114,6 +213,40 @@ export default {
...
+Une fonction appelée pour ajouter un [middleware de route](/fr/guides/route-data/) pour le site.
+
+La propriété `entrypoint` doit être un spécificateur de module pour le fichier middleware de votre module d'extension qui exporte une fonction `onRequest`.
+
+Dans l'exemple suivant, un module d'extension publié sous `@exemple/starlight-module-extension` ajoute un middleware de route utilisant un spécificateur de module npm :
+
+```js {6-9}
+// module-extension.ts
+export default {
+  name: '@exemple/starlight-module-extension',
```
trueberryless, line 229: I'm not native, so I'm not sure if `starlight-module-extension` is the same in English and French, but I'll leave this comment, maybe it has been overlooked to translate 👍 (probably not because `example` was translated)

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -163,70 +296,78 @@ L'exemple ci-dessus affichera un message qui inclut le message d'information fou
...
+Appelez `useTranslations()` avec une étiquette de langue BCP-47 pour générer une fonction utilitaire qui fournit l'accès aux chaînes d'interface utilisateur pour cette langue.
+`useTranslations()` retourne une fonction équivalente à l'API `Astro.locals.t()` disponible dans les composants Astro.
+Pour en savoir plus sur les APIs disponibles, consultez le [guide « Utiliser les traductions de l'interface utilisateur »](/fr/guides/i18n/#utiliser-les-traductions-de-linterface-utilisateur).
 
-```ts {6-13} /(injectTranslations)[^(]/
+```ts {6}
 // module-extension.ts
 export default {
-  name: 'plugin-avec-traductions',
+  name: 'plugin-use-translations',
```
trueberryless, line 310: same here, I think it is just a coincidence that it translates to the same (although `traductions` seems like a valid translation for `translations` too)... Probably best if the native speakers resolve these comments.

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -163,70 +296,78 @@ L'exemple ci-dessus affichera un message qui inclut le message d'information fou
...
+
+```js
+starlight({
+  title: 'Ma documentation',
+  defaultLocale: 'fr',
+  locales: {
+    // Documentation en anglais dans `src/content/docs/en/`
+    en: { label: 'English' },
+    // Documentation en français dans `src/content/docs/fr/`
+    fr: { label: 'Français', lang: 'fr' },
```
trueberryless, line 346: Since you changed the defaultLocale to French, we could think about removing the `lang` option here and define it in the `en` locale, so everything is kinda "flipped" (I noticed that you also changed the example below, where now the English UI-string is shown instead of the French one, like in the original docs)

But since the lang is an optional config, it's just a stylish recommendation...

trueberryless (approved):
Very good translations! I really like how you interchanged the `en` and `fr` locals at the end of plugin reference to demonstrate it exactly like the English version, but for all the French people ❤️ 

Left some thoughts about some parts where I - as a non-native - can't conclude whether something was overlooked or intentional.

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -41,12 +50,102 @@ Un module d'extension doit fournir un nom unique qui le décrit. Le nom est util
 
 ## `hooks`
 
-Les hooks sont des fonctions que Starlight appelle pour exécuter le code du module d'extension à des moments spécifiques. Actuellement, Starlight prend en charge un seul hook `setup`.
+Les hooks sont des fonctions que Starlight appelle pour exécuter le code du module d'extension à des moments spécifiques.
+
+Pour référencer le type des arguments d'un hook, utilisez le type utilitaire `HookParameters` et passez le nom du hook.
+Dans l'exemple suivant, le paramètre `options` est typé pour correspondre aux arguments passés au hook `config:setup` :
+
+In the following example, the `options` parameter is typed to match the arguments passed to the `config:setup` hook:
```
ArmandPhilippot, line 58: The sentence appears twice, translated and in English:
```suggestion
```

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -41,12 +50,102 @@ Un module d'extension doit fournir un nom unique qui le décrit. Le nom est util
...
+import type { HookParameters } from '@astrojs/starlight/types';
+
+function configSetup(options: HookParameters['config:setup']) {
+  options.useTranslations('fr');
+}
+```
+
+### `i18n:setup`
+
+Fontion de configuration de l'internationalisation du module d'extension appelée lorsque Starlight est initialisé.
```
ArmandPhilippot, line 70: typo:
```suggestion
Fonction de configuration de l'internationalisation du module d'extension appelée lorsque Starlight est initialisé.
```

docs/src/content/docs/fr/reference/plugins.md
ArmandPhilippot: The English version uses `@example/starlight-plugin` while `plugin` seems to be translated with `module d'extension` in the French version. So it should become `@exemple/starlight-module-d-extension` but with slugs or packages names, I think it's fine to omit the `d'`!

docs/src/content/docs/fr/reference/plugins.md
ArmandPhilippot: Probably an oversight, thanks trueberryless!

However it seems `plugin-use-translations` is used in multiple places across the file, so if we update the plugin name we might need to update the other occurrences as well.

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -163,70 +296,78 @@ L'exemple ci-dessus affichera un message qui inclut le message d'information fou
...
+    en: { label: 'English' },
+    // Documentation en français dans `src/content/docs/fr/`
+    fr: { label: 'Français', lang: 'fr' },
+  },
+});
+```
+
+Un module d'extension peut déterminer la langue d'un fichier en utilisant son chemin absolu :
+
+```ts {6-8} /fr/
```
ArmandPhilippot, line 353: Since you have update the path, we need to update the highlighting as well:
````suggestion
```ts {6-8} /en/
````

**Review by ArmandPhilippot — COMMENTED** [review]
Three little suggestions, otherwise LGTM!

File: docs/src/content/docs/fr/reference/plugins.md (hunk 2 above)
Comment (HiDeoo) [reply docs/src/content/docs/fr/reference/plugins.md:310]: This one is tricky (in my head ^^), I initially translated it, then thought the name was not about using translations but the `useTranslations()` function being documented in this section, so I partially translated it to `module-extension-use-translation`, ended up not liking it and reverting back to the English version.

😅

**Comment by HiDeoo** [comment]
Thanks for the great catches. Still not sure about this `use-translations` ^^ I'll sleep on it and decide something if someone doesn't have a strong opinion on it before.

File: docs/src/content/docs/fr/reference/plugins.md (hunk 2 above)
Comment (ArmandPhilippot) [reply docs/src/content/docs/fr/reference/plugins.md:308]: Oh, I didn't make the connection with `useTranslations`! 😆 For me, it was just to show the contrast with the other snippet l85: `plugin-with-translations`/`plugin-avec-traductions`. So, here I would use `plugin-use-translations`/`plugin-utilisant-traductions`. But, yeah, we could use `module-extension` to be consistent (l85 as well).

**Comment by trueberryless** [comment]
One thing which you could do about the `use-translations` topic:

I have reviewed the Korean and Russian version and written the German one, all have not translated any of the plugin names...
So you could also just not translate it I guess, but as you like, you know French better then I do 😅

File: docs/src/content/docs/fr/reference/plugins.md (hunk 2 above)
Comment (HiDeoo) [reply docs/src/content/docs/fr/reference/plugins.md]: Yeah, it was probably my brain seeing hidden things that weren't there like I said :sweat_smile: Went ahead and used your suggestion, thanks a lot :raised_hands:

File: docs/src/content/docs/fr/reference/plugins.md [hunk 7]
```diff
@@ -73,7 +170,7 @@ Dans l'exemple suivant, un nouveau compte de média [`social`](/fr/reference/con
 export default {
   name: 'ajout-twitter-plugin',
```
trueberryless, line 171: For consistency you should also not translate these if you chose to..

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -100,7 +197,7 @@ import react from '@astrojs/react';
 export default {
   name: 'plugin-utilisant-react',
```
trueberryless, line 198: and that

docs/src/content/docs/fr/reference/plugins.md
```diff
@@ -149,7 +280,7 @@ Tous les messages seront préfixés par le nom du module d'extension.
 export default {
   name: 'plugin-long-processus',
```
trueberryless, line 281: and one over here and so on...

HiDeoo:
> I have reviewed the Korean and Russian version and written the German one, all have not translated any of the plugin names...
> So you could also just not translate it I guess, but as you like, you know French better then I do 😅

Yeah, this is something up to translators and that has been consistently done in the French versions where it doesn't change a meaning or make it harder to understand. I personally think it adds a nice touch to the translation and shows that the translator has put some thought into it.

PR #2896, opened 2025-02-16
i18n(ko-KR): create `guides/route-data.mdx` and `reference/route-data.mdx`

jsparkdev:
#### Description

create `guides/route-data.mdx` and `reference/route-data.mdx`

### DIFF
--- a/docs/src/content/docs/ko/guides/route-data.mdx
+++ b/docs/src/content/docs/ko/guides/route-data.mdx
@@ -1,0 +1,137 @@
+---
+title: 경로 데이터
+description: Starlight의 페이지 데이터 모델이 페이지를 렌더링하는 데 어떻게 사용되는지, 그리고 이를 어떻게 사용자 정의할 수 있는지 알아보세요.
+---
+
+import { Steps } from '@astrojs/starlight/components';
+
+Starlight가 문서에서 페이지를 렌더링할 때, 먼저 해당 페이지의 내용을 나타내는 경로 데이터 객체를 생성합니다.
+이 가이드에서는 경로 데이터가 어떻게 생성되는지, 어떻게 사용하는지, 그리고 Starlight의 기본 동작을 수정하기 위해 어떻게 사용자 정의할 수 있는지 설명합니다.
+
+사용 가능한 속성의 전체 목록은 ["경로 데이터 참조"](/ko/reference/route-data/)를 확인하세요.
+
+## 경로 데이터란 무엇인가요?
+
+Starlight 경로 데이터는 단일 페이지를 렌더링하는 데 필요한 모든 정보를 담고 있는 객체입니다.
+여기에는 현재 페이지에 대한 정보뿐만 아니라 Starlight 구성에서 생성된 데이터도 포함됩니다.
+
+## 경로 데이터 사용하기
+
+Starlight의 모든 컴포넌트는 경로 데이터를 사용하여 각 페이지에 무엇을 렌더링할지 결정합니다. 예를 들어, [`siteTitle`](/ko/reference/route-data/#sitetitle) 문자열은 사이트 제목을 표시하는 데 사용되고, [`sidebar`](/ko/reference/route-data/#sidebar) 배열은 전역 사이드바 탐색을 렌더링하는 데 사용됩니다.
+
+Astro 컴포넌트에서 `Astro.locals.starlightRoute` 전역 변수를 통해 이 데이터에 접근할 수 있습니다.
+
+```astro title="example.astro" {2}
+---
+const { siteTitle } = Astro.locals.starlightRoute;
+---
+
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/ko/guides/route-data.mdx
```diff
@@ -0,0 +1,137 @@
...
+## 경로 데이터란 무엇인가요?
+
+Starlight 경로 데이터는 단일 페이지를 렌더링하는 데 필요한 모든 정보를 담고 있는 객체입니다.
+여기에는 현재 페이지에 대한 정보뿐만 아니라 Starlight 구성에서 생성된 데이터도 포함됩니다.
+
+## 경로 데이터 사용하기
+
+Starlight의 모든 컴포넌트는 경로 데이터를 사용하여 각 페이지에 무엇을 렌더링할지 결정합니다. 예를 들어, [`siteTitle`](/ko/reference/route-data/#sitetitle) 문자열은 사이트 제목을 표시하는 데 사용되고, [`sidebar`](/ko/reference/route-data/#sidebar) 배열은 전역 사이드바 탐색을 렌더링하는 데 사용됩니다.
+
+Astro 컴포넌트에서 `Astro.locals.starlightRoute` 전역 변수를 통해 이 데이터에 접근할 수 있습니다.
```
trueberryless, line 22: I think this is right, I just want to say, that my translation tool messes this part up really bad:

```
Astro Components `Astro.locals.starlightRoute`You can access this data through global variables.
```

docs/src/content/docs/ko/guides/route-data.mdx
```diff
@@ -0,0 +1,137 @@
...
+예를 들어, 특정 페이지의 사이드바 항목을 필터링하거나 제목을 사용자 정의할 수 있습니다. 이와 같은 변경 사항은 Starlight의 기본 컴포넌트를 수정할 필요 없이 해당 컴포넌트에 전달되는 데이터만 수정하면 됩니다.
+
+### 경로 데이터를 사용자 정의하는 방법
+
+"미들웨어"의 특수한 형태를 사용하여 경로 데이터를 사용자 정의할 수 있습니다.
+이는 Starlight가 페이지를 렌더링할 때마다 호출되는 함수로, 경로 데이터 객체의 값을 수정할 수 있습니다.
+
+<Steps>
+
+1. Starlight의 `defineRouteMiddleware()` 유틸리티를 사용하여 `onRequest` 함수를 내보내는 새 파일을 만듭니다.
```
trueberryless, line 54: okay, I think it always messes up when code quotes are used (so your translation is right i guess):

```
Starlight `defineRouteMiddleware()`Using the utility `onRequestCreate` a new file to export the function.
```

docs/src/content/docs/ko/guides/route-data.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+   ```js ins={9}
+   // astro.config.mjs
+   import { defineConfig } from 'astro/config';
+   import starlight from '@astrojs/starlight';
+
+   export default defineConfig({
+   	integrations: [
+   		starlight({
+   			title: 'My delightful docs site',
```
trueberryless, line 73: Can you maybe also translate the title into Korean please?

docs/src/content/docs/ko/guides/route-data.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+```js {9}
+// astro.config.mjs
+import { defineConfig } from 'astro/config';
+import starlight from '@astrojs/starlight';
+
+export default defineConfig({
+	integrations: [
+		starlight({
+			title: 'My site with multiple middleware',
```
trueberryless, line 114: Same here, I think it would be cool if the title is also translated 🙌

trueberryless (changes requested):
Very good job! Generally looks good to me, I left some suggestions (you can probably ignore the first two because that's just my translation tool doing something wrong, but I think the last two are more important) 👍

docs/src/content/docs/ko/guides/route-data.mdx
jsparkdev: Thank you for your review.

I have double-checked the translation of this section and it seems to be translated correctly.

docs/src/content/docs/ko/guides/route-data.mdx
jsparkdev: Good point. I've updated it!

docs/src/content/docs/ko/guides/route-data.mdx
jsparkdev: Thank you!

trueberryless (approved):
Thank you very much for the quick update! Looks perfect to me now! 👍

PR #2898, opened 2025-02-17
i18n(fr): fix highlighting in `reference/plugins.md`

ArmandPhilippot:
#### Description

Updates the French translation of `reference/plugins` to fix a code snippet highlighting.

I suggested a fix in https://github.com/withastro/starlight/pull/2886#discussion_r1957375034 but I didn't pay attention that we already have `en` in the string so the highlighting is currently wrong.

|Before|After|
|---|---|
|starlight-before|starlight-after|

I'm not used to Expressive Code so perhaps there is better a way... Looking at the docs it seems we need [capture groups](https://expressive-code.com/key-features/text-markers/#marking-capture-group-contents). The [recommended syntax](https://expressive-code.com/key-features/text-markers/#escaping-forward-slashes) is `/\/(en)\//` but Prettier removes the escape characters... Tested locally with the dev server, it also works without!

### DIFF
--- a/docs/src/content/docs/fr/reference/plugins.md
+++ b/docs/src/content/docs/fr/reference/plugins.md
@@ -348,7 +348,7 @@ starlight({
 
 Un module d'extension peut déterminer la langue d'un fichier en utilisant son chemin absolu :
 
-```ts {6-8} /en/
+```ts {6-8} //(en)//
 // module-extension.ts
 export default {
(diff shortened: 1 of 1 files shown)


delucis (approved):
Great catch — thank you @ArmandPhilippot!

And can’t believe this is your FIRST Starlight PR 🤯

PR #2911, opened 2025-02-23
Add a new dedicated theme showcase page

delucis:
#### Description

- Adds a new [`/resources/themes/`](https://deploy-preview-2911--astro-starlight.netlify.app/resources/themes/) page to show off Starlight themes with screenshots
- Removes the old themes section on the community plugins page


trueberryless (approved):
LGTM 🎉 

I was about to remind the author that we should also add docs to the Contribution file, then I saw who the author is and knew that Chris thought of everything.

docs/src/components/theme-grid.astro
```diff
@@ -0,0 +1,196 @@
...
+<FluidGrid minColumnWidth="25rem" gap="clamp(1rem, calc(0.75rem + 1vw), 1.5rem)">
+	{
+		themes.map(async ({ title, description, href, previews }) => {
+			if (!previews) return;
+			const inputName = `preview-${title.toLowerCase().replace(/[^a-z0-9_-]+/g, '')}`;
+			return (
+				<MediaCard>
+					<starlight-theme-preview slot="media">
+						<ThemeImage src={previews.light} class="light" />
+						<ThemeImage src={previews.dark} class="dark" />
```
HiDeoo, line 44: I understand why there is no `alt` attribute as it's a [graphical representation of some of the surrounding text](https://html.spec.whatwg.org/multipage/images.html#a-graphical-representation-of-some-of-the-surrounding-text), e.g. the theme title + description, altho on the other end, it feels a bit weird to me to have controls that only change these images with no feedback without even being announced that images are on this page 🤔

docs/src/components/theme-grid.astro
```diff
@@ -0,0 +1,196 @@
...
+								<span class="sr-only">{labels.light}</span>
+								<input class="sr-only light" name={inputName} value="light" type="radio" />
+							</label>
+						</fieldset>
+					</starlight-theme-preview>
+					<div class="meta sl-flex">
+						<a class="title" {href}>
+							{title}
+						</a>
+						{description && <p class="description" set:html={description} />}
```
HiDeoo, line 63:
```suggestion
						<p class="description" set:html={description} />
```

Don't think the description is optional

docs/src/components/theme-grid.astro
```diff
@@ -0,0 +1,196 @@
...
+								<Icon name="approve-check" />
+								<span class="sr-only">{labels.dark}</span>
+								<input class="sr-only dark" name={inputName} value="dark" type="radio" checked />
+							</label>
+							<label class="light">
+								<Icon name="approve-check" />
+								<span class="sr-only">{labels.light}</span>
+								<input class="sr-only light" name={inputName} value="light" type="radio" />
+							</label>
+						</fieldset>
```
HiDeoo, line 57: Considering some [well-known](https://bugs.webkit.org/show_bug.cgi?id=230824) screen reader issues with fieldsets, e.g. repeating everything twice, or even three times for me for the legend in Firefox, that's a lot of content for screen reader to go through before even getting to the theme name. I wonder if markup wise they should be inverted and swap only visually.

If compared to the video content showcase, these cards would just announce "Play" then the video title which is quite small in this case.

docs/src/components/theme-grid.astro
```diff
@@ -0,0 +1,196 @@
...
+					</div>
+				</MediaCard>
+			);
+		})
+	}
+</FluidGrid>
+
+<script>
+	/**
+	 * Progressive enhancement that syncs all theme previews when the light/dark option is toggled for a theme.
```
HiDeoo, line 73: I wonder if it could be nice if my current choice, either being a manually selected theme, or my OS fallback, would be respected. Would require a bit of inline code but only on that specific page.

docs/src/content/docs/resources/themes.mdx
```diff
@@ -0,0 +1,76 @@
...
+A theme is a Starlight plugin that changes the visual appearance of a site with [custom CSS](/guides/css-and-tailwind/), [component overrides](/guides/overriding-components/), or other new features.
+
+## Community Themes
+
+Install a theme built by the community to quickly customize the look and feel of your site.
+
+<ThemeGrid
+	labels={{
+		/** Accessible label for the theme toggle. */
+		toggle: 'Preview theme',
```
HiDeoo, line 24: Nit: a bit weird when using a screen reader imo, as it sounds like it's used for a control allowing you to preview the theme instead of selecting the theme of the previewed themes (yes, that's also a lof of "theme"s 😅 ). We use "Select theme" for the global theme picker, maybe we could use something similar here, e.g. "Select preview themes" or something similar?

docs/src/content/docs/resources/themes.mdx
```diff
@@ -0,0 +1,76 @@
...
+			href: 'https://delucis.github.io/starlight-theme-flexoki/',
+			previews: { light: 'flexoki-light.png', dark: 'flexoki-dark.png' },
+		},
+	]}
+/>
+
+:::tip[Add your own!]
+Have you built a Starlight theme?
+[Open a PR](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#themes) adding your theme to this page!
+:::
```
HiDeoo, line 76: In all resource pages, this block is always at the top of the page, is it expected to be at the bottom here?

docs/src/content/docs/resources/themes.mdx
```diff
@@ -0,0 +1,76 @@
...
+			previews: { light: 'ion-light.png', dark: 'ion-dark.png' },
+		},
+		{
+			title: 'Starlight Black',
+			description: 'Starlight theme inspired by shadcn docs.',
+			href: 'https://starlight-theme-black.vercel.app/',
+			previews: { light: 'black-light.png', dark: 'black-dark.png' },
+		},
+		{
+			title: 'Starlight Flexoki',
```
HiDeoo, line 64: Wonderful new addition 🤩

HiDeoo (commented):
Amazing new page 🤩  Left a few tiny comments/suggestions mostly around a11y but overall this is looking so great! :clap:

docs/src/content/docs/resources/themes.mdx
delucis: Mostly because with the wide layout I find it a bit distracting at the top? (As I side note: I think I’d be tempted to move them all down as I’m not sure they’re the most high priority content on these pages?)

HiDeoo: I guess the CTA at the top was more relevant before when these pages were brand new, not a lot of plugins/content, etc. but now I think they would be fine all at the bottom too?

docs/src/components/theme-grid.astro
delucis: Yeah, definitely tricky to get right. I’ve just pushed a commit that adds `aria-hidden="true"` on the previews. With VoiceOver in Firefox this doesn’t feel too bad, but in Safari and Edge it didn’t really help things. It would be nice to test with another screenreader if someone has the time.

- **Firefox + VoiceOver:** When tabbing, you still tab to the radio input but it is interpreted as the card receiving focus so the theme name and description is announced. The next tab stop takes you to the actual link. The inputs themselves are not announced at all.
- **Edge + VoiceOver:** Tabbing to the radio input still announces the input label and the fieldset legend.
- **Safari + VoiceOver:** Tabbing to the radio input still announces the input label, but the fieldset legend is no longer announced (so a regression to be honest).

I didn’t manage to find much information on this topic. Understandably most people want to make sure interactive form controls _are_ announced. If I don’t see a better solution, I might be tempted to switch to a single control above the previews so that even if it is announced and a little verbose, it is only announced once and not for every list item.

HiDeoo: On Windows using NVDA, the behaviors are mostly the same:

- **Firefox + NVDA**: Tabbing to the input seems to focus the card even though the input is focused. Announce the theme name and description.
**- Chrome + NVDA**: Tabbing to the input announce the "Preview theme" grouping + "Dark" radio button radio check 1 of 2. The next tab stop announces the actual link title.
- **Edge + NVDA**: Same as Chrome.

HiDeoo (approved):
Even tho I really liked the initial design, there is no doubt that the latest changes makes the experience way better for keyboard and screen reader users. Tested on multiple browsers and screen readers, really impressive improvements :clap:

:rocket:

trueberryless:
Okay, I'm very sorry that I didn't notice before, but during translation I noticed some odd things:

1. the plugins.mdx and themes.mdx pages have the same sidebar order.
2. All the accessible labels passed to the ThemeGrid component would already be available, either through the UI strings, and through the json files in the `content/i18n` directory: 
image

CC @delucis @HiDeoo I think the second point is rather intentional, but the sidebar order feels like it went unnoticed (even by me from the first review...)

delucis:
Well spotted @trueberryless!

> 1. the plugins.mdx and themes.mdx pages have the same sidebar order.

My intention was to have `order: 2` in the theme page but I guess I typo’d and never noticed because the end result is the same.

> 2. All the accessible labels passed to the ThemeGrid component would already be available, either through the UI strings, and through the json files in the `content/i18n` directory:

While this is true, the context here is quite different so there’s no guarantee that they would be translated the same for all languages in both contexts. You could argue for sure that light/dark are _likely_ to be the same, but for example “component preview” and “theme preview” are slightly different concepts so in some languages they might want to translate these independently. (Maybe no-one will, but still, in general I prefer having that flexibility just in case!)

trueberryless:
Shall I create a PR for the order thingy?😉

delucis:
> Shall I create a PR for the order thingy?😉

Sure, if you like! You can probably update all the translations at the same time and include `i18nIgnore` in the PR title.

PR #2917, opened 2025-02-24
starlight-view-transition plugin

martrapp:
#### Description

Adds the starlight-view-transition plugin of the Bag of Tricks to the list of community plugins.


docs/src/content/docs/resources/plugins.mdx
```diff
@@ -128,6 +128,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		title="starlight-kbd"
 		description="Document keyboard shortcuts with ease."
 	/>
+	<LinkCard
+		href="https://events-3bg.pages.dev/library/StarlightPlugin/"
+		title="starlight-view-transitions"
+		description="Add browser-native cross-document view transitions with just a config change."
```
trueberryless, line 134: I personally would keep the description short and remove the "with just a config change". Also because most other plugins are configurable via settings but they don't mention it. But not a must change...

```suggestion
		description="Add browser-native cross-document view transitions"
```

trueberryless (approved):
Very nice addition! I didn't even notice that the view transitions weren't linked to anywhere from Starlight. But it makes sense, now that a plugin with the most recent version is available!👍

Left one suggestion for the description...

docs/src/content/docs/resources/plugins.mdx
martrapp: Clearly, we're both enthusiastic about the route data middleware layer!

delucis (approved):
Let’s _transition_ this change into our live website! 🚀

PR #2918, opened 2025-02-25
Fix trailing slash inconsistency in sidebar links

HiDeoo:
#### Description

This PR fixes an inconsistency in the sidebar links when using the `trailingSlash: 'ignore'` Astro option (the default) between internal and auto-generated links.

When using this configuration value, [internal links](https://starlight.astro.build/guides/sidebar/#internal-links) specified using the `{ slug: 'constellations/andromeda' }` syntax or `'constellations/andromeda'` shorthand don't have a trailing slash while [auto-generated](https://starlight.astro.build/guides/sidebar/#autogenerated-groups) links do.

Starlight behavior is to use a trailing slash for this configuration value as many common hosting providers redirect to URLs with a trailing slash by default.

This PR ensures that internal links have a trailing slash in this case to match the auto-generated links.


delucis (approved):
Oh that was an easy code change in the end 😅 

Wish everything trailing slash related was this easy!

Thanks so much @HiDeoo 🙌

PR #2924, opened 2025-02-26
Remove trailing whitespace from `<Badge>` and `<Icon>` components

HiDeoo:
#### Description

In Astro, [components that include styles render with a trailing space](https://github.com/withastro/compiler/issues/1003). This behavior is not ideal for components like `<Badge>` and `<Icon>` that are often used in inline text.

For example, the following code:

```mdx
Feature (<Badge text="New" />)
```

will render as:

```
Feature (New )
```

This PR removes the trailing space from `<Badge>` and `<Icon>` components by wrapping them in [a fragment](https://docs.astro.build/en/reference/astro-syntax/#fragments) which prevents the trailing space from being rendered.

image

If we ship this change and its associated documentation change to accommodate the new behavior, we could potentially update the docs for all languages as it's purely structural changes and mark it as i18n ignored


SnowDingo (commented):
@HiDeoo, thank you for making this change.

I'm still a rookie coder so I can't give any suggestions on the code, but I feel like the new style without the trailing space is nice. I hope this PR will get merged.

-snowdingo

delucis (approved):
Changes all look great and the the changeset is really excellent and clear 👏 

I wish we didn’t need to rely on the weird fragment hack, but I’m glad we at least have good testing in place in case that ever breaks.

PR #2927, opened 2025-02-27
Add head content to route data

HiDeoo:
#### Description

This PR adds a new property `head` to route data which is an array of all tags to include in the `<head>` of the current page.

Previously, the `<Head>` component was responsible for generating a list of tags to include in the `<head>` of the current page and rendering them.
Now, the `<Head>` component is only responsible for rendering the tags available in `Astro.locals.starlightRoute.head`.

The PR also fixes a few issues at the same time that initially triggered the discussion around this change or spotted during the implementation.


HiDeoo:
Thanks for the review 🙌

PR #2928, opened 2025-02-27
fix: `themes.mdx` sidebar order [i18nIgnore]

trueberryless:
themes.mdx and plugins.mdx have the same sidebar order.


trueberryless:
> Thanks @trueberryless! I checked out the preview deployment and the sidebar order still looks correct 🙌 

😂 Imagine it would've looked different 🤣🤣🤣

trueberryless:
but yeah, better dribble check beforehand... You're right 👍

PR #2931, opened 2025-02-28
Support for `title`, `frame`, and `meta` Markdoc fence attributes

HiDeoo:
- Closes #2925

This PR adds support for the `title`, `frame`, and `meta` fence attributes to Markdoc code blocks.

The fence attribute syntax for Markdoc is an alternative to classic meta attributes after a fence block available in Markdown syntax (which are not supported in Markdoc):

````mdoc
```js {% title="editor.exe" frame="terminal" %}
console.log('Hello, world!');
```
````

Documentation note: not sure if we should document this as we don't document Markdoc syntax and at the same time, they are not even documenting this in the Markdoc documentation afaik and the only reference I could find is [this message](https://github.com/markdoc/markdoc/discussions/318#discussioncomment-4821979).

### DIFF
--- a/.changeset/healthy-rivers-divide.md
+++ b/.changeset/healthy-rivers-divide.md
@@ -1,0 +1,24 @@
+---
+'@astrojs/starlight-markdoc': minor
+---
+
+Adds support for the `title`, `frame`, and `meta` fence attributes to code blocks.
+ 
+These new optional attributes add support for Expressive Code [text & line markers](https://expressive-code.com/key-features/text-markers/). The following example renders a code block using a [terminal frame](https://expressive-code.com/key-features/frames/#terminal-frames) with a [title](https://expressive-code.com/key-features/frames/#code-editor-frames):
+
+```js {% title="editor.exe" frame="terminal" %}
+console.log('Hello, world!');
+```
+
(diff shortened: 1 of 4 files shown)


HiDeoo:
Added documentation examples for Markdoc.

So far I only added them using basic tabs to the "Expressive Code features" section which is the only one containing differences between MDX and Markdoc. Some questions that could come up are:

1. Should we do that for every example on this page for consistency even if it's not necessary 95% of the time?
1. Should we use the `<Preview>` component used for components? Would swap the order (preview → example vs example → preview).
   1. If we do, should we do the same for every example on this page too? Based on question 1, this would work as the component supports both a tabbed and a non-tabbed version. Altho, due to the switch in order, we may need to add some content to some sections to not only show the preview.

delucis:
> Added documentation examples for Markdoc.

Nice! These look great 👍 

> 1. Should we do that for every example on this page for consistency even if it's not necessary 95% of the time?

No, I don’t think so. I actually think it could be _more_ confusing because then you’re primed to look for a difference, when there is none.

> 2. Should we use the `<Preview>` component used for components? Would swap the order (preview → example vs example → preview).

Oh yeah, that’s a bigger question! I think it would indeed be nice now we have it, but we should do it comprehensively for the full page in a follow-up PR probably. And fixing the order definitely makes sense. This page was never properly revised from its days as a test for Markdown styles and the order has always bugged me.

PR #2935, opened 2025-03-04
docs: update `authoring-content` Markdown/MDX tab labels

HiDeoo:
This PR changes the labels of the tabs added in #2931 to `Markdown/MDX` instead of just `MDX` which is more accurate in the context of this page.


delucis (approved):
Perfect! Thanks for realising and fixing that so quickly 🚀

PR #2936, opened 2025-03-04

HiDeoo:
This PR updates the French version of the `guides/authoring-content` page with the changes from #2931 and #2935.

### DIFF
--- a/docs/src/content/docs/fr/guides/authoring-content.mdx
+++ b/docs/src/content/docs/fr/guides/authoring-content.mdx
@@ -245,6 +245,10 @@ Certaines des options les plus courantes sont présentées ci-dessous :
   }
   ```
 
+  <Tabs syncKey="content-type">
+
+  <TabItem label="Markdown/MDX">
+
   ````md
   ```js {2-3}
   function demo() {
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/authoring-content.mdx
```diff
@@ -329,37 +416,92 @@ Le titre optionnel d'un bloc de code peut être défini soit avec un attribut `t
...
   ```
 
+  <Tabs syncKey="content-type">
+
+  <TabItem label="Markdown/MDX">
+
+  ````md
+  ```js
+  // mon-fichier-de-test.js
+  console.log('Hello World!');
```
trueberryless, line 426: We could also translate this if you want...

trueberryless (approved):
LGTM 🎉 (non-native)

One small suggestions

docs/src/content/docs/fr/guides/authoring-content.mdx
trueberryless: Alto I see that in the French version non of those are translated...
(German and Korean translated...)

But I'll let the native people decide...

ArmandPhilippot: I know in Astro Docs it's usually translated, but I don't think it's necessary for `Hello world`. This is a common example that we also use in France (I don't know for other French-speaking countries). Personally this doesn't bother me. So it's up to you HiDeoo, you are more used to Starlight Docs than me!

ArmandPhilippot (approved):
LGTM! 💯

docs/src/content/docs/fr/guides/authoring-content.mdx
HiDeoo: Yeah, definitely a case where we decided keeping it as-is makes the most sense in this context 👍

PR #2947, opened 2025-03-06
feat: Add dark and light variants for images

dragomano:
#### Description

This PR adds dark and light variants for images in Markdown.

Source of inspiration: [the corresponding feature](https://squidfunk.github.io/mkdocs-material/reference/images/?h=light#light-and-dark-mode) in Material for MkDocs

#### Examples:

sshot-29
sshot-23

### DIFF
--- a/.changeset/empty-beds-heal.md
+++ b/.changeset/empty-beds-heal.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Make light and dark image variants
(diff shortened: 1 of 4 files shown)


dragomano:
In the print version (Firefox), I see only one image:

sshot-16

HiDeoo:
> In the print version (Firefox), I see only one image:

Yeah, altho the image can change depending based on if you have the dark theme specifically selected or if you're using the auto config.

When having the auto theme, you get the same preview as the one you shared, but if you specifically select the dark theme, you would get the following:

image

In the context of printing, I think it would make sense to enforce the light mode image.

dragomano:
Can you suggest how to do this?

delucis (commented):
Thanks for the PR @dragomano — I love this idea!

Unfortunately, it seems like we won’t be able to make this work consistently without some code fixes upstream in Astro. As @HiDeoo noted, currently appending a `#only-light` or `#only-dark` to an image path causes errors unless they are remote URLs like in your example in `authoring-content.mdx`.

If we were to support this, I think it’s important that it works for all variants of local and remote images, so users can always use it.

Given that, I think it probably makes sense to close this PR for now. If anyone is interested in investigating what it would take to fix the Astro image service to support fragment links like this, I’d love to see that! I think it may not be super complex, just something no-one has considered yet. Once that is done we can definitely bring back this idea.

PR #2948, opened 2025-03-06
update astro to the latest version

clintonsteiner:
* see https://github.com/withastro/astro/issues/12360 for details

#### Description

- Closes # 


delucis (approved):
Thanks for the PR @clintonsteiner! I took some time to update `astro` consistently across everywhere it is used in the monorepo which also required one small refactor to a test due to a type definition change (thanks for the help @HiDeoo), but this should be good to go now!

PR #2955, opened 2025-03-11
feat: add some IDE and Design app icons

trueberryless:
#### Description

- add 6 new icons: 
  - VS Code
  - Jetbrains
  - Zed
  - Vim
  - Figma
  - Sketch

- those icons where optimized with [SVGOMG](https://jakearchibald.github.io/svgomg/), Number precision: 2

- Closes #2944 (probably have to manually close, cause discussions dont get closed as issues do...)


packages/starlight/schemas/social.ts
HiDeoo, line 1: I think only Figma and Sketch should be included to link to design documents but the editors don't make a lot of sense here?

This should also probably be reflected in the changeset (and maybe have 2, one for the social options and one for the icons).

trueberryless: Sizing issue: Whoops, I think I didn't normalize the viewport of the SVGs before extracting the paths of them, gonna fix it right away.

About the other suggestion: I didn't know there were different "groups" of icons. Could you maybe explain to me what the difference is so I can edit my PR and changesets accordingly?

HiDeoo: It's not really groups of icons but more like all social media accounts configurable in Starlight (e.g. Twitter, Reddit, Discord) have an associated icons but not all icons have a social media account equivalent (e.g. C#)

When documenting a design system for example, it makes sense to link to a Figma or Sketch design document so adding these 2 social media account options makes sense.

For text editors, I don't think there is a social media link that would make sense, except if I'm wrong?

delucis: I’m not sure even the argument for Figma and Sketch being included as social icons is very strong? The social icons are for social media profiles in general and I don’t really think of design tools like this either.

In theory, someone could use it to link to a Figma plugin or profile or something, but I’d suggest not supporting that by default as I would guess you’d need a better accessible label than just “Figma”. I’m not sure what I’d expect encountering a link labelled “Figma” (compared to stuff labelled “GitHub” or “Twitter” etc. which I think are a bit clearer and more expected).

Then if people do need it in the social links they can always add it with an override as shown in the docs.

HiDeoo: I think that's a fair point too, no strong preference from me 👍 Would even simplify the changeset to something like:

> Adds 5 new icons: `figma`, `jetbrains`, `sketch`, `vscode`, and `zed`.

delucis (approved):
LGTM!

trueberryless:
Now it's right 👍😉

PR #2962, opened 2025-03-14
docs: add `starlight-auto-sidebar` plugin

HiDeoo:
#### Description

This PR adds the [`starlight-auto-sidebar`](https://github.com/HiDeoo/starlight-auto-sidebar) plugin to the list of community plugins.

PR #2963, opened 2025-03-14
i18n(fr): update `resources/plugins`

HiDeoo:
#### Description

This PR updates the French version of the `resources/plugins` page with the changes from #2962.

### DIFF
--- a/docs/src/content/docs/fr/resources/plugins.mdx
+++ b/docs/src/content/docs/fr/resources/plugins.mdx
@@ -133,6 +133,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-view-transitions"
 		description="Ajouter des transitions de vue natives entre vos pages."
 	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-auto-sidebar"
+		title="starlight-auto-sidebar"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -133,6 +133,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-view-transitions"
 		description="Ajouter des transitions de vue natives entre vos pages."
 	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-auto-sidebar"
+		title="starlight-auto-sidebar"
+		description="Modifier vos groupes de barre latérale générés automatiquement."
```
HiDeoo, line 139: A more literal translation of "Tweak" would be "Ajuster" but I didn't find it very fitting in this context regarding the purpose of the plugin.

trueberryless: Is the purpose of the plugin kinda to "optimise" working with the sidebar? Because in the German translation I also didn't find a 1:1 fitting translation, so I chose what translates to "Optimise ..."

jsparkdev (approved):
LGTM!

PR #2965, opened 2025-03-16

jsparkdev:
#### Description

- update `plugins.mdx`
- #2962


trueberryless:
Quick question: Did you find a nice translation for the English word "tweak"?

Because in the German version I used a word that translates to "optimise" and I wonder if you also chose some different translation...

trueberryless (approved):
Overall looks good to me 🙌

jsparkdev:
@trueberryless 

Thank you. Your question made me think again.

Although I'm not sure, judging from the context, I think "Tweak" here means to modify it as you want.

So I modified it again. Thank you!

trueberryless:
Perfect, this is wunderful! I too took HiDeoo's French version as an example: They also used "modifier" which means modify. Now we all did the same in each language 🤩

PR #2969, opened 2025-03-18
Add npm social icon

webpro:
#### Description

- Closes #  - did not create an issue first
- Add social icon for `npm`
- A work of art:

Screenshot 2025-03-18 at 1 55 39 PM

Can't make it any smaller ([SVGOMG](https://jakearchibald.github.io/svgomg/) makes it larger)


packages/starlight/schemas/social.ts
```diff
@@ -97,6 +98,7 @@ export const SocialLinksSchema = () =>
 					confluence: 'Confluence',
 					jira: 'Jira',
 					storybook: 'Storybook',
+					npm: 'npm',
```
delucis, line 101: I don’t know for sure because in testing with VoiceOver on macOS both “npm” and “NPM” were both read correctly, but I wonder if we should use the uppercase acronym style for the accessible label in case there are any screen readers that take that as a hint to read the letters separately vs trying to pronounce it as a word?

```suggestion
					npm: 'NPM',
```

delucis (commented):
Thanks @webpro! Was just thinking about this last week while setting up a docs site, so I think it’s a good addition.

Noting that we technically do already have an npm icon included although it looks a bit different as it’s the one from the Seti UI VS Code icon set:

npm icon labelled seti:npm

I guess the argument for the new `n` icon is that it’s better designed to be displayed in a small square icon space? Do we have any examples of sites using one or the other of these icons?

webpro:
> Thanks @webpro! Was just thinking about this last week while setting up a docs site, so I think it’s a good addition.
> 
> Noting that we technically do already have an npm icon included although it looks a bit different as it’s the one from the Seti UI VS Code icon set:
> 
> npm icon labelled seti:npm
> I guess the argument for the new `n` icon is that it’s better designed to be displayed in a small square icon space? Do we have any examples of sites using one or the other of these icons?

In my experience the "n" is pretty command. See the favicon and the icon down in the footer of the npm website itself, e.g. https://www.npmjs.com/about

Seeing that this bigger icon has a slightly rounded corner I updated the SVG.

packages/starlight/schemas/social.ts
webpro: Personally I cringe a bit at seeing uppercased npm (also see their about page), but I'm not an a11y expert so if you insist I'm happy to update it.

webpro:
Update:

Screenshot 2025-03-18 at 6 29 41 PM

packages/starlight/schemas/social.ts
delucis: Yeah. To be clear this text is never displayed visibly in UI, it is only used to provide an accessible label for those links. @HiDeoo was kind enough to test in NVDA and JAWS which also pronounce it correctly with the lowercase version, so I think we’re fine to leave it as-is and avoid you having to cringe 😄

delucis (approved):
Thanks again @webpro! Will merge this when we’re ready for our next patch release.

packages/starlight/schemas/social.ts
webpro: Sorry, that was an unfortunate way to express just personally not loving to see npm spelled uppercased, like, anywhere. But this isn't about me, it's about accessibility. Glad it's pronounced properly by assistive tech! Thanks to both of you for some high quality QA here ❤️

PR #2970, opened 2025-03-18
chore: add renovate to the repository

ematipico:
#### Description

This PR adds renovate to the repository, mostly to keep dependencies and GH actions in check.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,19 @@
+{
+	$schema: 'https://docs.renovatebot.com/renovate-schema.json',
+	extends: [
+		'config:recommended',
+		'schedule:weekly',
+		'group:allNonMajor',
+		':disablePeerDependencies',
+		'helpers:pinGitHubActionDigestsToSemver',
+	],
+	rangeStrategy: 'bump',
+	postUpdateOptions: ['pnpmDedupe'],
+	ignorePaths: ['**/node_modules/**'],
+	packageRules: [
+		{
+			groupName: 'github-actions',
+			matchManagers: ['github-actions'],
+		},
+	],
+}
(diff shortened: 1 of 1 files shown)


delucis:
> This PR adds renovate to the repository, mostly to keep dependencies and GH actions in check.

I think I have the same question as in https://github.com/withastro/docs/pull/11251: does this also update other dependencies? Or _only_ actions?

ematipico:
Also dependencies. I can remove it if needed

delucis:
Happy to hear different thoughts. I don’t have experience with renovate so I don’t know how to evaluate it. My past experience with dependency management tools was with Greenkeeper and then Dependabot and honestly, I mostly decided they were too annoying to be valuable, but maybe this is better. Has it been useful in the Astro repo?

trueberryless:
> This PR adds renovate to the repository, mostly to keep dependencies and GH actions in check.


WOW cool, what is renovate? I did some research and it sounds similar to Dependabot updates, but I have to agree with Chris, they too have always been to annoying to me in the past...

HiDeoo:
My personal experience with Renovate is that it's slightly better than some other similar solutions but can still be annoying, especially at our relatively small scale in the Starlight repo?

Maybe a good starting point could be to limit it to actions in the first place, as there is no denying that it's very valuable in this case, and then see later if we want to expand to other dependencies?

delucis:
> Maybe a good starting point could be to limit it to actions in the first place, as there is no denying that it's very valuable in this case, and then see later if we want to expand to other dependencies?

Sounds good to me. What do you think @ematipico?

ematipico:
I scrapped the `config:recommended` to have only the minimum:
- dependency dashaboard
- semantic startegy
- ignore modules and tests

delucis (approved):
Thanks @ematipico! Happy to try it out. We can always disable it again if we hate it.

PR #2985, opened 2025-03-19
Disable Renovate package manager updates

HiDeoo:
This PR should hopefully disable `packageManager` updates.


delucis (approved):
Thanks for digging into that!

PR #2991, opened 2025-03-20
feat(tailwind): support @astrojs/tailwind v6

florian-lefebvre:
#### Description

- What does this PR change? Updates the peer dep. I made it a patch since v6 doesn't change anything compared to v5

### DIFF
--- a/.changeset/five-avocados-hang.md
+++ b/.changeset/five-avocados-hang.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight-tailwind': patch
+---
+
+Adds support for `@astrojs/tailwind` v6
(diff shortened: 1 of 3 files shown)


trueberryless (approved):
LGTM, however in my opinion a minor bump should be the golden middle between "it's not a major change and we add a major peer dep change", but as the maintainers like...

delucis (approved):
LGTM! Thank you @florian-lefebvre and to @trueberryless for reviewing 🙌 

Re: minor vs patch bump — I think this is conceptually a bug fix as for people using v5 of the integration nothing changes, but it fixes things for anyone trying to use v6 (which itself doesn’t include any new functionality), so I’m happy leaving it a patch!

PR #2996, opened 2025-03-20
feat: Add Starlight Toc Overview Customizer plugin

trueberryless:
#### Description

- add my newest plugin to the list
- modify some other descriptions where I noticed some inconsistencies - mostly dots at end of sentence (please let me know if I should remove those changes if unwanted)

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -56,7 +56,7 @@ Extend your site with official plugins supported by the Starlight team and commu
 	<LinkCard
 		href="https://astro-ghostcms.xyz/intro/starlight/install/"
 		title="starlight-ghostcms"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/resources/plugins.mdx
```diff
@@ -56,7 +56,7 @@ Extend your site with official plugins supported by the Starlight team and commu
 	<LinkCard
 		href="https://astro-ghostcms.xyz/intro/starlight/install/"
 		title="starlight-ghostcms"
-		description="Add your GhostCMS blog posts alongside your Starlight Docs"
+		description="Add your GhostCMS blog posts alongside your Starlight docs."
 	/>
```
delucis, line 60: We can do this in a follow-up PR but this URL is a 404 and the plugin is deprecated so we should probably remove this.

delucis (approved):
Thanks @trueberryless — looks great!

PR #2998, opened 2025-03-21
Add Starlight Markdown Blocks to community plugins

delucis:
#### Description

- This PR adds [Starlight Markdown Blocks](https://delucis.github.io/starlight-markdown-blocks/) to the list of community plugins

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -148,6 +148,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		title="starlight-toc-overview-customizer"
 		description="Tweak Starlight's table of contents with customizable overview title."
 	/>
+	<LinkCard
+		href="https://delucis.github.io/starlight-markdown-blocks/"
+		title="starlight-markdown-blocks"
+		description="Extend Starlight’s Markdown asides syntax with custom block types."
+	/>
 </CardGrid>
 
 ## Community tools and integrations


docs/src/content/docs/resources/plugins.mdx
```diff
@@ -148,6 +148,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		title="starlight-toc-overview-customizer"
 		description="Tweak Starlight's table of contents with customizable overview title."
 	/>
+	<LinkCard
+		href="https://delucis.github.io/starlight-markdown-blocks/"
+		title="starlight-markdown-blocks"
+		description="Extend Starlight’s Markdown asides syntax with custom block types."
```
trueberryless, line 154: I recommend using the straight quotation mark here because then it is consistent with the other entries and Astro converts it to a smart apostrophe nonetheless...

```suggestion
		description="Extend Starlight's Markdown asides syntax with custom block types."
```

trueberryless (approved):
LGTM overall 🎉

Just left one very small suggestion...

docs/src/content/docs/resources/plugins.mdx
delucis: I don’t think this would get converted by smartypants as it’s not in Markdown?

For example, here it’s shown as a straight quote rather than an apostrophe:
starlight-toc-overview-customizer card where the description includes a straight quote

I type `’` and `“”` instinctively without thinking though, so appreciate you double checking!

trueberryless: > I don’t think this would get converted by smartypants as it’s not in Markdown?

Hmm, true. Didn't think about that. Maybe we should update the other apostrophes in this file and other components too then? 🤔

PR #2999, opened 2025-03-21
docs: remove ghostcms plugin

trueberryless:
#### Description

- remove ghostcms plugin because it's depricated (not available any more)


delucis (approved):
Thank you for following up on that! 🙌

PR #3001, opened 2025-03-21
[i18nIgnore] docs: Exchange apostrophes in plugins to be smart

trueberryless:
As noticed in #2998 there are some plugins that use `'` (straight apostrophe) instead of (`’`) smart apostrophes.

This is probably unintentional by the author of the PRs adding these plugins and changed to be smart in this PR.

Reference: https://github.com/withastro/starlight/pull/2998#discussion_r2007808876


delucis (approved):
Thanks for cleaning that up! I guess this might not require translation updates for everyone, so I’ll merge with an `i18nIgnore` keyword.

PR #3004, opened 2025-03-22
Add new `link` option to `LogoConfig`

mavam:
This PR adds a new `link` option to `LogoConfig` so that it becomes possible to choose a destination when clicking the logo.

This comes in handy when it's desired to link back from the docs to the product's website.

### DIFF
--- a/.changeset/long-trains-jog.md
+++ b/.changeset/long-trains-jog.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Allow for custom link target when clicking logo
--- a/docs/src/content/docs/guides/customization.mdx
+++ b/docs/src/content/docs/guides/customization.mdx
@@ -50,12 +50,14 @@ Adding a custom logo to the site header is a quick way to add your individual br
 By default, the logo will be displayed alongside your site’s `title`.
 If your logo image already includes the site title, you can visually hide the title text by setting the `replacesTitle` option.
 The `title` text will still be included for screen readers so that the header remains accessible.
+Use the `href` option to define a custom destination when clicking the logo, in case you prefer a value different from `site`.
 
-```js {5}
+```js {5-6}
 starlight({
   title: 'Docs With My Logo',
   logo: {
     src: './src/assets/my-logo.svg',
+    href: 'my-domain.example',
     replacesTitle: true,
   },
 }),
--- a/packages/starlight/components/SiteTitle.astro
+++ b/packages/starlight/components/SiteTitle.astro
@@ -4,7 +4,7 @@ import config from 'virtual:starlight/user-config';
 const { siteTitle, siteTitleHref } = Astro.locals.starlightRoute;
 ---
 
-<a href={siteTitleHref} class="site-title sl-flex">
+<a href={config?.logo?.href || siteTitleHref} class="site-title sl-flex">
 	{
 		config.logo && logos.dark && (
 			<>
--- a/packages/starlight/schemas/logo.ts
+++ b/packages/starlight/schemas/logo.ts
@@ -6,6 +6,8 @@ export const LogoConfigSchema = () =>
 			z.object({
 				/** Source of the image file to use. */
 				src: z.string(),
+				/** Link to follow when clicking the logo. */
+				href: z.string().optional(),
 				/** Alternative text description of the logo. */
 				alt: z.string().default(''),
 				/** Set to `true` to hide the site title text and only show the logo. */
@@ -16,6 +18,8 @@ export const LogoConfigSchema = () =>
 				dark: z.string(),
 				/** Source of the image file to use in light mode. */
 				light: z.string(),
+				/** Link to follow when clicking the logo. */
+				href: z.string().optional(),
 				/** Alternative text description of the logo. */
 				alt: z.string().default(''),
 				/** Set to `true` to hide the site title text and only show the logo. */


HiDeoo:
Thanks for the contribution.

Sharing a few early thoughts as I don't think this was discussed in a feature request yet ahead of the pull request, so we can get the conversation started:

- The current [`siteTitleHref` route data](https://starlight.astro.build/reference/route-data/#sitetitlehref) value controls the link used for the site title link.
- This value is already [customizable](https://starlight.astro.build/guides/route-data/#customizing-route-data) using a route data middleware, which is an approach that gives a lot of flexibility, e.g. different link for different pages, different link per locale, etc.
- If we were to add a new option, one potential question could be how internationalization is handled, e.g. should it be a string or an object with different values per locale, both, or something else?
- Implementation-wise, `siteTitleHref` from the route data object should remain the source of truth used in the `<SiteTitle>` component. If we were to add a new option, this would allow users to provide a custom value for the site title link using this option while still being overridable by other route data middlewares, e.g. to get a different link on a specific page.

mavam:
Thanks for the immediate response!

> This value is already [customizable](https://starlight.astro.build/guides/route-data/#customizing-route-data) using a route data middleware, which is an approach that gives a lot of flexibility, e.g. different link for different pages, different link per locale, etc.

But is this a user-friendly option? I understand that step 0 is agreeing that it's actually something we'd want to expose to users with a reasonable cost-complexity trade-off. Sorry for bypassing this discussion. But assuming we said it should be easy (I am a dumb user here), then this path would be harder and require switching from a user to hacker persona, as TypeScript is now needed.

> If we were to add a new option, one potential question could be how internationalization is handled, e.g. should it be a string or an object with different values per locale, both, or something else?

Intuitively, I would use the same type as `siteTitleHref`, which is `string`. Unless I'm mistaken, this should be sufficient for URLs. 

> Implementation-wise, siteTitleHref from the route data object should remain the source of truth used in the <SiteTitle> component. If we were to add a new option, this would allow users to provide a custom value for the site title link using this option while still being overridable by other route data middlewares, e.g. to get a different link on a specific page.

I see, so instead of modifying the component source, you'd want me to add a middleware that does the modification? If so, do you have an example where this is already done?

HiDeoo:
> But is this a user-friendly option?

An option is definitely more user-friendly, I was mostly describing the existing architecture rather than advocating for one solution over another.

Personally, it's not something I've changed a lot, so would be curious to hear what others think.

> Intuitively, I would use the same type as `siteTitleHref`, which is `string`. Unless I'm mistaken, this should be sufficient for URLs.

A similar case would be the `siteTitle` variable used in the [`<SiteTitle.astro>` component](https://github.com/withastro/starlight/blob/main/packages/starlight/components/SiteTitle.astro#L32). It is [a `string`](https://starlight.astro.build/reference/route-data/#sitetitle) but configured using either [a string or an object](https://starlight.astro.build/reference/configuration/#title-required) for multilingual sites and the end result value is computed based on the current locale and the configuration.

> I see, so instead of modifying the component source, you'd want me to add a middleware that does the modification? If so, do you have an example where this is already done?

The computation would need to be done when we actually compute the `siteTitleHref` value using the [`getSiteTitleHref()`](https://github.com/withastro/starlight/blob/3b962d1a55f2c8b90b278368012ef9c9f9a4e49c/packages/starlight/utils/routing/data.ts#L115-L117) helper. At the moment, it's currently returning a different value based on the current locale (which made me question if a potential option would also need to be locale-aware) and that would be the place to add the logic to consider a new option.

mavam:
I'm now not sure anymore what I can do to move this forward. We can also wait, as you said, for others to chime on the utility of this.

In the meantime, I'll use

```ts
import { defineRouteMiddleware } from '@astrojs/starlight/route-data';

export const onRequest = defineRouteMiddleware((context) => {
	context.locals.starlightRoute.siteTitleHref = 'https://my-domain.example';
});
```

as middleware.

delucis:
> is it common to have different values for different locales?

At least in our default behaviour, we do localise these links (i.e. for `example.com` docs, the French logo link might be `example.com/fr`, and the Hindi logo link might be `example.com/hi`), so I would assume if we were making this customisable here, the same should apply. The approach we use for the site title would probably make sense in that case.

I guess the only concern is that _could_ feel extremely repetitive:

```js
link: {
  en: 'https://example.com/en',
  fr: 'https://example.com/fr',
  hi: 'https://example.com/hi',
  'zh-CN': 'https://example.com/zh-cn',
  // etc.
}
```

In cases like that, a route middleware may even feel less annoying:
```js
import { defineRouteMiddleware } from '@astrojs/starlight/route-data';

export const onRequest = defineRouteMiddleware((context) => {
	context.locals.starlightRoute.siteTitleHref = 'https://example.com/' + context.currentLocale;
});
```

kevinzunigacuellar:
> In cases like that, a route middleware may even feel less annoying:
> ```js
> import { defineRouteMiddleware } from '@astrojs/starlight/route-data';
>
> export const onRequest = defineRouteMiddleware((context) => {
>	context.locals.starlightRoute.siteTitleHref = 'https://example.com/' + context.currentLocale;
>});
>```

I like the middleware approach, you have everything you may need in a typed object as input. I think having a another way to configure this might bring some confusion like: 
- Which one takes precedence over the other?
- Which is one is considered "best practice"?
- If you may require an additional piece of data (in this example the `currentLocale`) how would you have access to it?

mavam:
@kevinzunigacuellar the key question is whether you deem it user-facing. Someone without dev experience arguably could not configure the middleware without getting lost. You can give a config file to anyone, but not code.

I lack community insights, so probably the maintainers should call the shot.

delucis:
As there seem to be a few different opinions here and there is an existing workaround, I’m going to close this PR. I’d still love to see a new [feature request discussion](https://github.com/withastro/starlight/discussions/new?category=feature-requests) for this. That would allow others to share their opinions and upvote it if they share the use case.

mavam:
@delucis tracking this now in https://github.com/withastro/starlight/discussions/3042.

PR #3012, opened 2025-03-24
Add starlight theme nova

ocavue:
#### Description

This PR adds a new theme https://github.com/ocavue/starlight-theme-nova to the docs.

PR #3016, opened 2025-03-25
i18n(de): Update `themes.mdx`

trueberryless:
#### Description

- Update `themes.mdx`
- #3012


trueberryless:
Formating seems wrong, do you want me to add some autoformatting to Starlight PRs eg autofix which runs prettier?

HiDeoo:
> Formating seems wrong, do you want me to add some autoformatting to Starlight PRs eg [autofix](https://autofix.ci/) which runs prettier?

We already have a [workflow formatting the code](https://github.com/withastro/starlight/blob/main/.github/workflows/format.yml) except it's happening on the `main` branch so it'll automatically happen when the PR is merged so that branches (mostly) don't have to worry about formatting.

Altho, as you're doing frequent PRs and the repo is configured for it, I would make sure your IDE/text editor is properly configured to pick the repo configuration and use it, e.g. on save so everything you push is already formatted properly, or you manually run it with a shortcut, etc.

trueberryless:
> Altho, as you're doing frequent PRs and the repo is configured for it, I would make sure your IDE/text editor is properly configured to pick the repo configuration and use it, e.g. on save so everything you push is already formatted properly, or you manually run it with a shortcut, etc.

I already have that set up on my PC, but my enthusiasm leads me to creating such PRs from my phone where not even a nice Android IDE is available... 🤣 That's the reason why some of my PRs look ugly...

But thanks for the answer, gonna let the formatting happen only on `main`!

delucis (approved):
Yup, we’ll let the robots worry about tabs vs spaces so we can use our brains for other things. Thanks for the update @trueberryless! 🫡

PR #3018, opened 2025-03-25
Starlight Route Data Middleware must not conflict with Astro's middleware

trueberryless:
#### Description

As I noticed myself, the current documentation does not point out that Starlight's custom middleware may or may not conflict with [Astro's middleware](https://docs.astro.build/en/guides/middleware/). 

Related discussion: #3017 (pls comment your thoughts there, this PR is just an example)


packages/starlight/utils/user-config.ts
```diff
@@ -233,6 +233,22 @@ const UserConfigSchema = z.object({
...
 		.default([])
+		.refine(
+			(middlewares) => {
+				// Ensure the middleware paths don't match the invalid ones
+				const invalidPaths = [
+					'src/middleware.js',
+					'src/middleware.ts',
+					'src/middleware/index.js',
+					'src/middleware/index.ts',
+				];
```
HiDeoo, line 244: I guess technically for something like `src/middleware.js`, you can pass `src/middleware.js`, `./src/middleware.js`, or `/src/middleware.js` and same for all other cases (I don't think `.mjs`, `.cjs`, `.mts` and `.cts` files are supported for middlewares). Maybe it would make more sense to use an hoisted regexp in this case to match all of these cases in one go :thinking:

Maybe something like this: https://regex101.com/r/HgYWAm/1

Any thoughts?

trueberryless: Sounds cool! Also I didn't know it's possible to share RegEx like that, pretty neat! 

Do you think we should add a new test checking against the new RegEx then just to be sure? Because RegEx are always something I would check with integration tests imo...

Or do you think it's enough to maybe link to your RegEx101 in a comment inside the `refine`?

HiDeoo: An integration test seems quite much for this simple case, but we can definitely add a unit test similar to the ones in `packages/starlight/__tests__/basics/config-errors.test.ts` to make sure it catches an invalid value and shows the proper error message. We can also update the valid config one to check a valid value is accepted.

trueberryless: Oh, sry, I'm no export with testing in general, so I didn't know there was a difference, but now that you write it, I think integrations means end-to-end (?) and unit just testing a simple function or in this case RegEx config, is that right?

trueberryless:
Shall we include in the changelog that setting `srcDir` bypasses the validation?

packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -11,7 +11,10 @@ function parseStarlightConfigWithFriendlyErrors(config: StarlightUserConfig) {
 }
 
 test('parses valid config successfully', () => {
-	const data = parseStarlightConfigWithFriendlyErrors({ title: '' });
+	const data = parseStarlightConfigWithFriendlyErrors({
+		title: '',
+		routeMiddleware: './src/routeData.ts',
+	});
```
delucis, line 17: Part of the intention of this test (not well expressed in the test name) was to ensure a bare minimum configuration passes validation, so we should probably not also use it to test a valid `routeMiddleware` option. We can do that in an additional test (and just check we get the expected `'./src/routeData.ts'` to `['./src/routeData.ts']` data mapping rather than a full inline snapshot).

packages/starlight/utils/user-config.ts
```diff
@@ -233,6 +233,17 @@ const UserConfigSchema = z.object({
...
 		.default([])
+		.refine(
+			(middlewares) => {
+				// Regex pattern to match invalid middleware paths: https://regex101.com/r/kQH7xm/1
+				const invalidPathRegex = /^(?:.?\/)?src\/middleware(?:\/index)?\.[jt]s$/;
+				return !middlewares.some((middleware) => invalidPathRegex.test(middleware));
+			},
+			{
+				message:
+					'Middleware path cannot be one of the following: "src/middleware.js", "src/middleware.ts", "src/middleware/index.js", "src/middleware/index.ts" because they would conflict with Astro’s middleware. Read more about Astro’s middleware: https://docs.astro.build/en/guides/middleware/',
```
delucis, line 244: This is a great start for this error message but I think we can do even better!

We have some good guidance on writing error messages in the Astro repo: https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/README.md

In particular:

> - Begin with **what happened** and **why**. (ex: `Could not use {feature} because Server-side Rendering is not enabled.`)
> - Then, **describe the action the user should take**. (ex: `Update your Astro config with `output: 'server'` to enable Server-side Rendering.`)

and

> - Describe the _what_, _why_ and _action to take_ from the user's perspective. Assume they don't know Astro internals, and care only about how Astro is _used_. (ex: `You are missing...` vs `Astro/file cannot find...`)

Right now the message is good at the what/why but we can make it more from the user perspective, and we need answer “OK, what do I do now?” for them, e.g.

```suggestion
					'A `routeMiddleware` file path in your Starlight config conflicts with Astro’s middleware locations. It cannot be one of the following: "src/middleware.js", "src/middleware.ts", "src/middleware/index.js", "src/middleware/index.ts".\n\n' +
						'You should move your Starlight route middleware file somewhere else like `src/starlightRouteData.ts` and update the `routeMiddleware` file path to match.\n\n' +
						'- More about Starlight route middleware: https://starlight.astro.build/guides/route-data/#how-to-customize-route-data\n' +
						'- More about Astro middleware: https://docs.astro.build/en/guides/middleware/',
```

Some things I’m consciously doing here:

- Mention the _location_ of where things are going wrong:
  - Where’s the issue? “your Starlight config”
  - Which part? The “`routeMiddleware`” option
- Mention what action to take to resolve the issue

One more thing we could do to improve this potentially is to use the actual erroneous path to avoid listing all the possible variations. This isn’t possible with `.refine()` but is possible with `.superRefine()` which uses Zod’s `addIssue()` method to flag errors and can be built dynamically with the actual inaccurate path rather than using a static `message` string.

delucis (changes requested):
Nice work @trueberryless and @HiDeoo getting testing set up as well 🙌 

Suggesting some further details.

packages/starlight/__tests__/basics/config-errors.test.ts
trueberryless: Done! I also update the name so it says that the purpose is a bare minimum test!

packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -10,7 +10,7 @@ function parseStarlightConfigWithFriendlyErrors(config: StarlightUserConfig) {
 	);
 }
 
-test('parses valid config successfully', () => {
+test('parses bare minimum valid config successfully', () => {
```
trueberryless, line 13: This small specification can go with this PR I think...

packages/starlight/utils/user-config.ts
```diff
@@ -233,6 +233,21 @@ const UserConfigSchema = z.object({
 		.transform((string) => [string])
 		.or(z.string().array())
 		.default([])
+		.superRefine((middlewares, ctx) => {
+			// Regex pattern to match invalid middleware paths: https://regex101.com/r/kQH7xm/1
+			const invalidPathRegex = /^(?:.?\/)?src\/middleware(?:\/index)?\.[jt]s$/;
```
delucis, line 238: I think this RegEx is more accurate? `"src/middleware.ts"` would mean you were importing from the `src` npm package rather than a local path, which while probably wrong, is also not what we’re checking for. Instead we’re checking for relative paths like `./src/middleware.ts` or potentially for a path starting with `/` which Vite treats the same way:
```suggestion
			// Regex pattern to match invalid middleware paths: https://regex101.com/r/kQH7xm/2
			const invalidPathRegex = /^\.?\/src\/middleware(?:\/index)?\.[jt]s$/;
```

packages/starlight/utils/user-config.ts
```diff
@@ -233,6 +233,21 @@ const UserConfigSchema = z.object({
...
+			const invalidPaths = middlewares.filter((middleware) => invalidPathRegex.test(middleware));
+			for (const invalidPath of invalidPaths) {
+				ctx.addIssue({
+					code: 'custom',
+					message:
+						'A `routeMiddleware` file path in your Starlight config conflicts with Astro’s middleware locations. It cannot be one of the following: "src/middleware.js", "src/middleware.ts", "src/middleware/index.js", "src/middleware/index.ts".\n\n' +
+						`You should move your Starlight route middleware file \`${invalidPath}\` somewhere else like \`src/starlightRouteData.ts\` and update the \`routeMiddleware\` file path to match.\n\n` +
```
delucis, line 245: I think now we have the exact path we can rewrite with that in mind:

```suggestion
						`The \`"${invalidPath}"\` path in your Starlight \`routeMiddleware\` config conflicts with Astro’s middleware locations.\n\n` +
						`You should rename \`${invalidPath}\` to something else like \`./src/starlightRouteData.ts\` and update the \`routeMiddleware\` file path to match.\n\n` +
```

We don’t need general descriptions like “move your Starlight route middleware file” when we can say “rename `./src/middleware.js`”.

packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -253,3 +253,32 @@ test('errors with sidebar entry that includes `items` and `autogenerate`', () =>
...
+		routeMiddleware: './src/routeData.ts',
+	});
+	expect(data.routeMiddleware).toEqual(['./src/routeData.ts']);
+});
+
+test("errors if one route middleware is conflicting path with Astro' middleware", () => {
+	expect(() =>
+		parseStarlightConfigWithFriendlyErrors({
+			title: 'Test',
+			routeMiddleware: ['src/middleware.ts', 'src/routeData.ts'],
```
delucis, line 269: This should be updated for the new RegEx:

```suggestion
			routeMiddleware: ['./src/middleware.ts', './src/routeData.ts'],
```

packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -253,3 +253,32 @@ test('errors with sidebar entry that includes `items` and `autogenerate`', () =>
...
+		})
+	).toThrowErrorMatchingInlineSnapshot(
+		`
+		"[AstroUserError]:
+			Invalid config passed to starlight integration
+		Hint:
+			A \`routeMiddleware\` file path in your Starlight config conflicts with Astro’s middleware locations. It cannot be one of the following: "src/middleware.js", "src/middleware.ts", "src/middleware/index.js", "src/middleware/index.ts".
+			
+			You should move your Starlight route middleware file \`src/middleware.ts\` somewhere else like \`src/starlightRouteData.ts\` and update the \`routeMiddleware\` file path to match.
```
delucis, line 278: And the error message will need updating too:

```suggestion
			The \`"./src/middleware.ts"\` path in your Starlight \`routeMiddleware\` config conflicts with Astro’s middleware locations.
			
			You should rename \`./src/middleware.ts\` to something else like \`./src/starlightRouteData.ts\` and update the \`routeMiddleware\` file path to match.
```

delucis (changes requested):
🙌

trueberryless:
I just realized that the `routeMiddleware` configuration option is currently missing in the configuration reference docs. I added it in the latest commit, maybe double check the wording as I'm no docs expert, but I think we can add this in this PR as well...

delucis (approved):
Great work here @trueberryless — I think it’s ready to go!

Made a couple of tiny final tweaks:

- Now we have the config reference (well spotted that it was missing!) I removed the note from the route data guide about not conflicting with Astro’s middleware. My rationale for this is that in the context of a guide the focus is on accomplishing a task and these steps already suggest a specific non-conflicting file name for the middleware module. With the great new error message plus the note in config reference, I think it’s OK not to add this extra information in the middle of the goal-oriented guide.

- Tweaked some config reference wording just slightly to remove “custom” from “custom route middleware” as all route middleware is technically “custom”, so I think we could simplify there.

- Fixed a typo in a test name.

Thanks for the thoughtful way you went from experiencing issues yourself with the clash between middleware paths to thinking of this solution to providing clear errors to Starlight users. Lovely work!

PR #2261, opened 2024-08-27
Draft: Support custom icons

shubham-padia:
#### Description

- Closes https://github.com/withastro/starlight/discussions/2251

This PR adds the ability to pass icon attribute to the Aside component. This icon attribute can be any icon supported by starlight by default.

PR #3024, opened 2025-03-26
Add support for custom icons

HiDeoo:
#### Description

- Closes #967
- Closes #2807

This PR is a **very early draft** adding support for [Iconify](https://iconify.design/) and local icons in Starlight.

_The early draft PR is opened to provide some material for discussions and also evaluate potential changes in involved libraries. A lot of work, including more robust implementations for some parts, more tests, and documentation, is still needed._

Preview test page: https://deploy-preview-3024--astro-starlight.netlify.app/test/

SCR-20250326-pxwk

#### Implementation

Most of the implementation relies on the [Astro Icon](https://www.astroicon.dev/) library, altho Starlight does not only uses icons through components but also in different contexts, e.g. remark plugins.

A few high-level details about the implementation:

- No breaking changes are introduced for existing icons. Starlight's built-in icons are still available and can be used as before.
- Starlight does not ship with any Iconify icon set (built-in icons are still part of Starlight and not transformed in an Iconify icon set).
- Users can choose to install any Iconify icon set they want to use.
- Everywhere a built-in icon can be used, an Iconify or local icon can be used as well.
- Custom aside icons is built upon https://github.com/withastro/starlight/pull/2261 (which should be merged first for credits), refactored to also support Iconify icons.

#### Astro Icon configuration

Astro Icon is an integration that supports various [configuration options](https://www.astroicon.dev/reference/configuration/).

Just like [`expressiveCode`](https://starlight.astro.build/reference/configuration/#expressivecode), Astro Icon options are currently exposed using an `icons` (placeholder name) property in the Starlight configuration.

#### Astro Icon notes

Here are a few notes about Astro Icon that I gathered so far that I would like to discuss with the maintainers of the project (still need to put them in a more friendly format):

- The [issue](https://github.com/natemoo-re/astro-icon/issues/240) regarding icon sets not being found when icon sets are part of a library:
  - This issue seems to be related to how installed icon sets are determined which was changed in [this PR](https://github.com/natemoo-re/astro-icon/pull/175) to fix other issues, e.g. different results with different package managers or when `NODE_ENV` is set to `production`.
  - This issue does not affect the current implementation as Starlight does not ship with any Iconify icon set and users have to install them themselves in their projects.
  - This could be an issue if a Starlight plugin decides to ship with an Iconify icon set and use it in its components.
  - I didn't spend enough time yet on the issue, but it _seems_ like the [previous approach](https://github.com/natemoo-re/astro-icon/pull/175/files#diff-7ed62ba2c3bb3f700d1d0007e3db1c910366a968adbf54409095f7897229084b) may work by appending `--prod=false` to the list command from my initial and quick tests. Will need to investigate more.
- When [using the `@iconify/utils` helper `loadCollectionFromFS()`](https://github.com/natemoo-re/astro-icon/blob/85cfae2426b8a94ca7f25429dba307558f232345/packages/core/src/loaders/loadIconifyCollections.ts#L85) to load collections, by default this helper [uses `process.cwd()`](https://github.com/iconify/iconify/blob/3c594380b3410123fa429d3f13aab3c52083fca4/packages/utils/src/loader/fs.ts#L28) to load collections. Same goes for the [`importDirectory()` helper in `@iconify/tools`](https://iconify.design/docs/libraries/tools/import/directory.html) used to load local icons.
  - This can be an issue in some cases, e.g. in Starlight E2E tests where the current working directory is not the project root.
  - This PR includes a patch to Astro Icon using an absolute path.
- As Starlight uses icons in contexts where Astro Icon cannot be used, e.g. remark plugins, we also need to load local and Iconify icons in such contexts:
  - This PR ships (very) similar implementations to Astro Icon's Iconify and local icons loaders.

#### Remaining tasks

- [ ] Address all remaining `// TODO(HiDeoo)` comments
- [ ] Changesets covering all changes across all changed packages
  - [ ] `@astrojs/starlight`
  - [ ] `@astrojs/starlight-markdoc`
- [ ] Remove some test files
  - [ ] `docs/src/icons/test.svg`
  - [ ] `docs/src/content/docs/test.mdx`
  - [ ] `examples/markdoc/src/content/docs/test.mdoc`
- [ ] Uninstall `@iconify-json/mdi` from `starlight-docs` and `@example/starlight-markdoc`
- [ ] Documentation
  - [ ] Very early draft documentation just to expose what changed and should definitely be improved
  - [ ] In a lot of places, we mention "one of Starlight’s built-in icons" with a link. The sentence should be updated (maybe "supported icons"?) and the link should be updated too if needed.
- [ ] Merge https://github.com/withastro/starlight/pull/2261 before this PR


VisualEhrmanntraut:
are there still plans to finish this PR? it would be much better if the custom icon support was integrated into starlight than making copies of the components to integrate astro-icon manually

HiDeoo:
> are there still plans to finish this PR? it would be much better if the custom icon support was integrated into starlight than making copies of the components to integrate astro-icon manually

This is still planned. As mentioned in the PR, this was a very early draft to get some conversation going. I have plan to get back to it during the first quarter, probably after the release of Astro v6.

sanscontext:
Was there specific feedback/testing that was wanted on this PR?

PR #3025, opened 2025-03-26
Make `social` icon configuration more flexible

delucis:
#### Description

- This PR changes how social links are configured in the Starlight options.

- It switches from a shorthand object syntax to an array of explicit link items:

  ```diff
  - social: {
  -   github: 'https://github.com/withastro/starlight',
  -   discord: 'https://astro.build/chat',
  - },
  + social: [
  +   { icon: 'github', label: 'GitHub', href: 'https://github.com/withastro/starlight' },
  +   { icon: 'discord', label: 'Discord', href: 'https://astro.build/chat' },
  + ],
  ```

- This is a little unfortunate in that it’s a more verbose configuration approach, but it has a few advantages for long-term maintenance:

	- Icons no longer need dedicated built-in labels and code to be used as social links. Any icon can be used if people want (avoiding us needing to decide what counts as a valid social link and what doesn’t).

	- User control of labels means that less common links can be labelled accurately. For example, we avoided adding an `email` social icon because the label could vary depending on site language and context. This change allows people to use the existing `email` icon with a label of their choosing.

	- Following #3024, this should be updated to also support user-installed icons reducing the pressure on Starlight to contain icons for every platform in the known universe.

#### Not in scope

- Currently there’s no implementation of localized labels for social links in this PR. There are a few obstacles to implementing that nicely at the moment, and given labels are already not localizable (and for many cases don’t require it where the label is a platform name), we decided to skip that until we refactor some internals to make it easier to support that elegantly.

#### Also considered

- I also considered whether some hybrid of our current shorthand syntax and the new explicit syntax could make sense to keep config for common platforms terse, but I decided the slightly less verbose syntax didn’t justify users needing to understand two subtly different link shapes (e.g. some icons not needing an explicit `label`).


docs/src/components/social-links-type.astro
```diff
@@ -1,7 +1,4 @@
 ---
-import { socialLinks } from '../../../packages/starlight/schemas/social';
-
-const socials = [...socialLinks].sort((a, b) => a.localeCompare(b, 'en'));
 ---
 
-<code>{`Partial<Record<${socials.map((social) => `'${social}'`).join(' | ')}, string>>`}</code>
+<code set:text="Array<{ label: string; icon: string; href: string }>" />
```
delucis, line 4: I actually removed this component from the English docs now it no longer needs to generate this list. But as it’s still in use in other languages, I’ve updated the code here with a placeholder value until we can remove the component entirely.

packages/starlight/components/Head.astro
```diff
@@ -84,12 +84,13 @@ if (Astro.site) {
...
+if (twitterLink) {
 	headDefaults.push({
 		tag: 'meta',
 		attrs: {
 			name: 'twitter:site',
-			content: new URL(config.social.twitter.url).pathname,
+			content: new URL(twitterLink.href).pathname,
 		},
 	});
 }
```
delucis, line 96: This change will overlap with #2927 but thankfully it’s small and easy to update for whichever PR merges second.

packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -122,8 +122,7 @@ test('errors with bad social icon config', () => {
 		"[AstroUserError]:
 			Invalid config passed to starlight integration
 		Hint:
-			**social.unknown**: Invalid enum value. Expected 'twitter' | 'mastodon' | 'github' | 'gitlab' | 'bitbucket' | 'discord' | 'gitter' | 'codeberg' | 'codePen' | 'youtube' | 'threads' | 'linkedin' | 'twitch' | 'azureDevOps' | 'microsoftTeams' | 'instagram' | 'stackOverflow' | 'x.com' | 'telegram' | 'rss' | 'facebook' | 'email' | 'reddit' | 'patreon' | 'signal' | 'slack' | 'matrix' | 'openCollective' | 'hackerOne' | 'blueSky' | 'discourse' | 'zulip' | 'pinterest' | 'tiktok' | 'nostr' | 'backstage' | 'farcaster' | 'confluence' | 'jira' | 'storybook' | 'npm' | 'sourcehut' | 'substack', received 'unknown'
-			**social.unknown**: Invalid url"
+			**social**: Expected type \`"array"\`, received \`"object"\`"
```
HiDeoo, line 125: I guess technically the error map could for the `social` error path detect if the received type is `object` and in this case, show a more helpful message like the changeset altho I think considering how small the syntax change is and well described in the changelog, it's fine to keep it simple?

docs/src/components/social-links-type.astro
HiDeoo: Should `icon` link to the icon page in the current locale like the English version does?

HiDeoo (approved):
This looks amazing and definitely more flexible for users 👏 

Left a few link suggestions and thoughts but nothing blocking 👍

docs/src/components/social-links-type.astro
delucis: Skipped before out of laziness, but done in 8fa2f66. Thanks for the suggestion!

.changeset/gold-sloths-unite.md
```diff
@@ -0,0 +1,24 @@
...
+'@astrojs/starlight': minor
+---
+
+Makes `social` configuration more flexible.
+
+⚠️ **BREAKING CHANGE:** The `social` configuration option has changed syntax. You will need to update this in `astro.config.mjs` when upgrading.
+
+Previously, a limited set of platforms were supported using a shorthand syntax with labels built in to Starlight. While convenient, this approach was less flexible and required dedicated code for each social platform added.
+
+Now, you must specify the icon and label for each social link explicitly and you can use any of [Starlight’s built-in icons](https://starlight.astro.build/reference/icons/) for social links.
```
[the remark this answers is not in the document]
delucis: Thought about this and I _think_ I like linking to the top of the page for a bit more general information about icons rather than directly to just the list. It’s not much content to scroll if you’ve seen it before and want to find an icon, but it means you land somewhere with a few more signposts than if you were to go directly to the icon list.

packages/starlight/__tests__/basics/config-errors.test.ts
delucis: Very good idea. Added this in 69ed1a68 🙌

PR #3026, opened 2025-03-27
Fix list styling issue

HiDeoo:
#### Description

Initially reported in Discord by Chris and Sarah, this PR fixes a list styling issue that can happen when the last child element of a list item is a script tag.

This can naturally happen in Starlight, e.g. in a list of steps where the last element of a step is the first occurence of a set of tabs in the page. The issue is visible in the Starlight documentation of the `<Steps>` component:

|         | Before                                          | After           |
| ------- | ----------------------------------------------- | --------------- |
| URL     | https://starlight.astro.build/components/steps/ | // TODO(HiDeoo) |
| Preview | SCR-20250327-kkgd | SCR-20250327-kkkr |

We cannot really target the built-in tabs component to fix this issue, as we could have more components that could lead to the same issue in the future, or the user could have their own component that triggers the issue.

We also cannot fix this issue if only the last child element of a list item is a script tag, as even if rare, the last n child elements could be script tags too.

This leads to quite the complex selector to try to handle all cases, I tried to comment it as much as possible to make it more understandable.

### DIFF
--- a/.changeset/calm-cobras-breathe.md
+++ b/.changeset/calm-cobras-breathe.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a potential list styling issue if the last element of a list item is a `<script>` tag.
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -24,7 +24,29 @@
 
 .sl-markdown-content
 	li
-	> :last-child:not(li, ul, ol, a, strong, em, del, span, input, code, br, :where(.not-content *)) {
+	> :is(
+		:last-child:not(
+				li,
+				ul,
+				ol,
+				a,
+				strong,
+				em,
+				del,
+				span,
+				input,
+				code,
+				br,
+				script,
+				:where(.not-content *)
+			),
+			/**
+			 * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
+			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
+			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
+			 */
+		:not(script):has(~ script:last-child):not(:has(~ :not(script)))
+	) {
 	margin-bottom: 1.25rem;
 }
 


packages/starlight/style/markdown.css
```diff
@@ -24,7 +24,29 @@
...
+				br,
+				script,
+				:where(.not-content *)
+			),
+			/**
+			 * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
+			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
+			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
+			 */
+		:not(script):has(~ script:last-child):not(:has(~ :not(script)))
```
delucis, line 48: Noting that `:has()` is technically not covered by our [documented browser support matrix](https://github.com/withastro/starlight/blob/0edf4d7e3a315b9d88049faa65f83ca5a8298e88/CONTRIBUTING.md#L278) (not supported in Chrome/Edge 88–104 and Firefox 98–120). So far we only used it once in a non-critical hover style on `<details>`.

I think it’s OK for us to start to drop those older browsers (they’re releases from 2023 and earlier with <0.5% usage) but we should probably update our documented list (and maybe mention that in the changeset?)

It would also be possible here to keep the existing selector with the `:has()` variant added in addition as a progressive enhancement I guess? But not sure if it’s worth it.

delucis (commented):
Thanks for the fix. I think the selector is good and the comment is helpful 👍 

Left a comment for us to think about.

trueberryless:
> ## Pull Request Overview
> 
> This PR fixes a list styling issue that occurs when the last element of a list item is a <script> tag, which can interfere with proper styling in components like <Steps>.  
> - Added a changeset to update the '@astrojs/starlight' package with a patch  
> - Documented the fix for better traceability and reference in future debugging
> 
> <details>
> <summary>Files not reviewed (1)</summary>
> 
> * **packages/starlight/style/markdown.css**: Language not supported
> </details>
> 
> 
> 

What is THAT? Please don't tell me this is now enabled by default on GitHub... 🙏

HiDeoo:
> What is THAT? Please don't tell me this is now enabled by default on GitHub... 🙏

No, we just were curious, but it's manual https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review

packages/starlight/style/markdown.css
HiDeoo: Good call, it's already bumped in #2322 which would fix it for Firefox but not Chrome. I'll update the PR so that it's only 1 bump for both PRs 👍

HiDeoo: Updated our browserslist query to support both this PR and #2322 (which I'll need to update to remove the equivalent change). It's a 0.8% coverage decrease. I've also bumped the changeset to a minor version and also described the change in it (open to suggestion regarding the wording).

.changeset/calm-cobras-breathe.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/starlight': minor
+---
+
+Fixes a potential list styling issue if the last element of a list item is a `<script>` tag.
+
+⚠️ BREAKING CHANGE:
+
+This change slightly reduces browser compatibility coverage by 0.8% by dropping support for Chromium-based browsers prior to version 105 and Firefox-based browsers prior to version 121. You can find a list of supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+105%2C+Edge+%3E%3D+105%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all).
```
delucis, line 9: This wording is good, but I wonder if we should maybe downplay it _slightly_ at least given that for now even those older browsers will still mostly work (it’s only a small list styling detail that will regress).

```suggestion
This release drops official support for Chromium-based browsers prior to version 105 (released 30 August 2022) and Firefox-based browsers prior to version 121 (released 19 December 2023). You can find a list of currently supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+105%2C+Edge+%3E%3D+105%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all).

With this release, Starlight-generated sites will still work fine on those older browsers except for this small detail in list item styling, but future releases may introduce further breaking changes for impacted browsers, including in patch releases.
```

delucis (approved):
Looks good I think. Left a small suggestion on the changeset for your consideration.

.changeset/calm-cobras-breathe.md
HiDeoo: Love it (was not also liking a lot my initial version and that's why I mentioned being opened to suggestions 😆), really liking how it emphasizes the dates and how it's limited to this change at the moment :thumbsup:

PR #3030, opened 2025-04-02
fix(core): `isFallback` type is `true` instead of `boolean`

trueberryless:
#### Description

This is rather a question / discussion of the decision why the `isFallback` type is defined as type `true` instead of `boolean`:

https://github.com/withastro/starlight/blob/bdb05e704a6cf06a029367f99b6bf2f575e5a5f3/packages/starlight/utils/routing/types.ts#L61-L73

This has the consequence that my editor gives me the warning (when I use `<StarlightPage>`:

image

My var `isFallback` is of type boolean: 

image

I'm not sure if this is an oversight or has a specific reason, therefore I'm creating this as a draft PR...

### DIFF
--- a/.changeset/brown-bottles-stare.md
+++ b/.changeset/brown-bottles-stare.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Updates the type of the `isFallback` field in route data from `true` to `boolean`, keeping it optional but allowing `false` as a possible value.
--- a/packages/starlight/__tests__/basics/starlight-page-route-data.test.ts
+++ b/packages/starlight/__tests__/basics/starlight-page-route-data.test.ts
@@ -64,11 +64,13 @@ test('adds custom data to route shape', async () => {
 		hasSidebar: false,
 		dir: 'rtl',
 		lang: 'ks',
+		isFallback: true,
 	};
 	const data = await generateStarlightPageRouteData({ props, url: starlightPageUrl });
 	expect(data.hasSidebar).toBe(props.hasSidebar);
 	expect(data.entryMeta.dir).toBe(props.dir);
 	expect(data.entryMeta.lang).toBe(props.lang);
+	expect(data.isFallback).toBe(props.isFallback);
 });
 
 test('adds custom frontmatter data to route shape', async () => {
--- a/packages/starlight/utils/routing/types.ts
+++ b/packages/starlight/utils/routing/types.ts
@@ -68,7 +68,7 @@ export interface Route extends LocaleData {
 	/** The slug or unique ID if using the `legacy.collections` flag. */
 	id: string;
 	/** True if this page is untranslated in the current language and using fallback content from the default locale. */
-	isFallback?: true;
+	isFallback?: boolean;
 	[key: string]: unknown;
 }
 
--- a/packages/starlight/utils/routing/types.ts
+++ b/packages/starlight/utils/routing/types.ts
@@ -67,8 +67,8 @@ export interface Route extends LocaleData {
 	slug: string;
 	/** The slug or unique ID if using the `legacy.collections` flag. */
 	id: string;
-	/** True if this page is untranslated in the current language and using fallback content from the default locale. */
-	isFallback?: true;
+	/** Whether this page is untranslated in the current language and using fallback content from the default locale. */
+	isFallback?: boolean;
 	[key: string]: unknown;
 }
 
--- a/packages/starlight/utils/routing/types.ts
+++ b/packages/starlight/utils/routing/types.ts
@@ -68,7 +68,7 @@ export interface Route extends LocaleData {
 	/** The slug or unique ID if using the `legacy.collections` flag. */
 	id: string;
 	/** True if this page is untranslated in the current language and using fallback content from the default locale. */
-	isFallback?: boolean;
+	isFallback?: true;
 	[key: string]: unknown;
 }
 
--- a/packages/starlight/utils/starlight-page.ts
+++ b/packages/starlight/utils/starlight-page.ts
@@ -76,12 +76,19 @@ const validateSidebarProp = (
  */
 export type StarlightPageProps = Prettify<
 	// Remove the index signature from `Route`, omit undesired properties and make the rest optional.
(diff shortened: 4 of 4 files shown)


trueberryless:
@delucis Okay, is there a suggestion on how to define the property then if my own var is dynamic?

Like the worst case scenario would be to have duplicated code, like this (simplified, in my case that would be a lot of unnecessary duplicated code):

```html

{isFallback === true && <StarlightPage isFallback={isFallback} />}
{isFallback === false && <StarlightPage />}
```

trueberryless:
Okay, this looks ugly, but it does the job quite clean:

```html
<StarlightPage {...(isFallback ? { isFallback: true } : {})} />
```

trueberryless:
@delucis No, it doesnt like it that way, I tried it with `isFallback={isFallback ? true : undefined}` as well...

image

HiDeoo:
> Does `<StarlightPage isFallback={isFallback || undefined} />` pass the type check? Not opposed to loosening the type to `boolean` by the way if that’s a nicer experience with the component, just explaining the reason it’s typed the way it is.

This should change based on the user [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) setting. If set to `true`, you would need to conditionally add that to a props object and spread these props (this is what we do when generating route data for a `<StarlightPage>`).

We could change it only for `<StarlightPage>` tho.

trueberryless:
> This should change based on the user [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) setting. If set to `true`, you would need to conditionally add that to a props object and spread these props (this is what we do when generating route data for a `<StarlightPage>`).

Yeah, I use strictest (if this helps):

```json
{
  "extends": "astro/tsconfigs/strictest",
  "include": [".astro/types.d.ts", "**/*"],
  "exclude": ["dist"]
}
```

delucis:
> We could change it only for `<StarlightPage>` tho.

I think changing it everywhere is fine probably. The underlying code should handle `false` correctly I would assume?

HiDeoo:
> > We could change it only for `<StarlightPage>` tho.
> 
> I think changing it everywhere is fine probably. The underlying code should handle `false` correctly I would assume?

Probably, I guess we would just need to double check we don't have some things like:

```js
const thing = isFallback ?? 'something'
```

trueberryless:
Oh no, I have not been following the conversion, now maybe you wanna revert the recent commit...

trueberryless:
> There’s a reason we recommend issues before PRs 😂 That way we figure out the problem and solution _before_ going back and forth in a PR 🙂

Sorry about that, but as mentioned in the PR description, it was meant as a discussion (but with an code example -> therefore the PR). In the future I can open an issue with code blocks if that's more desired!

> So should be fine to globally set the type to `boolean | undefined`.

Changing the type to `isFallback: boolean | undefined` has some unwanted consequences, like you have to pass `isFrontmatter` in many places (which changes more code than we want) in order to pass the [similar root shape test](https://github.com/withastro/starlight/blob/b3796505c8337f3ff69108a93df3fe68810f8a65/packages/starlight/__tests__/basics/starlight-page-route-data.test.ts#L503-L516):

https://github.com/withastro/starlight/blob/b3796505c8337f3ff69108a93df3fe68810f8a65/packages/starlight/utils/routing/index.ts#L50-L56

https://github.com/withastro/starlight/blob/b3796505c8337f3ff69108a93df3fe68810f8a65/packages/starlight/utils/routing/data.ts#L148

So I changed the type to: `isFallback?: boolean` which solves those consequences.
Let me know if I missed something!

HiDeoo:
> So should be fine to globally set the type to `boolean | undefined`.

Small question: if the change is made globally, is there any specific reason to go for `boolean | undefined`, which means 2 different values indicate a fallback page (`false` or `undefined`) rather than just `boolean` where `false` indicates a fallback page and `true` indicates a non-fallback page?

My initial thinking is that it could mean less change on our side but maybe a slightly more-confusing route data value?

trueberryless:
> Small question: if the change is made globally, is there any specific reason to go for boolean | undefined, which means 2 different values indicate a fallback page (false or undefined) rather than just boolean where false indicates a fallback page and true indicates a non-fallback page?

The main reason as you said is less changes, because there are some parts where the route data is initialized without the `isFallback` property (because it should be `false`). We can of course change it to `isFallback: boolean`, but this means that you have to pass it everywhere. I think this would also mean you have to pass it to the `<StarlightPage>` (_I think!_) even if you don't care, which would be a breaking change? 🤔

Just to make one thing sure: Did you accidentally swap the meaning in the question? Because `false` (and `undefined`) define a non-fallback page and `true` defines a fallback page. Maybe this is now unclear in the TS docs of the Route interface...

HiDeoo:
> Did you accidentally swap the meaning in the question?

Thanks, fixed it :raised_hands:

Yeah, the question was mostly related to the quote at the beginning from Chris. It would still be possible to have it working in `<StarlightPage>` [as currently documented](https://starlight.astro.build/guides/pages/#isfallback) (`boolean` with a default value of `false`). My concern is mostly around the fact that 2 values indicate the same thing rather than one like it's the case currently so I'm wondering if maybe I'm missing something that Chris may have thought about.

> which would be a breaking change?

I think the current changes are already breaking. A user or plugin reading route data and having something like `if (isFallback === undefined)` would no longer be working as before.

delucis:
> if the change is made globally, is there any specific reason to go for `boolean | undefined`, which means 2 different values indicate a non-fallback page (`false` or `undefined`) rather than just `boolean` where `false` indicates a non-fallback page and `true` indicates a fallback page?

I think that adding `false` as an option is OK yeah. It is _technically_ breaking like you mentioned, but I feel like people checking specifically for `isFallback === undefined` seems unlikely in this case? And boolean checks like `isFallback: someValue === fallbackValue` are pretty common for flags like this so it seems silly for us to insist on `isFallback: someValue === fallbackValue ? true : undefined` instead.

I’m happy releasing this in a patch and we can revert if anyone complains.

delucis (approved):

PR #3034, opened 2025-04-04
docs: add `credits` type to configuration reference

HiDeoo:
#### Description

This PR adds the type and default value for the `credits` configuration option in the documentation. Noticed that it was missing while working on a different PR.

The type is fairly obvious, but this should still be documented nevertheless, just for consistency and completeness.


delucis (approved):
Thank you for spotting it! I blame releasing this feature live on stream for forgetting this 😁

PR #3039, opened 2025-04-06
Add home icon and Homepage social

voidxno:
PR adding home icon and Homepage social

Reason:
- Need it myself. Have observed the other PR's around extra icon collections (very good). Independent of that, think a 'home' icon should be in standard Starlight collection.

Sample:

Created/drawn from scratch by me (screenshot below). Give all rights to Starlight, unconditionally. Basically used existing standard Starlight icons style as template. Baseline, alignment, size logic, thickness, curves, so on:
- Drawn in Corel Designer, exported SVG (24x24), tuned/optimized SVG ([svgo](https://github.com/svg/svgo)).

_PS: Sorry if I have missed elements that should be in PR. Please point me in correct direction, will fix._


delucis:
Thanks for your PR @voidxno! I think we won’t accept this PR in this form for a couple of reasons although I appreciate this nonetheless and you taking the time to draw the icon.

We are currently in the process of making Starlight’s icon and social link configuration more flexible and I think that work would already unlock what you want to do. Relevant PRs:

- #3024 
- #3025 

We also so far have always added icons from the open source Unicons icon set when adding generic icons, so for consistency if adding a built-in “home” icon, we would probably do the same to ensure the design matches between different icons.

In the meantime, I’d recommend using a component override to do this. There’s an example of exactly your use case with a custom SVG here in this thread: https://github.com/withastro/starlight/discussions/3014#discussioncomment-12633054

PR #2949, opened 2025-03-09
i18n(de): change translation of icon

trueberryless:
#### Description

- The German translation currently uses `Symbol` as a translation of `Icon`. Maybe we should change it to `Icon` because it is also a commonly used work in German.
- Native speaker needed!

Some words read a little bit weird now tho:

- `Dateiicon` (maybe `Datei-Icon`?)

Okay, maybe it's only one word :)

PR #3060, opened 2025-04-07
fix(de): replace translation of "icon" in german language [i18nIgnore]

trueberryless:
#### Description

- This PR replaces translations of the English word `icon` from the German translation `Icon` to `Symbol` so it is consistent across all pages. Therefore the i18nIgnore, because this should not mark pages as completed in any way.
- This is a follow-up PR of #2949 where Chris and I agreed on making the translation consistent the other way around (initially I wanted to replace all `Symbol` with `Icon`, but [Wikipedia](https://de.wikipedia.org/wiki/Icon_(Computer)) says, `Symbol` is the generally more acceptable translation of `icon`
- I again ask a native speaker to review this PR, maybe @delucis wants to step up again and help out as always 💜 (I think I didn't miss any `Icon`s with my global VS Code search...)

### DIFF
--- a/docs/src/content/docs/de/reference/overrides.md
+++ b/docs/src/content/docs/de/reference/overrides.md
@@ -115,7 +115,7 @@ So kannst du eine Benutzeroberfläche für alternative Suchanbieter hinzufügen,
 
 **Standardkomponente:** [`SocialIcons.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/SocialIcons.astro)
 
-Diese Komponente wird in der Kopfzeile der Website gerendert und enthält soziale Symbol-Links.
+Diese Komponente wird in der Kopfzeile der Website gerendert und enthält Symbol-Links zu sozialen Netzwerken.
 Die Standard&shy;implementierung verwendet die Option [`social`](/de/reference/configuration/#social) in der Starlight-Konfiguration, um Symbole und Links darzustellen.
 
 #### `ThemeSelect`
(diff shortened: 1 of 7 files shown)


docs/src/content/docs/de/reference/overrides.md
```diff
@@ -115,8 +115,8 @@ So kannst du eine Benutzeroberfläche für alternative Suchanbieter hinzufügen,
 
 **Standardkomponente:** [`SocialIcons.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/components/SocialIcons.astro)
 
-Diese Komponente wird in der Kopfzeile der Website gerendert und enthält Links zu sozialen Symbolen.
-Die Standard&shy;implementierung verwendet die Option [`social`](/de/reference/configuration/#social) in der Starlight-Konfiguration, um Icons und Links darzustellen.
+Diese Komponente wird in der Kopfzeile der Website gerendert und enthält soziale Symbol-Links.
```
delucis, line 118: Do you think being more explicit here could help make it more understandable?

```suggestion
Diese Komponente wird in der Kopfzeile der Website gerendert und enthält Symbol-Links zu soziale Netzwerke.
```

delucis (commented):
Looks great! 🙌 

Just left one query.

docs/src/content/docs/de/reference/overrides.md
trueberryless: Yep, sounds good, but shouldn't it be: "zu sozialen Netzwerken", "zu mir", 3. Fall...?

PR #3061, opened 2025-04-07
i18n(de): update `components/badges`

trueberryless:
#### Description

This PR updates the German version of the `components/badges` page with the changes from #2924.


HiDeoo (approved):
Thanks for the update 👍

PR #3064, opened 2025-04-08
i18n(ru): update `reference/route-data.mdx` and `reference/overrides.md`

dragomano:


docs/src/content/docs/ru/reference/route-data.mdx
HiDeoo, line 1: It looks like the changes from https://github.com/withastro/starlight/pull/3046/files are missing in this file.

HiDeoo (changes requested):

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://contribute.docs.astro.build/reviewers/reviewing-translations/). Except for a very tiny missing change, everything else looks great.

HiDeoo (approved):
Thanks for the quick fix 🚀

PR #3067, opened 2025-04-08
i18n(de): update `guides/customization`

trueberryless:
#### Description

This PR updates the German translation of `guides/customization` with the changes from #3025


delucis (approved):
Thanks for all the updates @trueberryless!

PR #3081, opened 2025-04-10
i18n(ko-KR): update `route-data.mdx`

jsparkdev:
#### Description

- #2927


HiDeoo (approved):
And another great update to end another great round of translations :star2: 🚀

PR #3090, opened 2025-04-11
Update dependencies

delucis:
Updates Astro and related dependencies across the monorepo


HiDeoo (approved):
The reasoning makes sense and everything seems in order to me (also tested locally).

PR #3097, opened 2025-04-12
feat: update environment comparisons data

trueberryless:
#### Description

This PR updates the comparison data of different docs frameworks visible on [the environmental impact page](https://starlight.astro.build/environmental-impact/#comparisons) with up-to-date 12. April 2025 data fetches.

There are some things to consider:

- [ ] Shall we include all these frameworks or leave some out to make the table smaller? If we drop some, which ones?
- [ ] Either the Zola calculation is bugged or really respectable, which is very plausible cause it's built with Rust :)

### DIFF
--- a/docs/src/content/docs/environmental-impact.md
+++ b/docs/src/content/docs/environmental-impact.md
@@ -91,31 +91,37 @@ These tests with the [Website Carbon Calculator][wcc] compare similar pages buil
 
 | Framework                   | CO₂ per page visit | Rating |
 | --------------------------- | ------------------ | :----: |
+| [Zola][zo-carbon]           | 0.00g              |   A+   |
 | [Starlight][sl-carbon]      | 0.01g              |   A+   |
-| [Read the Docs][rtd-carbon] | 0.05g              |   A+   |
-| [Sphinx][sx-carbon]         | 0.06g              |   A+   |
+| [Read the Docs][rtd-carbon] | 0.07g              |   A+   |
+| [Sphinx][sx-carbon]         | 0.07g              |   A+   |
 | [VitePress][vp-carbon]      | 0.07g              |   A+   |
-| [Docus][dc-carbon]          | 0.09g              |   A+   |
-| [docsify][dy-carbon]        | 0.10g              |   A    |
-| [Nextra][nx-carbon]         | 0.11g              |   A    |
-| [MkDocs][mk-carbon]         | 0.19g              |   B    |
-| [Docusaurus][ds-carbon]     | 0.21g              |   B    |
-| [GitBook][gb-carbon]        | 0.43g              |   C    |
-| [Mintlify][mt-carbon]       | 1.22g              |   F    |
+| [Docus][dc-carbon]          | 0.10g              |   A    |
+| [docsify][dy-carbon]        | 0.11g              |   A    |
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/environmental-impact.md
```diff
@@ -91,31 +91,37 @@ These tests with the [Website Carbon Calculator][wcc] compare similar pages buil
 
 | Framework                   | CO₂ per page visit | Rating |
 | --------------------------- | ------------------ | :----: |
+| [Zola][zo-carbon]           | 0.00g              |   A+   |
```
delucis, line 94: Zola isn’t specifically a docs tool as far as I can tell? Obviously any static site generator _can_ be used to build a docs site, but the intention here is to compare equivalent docs tools so I don’t think it makes sense to include this. Otherwise we’d also include other generic site builders.

The Website Carbon Calculator measures client-side weight (and server electricity source when known), which are not directly related to the technology of a static site generator, but much more closely associated with the template and assets it generates (so for example a tool being built with Rust or Python etc. has no impact on this measure). You can build a heavy website with Astro, Jekyll, 11ty, Next.js, Zola, or any other tool. Astro has some helpful features which Starlight makes good use of to reduce impact, but comparing build tools with this calculator doesn’t really make sense. The result would entirely depend on the template you use.

trueberryless: Totally fair enough, I myself wasn't sure what Zola exactly is and thought it might be bugged... Thanks very much for the clarification!

As a reference I thought when I update the stats I might as well include some new technologies from this list: https://dev.to/silviaodwyer/10-open-source-documentation-frameworks-to-check-out-331f
Not sure if this is trustworthy source tho, so I created the PR as draft to discuss the changes!😉👍

delucis (approved):
I think it’s good to add Fumadocs and mdBook 👍 Thanks for thinking of this @trueberryless 

> Would be interesting to know why others are dropping down the list, eg GitBook more than doubled... 🤔

That’s a complex thing to answer as there are probably many reasons! There are technical aspects (how good a tool’s chosen tech stack is at avoiding large or unnecessary data being included in page loads for example), but there’s also a question of culture. Hard to speculate about other projects where I’m not familiar with their processes, but to speak about Starlight: we intentionally track asset size with every PR so we can easily see if we introduce large increases in CSS or JS size. We also have the architectural advantage of shipping server-rendered HTML with minimal JS, which not everyone has. So just _monitoring_ and having that feedback as part of your process can be part of it. As an example, the GitBook URL appears to currently be loading a bunch of images via its client-side JavaScript which are not actually displayed anywhere. Assuming there’s not a good reason for that, one could speculate that the team is not regularly monitoring how the site behaves to pick up unnecessary resource requests like that and perhaps technical decisions make it easier to accidentally include resources. But hard to say for sure without knowing how they do stuff.

Regarding the concept of “performance culture” (which overlaps a lot with this page as it is mostly discussing the impact of _what_ a site serves, more than _how_ it is built/served, because that is harder to measure), I can recommend the [“Performance culture through the looking-glass”](https://youtu.be/t3smBkiGUWw?si=NhGQ9AkF13RiPhrB) talk by Dora Militaru from performance.now() a couple of years ago.

Anyway. Happy to merge this update ✅

PR #3098, opened 2025-04-12
Lower CSS specificity by wrapping styles in `@layer`

felix-bohlin:
#### Description

Lowers CSS specificity by wrapping the styles in `@layer`. In my project (a component library) I'd like a lot of control over the styles and this solution would make it easier to override some of the Starlight styles.


delucis:
Thanks for the PR @felix-bohlin. We already have a PR for this that we’re working on so will close this as a duplicate of #2322.

PR #3109, opened 2025-04-13
Update Expressive Code and related packages

### LINKED ISSUE #3108: Update `astro-expressive-code` to v0.41.0

0.33.1

5.6.1

pnpm

Mac

Chrome

### Describe the Bug

`astro-expressive-code` v0.41.0 has been released. Since it is pre-v1, it is not being installed unless updated here.

https://github.com/withastro/starlight/blob/9b39a06324603fee7cb288d35acedfe82a9b73df/packages/starlight/package.json#L200

-

### Participation

dhruvkb:
#### Description

- Closes #3108 
- This PR updates the version of the package `astro-expressive-code` from v0.40 to v0.41. This also replaces Shiki v1 dependencies with Shiki v3.

### DIFF
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -197,7 +197,7 @@
     "@types/hast": "^3.0.4",
     "@types/js-yaml": "^4.0.9",
     "@types/mdast": "^4.0.4",
-    "astro-expressive-code": "^0.40.2",
+    "astro-expressive-code": "^0.41.0",
     "bcp-47": "^2.1.0",
     "hast-util-from-html": "^2.0.1",
     "hast-util-select": "^6.0.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -183,8 +183,8 @@ importers:
         specifier: ^4.0.4
         version: 4.0.4
       astro-expressive-code:
-        specifier: ^0.40.2
-        version: 0.40.2(astro@5.6.1)
+        specifier: ^0.41.0
+        version: 0.41.0(astro@5.6.1)
       bcp-47:
         specifier: ^2.1.0
         version: 2.1.0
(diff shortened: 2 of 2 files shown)


dhruvkb:
So it's going to be a bit more complex than just a version bump.

HiDeoo:
> So it's going to be a bit more complex than just a version bump.

Not at a computer right now to take a very close look, but sharing a few early pointers regarding the issues:

- The type errors are types tests in the Starlight code base that ensures feature parity between the Expressive Code `<Code>` component and its equivalent in the Starlight Markdoc preset. I'm assuming some props were updated/added in the `<Code>` component in this new version, and we probably need to reflect these changes in the Markdoc preset.
- The accessibility warnings may be a bit more concerning, would probably need some investigation to see if it's a regression in the new version, some adjustments that need to be made in our EC configuration or something else.

Let me know if you want to me to take a closer look at any of these issues, give you a bit more guidance around them or if you want to take a stab at investigating them yourself. Thanks again for the contribution.

dhruvkb:
The accessibility warnings are from upstream. Created this issue to track: https://github.com/expressive-code/expressive-code/issues/323

hippotastic:
Thank you for the ping! I'll release a fix in the next few minutes.

I think this shows how great it is to have automated a11y checks! I have some of my own in EC already, but apparently I'll need to upgrade them to use the same tool Starlight uses as well. :)

hippotastic:
The patch is live now as EC version 0.41.1 on NPM. :)

dhruvkb:
Thanks @HiDeoo for looking into the PR and for the quick feedback on the failures, thanks @delucis for pointing out the new `hangingIndent` prop and thanks @hippotastic for patching and releasing v0.41.1 so quickly.

All checks are passing now, excited to be able to use this update!

hippotastic:
> All checks are passing now, excited to be able to use this update!

Great work, this PR looks good to me!

In case you want to use your new EC feature before the next Starlight release, I'd like to point out that you can use the following method explained in the EC docs to use the latest version earlier:

https://expressive-code.com/upgrading/#overriding-expressive-code-within-starlight

dhruvkb:
Thanks @hippotastic, guess I should've read the manual. I could've avoided all this work 🤦‍♂️ for myself and everyone else here.

delucis (approved):
Yes, it’s helpful for us to have PRs like this! We discovered a bug and @hippotastic and @dhruvkb fixed it 🎉 

Thanks all around 🙌

PR #3122, opened 2025-04-16
Don’t set defaults for `attrs` and `content` in head entries

delucis:
#### Description

- Removes default values for `attrs` and `content` from the schema used to parse head entries (previously these were `{}` and `""` respectively).
- This makes it simpler to add new entries directly to the parsed `head` array without types saying you must set these default values.
- Made this a minor as it could technically be breaking for users or plugins who are overriding or processing `head` in some way, although for the average site this should be basically an internal API change.

### DIFF
--- a/.changeset/chilly-dolphins-clap.md
+++ b/.changeset/chilly-dolphins-clap.md
@@ -1,0 +1,24 @@
+---
+'@astrojs/starlight': minor
+---
+
+Removes default `attrs` and `content` values from head entries parsed using Starlight’s schema.
+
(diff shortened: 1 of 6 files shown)


HiDeoo (approved):
Amazing, was planning to do this during lunch after noticing it while updating guides on my website for Starlight `0.33` but you already did it :tada: :raised_hands:

Looks perfect to me, easier for users, easier for us in our tests, only wins all around.

I'll just have to update my guides again :smile:

PR #3125, opened 2025-04-16
i18n(fr): update `reference/configuration`

HiDeoo:
#### Description

This PR updates the French version of the `reference/configuration` page with the changes from #3113,  https://github.com/withastro/starlight/commit/59497507ef52ffe871a6a485547e048c1b34f0ff and #3033.

PR #3126, opened 2025-04-16

HiDeoo:
#### Description

This PR updates the French version of the `guides/authoring-content` page with the changes from #3033.

Links validation should be fixed once #3125 is merged and this PR updated.

### DIFF
--- a/docs/src/content/docs/fr/guides/authoring-content.mdx
+++ b/docs/src/content/docs/fr/guides/authoring-content.mdx
@@ -649,3 +649,22 @@ Si vous disposez déjà d'un site Starlight et que vous souhaitez ajouter Markdo
 </Steps>
 
 Pour en savoir plus sur la syntaxe et les fonctionnalités de Markdoc, consultez la [documentation Markdoc](https://markdoc.dev/docs/syntax) ou le [guide de l'intégration Astro Markdoc](https://docs.astro.build/fr/guides/integrations-guide/markdoc/).
+
+### Paramétrage de la préconfiguration Markdoc
+
+La préconfiguration `starlightMarkdoc()` accepte les options de configuration suivantes :
+
+#### `headingLinks`
+
+**Type :** `boolean`  
+**Par défaut :** `true`
+
+Contrôle si les en-têtes sont affichés avec des liens d'ancrage cliquables ou non.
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/authoring-content.mdx
```diff
@@ -649,3 +649,22 @@ Si vous disposez déjà d'un site Starlight et que vous souhaitez ajouter Markdo
 </Steps>
 
 Pour en savoir plus sur la syntaxe et les fonctionnalités de Markdoc, consultez la [documentation Markdoc](https://markdoc.dev/docs/syntax) ou le [guide de l'intégration Astro Markdoc](https://docs.astro.build/fr/guides/integrations-guide/markdoc/).
+
+### Paramétrage de la préconfiguration Markdoc
```
HiDeoo, line 653: To avoid "Configuration de la préconfiguration Markdoc" : https://dictionnaire.lerobert.com/definition/parametrage

ArmandPhilippot: That suits me!

To avoid `préconfiguration`/`options de configuration` in the description below (kind of repeat), we could use `préréglage` instead of `préconfiguration`. So, we could use `Configuration du préréglage Markdoc` here and ``Le préréglage `starlightMarkdoc()` accepte les options de configuration suivantes :`` below.

But this is just a nit, and it seems the rest of the docs uses `préconfiguration` so I'm fine with your choice!

ArmandPhilippot (approved):
LGTM!

docs/src/content/docs/fr/guides/authoring-content.mdx
HiDeoo: Oh, I quite like "préréglage" a lot and it would definitely free us from having to think every time we talk about configuration about it (which would happen more and more as it's most likely that we end up adding options than removing them).

That would definitely be a follow-up PR changing every instances in the docs so adding a todo to my list or if you want to make a PR for it at some point, I would be happy to review it 👍

ArmandPhilippot: Your Starlight todo list must already be long 😅 so I don't mind doing this!

PR #3130, opened 2025-04-16
i18n(fr): update `guides/pages`

HiDeoo:
#### Description

This PR updates the French version of the `guides/pages` page with the changes from #3033.

### DIFF
--- a/docs/src/content/docs/fr/guides/pages.mdx
+++ b/docs/src/content/docs/fr/guides/pages.mdx
@@ -3,6 +3,8 @@ title: Pages
 description: Apprenez à créer et à gérer les pages de votre site de documentation avec Starlight.
 sidebar:
   order: 1
+tableOfContents:
+  maxHeadingLevel: 4
 ---
 
 Starlight génère les pages HTML de votre site en fonction de votre contenu, avec des options flexibles fournies par le biais du frontmatter en Markdown.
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -177,3 +202,33 @@ Définit l'étiquette d’identification BCP-47 pour le contenu de cette page, p
...
+#### Composant `<AnchorHeading>`
+
+Le composant `<AnchorHeading>` affiche un élément d'en-tête HTML avec un lien d'ancrage cliquable avec une apparence équivalente aux styles Markdown de Starlight.
+
+```astro
+---
+import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
+---
+
+<AnchorHeading level="2" id="sub-heading">Sous paragraphe</AnchorHeading>
```
trueberryless, line 215: Same here...

trueberryless (changes requested):
I think we should translate the id property of the new AnchorLink component, since the fragment of markdown headings in Starlight is automatically localized as well...

docs/src/content/docs/fr/guides/pages.mdx
ArmandPhilippot, line 215: I know we already had this discussion regarding the translation of heading... so I don't mind `en-tête`, but it feels a bit odd to have `Sous paragraphe` for `Sub heading`... I mean this is rendered as an `<hX>` tag not a `<p>` tag, right? `Sous-titre` would have be an easier choice but since we use `en-tête`, `Sous rubrique` maybe? Not sure...

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -73,28 +75,51 @@ Pour en savoir plus, consultez le guide [« Pages » dans la documentation d'Ast
...
 import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
+import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
 import CustomComponent from './CustomComponent.astro';
 ---
 
 <StarlightPage frontmatter={{ title: 'Ma page personnalisée' }}>
 	<p>Il s'agit d'une page personnalisée avec un composant personnalisé :</p>
 	<CustomComponent />
+
+	<AnchorHeading level="2" id="learn-more">En savoir plus</AnchorHeading>
```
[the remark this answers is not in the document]
HiDeoo: Interesting question. I initially translated them and went back on my decision before committing for a few reasons (not saying they're good, just my reasoning):

- I thought it could be used to emphasize this is not just a duplicate of the text, with [specific rules](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) like no spaces, etc.
- I don't think a lot of people in this specific context would end up translating them.

I think I'll end up translating them while adding Armand suggestion, altho definitely not something that should be enforced and we can let translators decide.

docs/src/content/docs/fr/guides/pages.mdx
HiDeoo: I was blocked on this one while for quite a bit before falling back to "Sous paragraphe" which I did not like a lot but could not find anything better at the time. I didn't think about "Sous rubrique" which I quite like, at least it's way better than "Sous paragraphe" so I'll use that for now. Thanks as always for the thorough review.

ArmandPhilippot (approved):

PR #3150, opened 2025-04-22
Ensure layer order in custom pages

HiDeoo:
#### Description

In a Starlight content page, we have full control over the CSS import order and [we are making sure](https://github.com/withastro/starlight/blob/6babbc1b1c2a175b31c859b22ad0ce1181124323/packages/starlight/components/Page.astro#L2-L6) that CSS files are imported in the correct order to properly support cascade layers.

However, in custom pages using `<StarlightPage>`, due to how [import order](https://docs.astro.build/en/guides/styling/#import-order) for stylesheets works in Astro, we are not enforcing such order.

This means that the same custom page can ends up with broken styles depending on the order in which components are imported.

For example, the following custom page work as expected as `<StarlightPage>` is imported first which will import `<Page>` which will import the CSS files linked previously:

```astro
---
import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
import { Aside } from '@astrojs/starlight/components';
---
```

However, having the `<AnchorHeading>` or `<Aside>` component imported first will break the expected order as these components include styles using layers which will end up being ordered by order of usage/import.

To prevent this issue in custom pages, this PR moves the user and Starlight laer definitions to a new file which is now imported first in the `<AnchorHeading>` and the `components.ts` barrel file.

To reproduce the issue locally, comment the changes in `packages/starlight/components.ts` and `packages/starlight/components/AnchorHeading.astro` and visit the test pages included in this PR at [`http://localhost:4321/tests/test2/`](http://localhost:4321/tests/test2/) and [`http://localhost:4321/tests/test3/`](http://localhost:4321/tests/test3/). You should see the styles broken in both pages. After applying the changes, the styles should be fixed.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -70,6 +70,23 @@ export default defineConfig({
 			customCss: ['./src/assets/landing.css'],
 			locales,
 			sidebar: [
+				// TODO(HiDeoo) Remove this sidebar group
+				{
+					label: 'Tests',
+					items: [
+						{ link: '/tests/test1/', label: 'Test 1', badge: { text: 'OK', variant: 'success' } },
+						{
+							link: '/tests/test2/',
+							label: 'Test 2',
+							badge: { text: 'Broken', variant: 'danger' },
+						},
+						{
+							link: '/tests/test3/',
+							label: 'Test 3',
+							badge: { text: 'Broken', variant: 'danger' },
+						},
+					],
+				},
 				{
 					label: 'Start Here',
 					translations: {
--- a/docs/src/pages/tests/test1.astro
+++ b/docs/src/pages/tests/test1.astro
@@ -1,0 +1,22 @@
+---
+import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
+import { LinkButton } from '@astrojs/starlight/components';
+import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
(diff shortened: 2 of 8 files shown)


delucis:
Sharing some testing notes here in public so I don’t lose them:

- This PR’s a11y checks were failing because the changes here impacted the order of styles when bundled for production, changing how cascade layers were ordered. This made our `starlight.content` layer take precedence over `starlight.components`.
- From a quick look at the deployment preview, the order in `<head>` in this PR is now:
  1. Link to print stylesheet.
  2. Link to main stylesheet, which includes most of our `base`, `reset`, `utils`, and `core` layer styles.
  3. Inline `<style>` that includes (in order): user custom CSS, our layer order statement, and then the `components` layer.

I guess basically something like this is happening:

- When there are no custom pages, Vite/Astro bundles *all* CSS into one file because all routes using Starlight styles are considered a “single” route (our `[...slug]` route). In this case, the order we set up in the `Page` component is respected.

- When there *are* custom pages, Vite/Astro splits CSS into chunks to only include component styles on the routes that use them (in this case, the custom page). When this happens, it’s possible for some other styles to somehow end up in the wrong one of these chunks and that appears to be happening to user CSS and our `@layer` ordering, which are in the inline “component” chunk.

- This has presumably always happened, but the change in order didn’t have the same side effect that it now does with `@layer`.

PR #3160, opened 2025-04-29
docs(showcase): add create.bingo and octo.guide

JoshuaKGoldberg:
#### Description

Adds them both because I'd forgotten to send create.bingo when I first made it earlier this year. 😄


delucis (approved):
So cute!

Issue #3154, opened 2025-04-25
`spawnSync` usage in `getAllNewestCommitDate` may hit default `maxBuffer` limit


KianNH:
### What version of `starlight` are you using?

0.32.5

### What version of `astro` are you using?

5.6.1

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Safari

### Describe the Bug

On large repositories, the usage of `git log --format=t:%ct --name-status -- ...` may hit the default `1024 * 1024` buffer limit of `spawnSync`:

```
Error: spawnSync git ENOBUFS
    at Object.spawnSync (node:internal/child_process:1120:20)
    at spawnSync (node:child_process:868:24)
    at getAllNewestCommitDate (file:///Users/kian/cf-repos/GitHub/cloudflare-docs/index.js:4:17)
    at file:///Users/kian/cf-repos/GitHub/cloudflare-docs/index.js:61:13
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:547:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5) {
  errno: -55,
  code: 'ENOBUFS',
  syscall: 'spawnSync git',
  path: 'git',
  spawnargs: [
    'log',
    '--format=t:%ct',
    '--name-status',
    '--',
    'src/content/docs/'
  ]
}
```

I think it would be safe to increase the default, but if not then it could be made configurable:

https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options

https://nodejs.org/api/child_process.html#maxbuffer-and-unicode

### Link to Minimal Reproducible Example

N/A

### Participation

- [x] I am willing to submit a pull request for this issue.

PR #3165, opened 2025-05-01
Increase `maxBuffer` in `getAllNewestCommitDate`'s `spawnSync` usage

KianNH:
#### Description

Closes https://github.com/withastro/starlight/issues/3154

This could technically be tested, possibly by mocking `execSync` to return a >1MB (the default `maxBuffer` value) output but this shouldn't be a user-visible change unless they were already hitting the limit.

### DIFF
--- a/.changeset/slow-rivers-attack.md
+++ b/.changeset/slow-rivers-attack.md
@@ -2,6 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks @KianNH! Added a comment to the code for people to reference if we hit this limit again in the future.

PR #3167, opened 2025-05-01
Update CSS guide to clarify cascade layer references

niallobrien:
#### Description
Updates CSS guide to clarify cascade layer references as discussed on today's call.


HiDeoo (approved):
Thanks a lot for jumping in and making a PR after the Talking & Doc'ing session, really appreciated :star2:

The changes look exactly like what we discussed.

PR #3168, opened 2025-05-02
i18n(ko-KR): update Korean UI strings

jsparkdev:
#### Description

update Korean UI strings (and `404.md`, sidebar)


HiDeoo (dismissed):
As a non-native speaker reviewing using the [Docs Translation Review Checklist](https://contribute.docs.astro.build/reviewers/reviewing-translations/), the changes look great to me 🌟 

Considering `packages/starlight/translations/ko.json` is updated, which is a file shipped with Starlight to end users, we would need to [add a changeset](https://contribute.docs.astro.build/docs-for-code-changes/changesets/#format).

Would you be up for that? I can also help you with that if you want, altho, as a non-native speaker, I'm not sure if the PR only improves the translation or if it also fixes some typos or grammar issues so we can best describe the changes in the changeset.

jsparkdev:
@HiDeoo 

Thanks, I had forgotten about that

And this PR is about making the existing translation even better!

packages/starlight/translations/ko.json
```diff
@@ -1,30 +1,30 @@
...
   "search.ctrlKey": "Ctrl",
   "search.cancelLabel": "취소",
-  "search.devWarning": "검색 기능은 프로덕션 환경에서만 사용할 수 있습니다. \n사이트를 빌드하고 미리 보기를 실행하여 로컬에서 테스트하세요.",
+  "search.devWarning": "검색 기능은 프로덕션 빌드에서만 작동합니다. \n로컬에서 테스트하려면 사이트를 빌드하고 미리 보기를 실행하세요.",
   "themeSelect.accessibleLabel": "테마 선택",
-  "themeSelect.dark": "다크",
-  "themeSelect.light": "라이트",
+  "themeSelect.dark": "어두운 테마",
+  "themeSelect.light": "밝은 테마",
   "themeSelect.auto": "자동",
```
delucis, line 10: Definitely trust you to make the best translation, but just noting that in most languages this is _only_ the colour, e.g. “Light” or “Dark” and doesn’t also include the word “theme”, so you have the accessible “테마 선택” label followed by the theme name and it looks like for the “auto” variant, this is still the case: “자동”, even though the light and dark translations include “theme” (“테마”).

That is totally fine if the meaning here is otherwise unclear and “테마” helps to clarify it but wanted to note it.

packages/starlight/translations/ko.json
```diff
@@ -1,30 +1,30 @@
...
+  "i18n.untranslatedContent": "이 콘텐츠는 아직 번역되지 않았습니다.",
+  "page.editLink": "페이지 편집",
   "page.lastUpdated": "마지막 업데이트:",
-  "page.previousLink": "이전",
-  "page.nextLink": "다음",
-  "page.draft": "이 콘텐츠는 초안이며 프로덕션 빌드에는 포함되지 않습니다.",
-  "404.text": "페이지를 찾을 수 없습니다. URL을 확인하거나 검색창을 사용해 보세요.",
-  "aside.note": "노트",
+  "page.previousLink": "이전 페이지",
+  "page.nextLink": "다음 페이지",
```
delucis, line 20: And similarly here: most languages just include “Next”/“Previous” without specifying “page”. But if “페이지” here helps to clarify the meaning, then it’s fine 👍

.changeset/tall-lies-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+update translations of Korean UI strings
```
delucis, line 5: Suggesting a small tweak to specify that we’re improving as well as adding the missing translations in this patch:

```suggestion
Updates Korean langage support with improvements and missing translations
```

delucis (commented):
Thanks @jsparkdev! Just left some quick questions — up to you if you want to change anything in the translation or if you think it’s good like this.

packages/starlight/translations/ko.json
jsparkdev: That's a very reasonable question, and it's something I also considered carefully.

You're right that in English, 'Dark' and 'Light' are common conventions likely understood as 'Dark Theme' and 'Light Theme'.

However, in Korean, using just the adjective form – like '어두운' (eodun - meaning 'dark') or '밝은' (balgeun - meaning 'light') – as a standalone UI option often feels grammatically awkward or incomplete. It leaves the user wondering, "Dark what?". This might be a key difference compared to how adjectives function in English UI labels.

I also tried using the loanwords '다크' (Dark) and '라이트' (Light), but they similarly felt somewhat unnatural or abrupt when used alone in this context.

Therefore, to ensure the meaning is absolutely clear and feels natural to Korean users, I decided to explicitly add the noun '테마' (Theme) after the adjective, resulting in '어두운 테마' (Dark Theme) and '밝은 테마' (Light Theme). Although this deviates slightly from the structure in other languages by adding a word, I believe prioritizing clarity and naturalness for the target audience is the best approach here.

(Optional addition regarding 'Auto'): For the 'Auto' option, the Korean word '자동' (jadong) works perfectly well on its own because it describes the method of theme selection (automatic) rather than a specific visual style like 'dark' or 'light'.

.changeset/tall-lies-wonder.md
jsparkdev: Thanks for good suggestion.

delucis (approved):
Thank you for the detailed and thoughtful explanations @jsparkdev! Great that the translations are in your careful hands 💖

PR #3178, opened 2025-05-05
Add constants to exports

jiv-ropo:
#### Description

- Adds constants to exports.
- Overriding PageTitle component was problematic because I could not find a good way to import constants from the Starlight package.

### DIFF
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -28,6 +28,7 @@
   "exports": {
     ".": "./index.ts",
     "./locals": "./locals.ts",
+    "./constants": "./constants.ts",
     "./components": "./components.ts",
     "./components/LanguageSelect.astro": {
       "types": "./components/LanguageSelect.astro.tsx",


HiDeoo:

Indeed, at the moment, the `PAGE_TITLE_ID` constant is not exported and we do [document for the `<PageTitle>` component](https://starlight.astro.build/reference/overrides/#pagetitle) that `_top` should be used.

While having this constant exposed is definitely a good idea, would you mind first opening an issue so we can discuss the change? This is something that has [already been mentioned](https://github.com/withastro/starlight/issues/860#issuecomment-1757457206) in the past, and different approaches could be used.

Exposing this file using a dedicated export could be one of them, but force us to document it and consider this entire file as a public API. Another suggested approach was to expose such contant in the [route data](https://starlight.astro.build/guides/route-data/) object containing all the information required to render a single page. This approach could allow us to expose a way less surface to the user, while still allowing to use the constant in a more reliable way.

Discussing this in an issue would allow us to gather more feedback from the community and other maintainers, and decide on the best approach to take before we move forward with a pull request. Thanks for your understanding!

PR #3193, opened 2025-05-11
docs(showcase): add fluid-dnd showcase

carlosjorger:
#### Description

- Add [fluid-dnd](https://fluid-dnd.netlify.app/) showcase.


HiDeoo (approved):
Thanks for the new showcase entry, looks amazing to me :star2:

PR #3194, opened 2025-05-12
docs(plugins) Add a link to starlight-scroll-to-top plugin

frostybee:
This PR adds to the official documentation website a link to the scroll-to-top plugin.

PR #3195, opened 2025-05-12
i18n(fr): update `resources/plugins`

ArmandPhilippot:
#### Description

Adds changes from #3194 to the French translation of `resources/plugins`.

### DIFF
--- a/docs/src/content/docs/fr/resources/plugins.mdx
+++ b/docs/src/content/docs/fr/resources/plugins.mdx
@@ -153,6 +153,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-fullview-mode"
 		description="Réduire les barres latérales et élargir le contenu pour une expérience plein écran."
 	/>
+	<LinkCard
+		href="https://github.com/frostybee/starlight-scroll-to-top"
+		title="starlight-scroll-to-top"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -153,6 +153,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-fullview-mode"
 		description="Réduire les barres latérales et élargir le contenu pour une expérience plein écran."
 	/>
+	<LinkCard
+		href="https://github.com/frostybee/starlight-scroll-to-top"
+		title="starlight-scroll-to-top"
+		description="Ajouter un bouton de retour en haut de page à votre site de documentation."
```
ArmandPhilippot, line 159: The literal translation for `scroll-to-top button` would be `bouton de défilement (or déplacement) vers le haut` but it sounds weird. I think we commonly call this feature `retour en haut de page` but if you have another idea, don't hesitate.

HiDeoo: I think that's a great call, it's even called like that in https://www.systeme-de-design.gouv.fr/composants-et-modeles/composants/retour-en-haut-de-page/ :thumbsup:

HiDeoo (approved):
Looks perfect to me 🚀

Issue #3162, opened 2025-04-29
Cascade layer issues/feedbacks


HiDeoo:
This issue is more of an umbrella for the various issues and feedback gathered regarding the style changes from [Starlight `0.34.0`](https://github.com/withastro/starlight/releases/tag/%40astrojs%2Fstarlight%400.34.0) and the switch to cascade layers.

##### ~~1. Missing styles~~

Workaround shipped in [Starlight `0.34.2`](https://github.com/withastro/starlight/releases/tag/%40astrojs%2Fstarlight%400.34.2)

<details>
<summary>History</summary>

Not technically related to the changes in `0.34.0`, but seems to be more easily reproducible due to some refactoring in this release. If an integration or Starlight plugin injecting dynamic routes that also render the `<StarlightPage>` component, due to an head propagation issue, some styles can end up missing in development mode.

- Tracked in https://github.com/withastro/astro/issues/13724
- Initially, I opened a [workaround PR](https://github.com/withastro/starlight/pull/2905) before investigating the issue
- After having investigated the issue, I implemented a better and documented workaround with an E2E test that we can decided to ship if the Astro issue is not something that can be easily fixed

</details>

##### ~~2. Custom pages layer order~~

Documentation updated in https://github.com/withastro/starlight/pull/3199 and we will monitor for extra feedback to see if we need to improve the documentation further.

<details>
<summary>History</summary>

For custom pages rendered using the `<StarlightPage>` component, due to how [import order](https://docs.astro.build/en/guides/styling/#import-order) for stylesheets works in Astro, importing `<AnchorHeading>` or any Starlight component before `<StarlightPage>` breaks the expected layer order as we cannot control the order in this case like we do for Starlight pages.

- Opened a [tentative PR](https://github.com/withastro/starlight/pull/3150) to fix this issue but ended up fixing the issue in dev but breaking in production
- Need to investigate further if we can somehow reliably enforce the order in such cases
- If not, we would need to document this behavior and suggest to import `<StarlightPage>` before any other Starlight component

</details>

##### ~~3. Do I have to use cascade layers?~~

Documentation updated in https://github.com/withastro/starlight/pull/3167 and we will monitor for extra feedback to see if we need to improve the documentation further.

<details>
<summary>History</summary>

It looks like some users could be confused by the fact that we are using cascade layers in Starlight and that they may also need to use them in their own stylesheets. The "any custom unlayered CSS will override the default styles" may not be clear enough.

[...]

PR #3198, opened 2025-05-13
docs: add Tailwind instructions on how to use multiple configs

HiDeoo:
#### Description

This PR updates the Tailwind instructions to help with issue 6 in https://github.com/withastro/starlight/issues/3162:

> When using custom pages or Starlight at a subpath and want to use Tailwind on both your main site/custom pages and Starlight, you propably want to use a default Tailwind setup for your main website/custom pages and a different one for Starlight which would use the Starlight’s Tailwind compatibility package.

1. My initial assessment was only focusing on the case of using Starlight at a subpath.
2. My initial idea was to add a small sentence in the existing Tailwind instructions, e.g.:

   > If you are using [Starlight at a subpath](https://starlight.astro.build/manual-setup/#use-starlight-at-a-subpath), you can use a different [Tailwind CSS file](https://docs.astro.build/en/guides/styling/#add-tailwind-4) for your main website often imported in a layout component.

3. After discussing the topic during Talking and Doc'ing, the idea of creating a entirely new sub-section "Add site-wide Tailwind support with Starlight at a subpath" was identified as a better approach. Even though it was vaguely similar to the existing ["Add Tailwind to an existing project" section](https://starlight.astro.build/guides/css-and-tailwind/#add-tailwind-to-an-existing-project), it gives the user better instructions, and in the "Tailwind CSS" intro section, we can link to the various possible setups.

4. When working on the PR and checking various support threads, I realized that this setup may not only be useful for Starlight at a subpath, but also for users who want to use [custom pages](https://starlight.astro.build/guides/pages/#custom-pages).

This PR keeps the new idea of creating a new sub-section but tries to cover both use cases: custom pages and Starlight at a subpath.

Other ideas discarded during T&D:

- Not re-using `src/styles/global.css` in our "Add Tailwind to an existing project" section but instead creating a new `src/styles/starlight.css` file so that it's more obvious that this file is only used for Starlight and the other one is only used for the main site/custom pages.

  This was discarded because `src/styles/global.css` is automatically scaffolded by Astro and this use-case is still a small percentage of users. That would mean that most people would end up with an `src/styles/global.css` file that they don't need.

### DIFF
--- a/docs/src/content/docs/guides/css-and-tailwind.mdx
+++ b/docs/src/content/docs/guides/css-and-tailwind.mdx
@@ -210,7 +210,7 @@ If you already have a Starlight site and want to add Tailwind CSS, follow these 
 
 ### Add site-wide Tailwind support to an existing project
 
-If you are using [custom pages](/guides/pages/#custom-pages) with you Starlight site, or if you are using [Starlight at a subpath](/manual-setup/#use-starlight-at-a-subpath), and want to add site-wide Tailwind support with different Tailwind CSS configurations, follow these steps.
+If you are using [custom pages](/guides/pages/#custom-pages) with your Starlight site, or if you are using [Starlight at a subpath](/manual-setup/#use-starlight-at-a-subpath), and want to add site-wide Tailwind support with different Tailwind CSS configurations, follow these steps.
 
 <Steps>
 
--- a/docs/src/content/docs/guides/css-and-tailwind.mdx
+++ b/docs/src/content/docs/guides/css-and-tailwind.mdx
@@ -71,6 +71,8 @@ Any styles in the `my-overrides` layer would take precedence over Starlight’s 
 Tailwind CSS support in Astro projects is provided by the [Tailwind Vite plugin](https://tailwindcss.com/docs/installation/using-vite).
 Starlight provides complementary CSS to help configure Tailwind for compatibility with Starlight’s styles.
 
+You can [create a new project with Tailwind](#create-a-new-project-with-tailwind), [add Tailwind to an existing project](#create-a-new-project-with-tailwind), or [add site-wide Tailwind support to an existing project with different Tailwind CSS configurations](#add-site-wide-tailwind-support-to-an-existing-project) when using custom pages or Starlight at a subpath.
+
 The Starlight Tailwind CSS applies the following configuration:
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/css-and-tailwind.mdx
```diff
@@ -206,6 +208,81 @@ If you already have a Starlight site and want to add Tailwind CSS, follow these
 
 </Steps>
 
+### Add site-wide Tailwind support to an existing project
+
+If you are using [custom pages](/guides/pages/#custom-pages) with you Starlight site, or if you are using [Starlight at a subpath](/manual-setup/#use-starlight-at-a-subpath), and want to add site-wide Tailwind support with different Tailwind CSS configurations, follow these steps.
```
awhitford, line 213: "with _you_ Starlight site" --> "with _your_ Starlight site"

awhitford:
Is the [Starlight with Tailwind example](https://github.com/withastro/starlight/blob/main/examples/tailwind/src/styles/global.css) going to be updated to match these instructions?  It sounds like we should get `global.css` and `starlight.css` out of the box.

Just like someone might add Starlight to an Astro project, the other way around is also logical; I should be able to easily add custom Astro pages to a Starlight project.

docs/src/content/docs/guides/css-and-tailwind.mdx
HiDeoo:
```suggestion
If you are using [custom pages](/guides/pages/#custom-pages) with your Starlight site, or if you are using [Starlight at a subpath](/manual-setup/#use-starlight-at-a-subpath), and want to add site-wide Tailwind support with different Tailwind CSS configurations, follow these steps.
```

HiDeoo:
> Is the [Starlight with Tailwind example](https://github.com/withastro/starlight/blob/main/examples/tailwind/src/styles/global.css?rgh-link-date=2025-05-16T06%3A10%3A04Z) going to be updated to match these instructions? It sounds like we should get `global.css` and `starlight.css` out of the box.

Not something planned with this approach as described at the end of the PR body.

I'm working on another PR as an alternative to this one doing that and taking another approach for the docs that I plan to share with the team to discuss the pros/cons of both approaches.

delucis:
Closing in favour of #3273!

PR #3199, opened 2025-05-13
docs: add `<StarlightPage>` import order guidelines

HiDeoo:
#### Description

This PR updates the `<StarlightPage>` component documentation to help with issue 2 in https://github.com/withastro/starlight/issues/3162:

After trying many possible workarounds to find a way around what seems to be an unsolvable issue, which depends on how [import order](https://docs.astro.build/en/guides/styling/#import-order) works in Astro and also many variables e.g. the number of routes, etc., this PR adds guidelines to the documentation regarding the import order of the `<StarlightPage>` component and its role in setting up cascade layers.

This first attempt is based on recommendations from Sarah during a Talking & Doc'ing session mentioning the limitations using text and code example comments.

I tried to avoid an aside even tho we use one in the Astro Docs:

image

Considering how tricky this is, maybe it should be an aside but I wanted to try a first approach without it. Open to any suggestions.


alvinometric:
Not sure if this needs a review by @sarah11918 ?

docs/src/content/docs/guides/pages.mdx
```diff
@@ -77,13 +77,16 @@ Read more in the [“Pages” guide in the Astro docs](https://docs.astro.build/
...
+This component must be the first import in your file to set up [cascade layers](/guides/css-and-tailwind/#cascade-layers) and ensure a predictable CSS order.
 
 To add anchor links to headings that match Starlight’s Markdown anchor link styles, you can use the [`<AnchorHeading>` component](#anchorheading-component) in your custom pages.
 
 ```astro
 ---
 // src/pages/custom-page/example.astro
+// Import the `<StarlightPage>` component first to set up cascade layers.
+// Import any other components you want to use in your custom page.
```
sarah11918, line 90:
```suggestion

// Import any other components you want to use in your custom page.
```
Extra visual separation to emphasize? (and break up wall of code?)

delucis (approved):
Seems fair to document this workaround for now. Hopefully we can fix this properly soon!

PR #3205, opened 2025-05-20
fix userCollections?.docs

sgalcheung:
- check `userCollections?.docs`
- add starlight-page example for test

#### Description

- Closes https://github.com/withastro/starlight/discussions/3192

Capture docs as undefined.
```
// starlight-page.ts
// before
return userCollections?.docs.schema ?? docsSchema();

// after
const userSchema = userCollections?.docs?.schema;

  if (typeof userSchema === 'function') {
    return userSchema;
  } else if (userSchema) {
    return () => userSchema;
  } else {
    return docsSchema();
  }
```


packages/starlight/utils/starlight-page.ts
```diff
@@ -211,5 +211,13 @@ async function getUserDocsSchema(): Promise<
...
-	return userCollections?.docs.schema ?? docsSchema();
+  const userSchema = userCollections?.docs?.schema;
+
+  if (typeof userSchema === 'function') {
+    return userSchema;
+  } else if (userSchema) {
+    return () => userSchema;
+  } else {
+    return docsSchema();
+  }
```
delucis, line 222: Wouldn’t this be the simpler fix?
```suggestion
  return userCollections?.docs?.schema ?? docsSchema();
```

delucis (changes requested):
Thanks for the PR @sgalcheung! I don’t think we need to add a full example for this though, so we should remove the `examples/starlight-page/` addition.

sgalcheung:
> Thanks for the PR @sgalcheung! I don’t think we need to add a full example for this though, so we should remove the `examples/starlight-page/` addition.

In this case, this doesn't have any huge effect, but we could use this example for StarlightPage in the future.
Otherwise, I don't know how to write a better UT for it.

packages/starlight/utils/starlight-page.ts
sgalcheung: This needs a UT for help.

sgalcheung:
Do we consider using [astro-tests](https://inox-tools.fryuni.dev/astro-tests)?

sgalcheung:
Thank you for improving the unit test @HiDeoo ! Could you commit this test code?

HiDeoo (commented):
> Thank you for improving the unit test @HiDeoo ! Could you commit this test code?

Sorry, didn't get the time to get back to you earlier, but it looks like you already did and also included [Chris's suggestion](https://github.com/withastro/starlight/pull/3205#discussion_r2098033524) :raised_hands:

Thanks for updating the PR, this is looking great. I'll let Chris make a final review as he already had a look at it, but we should be able to release this change in the next patch version :tada:

delucis (approved):
Yes, this looks great — thanks @sgalcheung! 🚀

PR #3215, opened 2025-05-26
docs: update yummacss thumbnail

rrenildopereiraa:
- This change updates the yummacss.com.png

The old thumbnail shows an outdated landing page, so we replaced it with a current one.

Comparison


HiDeoo (approved):
Thanks for the lovely update 🤩

PR #3216, opened 2025-05-26
docs: correct environmental impact percentage

scriptcoded:
#### Description

This might be the most mundane and picky PR I ever have and ever will open. But if anything should be correct, this would be it. Then test score from the Website Carbon Calculator has dropped from 99% to 98%. That's it. I'm sorry.


HiDeoo (approved):
Thanks for spotting the change, no need to be sorry, and also thanks for taking the time to fix it! :raised_hands:

PR #3235, opened 2025-06-07
docs: Update plugins.mdx to add astro-mermaid

joesaby:
#### Description

Support for mermaid & browser rendering in astro. Demo site https://starlight-mermaid-demo.netlify.app/


docs/src/content/docs/resources/plugins.mdx
```diff
@@ -212,4 +212,9 @@ These community tools and integrations can be used to add features to your Starl
 		title="astro-plantuml"
 		description="Render PlantUML content in Markdown code blocks."
 	/>
+	<LinkCard
+		href="https://github.com/joesaby/astro-mermaid"
+		title="astro-mermaid"
+		description="Render Mermaid content in Markdown code blocks."
```
HiDeoo:
```suggestion
		description="Client-side render Mermaid content in Markdown code blocks."
```

I think it may be important to clarify that the integration only supports client-side rendering of Mermaid diagrams. This may be a limitation for many documentation projects due to the trade-off of the approach so making it clear in the wording seems like a good idea.

joesaby:
@HiDeoo - sure, make sense. In our own personal repo, we use rehype to render it during build time (also works in dev mode). Problem with it is the usage of `playwright` dependency in the build container. 

So can create a new plugin for it or add conditional support. Would be far easier to create a new plugin for it given its specific requirement with imports and how the plugin gets consumed in a build env

HiDeoo (approved):
Thanks for updating the PR.

> Problem with it is the usage of `playwright` dependency in the build container.

Yeah, that's definitely another downside of Mermaid. I guess it could be handled like it's done in [`rehype-mermaid`](https://github.com/remcohaszing/rehype-mermaid/blob/main/package.json#L59-L67) with an optional peer dependency and instructions to install it when you want to server render the diagrams.

joesaby:
@HiDeoo. I will try to look at both of your suggestions

docs/src/content/docs/resources/plugins.mdx
```diff
@@ -212,4 +212,9 @@ These community tools and integrations can be used to add features to your Starl
...
+	/>
 </CardGrid>
```
Randy88-art, line 220: docs/src/content/docs/resources/plugins.mdx

joesaby: was there something that needed to be done?

joesaby:
> Thanks for the new addition to the showcase 🙌
> 
> Mermaid is definitely an interesting feature for many documentation projects. I suggested a wording change to more closely match the integration behavior. Let me know what you think.
> 
> Are you also planning on supporting server rendering too, to potentially mitigate with client-side rendering issues?
> 
> If you're also interested, here is some feedback I quickly noticed while checking out the integration:
> 
> * It looks like Mermaid is imported in all pages, even the ones that don't use it.
> * There are some layout shifts when loading pages, e.g. https://starlight-mermaid-demo.netlify.app/ before and after diagrams are rendered, altho I assume supporting server rendering would fix this.
> * It looks like https://starlight-mermaid-demo.netlify.app/examples/git/ is erroring?
> 
> Thanks again for your work on this integration.

https://github.com/joesaby/astro-mermaid/commit/4ea7f968f4a923d035d697ba7ae05c544c5b342d
https://starlight-mermaid-demo.netlify.app/examples/git/

docs/src/content/docs/resources/plugins.mdx
HiDeoo: > was there something that needed to be done?

No, this is probably just spam 😅

joesaby:
@HiDeoo. Created https://github.com/joesaby/astro-mermaid/issues/10

PR #3242, opened 2025-06-11
i18n(fr): fix link in `reference/configuration` + small rewording

ArmandPhilippot:
#### Description

Updates the French translation of `reference/configuration` to:
* fix a link that has been updated in Astro Docs (see https://github.com/withastro/docs/pull/11874)
* use French quotes instead of double quotes
* translate the only untranslated example
* reword some sentences following changes in the French i18n guide in Astro Docs:
  * `built on top`: `basée sur` instead of `construite sur`. Not sure if this is the best translation but I always find that `construite sur` sounds odd.
  * `repo` > `dépôt`
  * `docs` > `documentation`
  * `supporter` > `prendre en charge`
  * `styliser` > `mettre en forme`

### DIFF
--- a/docs/src/content/docs/fr/reference/configuration.mdx
+++ b/docs/src/content/docs/fr/reference/configuration.mdx
@@ -5,7 +5,7 @@ description: Une vue d'ensemble de toutes les options de configuration prises en
 
 ## Configuration de l'intégration `starlight`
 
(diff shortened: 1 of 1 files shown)


HiDeoo (approved):
Thanks for the update, looks perfect to me :raised_hands:

I'm glad that link needed to be updated so we can get all these great improvements at the same time :laughing:

PR #3256, opened 2025-06-26
i18n(de): update guides/pages.mdx

trueberryless:
#### Description

- update guides/pages.mdx
- #3199

### DIFF
--- a/docs/src/content/docs/de/guides/pages.mdx
+++ b/docs/src/content/docs/de/guides/pages.mdx
@@ -84,10 +84,12 @@ Um Ankerlinks zu Überschriften hinzuzufügen, die den Markdown-Ankerlinkstilen 
 ```astro
 ---
 // src/pages/custom-page/example.astro
-// Importiere zuerst die Komponente `<StarlightPage>`, um die Kaskadenschichten einzurichten.
+// Importiere zuerst die Komponente `<StarlightPage>`,
+// um die Kaskadenschichten einzurichten.
 import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
 
-// Importiere alle anderen Komponenten, die du auf deiner benutzerdefinierten Seite verwenden möchtest.
+// Importiere alle anderen Komponenten, die du auf
+// deiner benutzerdefinierten Seite verwenden möchtest.
 import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
 import CustomComponent from './CustomComponent.astro';
 ---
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/de/guides/pages.mdx
```diff
@@ -77,39 +77,41 @@ Mehr dazu findest du in der [„Seiten“-Anleitung in der Astro-Dokumentation](
 
 Um das Starlight-Layout in benutzerdefinierten Seiten zu verwenden, umhüllst du deinen Seiteninhalt mit der [Komponente `<StarlightPage />`](#starlightpage-komponente).
 Das kann hilfreich sein, wenn du Inhalte dynamisch generierst, aber trotzdem das Starlight-Design verwenden willst.
+Diese Komponente muss als erstes in deine Datei importiert werden, um [Kaskadenschichten](/de/guides/css-and-tailwind/#kaskadenschichten) einzurichten und eine vorhersehbare CSS-Reihenfolge sicherzustellen.
 
 Um Ankerlinks zu Überschriften hinzuzufügen, die den Markdown-Ankerlinkstilen von Starlight entsprechen, kannst du die [`<AnchorHeading>`-Komponente](#anchorheading-komponente) in deinen eigenen Seiten verwenden.
 
+{/* prettier-ignore-start */}
```
HiDeoo: I'm not sure to understand why this was added?

Removing them locally and letting the formatter run seems to work fine:

image

Altho, considering the length of the added comments, maybe they could span 2 lines to avoid scrolling on larger viewports.

trueberryless: When I don't have them, my local format on save option always creates some weird newlines:

image

So I think it's more safe to include them. Or do you think I somehow have the wrong prettier setup, although there is a `.prettierrc`?

HiDeoo: Oh yeah, that's the default behavior, I just manually added in my screenshot an extra new line before and after the text of the link so it's formatted like my screenshot and read a bit better, but both are valid anyway 👍

trueberryless: Oh, wait, maybe the best solution to this inconcistency is, that I create another PR which specifies the HTML whitespace sensitivity: https://prettier.io/docs/options#html-whitespace-sensitivity

Because this option is not included in Starlights prettier config yet and seems to create this inconsistency because I guess Prettier then relies on the users setting... 🤔 

What do you think about that solution? If you agree, what option should I set it to?

HiDeoo: This is not an inconsistency, the default behavior is the screenshot you shared and as soon as you explicitly add new lines before and after the link text, you get my screenshot as the formatter see the new lines and preserve them.

trueberryless: Ah alright. Sorry the previous message was sent before I could see your explanation!

I'm gonna add the newlines then and remove the prettier ignore comments 👍

HiDeoo: No worries, sometime GitHub updates live, sometime it does not, who even know at this point 😆

HiDeoo: Btw, what do you think on splitting the new comments on 2 lines to avoid scrolling on larger desktop? They seems quite larger compared to the English version.

trueberryless: Sure, that's a good catch 👌

PR #3258, opened 2025-06-27
i18n(fr): update `guides/pages`

ArmandPhilippot:
#### Description

Updates the French translation of `guides/pages.mdx` to:
* add changes from #3199 
* improves the consistency across the docs based on the updated French Astro i18n guide:
  * replace `support` with `prise en charge`
  * while I don't mind `titres` for `headings` 😄 Starlight uses `en-têtes` so let's stay consistent
  * replace `construire` with `créer` for `build pages`
  * some other small changes where a comment on the spot will perhaps be more understandable than a description here

### DIFF
--- a/docs/src/content/docs/fr/guides/pages.mdx
+++ b/docs/src/content/docs/fr/guides/pages.mdx
@@ -7,7 +7,7 @@ tableOfContents:
   maxHeadingLevel: 4
 ---
 
-Starlight génère les pages HTML de votre site en fonction de votre contenu, avec des options flexibles fournies par le biais du frontmatter en Markdown.
+Starlight génère les pages HTML de votre site en fonction de votre contenu, avec des options flexibles fournies via le frontmatter du fichier Markdown.
 En outre, les projets Starlight bénéficient d'un accès complet aux [puissants outils de génération de pages d'Astro](https://docs.astro.build/fr/basics/astro-pages/).
 Ce guide montre comment fonctionne la génération de pages dans Starlight.
 
@@ -40,7 +40,7 @@ import { FileTree } from '@astrojs/starlight/components';
 
 ### Frontmatter avec sûreté du typage
 
-Toutes les pages Starlight partagent un [ensemble commun de propriétés du frontmatter](/fr/reference/frontmatter/) personnalisable qui permet de contrôler l'apparence de la page :
+Toutes les pages Starlight partagent un [ensemble commun de propriétés de frontmatter](/fr/reference/frontmatter/) personnalisable pour contrôler l'apparence de la page :
 
 ```md
 ---
@@ -54,8 +54,8 @@ Si vous oubliez quelque chose d'important, Starlight vous le fera savoir.
 ## Pages personnalisées
 
 Pour les cas d'utilisation avancés, vous pouvez ajouter des pages personnalisées en créant un répertoire `src/pages/`.
-Le répertoire `src/pages/` utilise le [routage basé sur les fichiers d'Astro](https://docs.astro.build/fr/basics/astro-pages/#routage-basé-sur-les-fichiers) et inclut le support des fichiers `.astro` parmi d'autres formats de pages.
-Ceci est utile si vous avez besoin de construire des pages avec une mise en page complètement personnalisée ou de générer une page à partir d'une source de données alternative.
+Le répertoire `src/pages/` utilise le [routage basé sur les fichiers d'Astro](https://docs.astro.build/fr/basics/astro-pages/#routage-basé-sur-les-fichiers) et inclut la prise en charge des fichiers `.astro` parmi d'autres formats de pages.
+Ceci est utile si vous avez besoin de créer des pages avec une mise en page complètement personnalisée ou de générer une page à partir d'une source de données alternative.
 
 Par exemple, ce projet mélange du contenu Markdown dans `src/content/docs/` avec des routes Astro et HTML dans `src/pages/` :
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -7,7 +7,7 @@ tableOfContents:
   maxHeadingLevel: 4
 ---
 
-Starlight génère les pages HTML de votre site en fonction de votre contenu, avec des options flexibles fournies par le biais du frontmatter en Markdown.
+Starlight génère les pages HTML de votre site en fonction de votre contenu, avec des options flexibles fournies via le frontmatter du fichier Markdown.
```
ArmandPhilippot, line 10: While reading `frontmatter en Markdown`, from my point of view, it means the frontmatter uses Markdown syntax while it actually uses either YAML or TOML. So I've slightly reworded the sentence (and I used `via` from the original version and which is also valid in French to avoid `du... du` which sounds a bit heavy otherwise, I think).

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -40,7 +40,7 @@ import { FileTree } from '@astrojs/starlight/components';
 
 ### Frontmatter avec sûreté du typage
 
-Toutes les pages Starlight partagent un [ensemble commun de propriétés du frontmatter](/fr/reference/frontmatter/) personnalisable qui permet de contrôler l'apparence de la page :
+Toutes les pages Starlight partagent un [ensemble commun de propriétés de frontmatter](/fr/reference/frontmatter/) personnalisable pour contrôler l'apparence de la page :
```
ArmandPhilippot, line 43: We don't talk about a particular frontmatter but any frontmatter, so I think `de` is more appropriate here. Also, the original version uses `to control` so I think we should stay direct and use `pour contrôler` instead of `qui permet de contrôler` (unnecessary heaviness).

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -94,9 +98,7 @@ import CustomComponent from './CustomComponent.astro';
 
 	<AnchorHeading level="2" id="en-savoir-plus">En savoir plus</AnchorHeading>
 	<p>
-		<a href="https://starlight.astro.build/">
-			Consulter la documentation de Starlight
-		</a>
+		<a href="https://starlight.astro.build/">Consulter la documentation de Starlight</a>
```
ArmandPhilippot: In the original version, all this is on a single line, and I think this is the correct format. If we put `Consulter...` on a new line, the output would be `<a> Consulter... </a>` (unwanted extra spaces around the text).

HiDeoo: While true, considering [whitespaces at the start and end of elements](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#html_largely_ignores_whitespace) are ignored in HTML, I personally tend to favor readability of the code snippet on large viewport without scrolling.

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -75,15 +75,19 @@ Pour en savoir plus, consultez le guide [« Pages » dans la documentation d'Ast
...
 
 Pour ajouter des liens d'ancrage vers des en-têtes avec une apparence équivalente aux liens d'ancrage Markdown de Starlight, vous pouvez utiliser le [composant `<AnchorHeading>`](#composant-anchorheading) dans vos pages personnalisées.
 
 ```astro
 ---
 // src/pages/page-perso/exemple.astro
+// Importer le composant `<StarlightPage>` en premier pour configurer les couches de cascade.
+
+// Importer tous les autres composants que vous souhaitez utiliser dans votre page personnalisée.
```
HiDeoo:
```suggestion
// Importer le composant `<StarlightPage>` en premier pour configurer les
// couches de cascade.
import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';

// Importer tous les autres composants que vous souhaitez utiliser dans votre
// page personnalisée.
```

Same here, we can make it easier to read on large viewports.

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -107,6 +109,7 @@ Le composant `<StarlightPage />` affiche une page complète de contenu en utilis
 
 ```astro
 ---
+// Importer le composant `<StarlightPage>` en premier pour configurer les couches de cascade.
```
HiDeoo:
```suggestion
// Importer le composant `<StarlightPage>` en premier pour configurer les
// couches de cascade.
```

And same here

docs/src/content/docs/fr/guides/pages.mdx
HiDeoo: I guess it depends how you read it, it can be read as a particular frontmatter, the one from the schema supported by Starlight. Altho, I don't think I personally have a strong opinion about either one of them so happy to go with what you feel is best 👍

docs/src/content/docs/fr/guides/pages.mdx
```diff
@@ -75,15 +75,19 @@ Pour en savoir plus, consultez le guide [« Pages » dans la documentation d'Ast
 
 ### Utiliser le design de Starlight dans des pages personnalisées
 
-Pour utiliser la mise en page Starlight dans des pages personnalisées, englobez le contenu de votre page avec le [composant `<StarlightPage />`](#composant-starlightpage).
+Pour utiliser la mise en page de Starlight dans des pages personnalisées, englobez le contenu de votre page avec le [composant `<StarlightPage />`](#composant-starlightpage).
 Cela peut s'avérer utile si vous générez du contenu de manière dynamique, mais que vous souhaitez tout de même utiliser le design de Starlight.
+Ce composant doit être la première importation dans votre fichier pour configurer les [couches de cascade](/fr/guides/css-and-tailwind/#couche-de-cascade) et garantir un ordre prévisible pour le CSS.
```
HiDeoo, line 80: This "première importation" reads super weird to me. Looks like it's in the french guide altho I wonder if we should follow [MDN](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Statements/import) on this one and just use "premier import" 🤔

docs/src/content/docs/fr/guides/pages.mdx
ArmandPhilippot: It seems even MDN is inaccurate. 😅 In their example the whitespaces at the beginning are indeed removed but the ones at the end are merged into a single space `Hello World! `.

I quickly checked with a minimal Astro project, and indeed I get the same behavior.  I thought even the first whitespace was visible and so it could be annoying. But, yeah, maybe this is a not a big deal in this case. I'm just so used to people trying to set a coma and seeing an "unexpected" space in between.

I don't have a strong opinion on this or the horizontal overflow, I'm just used to Astro Docs where we tend to keep the same line numbers as the original to quickly find the updates. So I don't mind reverting this if you prefer! 😄

docs/src/content/docs/fr/guides/pages.mdx
ArmandPhilippot: Well, this is the [French word](https://www.larousse.fr/dictionnaires/anglais-francais/import/588150). MDN also uses it, for example in [Exemples](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Statements/import#exemples) or [Importations des défauts](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Statements/import#importation_des_d%C3%A9fauts). It just seems they are mixing both `import` and `importation` (e.g. [Imports dynamiques](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Statements/import#imports_dynamiques)).
Using `import` would be an anglicism... but if you don't like `importation`, we could maybe say something like `` la première instruction `import` `` (ie. using import as inline code)?

HiDeoo: Yeah, even tho I know it's the proper word, I guess it's the first time I'm seeing/hearing it in dev-related context (I don't think we ever ever documented anything related to import in the Starlight Docs before, or I don't remember it). It's probably just me being used too much to the "import" word to the point I believe "importation" is too weird, and let's avoid mixing them like MDN so happy to go with it 👍

HiDeoo (approved):
Perfect, as usual 🚀 

Thanks for updating the PR and following up on my thoughts.

docs/src/content/docs/fr/guides/pages.mdx
ArmandPhilippot: Yeah, if you are used to read docs in English that explain it! Me too by the way, but I know this is used in other docs. For example, with a quick search [the French Python docs](https://docs.python.org/fr/3.13/reference/import.html) or [the French PHP docs](https://www.php.net/manual/fr/language.namespaces.importing.php). So, I don't think this is so uncommon and it doesn't shock me unlike the translations for `middleware` or `framework`... 😅

PR #3266, opened 2025-07-01
Autogenerated link custom attributes

HiDeoo:
#### Description

- Closes #3264

This PR adds support for custom HTML attributes on autogenerated sidebar links using the `autogenerate.attrs` option.

- The `autogenerate.attrs` is used rather than just `attrs` as `attrs` is only supported for links and not groups at the moment. This could be confusing to support `attrs` on autogenerated groups but not on manually defined groups, so `autogenerate.attrs` is used to clarify that this is only for links in an autogenerated groups.
- Using `attrs` on sidebar groups now results in a type error on top of the existing runtime error. I added some type tests to ensure this is the case.
- Individual pages can override custom attributes using the `sidebar.attrs` frontmatter field.


alvinometric:
🥳 Amazing !

packages/starlight/utils/navigation.ts
```diff
@@ -268,12 +274,12 @@ function treeify(routes: Route[], locale: string | undefined, baseDir: string):
...
 
 /** Create a link entry for a given content collection entry. */
-function linkFromRoute(route: Route): SidebarLink {
+function linkFromRoute(route: Route, attrs?: LinkHTMLAttributes): SidebarLink {
 	return makeSidebarLink(
 		slugToPathname(route.slug),
 		route.entry.data.sidebar.label || route.entry.data.title,
 		route.entry.data.sidebar.badge,
-		route.entry.data.sidebar.attrs
+		Object.keys(route.entry.data.sidebar.attrs).length > 0 ? route.entry.data.sidebar.attrs : attrs
```
delucis: I wonder if some people would expect the `attrs` to be combined rather than overridden? Like if you have `autogenerate: { attrs: { 'data-special': true } }` and then within that a page with frontmatter like:

```yml
sidebar:
  attrs:
    class: experimental
```

Would you expect to get a `<a data-special class="experimental">`?

The argument against is that it makes it a bit harder to _remove_ attributes. I guess you’d need something like this to remove an attribute:

```yml
sidebar:
  attrs:
    data-special:
```

Assuming something like this:

```suggestion
		{ ...attrs, ...route.entry.data.sidebar.attrs }
```

kevinzunigacuellar: Nice catch! I'd expect the attributes to be merged, with page-level attributes taking precedence over others.

HiDeoo: Good question. When working on the PR, I actually tried both approaches and found the overriding one more to my liking in the scenarios I was testing but I can definitely see how when you expect one approach, the other can be confusing (that's also why I mentioned it explicitly in the docs).

Happy to update the PR if more people think merging is the better approach, I'll do that tomorrow morning.

Thanks for the feedback 🙌

delucis: Would you mind sharing a bit about the detail of those scenarios? I always like to hear the concrete use cases as it can help understand better. My thinking was mostly based on a scenario like:

- `autogenerate.attrs` would be used for a unified control over a group of links, e.g. maybe these are your experimental docs, or reference, or some other category, and get a specific custom style to indicate that
- `sidebar.attrs` would be used for adding additional attributes for a specific page link, e.g. to flag a new page perhaps
- The group identity and the page-specific attributes aren’t necessarily exclusive: you’d want to keep both.

Admittedly, if that example was setting `class` for both, nothing would change here, because `sidebar.attrs.class` would still need to include the group class names, but for some other attributes it might make more sense, e.g. maybe there’s some group you set `target="_blank" rel="noopener"` for but don’t want to have to set those again in frontmatter just to add a `class`. (And to maintain — changes to the group config would need adding to all overridden pages and remembering that there _are_ overridden pages.)

Open to convincing either way though — sounds like you had some specific scenarios in mind, so those might be valid.

HiDeoo: My initial decision to override attributes was not really tied to the scenarios I tested but more to the overall experience when playing with them and overriding attributes at the page level:

- I found overriding easier as I was starting from a blank state, and just having to focus on "what exactly do I want as custom attributes for this page?" vs "ok, I already have these attributes applied defined in another file, here is what I want in the end, what do I need to remove and add to get there?"
- Removing attributes is not the obvious thing to do when merging attributes
- One example I tried was with the `style` attributes while the configuration was applying a `class` attribute. I kinda liked that my custom class was automatically removed when I set a `style` attribute, as it was not relevant anymore.

I've updated the PR to use a merge strategy so that we can play with it more and see how it feels.

One small note, I had to allow `null` values to properly support removing attributes from the frontmatter:

```md
sidebar:
  attrs:
    data-special:
    data-special: ~
    data-special: null
```

All of these would result in a `null` value which we were not previously allowing.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -244,7 +244,7 @@ The following sidebar will be generated:
 
 Use the [`sidebar` frontmatter field](/reference/frontmatter/#sidebar) in individual pages to customize autogenerated links.
 
-Sidebar frontmatter options allow you to set a [custom label](/reference/frontmatter/#label) or add a [badge](/reference/frontmatter/#badge) to a link, [hide](/reference/frontmatter/#hidden) a link from the sidebar, or define a [custom sort weighting](/reference/frontmatter/#order).
+Sidebar frontmatter options allow you to set a [custom label](/reference/frontmatter/#label), use [custom attributes](/reference/frontmatter/#attrs), or add a [badge](/reference/frontmatter/#badge) to a link, [hide](/reference/frontmatter/#hidden) a link from the sidebar, or define a [custom sort weighting](/reference/frontmatter/#order).
```
trueberryless: I think this "or" is one too much.

```suggestion
Sidebar frontmatter options allow you to set a [custom label](/reference/frontmatter/#label), use [custom attributes](/reference/frontmatter/#attrs), add a [badge](/reference/frontmatter/#badge) to a link, [hide](/reference/frontmatter/#hidden) a link from the sidebar, or define a [custom sort weighting](/reference/frontmatter/#order).
```

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -436,6 +436,72 @@ The configuration above generates the following sidebar:
 	]}
 />
 
+### Custom HTML attributes for autogenerated links
+
+Customize HTML attributes of all links in [autogenerated groups](#autogenerated-groups) by defining the `attrs` property in the `autogenerate` configuration.
+Individual pages can specify custom attributes using the [`sidebar.attrs` frontmatter field](/guides/sidebar/#customizing-autogenerated-links-in-frontmatter) which will be merged with the `autogenerate.attrs` configuration.
```
trueberryless: Whole link not necessary, we are in the correct file already:

```suggestion
Individual pages can specify custom attributes using the [`sidebar.attrs` frontmatter field](#customizing-autogenerated-links-in-frontmatter) which will be merged with the `autogenerate.attrs` configuration.
```


UPDATE: I think it is generally speaking even better, if we link to the place, where the `sidebar.attrs` really exists and not to a place in the docs, where it's not even mentioned (just behind the reference link):

```suggestion
Individual pages can specify custom attributes using the [`sidebar.attrs` frontmatter field](/reference/frontmatter/#attrs) which will be merged with the `autogenerate.attrs` configuration.
```

trueberryless (approved):

docs/src/content/docs/reference/frontmatter.md
```diff
@@ -386,6 +386,7 @@ sidebar:
 **type:** `Record<string, string | number | boolean | undefined>`
 
 HTML attributes to add to the page link in the sidebar when displayed in an autogenerated group of links.
+Such attributes will be merged with the [`autogenerate.attrs` option](/guides/sidebar/#custom-html-attributes-for-autogenerated-links) of the autogenerated group this page belongs to.
```
delucis: Small wording suggestion:

```suggestion
If [`autogenerate.attrs`](/guides/sidebar/#custom-html-attributes-for-autogenerated-links) is set on the autogenerated group this page belongs to, frontmatter attributes will be merged with the group attributes.
```

Mainly inspired by trying to avoid “such”, but ended up doing more to get to that goal 😄 

Considered whether we should explicitly add sommething like “If a property is set in both `autogenerate.attrs` and `sidebar.attrs`, the property in `sidebar.attrs` will override the `autogenerate.attrs` property.” But it’s a bit of a mouthful and hopefully “merge” is clear enough on its own.

trueberryless: In my opinion "merge" is fine, but at the same time feels like hiding a little but what takes _priority_ (I think that's not the right word, but you know what I mean... Was the word "presendecy" what I was looking for? No, this gets underlined red...) over what... But I think here we can expect people to understand that the more specific setting overrides the general one...

PR #3271, opened 2025-07-02
Style Pagefind root excerpt

### LINKED ISSUE #3267: Search display broken due to code marks and other elements

0.34.4

5.10.1

npm

Mac

Firefox (Zen Browser)

### Describe the Bug

Search results for elements found in code blocks or `<mark>` elements seem to break the display output of the search results. See below.

Image

The documentation is publicly available. You can run the build and preview actions and then search for `w`.

https://github.com/DuendeSoftware/docs.duendesoftware.com/

### Participation

HiDeoo:
#### Description

- Closes #3267

In Pagefind, when searching the [indexed content section](https://pagefind.app/docs/indexing/#limiting-what-sections-of-a-page-are-indexed), if a result is [found before the first heading](https://github.com/Pagefind/pagefind/blob/d7d0b3a0f0eb12661cd2eb894ad02f10687a4ca0/pagefind_ui/default/svelte/result_with_subs.svelte#L31-L32), an extra excerpt is [shown below the page title](https://github.com/Pagefind/pagefind/blob/d7d0b3a0f0eb12661cd2eb894ad02f10687a4ca0/pagefind_ui/default/svelte/result_with_subs.svelte#L66-L68).

At the moment, we do not style this excerpt at all in Starlight altho it can appear under some conditions, e.g. having a banner with content that is a potential match, and as we render `<Banner>` in `<main>` (our `data-pagefind-body`), before the `<PageTitle>`, it'll be detected as root sub result and the excerpt will be shown.

This PR adds some basic styling to the root excerpt so that it is visually consistent with the rest of the search results altho I'm personally not really happy with the result, as it kinda looks like a result rather than just an excerpt for a root result? I'm definitely not the best person when it comes to design, but I think we may be able to improve it further? A difference from nested results is that the title/link and excerpt are not grouped together with a parent element, they're just siblings in the entire page results.

|                    | Preview  |
| ------------------ | -------- |
| With subresults    | SCR-20250702-kqta |
| With no subresults | SCR-20250702-kqvy |

This PR includes a temporary route data middleware adding a banner to every page so searching something like `test` will show the root excerpt.

#### Remaining tasks

- [ ] Add a changeset when happy with the PR so we have more context about how to describe the change

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -35,6 +35,8 @@ export default defineConfig({
 	trailingSlash: 'always',
 	integrations: [
 		starlight({
+			// TODO(HiDeoo) Remove
+			routeMiddleware: './src/routeData.ts',
 			title: 'Starlight',
 			logo: {
 				light: '/src/assets/logo-light.svg',
--- a/docs/src/routeData.ts
+++ b/docs/src/routeData.ts
@@ -1,0 +1,9 @@
+// TODO(HiDeoo) Remove
+
+import { defineRouteMiddleware } from '@astrojs/starlight/route-data';
+
+export const onRequest = defineRouteMiddleware((context) => {
+	context.locals.starlightRoute.entry.data.banner = {
+		content: 'This is a <a href="https://starlight.astro.build/">test banner</a>.',
+	};
+});
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -361,13 +361,15 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 			padding: 0;
 		}
 
-		#starlight__search .pagefind-ui__result-nested {
+		#starlight__search .pagefind-ui__result-nested,
+		#starlight__search .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *)) {
 			position: relative;
 			padding: var(--sl-search-result-nested-pad-block) var(--sl-search-result-pad-inline-end);
 			padding-inline-start: var(--sl-search-result-pad-inline-start);
 		}
 
 		#starlight__search .pagefind-ui__result-title:not(:where(.pagefind-ui__result-nested *)),
+		#starlight__search .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *)),
 		#starlight__search .pagefind-ui__result-nested {
 			position: relative;
 			background-color: var(--sl-color-black);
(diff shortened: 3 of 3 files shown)


delucis (commented):
Oh yeah, I see the issues with this like you mentioned: the excerpt is not clickable and not clearly associated with the page title as one might expect.

I wonder if we can improve on that styling.

An additional thought: if the most common cause of this is our own banner component — we should maybe consider ignoring that by default? Don’t think it’s expected for that to appear in search results.

delucis:
Couple of quick experiments:

- Moving the excerpt up and removing top padding visually links the excerpt with the title a bit better:

	```css
	/* Move excerpts up closer to the result title */
	#starlight__search .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *)) {
	    margin-top: -0.75em; /* Magic number — would be good to derive this instead. */
	    padding-top: 0;
	}
	
	/* When the title is focused, also highlight the excerpt */
	#starlight__search .pagefind-ui__result-title:focus-within + .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *)) {
	    background-color: var(--sl-color-accent-low);
	}
	```

	search UI

- But this introduces issues of course with hover states:

	search UI

- We can’t use `outline` because of the overlap, but we can simulate it using border on a pseudo element attached to the excerpt:

	```css
	#starlight__search .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *))::after {
	    content: "";
	    position: absolute;
	    inset: -1px;
	    border: 1px solid transparent;
	    border-top: 0;
	}
	```
	
	Which we reveal when the title is hovered or focused:

	```css
	#starlight__search .pagefind-ui__result-title:where(:hover,:focus-within) + .pagefind-ui__result-excerpt:not(:where(.pagefind-ui__result-nested *))::after {
	    border-color: var(--sl-color-accent-high);
	}
	```
	
	Hover version:

	search UI
	
	Focus version:

	search UI


The weird thing about this approach is that the excerpt is still not actually hoverable to click the link over the excerpt — only on the title. Not sure if there’s a reliable way to extend the interactive area down over the excerpt without hard coding the height. Anchor positioning _might_ help? But it doesn’t have broad support yet.

delucis:
Chatted with @HiDeoo and we’re closing this in favour of our project rewriting the search UI from scratch which he is currently working on.

Any solution we try here will likely feel hacky and not fully work. #3276 should reduce the risk of users hitting this, so also reduces the urgency of a hacky fix like this.

PR #3272, opened 2025-07-02
Support passing `generateId` to `docsLoader()`

### LINKED ISSUE #3227: Using locales with regions (es-AR) makes build output more nested directories than necessary

^0.34.3

^5.8.1

npm

Mac

Chrome

### Describe the Bug

Here are my locales:

```
defaultLocale: 'root',
locales: {
        root: {
          label: 'English',
          lang: 'en', // lang is required for root locales
        },
        'id': { label: 'Bahasa Indonesia', lang: 'id' },
        'ms': { label: 'Bahasa Malaysia', lang: 'ms' },
        'bg': { label: 'Български', lang: 'bg' },
        'ca': { label: 'Català', lang: 'ca' },
        'cs': { label: 'Čeština', lang: 'cs' },
        'da': { label: 'Dansk', lang: 'da' },
        'de': { label: 'Deutsch', lang: 'de' },
        'et': { label: 'Eesti', lang: 'et' },
        'es': { label: 'Español', lang: 'es' },
        'es-AR': { label: 'Español (AR)', lang: 'es-AR' },
        'es-MX': { label: 'Español (MX)', lang: 'es-MX' },
        'tl': { label: 'Filipino', lang: 'tl' },
        'fi': { label: 'Suomi', lang: 'fi' },
        'fr': { label: 'Français', lang: 'fr' },
        'el': { label: 'Ελληνικά', lang: 'el' },
        'he': { label: 'עברית', lang: 'he' },
        'hi': { label: 'हिन्दी', lang: 'hi' },
        'hr': { label: 'Hrvatski', lang: 'hr' },
        'hu': { label: 'Magyar', lang: 'hu' },
        'is': { label: 'Íslenska', lang: 'is' },
        'it': { label: 'Italiano', lang: 'it' },
        'ja': { label: '日本語', lang: 'ja' },
        'ko': { label: '한국어', lang: 'ko' },
        'lv': { label: 'Latviski', lang: 'lv' },
        'lt': { label: 'Lietuvių', lang: 'lt' },
        'nl': { label: 'Nederlands', lang: 'nl' },
        'no': { label: 'Norsk', lang: 'no' },
        'pl': { label: 'Polski', lang: 'pl' },
        'pt-BR': { label: 'Português (BR)', lang: 'pt-BR' },
        'pt-PT': { label: 'Português (PT)', lang: 'pt-PT' },
        'ro': { label: 'Română', lang: 'ro' },
        'ru': { label: 'Русский', lang: 'ru' },
        'sk': { label: 'Slovenčina', lang: 'sk' },
        'sl': { label: 'Slovenščina', lang: 'sl' },
        'sr': { label: 'Srpski', lang: 'sr' },
        'sv': { label: 'Svenska', lang: 'sv' },
        'th': { label: 'ภาษาไทย', lang: 'th' },
        'tr': { label: 'Türkçe', lang: 'tr' },
        'uk': { label: 'Українська', lang: 'uk' },
        'vi': { label: 'Tiếng Việt', lang: 'vi' },
        'ar': { label: 'العربية', lang: 'ar' },
        'zh-CN': { label: '简体中文', lang: 'zh-CN' },
        'zh-TW': { label: '繁體中文', lang: 'zh-TW' },
      },
```

Here's my directory structure under src/content/docs:
Image

When I build this, every locale directory that outputs to dist/ gets nested locale directories of all the locales that have region on them. 

Image

Could have something to do with the fact that the region on the locale folder is not upper case in dist.

https://github.com/wink-travel/docs

### Participation

delucis:
Closes #3227

This PR makes it possible to pass `generateId` to Starlight’s `docsLoader()`.

Currently, using `generateId` to customize content slugs is only possible by ejecting back up to Astro’s generic `glob()` loader, but this requires figuring out the `pattern` and `base` Starlight uses if you want to mirror the `docsLoader()` behaviour, e.g.:

```diff
-   loader: docsLoader(),
+   loader: glob({
+     pattern: "**/[^_]*.{md,mdx}",
+     base: "./src/content/docs",
+     generateId: customMethod
+   }),
```

After this PR you can keep using `docsLoader()`, and just add your ID method:

```diff
-   loader: docsLoader(),
+   loader: docsLoader({ generateId: customMethod }),
```

I did not also add an equivalent method to the `i18nLoader` as customizing slugs for languages seems much less common, and it would be more liable to break given how we load translation files in non-content collection contexts.

A big part of this PR was also figuring out how best to document this, as we currently only document our loaders in passing. This PR adds a “Configuring content collections” section to the config reference which documents all our loaders and schemas. I kept this minimal for now and linked out to existing relevant docs rather than duplicating or moving things around, but I’d welcome feedback on this new section.

---

It’s perhaps worth noting that there is _some_ risk in supporting custom `generateId` methods. Especially for multilingual sites.

There are a few code paths running outside of the Astro/Vite context where Starlight uses raw file paths directly rather than the content collection ID. This is notably the case for remark/rehype plugins. For example, we use the file path to work out the current language for remark plugins. This means it won’t be possible to use `generateId` for entirely novel structures like storing localized content in sub-directories or something.

For example, moving the locale from a file name to the front of the ID like this wouldn’t work for our remark plugins:

```js
// Example input: { entry: 'tutorials/intro/en.md' }
generateId: ({ entry }) => {
  const extensionlessPath = entry.split('.').slice(0, -1).join('.'); // "tutorials/intro/en"
  const segments = extensionlessPath.split('/');                     // ["tutorials", "intro", "en"]
  return [segments.at(-1), ...segments.slice(0, -1)].join('/');      // "en/tutorials/intro"
},
```

But for common scenarios like preserving uppercase characters or special characters like `.` inside IDs, it should work pretty OK although it is likely to break compatibility with plugins like https://github.com/HiDeoo/starlight-links-validator that rely on being able to predict the ID format based on the filepath.


docs/src/content/docs/reference/configuration.mdx
```diff
@@ -673,3 +673,107 @@ starlight({
...
+```
+
+##### Options
+
+###### `generateId()`
+
+**type:** `({ entry: string; base: URL; data: Record<string, unknown> }) => string`
+
+By default, pages generated using `docsLoader()` process your file names using a sluggifier, which removes special characters and lowercases the file name.
+If you want to override this default, provide your own custom `generateId()` function.
```
HiDeoo, line 721: I wonder if we should/could be helpful to provide a bit more "guidance/use-cases" here, e.g.

> If you want to override this default, e.g. to preserve the original case or some special characters in the slug, provide your own custom `generateId()` function.

Maybe it could be a hint that totally transforming the structure may not work at all for some cases/features?

HiDeoo (approved):
Impressive how exposing such feature can be done quite easily and beautifully 👏 Like you mentioned, the bulk of the work is definitely the documentation (which imo is perfect).

Left a tiny comment, also wonder if there are some places where it could be relevant to link to this new documentation section, e.g.:

- ["Configure content collections" section](https://starlight.astro.build/manual-setup/#configure-content-collections), maybe linking `docsLoader` and `docsSchema`?
- ["Extend translation schema" section](https://starlight.astro.build/guides/i18n/#extend-translation-schema), maybe linking `extend`?
- ["Customize frontmatter schema" `extend` section](https://starlight.astro.build/reference/frontmatter/#extend), maybe linking `extend`?

Nothing blocking, just some thoughts.

Like you mentioned and as we discussed earlier, there are definitely some potential limits to this feature or potential issues that could arise, but I still think it is a great addition, and in the case of plugins, if it's ever an issue, it should be relatively easy to support an equivalent option at the plugin level, to use a custom function instead of the default assumed `github-slugger` one.

docs/src/content/docs/reference/configuration.mdx
delucis: I’m not 100% sure what the common use cases are although casing and special characters are indeed what I’ve seen before. Maybe something like this?

````suggestion
If you want to override this default, provide your own custom `generateId()` function.

For example, this can be useful to preserve special characters that would be removed.
By default, `Example.File.md` would be served at `/examplefile`.
If you wanted to serve it at `/Example.File`, you could do so by defining a custom `generateId()` function:

```js
docsLoader({
	// Remove the `.md` or `.mdx` extension, but otherwise don’t process filenames.
	generateId: ({ entry }) => entry.split('.').slice(0, -1).join('.'),
}),
```
````

HiDeoo: Oh, I personally really like that. It adds a small and really clear example to a section that did not have one before, the comment makes it obvious what is being done while showing valid a valid use case.

Totally on board with this personally :clap:

delucis:
OK, perfect — thanks for the helpful feedback!

I decided against linking the two `extend` places you spotted (thanks for hunting those down) as the reference material for those is really thin, and don’t add to what you already have in those sections. But I added the links for `docsLoader` and `docsSchema` as those seem more helpful.

PR #3241, opened 2025-06-11
Support Astro 5.9 new API renderMarkdown

sgalcheung:
#### Description

- Closes #3239
Create a [‘createMarkdownTestHelper`](https://github.com/sgalcheung/starlight/blob/b4c829dabc01861e9155747335ba1c0242db0930/packages/starlight/__tests__/remark-rehype/setupMarkdown.ts#L10) method to reduce repeated creation of createMarkdownProcessor, and this method could initialize different parameters in the test method, so we get all tests passed!

before
image

after
image

Next, I will create a test to reproduce this bug, so that we can better verify the modified code logic.

PR #3181, opened 2025-05-06
Prevent heading anchor links on non-Starlight content

HiDeoo:
#### Description

- Closes #3175

In Starlight [`0.34.0`](https://github.com/withastro/starlight/releases/tag/%40astrojs%2Fstarlight%400.34.0), support for generating [clickable anchor links for headings was added](https://github.com/withastro/starlight/pull/3033). For Markdown and MDX content, a rehype plugin is used to generate the links.

When the feature is not disabled, this plugin runs unconditionally, which means that these links are generated for non-Starlight content like [individual Markdown page](https://docs.astro.build/en/guides/markdown-content/#individual-markdown-pages) or content not from the Starlight `docs` content collection.

This PR prevents this behavior by checking if the virtual file path is part of the Starlight `docs` content collection first, and bails if it is not.

For testing, I relied on E2E tests as it seems that on Windows, file path separators are inconsistent when the rehype plugin is used on Markdown vs MDX files. I'll investigate and open a separate issue/PR for that on the Astro repo.

Here is a preview of the changes before and after the fix for an individual Markdown page:

| Before          | After           |
| --------------- | --------------- |
| SCR-20250506-pgpd | SCR-20250506-pgre |

PR #3274, opened 2025-07-03
Restrict remark/rehype plugins usage

### LINKED ISSUE #3239: Support Astro 5.9 new API renderMarkdown

0.34.3

5.9.1

pnpm

Mac

Edge

### Describe the Bug

My project uses a custom Loader and docs, when I add this new API [renderMarkdown](https://astro.build/blog/astro-590/#rendering-markdown-in-content-loaders), I met an error:
```
...
BC%89/WebFramework/Astro/starsquid/node_modules/.pnpm/trough@2.2.0/node_modules/trough/lib/index.js:204:5)
Failed to parse Markdown file "undefined":
Cannot read properties of undefined (reading 'replace')
  Location:
    eval at runInlinedModule 
...
```

I found this error because the starlight does not update the dependency.
```
…/Starlight/starlight main ❯ pnpm why trough
Legend: production dependency, optional only, dev only


devDependencies:
astro 5.6.2
├─┬ @astrojs/markdown-remark 6.3.1
│ ├─┬ rehype-stringify 10.0.1
│ │ └─┬ unified 11.0.5
│ │   └── trough 2.1.0
│ ├─┬ remark-gfm 4.0.1
│ │ ├─┬ remark-parse 11.0.0
│ │ │ └─┬ unified 11.0.5
│ │ │   └── trough 2.1.0
│ │ ├─┬ remark-stringify 11.0.0
│ │ │ └─┬ unified 11.0.5
│ │ │   └── trough 2.1.0
│ │ └─┬ unified 11.0.5
│ │   └── trough 2.1.0
│ ├─┬ remark-parse 11.0.0
│ │ └─┬ unified 11.0.5
│ │   └── trough 2.1.0
│ ├─┬ remark-rehype 11.1.1
│ │ └─┬ unified 11.0.5
│ │   └── trough 2.1.0
│ ├─┬ remark-smartypants 3.0.2
│ │ ├─┬ retext 9.0.0
│ │ │ ├─┬ retext-latin 4.0.0
│ │ │ │ └─┬ unified 11.0.5
│ │ │ │   └── trough 2.1.0
│ │ │ ├─┬ retext-stringify 4.0.0
│ │ │ │ └─┬ unified 11.0.5
│ │ │ │   └── trough 2.1.0
│ │ │ └─┬ unified 11.0.5
│ │ │   └── trough 2.1.0
│ │ └─┬ unified 11.0.5
│ │   └── trough 2.1.0
│ └─┬ unified 11.0.5
│   └── trough 2.1.0
└─┬ rehype 13.0.2
  ├─┬ rehype-parse 9.0.0
  │ └─┬ unified 11.0.5
  │   └── trough 2.1.0
  ├─┬ rehype-stringify 10.0.1
  │ └─┬ unified 11.0.5
  │   └── trough 2.1.0
  └─┬ unified 11.0.5
    └── trough 2.1.0
```

https://stackblitz.com/edit/withastro-astro-3nw2q24q?file=astro.config.mjs

### Participation

Reference: [Upgrade @starlight dependency Astro to 5.9 to support new API renderMarkdown](https://discord.com/channels/830184174198718474/1381890313546891265)

HiDeoo:
#### Description

- Closes #3239

This PR builds on top of #3241 where I cherry-picked and co-auhored the initial fix preventing errors when using the Astro [`renderMarkdown()`](https://docs.astro.build/en/reference/content-loader-reference/#rendermarkdown) content loader API.

This PR also:

- Refactor such fix to make it easier to check if we should transform a file or not using remark and rehype plugins.
- Apply such check to other Starlight remark and rehype plugins like #3181 initially did for heading anchor links, e.g. at the moment, Starlight asides are rendered in non-Starlight pages.


.changeset/unlucky-bananas-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue where some Starlight remark and rehype plugins were transforming Markdown and MDX content in non-Starlight pages.
```
delucis, line 5: Would this break usage of our plugins in e.g. Starlight Blog? I guess not because that uses `src/content/docs/` still?

I’m a bit hesitant about this part. One the one hand it makes sense — I can totally see someone not wanting to have Starlight’s Markdown processing applying in other Markdown content, like we’ve occasionally heard for Expressive Code (not covered in this PR) — but on the other hand, I can also see scenarios where some people might already depend on our pipeline running for another collection? Even something simple like having a `src/content/partials/` that get reused and could contain asides for example.

We might need to offer some configuration control potentially?

HiDeoo: > Would this break usage of our plugins in e.g. Starlight Blog? I guess not because that uses `src/content/docs/` still?

Yeah, this would not break anything in this case.

I was a bit hesitant too altho a few reasons made me go for it initially:

- Why transforming the HTML to add heading anchor links would be disabled in non-Starlight pages but transforming directives to an aside element with an icon would not?
- We got lots of reports right after the heading anchor links release and we never got any for the aside directives. I think this is because heading are a common thing people use in their content, while nobody try to use Starlight directives in non-Starlight pages?
- You don't get any styles if you rely on using something like asides in non-Starlight pages, e.g. in an individual Markdown pages, so not even sure if this would be useful in practice.


Although the partials example is a good one, but I'm still unsure if asides should be any different from headings in this regard :thinking:

A configuration could indeed be added, altho I wonder what kind of configuration would be useful here, e.g. I want Starlight processing in my partials but I don't want it in my custom pages? Would this need to be per-Starlight remark/rehype plugin, e.g. I want Starlight asides in my custom pages but not Starlight headings?

I think we definitely need to think about this more (at least I think the PR is useful to identify the various points that may need handling regarding #3239) altho I guess no matter what we decide, not making it a patch may be a good idea.

delucis: Hmm, that’s true — asides only kick in for that syntax, so you’re less likely to hear that it’s messing with regular Markdown. I guess that makes sense. And the point about styles missing is a good one (although it could apply for example in a custom route that uses `<StarlightPage>` to render some Markdown files, where currently it would work no problem).

OK, so then this would be a minor “breaking” change and in the changeset we can ask people to let us know if they have use cases where they need the Starlight processing in other places.

And yes, controlling _where_ could indeed be tricky. I think the only flexible way we could have would be to allow some kind of glob pattern to define the directories for Markdown to be processed?

HiDeoo: I'll think about it more tonight and try to update the changeset tomorrow morning and see what it would look like to:

- Release in a minor
- Present it as a potential breaking change
- Making it really obvious in which scenarios it could break
- Indicate that we are open to any feedback regarding such change

.changeset/unlucky-bananas-accept.md
```diff
@@ -0,0 +1,11 @@
...
+
+⚠️ **BREAKING CHANGE:**
+
+Previously, some of Starlight’s remark and rehype plugins, most notably the plugin transforming Starlight's custom Markdown syntax for [rendering asides](https://starlight.astro.build/guides/authoring-content/#asides), were applied to all Markdown and MDX content. This included content from [individual Markdown pages](https://docs.astro.build/en/guides/markdown-content/#individual-markdown-pages) and content from [content collections](https://docs.astro.build/en/guides/content-collections/) other than the `docs` collection used by Starlight.
+
+This change restricts the application of Starlight’s remark and rehype plugins to only Markdown and MDX content loaded using Starlight's [`docsLoader()`](https://starlight.astro.build/reference/configuration/#docsloader). While never intended, if you were relying on this behavior, please let us know in the dedicated `#starlight` channel in the [Astro Discord](https://astro.build/chat/) or by [opening an issue](https://github.com/withastro/starlight/issues/new?template=---01-bug-report.yml).
```
HiDeoo: The `docsLoader()` link reference the new section added in #3272.

delucis: Maybe a slight tweak here? But otherwise the changeset looks great 👍 

```suggestion
This change restricts the application of Starlight’s remark and rehype plugins to only Markdown and MDX content loaded using Starlight's [`docsLoader()`](https://starlight.astro.build/reference/configuration/#docsloader). If you were relying on this behavior, please let us know about your use case in the dedicated `#starlight` channel in the [Astro Discord](https://astro.build/chat/) or by [opening an issue](https://github.com/withastro/starlight/issues/new?template=---01-bug-report.yml).
```

delucis (approved):
This looks great — thank you @HiDeoo!

One thought seeing this is whether a future refactor might be to move our Markdown plugins into some kind of “parent” plugin perhaps that can manage running checks like the file path one for all plugins. That would avoid running the checks three times per each file and avoid needing to remember to duplicate the check when adding something new.

PR #3278, opened 2025-07-06
docs: add Action Continuous Translation integration

trueberryless:
#### Description

This PR adds the [Action Continuous Translation](https://github.com/pelikhan/action-continuous-translation) integration to the plugins page / integration sector. I do not mention in the description of the Card that it has Starlight support but since it will be listed on the Starlight Integration page, I think people can assume that it has something to do with Starlight. If you want me to specify further that there is Starlight support, or if you want that I directly link to the [Astro Starlight page of the integration](https://pelikhan.github.io/action-continuous-translation/reference/astro-starlight/), just ping me to let me know! 🙌

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -158,6 +158,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		title="starlight-scroll-to-top"
 		description="Add a scroll-to-top button to your documentation site."
 	/>
+	<LinkCard
+		href="https://github.com/frostybee/starlight-codeblock-fullscreen"
+		title="starlight-codeblock-fullscreen"
+		description="Add fullscreen toggle functionality to Expressive Code blocks in your documentation."
+	/>
 </CardGrid>
 
 ## Community tools and integrations


delucis (commented):
Thanks for the PR @trueberryless! This doesn’t seem to be Starlight or even Astro-specific right? Not sure if we should be aiming to include generic tools like this here.

trueberryless:
As I see it, it is Starlight related, just not purely made _for_ Starlight. The integration provides clear instruction on how to set it up for a Starlight project: https://pelikhan.github.io/action-continuous-translation/reference/astro-starlight/

delucis:
While it’s true this has a Starlight “recipe”, we generally focus on including Starlight-specific tools as otherwise there is almost an infinity of things we could include. We have in the past made some exceptions for Astro integrations like the PlantUML and Mermaid ones because they provide something people ask for a lot in relation to Starlight, but on the whole, automated translation is not a common user need and also not a best practice for documentation where it’s preferable to allow and end user to choose automated translation (e.g. via their browser), giving them full context on judging the reliability of what they are reading rather than offering something automated where a user doesn’t have that context.

trueberryless:
Alright. For me that's a little bit sad, because we are kinda putting much work into supporting all Starlight features, but I understand the limitation because of overflows. Do you know any other resources where we maybe could provide a link to it? Like Astro integrations? Astrolicious? 👀 I would really appreciate spreading the words somehow because imo it's really cool addition to documentation in general. 💜

PR #3281, opened 2025-07-08
Made working with sidebars a little easier

alvinometric:
## Description

- Closes #3264
- Added a new `'sl-sidebar-link'` class to sidebar links.
- Exported the `SidebarEntry` type

### DIFF
--- a/packages/starlight/types.ts
+++ b/packages/starlight/types.ts
@@ -5,3 +5,4 @@ export type {
 	HookParameters,
 } from './utils/plugins';
 export type { StarlightIcon } from './components/Icons';
+import type { SidebarEntry } from './utils/routing/types';
--- a/packages/starlight/types.ts
+++ b/packages/starlight/types.ts
@@ -5,4 +5,4 @@ export type {
 	HookParameters,
 } from './utils/plugins';
 export type { StarlightIcon } from './components/Icons';
-import type { SidebarEntry } from './utils/routing/types';
+export type { SidebarEntry } from './utils/routing/types';
--- a/packages/starlight/components/SidebarSublist.astro
+++ b/packages/starlight/components/SidebarSublist.astro
@@ -21,7 +21,7 @@ const { sublist, nested } = Astro.props;
 					<a
 						href={entry.href}
 						aria-current={entry.isCurrent && 'page'}
-						class:list={[{ large: !nested }, entry.attrs.class]}
+						class:list={['sl-sidebar-link', { large: !nested }, entry.attrs.class]}
 						{...entry.attrs}
 					>
 						<span>{entry.label}</span>


packages/starlight/types.ts
```diff
@@ -5,3 +5,4 @@ export type {
 	HookParameters,
 } from './utils/plugins';
 export type { StarlightIcon } from './components/Icons';
+export type { SidebarEntry } from './utils/routing/types';
```
delucis: Could you share your use case for needing access to `SidebarEntry` — it’s an internal type, so I’d like to learn where you find you’re needing to interact with these objects and need the type. (Might help us improve the API in more fundamental ways.)

delucis (commented):
Thanks for the PR @alvinometric! Left one small question.

Would you also be able to add changesets to the PR describing the changes for our changelog. I think these would both be OK as `patch` changes and each change should get its own changeset. You can add them by running `pnpm changeset` in the repo root or clicking the “Add a changeset” link in the bot comment in this PR.

packages/starlight/types.ts
HiDeoo: Depending on the use case, it may also already be accessible using both [`ComponentProps`](https://starlight.astro.build/components/using-components/#component-props) or [`StarlightRouteData`](https://starlight.astro.build/reference/route-data/#starlightroutedata-type) types.

alvinometric: I'm building a sidebar based on data, and my code looks a bit like this:

```ts

  // Build sidebar sections in the order of spec.tags
  const sidebar = spec.tags
    .map((t: any) => {
      const name = t.name;
      return {
        label: name,
        items: groups[name],
      };
    })
    .toSorted((a: SidebarGroup, b: SidebarGroup) => {
      return a.label.localeCompare(b.label);
    });
  ```
  
  I **had to make** the `SidebarGroup` type. TypeScript/Cursor couldn't find it

Comment (alvinometric) [reply packages/starlight/types.ts]: I tried again, and TS could find the `StarlightRouteData` type but nothing for sidebar

Comment (HiDeoo) [reply packages/starlight/types.ts]: Starlight route data exposes [many properties](https://starlight.astro.build/reference/route-data/#starlightroute), so you would need to cherry-pick the one you need, for example:

```ts
import type { StarlightRouteData } from '@astrojs/starlight/route-data';

// This type represents a single entry in the `sidebar` array of the Starlight route data.
type SidebarEntry = StarlightRouteData['sidebar'][number];
```

Something to note, even if the `SidebarEntry` type is directly exported like in the PR or grabbed from the Starlight route data, to specifically target `SidebarGroup` or `SidebarLink`, you would need at the moment some TypeScript code to [extract](https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union) those types, e.g.:

```ts
type SidebarGroup = Extract<SidebarEntry, { type: 'group' }>;
type SidebarLink = Extract<SidebarEntry, { type: 'link' }>;
```

alvinometric: What do I have to do so that the type is available just like `import type { SidebarEntry } from '@astrojs/starlight/'` ?

delucis: > What do I have to do so that the type is available just like `import type { SidebarEntry } from '@astrojs/starlight/'` ?

By “the type”, do you mean `SidebarGroup`/`SidebarLink`? You’d need to export those too.

alvinometric: Ah ok, fixed it 🙏

delucis: Would it be OK if we split the type changes into a different PR for discussion? We can get the previously discussed class name change released quickly, but I think we might want some more time to discuss about types. As HiDeoo mentioned, we already expose these types via `StarlightRouteData`, but there are a _lot_ of interfaces in there, so if we’re going to start exporting things, we probably need to do it systematically, for example to answer questions like:

- Do we export all types?
- Should we document each type?

I wouldn’t want to block the class name change behind that discussion, so separate PRs would be helpful I think.

delucis: > Would it be OK if we split the type changes into a different PR for discussion?

Going to go ahead and do this so we can get the class name change released.

delucis:
Er… very sorry, but I jumped the gun on this one. Explained more reasoning in #3291, but we’re going to revert this. Sorry for first encouraging the PR and then walking that back — I hadn’t considered the big picture.

alvinometric:
@delucis 😭

In all honesty I'm OK if this https://github.com/withastro/starlight/pull/3266 gets merged, and that the "built-in way" is properly documented.

PR #3287, opened 2025-07-10
docs: fix `t.exists()` documentation + example

HiDeoo:
#### Description

This PR updates the documentation for the `t.exists()` function in the i18n guide. It clarifies (and fixes) the usage of the function, including how to check if a translation key exists in a specific locale without considering fallback logic (which would be the behavior when using the `lng` option).

### DIFF
--- a/docs/src/content/docs/guides/i18n.mdx
+++ b/docs/src/content/docs/guides/i18n.mdx
@@ -360,15 +360,15 @@ const allStrings = Astro.locals.t.all();
 
 #### `t.exists()`
 
-To check if a translation key exists for a locale, use the `locals.t.exists()` function with the translation key as first argument.
-Pass an optional second argument if you need to override the current locale.
+To check if a translation key exists, use the `locals.t.exists()` function with the translation key as first argument.
+Pass an optional second argument if you need to check if a key exists in a specific locale without considering any fallback logic.
 
 ```astro
 ---
 // src/components/Example.astro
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/i18n.mdx
```diff
@@ -360,15 +360,15 @@ const allStrings = Astro.locals.t.all();
 
 #### `t.exists()`
 
-To check if a translation key exists for a locale, use the `locals.t.exists()` function with the translation key as first argument.
-Pass an optional second argument if you need to override the current locale.
+To check if a translation key exists, use the `locals.t.exists()` function with the translation key as first argument.
+Pass an optional second argument if you need to check if a key exists in a specific locale without considering any fallback logic.
```
delucis: Maybe this might be a clearer way to say this?

```suggestion
Pass an optional second argument if you need to check if a translation exists for a specific locale.
```

“fallback logic” is kind of how we think about it internally, but I guess from a more user-facing POV, the thing you’re checking is if a _translation_ is available?

delucis (commented):
Thanks for updating this based on your experiences! Just left one small suggestion.

docs/src/content/docs/guides/i18n.mdx
HiDeoo: Oh, I really like that, so much better 🧠 Thanks

delucis (approved):
LGTM!

```js
t.exists('better.docs') === true // 🥳
```

PR #3298, opened 2025-07-14
Fix `absolutePathToLang()` issue

### LINKED ISSUE #3297: When the starlight project is located on a path with blank, some test run fails
### Describe the Bug

> This bug was found in #3241 

Image

I find these three run fail tests related to fr, and I run another machine, all test runs pass, GitHub action also, so this may be related to the local environment, so I do the below debugging to confirm it.

Image

Image

That's because `absolutePathToLang()` doesn't work well when the path location is in a bank path.

undefined

### Participation

HiDeoo:
#### Description

- Closes #3297

This PR fixes an issue with the `absolutePathToLang()` helper function not handling absolute paths with spaces correctly.

- `absolutePathToLang()` is meant to be called with the `path` property of a virtual file (altho not mandatory), which is always a POSIX-style path (using `/` as a separator) no matter the operating system.
- It calls `getCollectionPath()` that relies on an URL `pathname` which will also be POSIX-style but also encoded, e.g. spaces are replaced with `%20`.
- A virtual file's `path` property does not encodes special characters like spaces.

This can lead to issues when the path contains spaces which can be solved in different approaches:

- Using `fileURLToPath()` in `getCollectionPath()` to convert the URL to a file path, altho depending on the OS, different separators will be used and would require an additional step to convert it again to a POSIX-style path.
- Relying on `pathToFileURL()` on the input path of `absolutePathToLang()` to convert it to a URL-encoded path also using POSIX-style separators and using its `pathname`.

I went with the second approach which replaces an existing step instead of adding a new one.

On top of the added test, I also manually tested on Windows.

### DIFF
--- a/.changeset/dry-tomatoes-camp.md
+++ b/.changeset/dry-tomatoes-camp.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a potential issue with `absolutePathToLang()` not handling paths with spaces correctly.
--- a/packages/starlight/__tests__/plugins/translations.test.ts
+++ b/packages/starlight/__tests__/plugins/translations.test.ts
@@ -68,4 +68,6 @@ test('can infer langs from an absolute path in the plugin context using absolute
 	expect(langs[1]).toBe('pt-BR');
 	// A page not matching any language defaults to the default language.
 	expect(langs[2]).toBe('en');
+	// A known language and an absolute path containing spaces.
+	expect(langs[3]).toBe('ar');
 });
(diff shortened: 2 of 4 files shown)


delucis (commented):
Thanks @HiDeoo, this looks good.

One question: `absolutePathToLang()` is a [public API](https://starlight.astro.build/reference/plugins/#absolutepathtolang). Would we expect this change to impact Starlight plugin authors at all? I notice we have the test file changing the direct use of `URL#pathname`, but I guess that was a slightly unusual usage and not something we’d expect real-world users to be doing?

HiDeoo:
Following your comment, I pushed 2 changes that I thought would be useful:

- I mentioned in the changeset that this function is part of the plugin API.
- I added a comment on why we are specifically using `fileURLToPath()` in our tests.

To answer your question, the only way this change would impact a plugin API authors would be if they were somehow relying on the previous behavior that would result in an invalid language in some cases, e.g. with spaces in the path (I cannot imagine why or how). Otherwise, it's a fix that would benefit all plugins using that API.

delucis (approved):
Perfect — thanks very much @HiDeoo! 🌟

PR #3300, opened 2025-07-15
Removed empty footer in Splash template

alvinometric:
#### Description

In `template: "splash"` an empty footer is added. The problem is it creates an empty margin at the bottom

image

So, this PR removes the footer in that template.

### DIFF
--- a/packages/starlight/components/Page.astro
+++ b/packages/starlight/components/Page.astro
@@ -101,7 +101,6 @@ if (pagefindEnabled) mainDataAttributes['data-pagefind-body'] = '';
 								<MarkdownContent>
 									<slot />
 								</MarkdownContent>
-								<Footer />
 							</ContentPanel>
 						) : (
 							<>


delucis:
The footer is intentionally present there. Although empty in our landing page case, just because a page doesn’t use the `hero` frontmatter, doesn’t guarantee a page doesn’t have footer content. Pages with `hero` can include any of pagination, edit links, last-updated dates, for example. Plus they can also include additional UI users may have added in a component overrides, so we can’t assume that it should be removed.

For just one example, see the [Astro docs landing page](https://docs.astro.build/en/getting-started/).

alvinometric:
Right, then could I make another PR that adds a `template: "empty"` ?

This would just have the navbar, but no title, sidebar, or footer.

delucis:
Could you open a [feature request](https://github.com/withastro/starlight/discussions/categories/feature-requests) first? We’ve had various similar requests before (see e.g. https://github.com/withastro/starlight/discussions/1112), but I don’t think those would have expected to exclude the footer, so not sure it’s a clear case for what should/shouldn’t be included for an “empty” template.

I’m also a bit confused: if the issue was to do with `margin` on a component (something it’s possible to customise using some [custom CSS styles](https://starlight.astro.build/guides/css-and-tailwind/)), then reaching for an entirely new `template` option feels like a big hammer. So probably good to actually walk through your use case and issues to figure out the correct solution. That’s what GH discussions are great for.

alvinometric:
There you go, here's the discussion https://github.com/withastro/starlight/discussions/1112

PR #3302, opened 2025-07-15
docs: showcase `astro-d2` and "Starlight Plugins by Example"

HiDeoo:
#### Description

This PR adds the following resources to the documentation:

- The [`astro-d2`](https://github.com/HiDeoo/astro-d2) integration, which provides another alternative to the PlantUML and Mermaid formats already linked in the documentation.
- A link to the [Starlight Plugins by Example](https://hideoo.dev/notebooks/starlight-plugins-by-example) notebook I wrote.

PR #3318, opened 2025-07-16
i18n(fr): update `resources/plugins.mdx` & `resources/community-content.mdx`

HiDeoo:
#### Description

This PR updates the French version of the `resources/plugins` & `resources/community-content` pages with the changes from #3302.

### DIFF
--- a/docs/src/content/docs/fr/resources/community-content.mdx
+++ b/docs/src/content/docs/fr/resources/community-content.mdx
@@ -109,6 +109,11 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 		title="Exemples Starlight"
 		description="Collection de ressources StackBlitz démontrant par la pratique comment mettre en place diverses fonctionnalités dans des sites de documentation Starlight."
 	/>
+	<LinkCard
+		href="https://hideoo.dev/notebooks/starlight-plugins-by-example"
+		title="Modules d'extension Starlight par l'exemple"
+		description="Une collection de notes et d'exemples sur les modules d'extension Starlight et techniques courantes utilisées pour les construire."
+	/>
 </CardGrid>
 
 ## Contenu vidéo
--- a/docs/src/content/docs/fr/resources/community-content.mdx
+++ b/docs/src/content/docs/fr/resources/community-content.mdx
@@ -112,7 +112,7 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 	<LinkCard
 		href="https://hideoo.dev/notebooks/starlight-plugins-by-example"
 		title="Modules d'extension Starlight par l'exemple"
-		description="Une collection de notes et d'exemples sur les modules d'extension Starlight et techniques courantes utilisées pour les construire."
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/fr/resources/community-content.mdx
```diff
@@ -109,6 +109,11 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 		title="Exemples Starlight"
 		description="Collection de ressources StackBlitz démontrant par la pratique comment mettre en place diverses fonctionnalités dans des sites de documentation Starlight."
 	/>
+	<LinkCard
+		href="https://hideoo.dev/notebooks/starlight-plugins-by-example"
+		title="Modules d'extension Starlight par l'exemple"
```
HiDeoo, line 114: Reference: https://www.amazon.fr/s?k=par+l%27exemple

docs/src/content/docs/fr/resources/community-content.mdx
```diff
@@ -109,6 +109,11 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
...
+		description="Une collection de notes et d'exemples sur les modules d'extension Starlight et techniques courantes utilisées pour les construire."
```
HiDeoo: Slightly different from the English version but I was not happy with words like "motifs", etc. I think this translation reflects more what I wanted to say, and considering I'm the author, I guess that's fine? 😅

docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -222,4 +222,9 @@ Ces outils et intégrations communautaires peuvent être utilisés pour ajouter
 		title="astro-mermaid"
 		description="Générer du contenu Mermaid côté client dans des blocs de code Markdown."
 	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/astro-d2"
+		title="astro-d2"
+		description="Transformer des blocs de code Markdown D2 en diagrammes."
```
ArmandPhilippot: I'm fine with this. Just a note as a non-aware of what is D2: I was wondering - naively - if D2 was a variant of Markdown while this is a different syntax to represent diagrams. So, I guess we could say `Transformer des blocs de code Markdown utilisant D2 en diagrammes`. But, this is just a nit, not sure if this is better... Maybe D2 is well-know, or those looking for this would be fine with "Markdown D2".

docs/src/content/docs/fr/resources/community-content.mdx
ArmandPhilippot: Oh I didn't see this in the files preview because of the update, but I think this match pretty well the English version!
Yeah, I don't think the i18n guide in the Astro docs repo covers all the variants for `patterns`, so it was more to incite translators to find better alternatives than a literal translation. In the tutorial I choose `formule` because of the context, and here I think `technique` is the perfect word!

docs/src/content/docs/fr/resources/plugins.mdx
HiDeoo: Oh, really interesting point 👍 Used your suggestion which workaround that quite nicely imo.

PR #3327, opened 2025-07-18
Fix a routing issue with unnormalized content collection slugs

delucis:
#### Description

- This PR fixes mapping between `Astro.params` and content collection entries for entries which have a filename with non-normalized composition.
- For example, `اللُّغَةُ-الْعَرَبِيَّةُ/الْآجُرُومِيَّةُ.md` uses a mix of precomposed and decomposed diacritics. However, Astro’s router [always normalizes URL parameters](https://github.com/withastro/astro/blob/687d25365a41ff8a9e6da155d3527f841abb70dd/packages/astro/src/core/routing/manifest/generator.ts#L13), which led to a mismatch between Starlight’s routes and Astro’s params, resulting in a 404 for a file with that name.
- Starlight’s internal route map now normalizes slugs to match Astro’s behaviour.
- One issue this can’t fix is differences between the composition of a user-authored link in Markdown and the actual built URL. For example, this change fixes the 404 issue but this means the output HTML file name is now differently composed than the input Markdown file name. If a user-authored link uses the same composition as the input file name, that will still be a 404, because it doesn’t match the output composition. Long story short, maybe people should avoid risking diacritics in URLs 😅 But at least we’re now matching Astro’s behaviour so there is a path to doing this without manually normalizing your file names.

**N.B.** The test changes in this PR look big because I wrapped existing tests in a `describe()` block, but apart from that those haven’t actually changed. I just renamed `components.test.ts` to `basics.test.ts` to match and reuse the basics fixture and avoid spinning up a new fixture just for this.

### DIFF
--- a/packages/starlight/utils/slugs.ts
+++ b/packages/starlight/utils/slugs.ts
@@ -39,6 +39,14 @@ function localeToDir(locale: string | undefined): 'ltr' | 'rtl' {
 	return dir || config.defaultLocale.dir;
 }
 
+/**
+ * Convert a content collection slug to a param as expected by Astro’s router.
+ * This utility handles stripping `index` from file names and matches
+ * [Astro’s param sanitization logic](https://github.com/withastro/astro/blob/687d25365a41ff8a9e6da155d3527f841abb70dd/packages/astro/src/core/routing/manifest/generator.ts#L4-L18)
+ * by normalizing strings to their canonical representations.
+ * @param slug Content collection slug
+ * @returns Param compatible with Astro’s router
+ */
 export function slugToParam(slug: string): string | undefined {
 	return slug === 'index' || slug === '' || slug === '/'
 		? undefined


packages/starlight/__tests__/basics/slugs.test.ts
```diff
@@ -43,6 +43,9 @@ describe('slugToParam', () => {
 		expect(slugToParam('dir/page')).toBe('dir/page');
 		expect(slugToParam('dir/sub-dir/page')).toBe('dir/sub-dir/page');
 	});
+	test('normalizes unnormalized strings', () => {
+		expect(slugToParam('اللُّغَةُ-الْعَرَبِيَّةُ')).toBe('اللُّغَةُ-الْعَرَبِيَّةُ');
```
HiDeoo, line 50: Should we provide a bit more details, e.g. for someone without context reading the test thinking the input and output are identical and should match? Maybe something like?

```suggestion
		// The input and output contain visually indistinguishable glyphs with different code point
		// representation that are canonically equivalent that should be normalized using NFC
		// normalization matching Astro's behavior.
		expect(slugToParam('اللُّغَةُ-الْعَرَبِيَّةُ')).toBe('اللُّغَةُ-الْعَرَبِيَّةُ');
```

HiDeoo (approved):
Left a tiny suggestion to add a comment, up to you if you want to add it or not, but otherwise, looks great and makes sense (as much as it can considering the topic ^^) to me.

Great investigation and fix :clap:

packages/starlight/__tests__/basics/slugs.test.ts
delucis: Oh, very good point. And same goes for the `slugToParam` use of normalize itself. I’ll add a comment there too.

PR #3332, opened 2025-07-22
Relax remark/rehype plugins usage

HiDeoo:
#### Description

This draft PR is a follow-up to #3274 which restricted the usage of Starlight remark and rehype plugins to only Markdown and MDX content loaded using Starlight's `docsLoader()`. As initially discussed and based on feedback, this PR relaxes that restriction using a new `markdown.processedDirs` (temporary name to be discussed) option that allows specifying additional directories where Markdown content should be processed by Starlight's remark and rehype plugins.

A few key points:

- As initially [pointed out](https://github.com/withastro/starlight/pull/3274#pullrequestreview-2986640748) by Chris, checks are currently run by all Starlight remark and rehype plugins. If we're to potentially add more checks by allowing more directories, this PR felt like a good opportunity to refactor and limit the number of checks. Two ideas were envisaged:
  - Have some kind of "parent" plugin that runs the checks and then calls the other plugins if needed (this means 1 check for remark and 1 for rehype).
  - Have a remark plugin that runs before any other and using an approach similar to Astro saving a field in the vfile data, e.g. `file.data.starlight.skip` or something like that with a symbol, which plugins running after could check and decide if processing should continue or not (this means 1 check for both remark and rehype).
  - This PR explores the first option in a dedicated commit so we can easily revert/modify it if needed.
- I went with directories paths relative to the project root instead of something like globs for this first iteration. I still wonder if we need the extra complexity/flexibility/cost of globs or just specifying directories is enough.


alvinometric:
`markdown.includes` ? Rather than `processedDirs`

LasseRosenow:
Is this still being worked on? We depend on it for our documentation, because we would like to make our guides not so much hidden inside the starlight folder, to make contribution more accessible :)

Right now we are working around it using symlinks, but it also creates problems of their own, see: #3470

HiDeoo:
> Is this still being worked on?

Yes, this is not something we've been able to prioritize recently, but we do plan to get back to it and ship it in the next minor release hopefully.

LasseRosenow:
That sounds amazing, thank you so much :)

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -399,6 +399,28 @@ starlight({
...
+#### `processedDirs`
+
+**type:** `string[]`  
+**default:** `[]`
+
+Define additional directories where files should be processed by Starlight’s Markdown pipeline.
+By default, only Markdown and MDX content loaded using Starlight's [`docsLoader()`](/reference/configuration/#docsloader) is processed.
+Supports local directories relative to the root of your project, e.g. `'./src/data/comments/'`.
+
+Such processing includes support for [clickable heading anchor links](#headinglinks), [asides Markdown directive syntax](/guides/authoring-content/#asides), and RTL support for code blocks.
```
delucis: Small nit:

```suggestion
Starlight’s processing includes support for [clickable heading anchor links](#headinglinks), [asides Markdown directive syntax](/guides/authoring-content/#asides), and RTL support for code blocks.
```

packages/starlight/__e2e__/fixtures/basics/src/content.config.ts
```diff
@@ -5,9 +5,13 @@ import { glob } from 'astro/loaders';
...
-	// A collection not handled by Starlight.
+	// Collections not handled by Starlight.
 	reviews: defineCollection({
 		loader: glob({ base: './src/content/reviews', pattern: `**/[^_]*.{md,mdx}` }),
 		schema: z.object({ title: z.string() }),
 	}),
+	comments: defineCollection({
+		loader: glob({ base: './src/content/comments', pattern: `**/[^_]*.{md,mdx}` }),
+		schema: z.object({ title: z.string() }),
+	}),
```
delucis: Should the comments here reflect the Markdown usage?

```suggestion
	// A collection not handled by Starlight.
reviews: defineCollection({
loader: glob({ base: './src/content/reviews', pattern: `**/[^_]*.{md,mdx}` }),
schema: z.object({ title: z.string() }),
}),
	// A collection opted in to Starlight’s Markdown processing.
	comments: defineCollection({
		loader: glob({ base: './src/content/comments', pattern: `**/[^_]*.{md,mdx}` }),
		schema: z.object({ title: z.string() }),
	}),
```

packages/starlight/utils/user-config.ts
```diff
@@ -255,6 +255,23 @@ const UserConfigSchema = z.object({
...
 				),
+			/**
+			 * Define additional directories where files should be processed by Starlight’s Markdown pipeline which includes support for:
+			 *
+			 *  - Clickable heading anchor links
+			 *  - Asides Markdown directive syntax
+			 *  - RTL support for code blocks
+			 *
+			 * Supports local directories relative to the root of your project, e.g. './src/data/comments/'.
+			 * Content of the `docs` content collection is always processed by Starlight’s Markdown pipeline.
```
delucis: This is very helpful, but I’m a bit nervous about listing the supported features here, because I suspect we’d quickly forget to update it when adding more Markdown features. Maybe we can omit it?
```suggestion
			 * Define additional directories where files should be processed by Starlight’s Markdown pipeline.
			 *
			 * Supports local directories relative to the root of your project, e.g. './src/data/comments/'.
			 * Content of the `docs` content collection is always processed by Starlight’s Markdown pipeline.
```

packages/starlight/index.ts
```diff
@@ -115,31 +118,22 @@ export default function StarlightIntegration(
...
-							...starlightAutolinkHeadings({
-								starlightConfig,
-								astroConfig: config,
-								useTranslations,
-								absolutePathToLang,
-							}),
-						],
+						remarkPlugins: [...starlightRemarkPlugins(remarkRehypeOptions)],
+						rehypePlugins: [...starlightRehypePlugins(remarkRehypeOptions)],
 					},
```
delucis, line 137: Given we’re already abstracting these, do you think it might make sense to move the whole config behind an abstraction?

i.e. `markdown: starlightMarkdownConfig(remarkRehypeOptions)`

HiDeoo: Interesting, did not think of that. Just tried it locally and it definitely simplifies the code a bit in this file (`packages/starlight/index.ts`), I am not a big fan of the impact on tests.

For example, in some tests, we dynamically add plugins after Starlight's own plugins. With this change, this now means we need to first get the full markdown config from `starlightMarkdownConfig()`, then spread specific type of plugins and add our own:

```ts
const markdownConfig = starlightMarkdownConfig(
	await createRemarkRehypePluginTestOptions(starlightConfig)
);

const processor = await createMarkdownProcessor({
	...markdownConfig,
	remarkPlugins: [
		...markdownConfig.remarkPlugins,
		// The restoration plugin is run after the asides and any other plugin that may have been
		// injected by Starlight plugins.
		remarkDirectivesRestoration,
	],
});
```

We also have some specific tests that currently focus on a type of plugin and expect to not run with the other type, e.g. we have a test only running remark plugins with the `## Environment variables (astro:env)` input to test the aside directive handling. In this case, we expect the output to be `<h2 id="environment-variables-astroenv">Environment variables (astro:env)</h2>` which is simple enough to assert. But if we run with the full markdown config, the rehype plugins will also run and add heading links.

We could of course adjust the tests to account for that, but it makes the assertions more complex and less focused on the specific feature being tested. We could also override the rehype plugins to an empty array in these specific tests, but that feels a bit less practical.

Another alternative, instead of only exporting `starlightMarkdownConfig()`, would be also to keep the current exports of `starlightRemarkPlugins()` and `starlightRehypePlugins()` for tests, but this feels like we're not fully committing and testing this abstraction.

I have kept the exploration of this suggestion in a stash if ever needed, but I'm not sure it's worth it given the above. What do you think?

HiDeoo:

Updated the PR with the suggested changes, and commented on one of them but happy to discuss further!

I've also added a changeset for this new configuration option.

packages/starlight/index.ts
delucis: That sounds good to me! Sometimes it’s worth not simplifying in one place if it creates more complexity elsewhere 👍

delucis (approved):
Took the liberty of moving this out of draft so I can approve it! I think it’s not just ready for review, it’s ready to merge! 🚀 (Once we’re ready for the next minor of course.)

PR #3276, opened 2025-07-04
Exclude banner content from Pagefind indexing

delucis:
#### Description

- Closes #3267
- Partially addresses the issue raised in #3267 where a banner’s content would end up indexed by Pagefind and displayed in search results by excluding the default `Banner.astro` component from indexing.
- #3271 will address the styling issues more fully for cases where this still occurs, but I think it makes sense for us to exclude banner content nonetheless as it’s often unexpected for this content to show up in search results, especially for sites where people set banners to show up on all pages for example.

PR #3335, opened 2025-07-23
New Pagefind UI

HiDeoo:
#### Description

This draft PR is a very early version of a rewrite of the Pagefind UI that is used in Starlight.

This is a work in progress and is not ready for production or even for a full review yet. A lot of work is still needed to get it to a point where it can be used in Starlight. Most of the work has only been done in 1 browser and parts of the UI have been tested in isolation against only 1 screen reader so far so it'll require a lot of testing and improvements before it can be considered ready in its entirety.

So far, this PR only focuses on features and options that can be used or overridden in Starlight with the goal of having more control over the search UI and improving the accessibility of the search experience in Starlight.

A few important points to note:

- [Filters](https://pagefind.app/docs/filtering/) are always visible as some have been configured in the content which is not the case in the live version of the documentation.
- Translations are now part of Starlight although I did not port the [existing ones](https://github.com/Pagefind/pagefind/tree/main/pagefind_ui/translations) from the original Pagefind UI. That was my original idea although after checking the French one, I was not pleased with some of the translations. We can think about it later.
- The UI aims at first to mimic the original Pagefind UI as closely as possible to iterate faster on a working veersion. Only a few details may have changed, e.g. filters disclosure icon now matching the one used in the Starlight Markdown UI, etc.
- The current UI used in the live version of the documentation was potentially limited due to the lack of control when using the original Pagefind UI. Something to potentially consider later is if there are any changes to the original UI that we would like to see in this version.
  - Something coming to mind could be a sticky query input but that's for later.

An interesting search is `upgrade`:

- It displays some [metadata](https://pagefind.app/docs/metadata/).
- With this branch not being updated with #3276, the banner content of the homepage is still indexed and this search will display it as a single root result with an excerpt and no sub-results (which is not broken like in #3267).

Development notes:

- This PR adds the `pnpm pagefind` command to the `docs/` directory which should be used when iterating on this PR:
  - It simultaneously runs a dev version of the docs running on the `4321` port and a prod version on the `4322` port.
  - When developing, visiting `http://localhost:4321` will use the search indexes from the prod version which allows for using/testing/iterating on the search UI (note that changes that would result in a change to the search indexes will require running the `pnpm pagefind` command again).

#### Remaining todos

- [ ] Look for any `TODO(HiDeoo)` remaining comments and address them.
- [ ] Add changeset(s)


HiDeoo:
Alongside some minor changes and improvements, this PR now also includes an always-visible query input. It does not looks great and it's part of the points I initially mentioned in the description of this PR that should be discussed regarding potential changes to the original Pagefind UI.

The main reason I quickly implemented it is to make testing easier as the original design was not very convenient when using the search UI a lot. It's in a dedicated commit so it can be easily reverted later if needed.

HiDeoo:
Updated the PR to adds support for new features which are currently not supported at all in Starlight. To decide which features to implement first, I looked at the various feature requests that have been opened and also what is currently being done in the ecosystem.

- [Adds support for programmatically controlling Pagefind](https://pagefind.app/docs/ui-usage/#programmatically-controlling-the-pagefind-ui)
  - This adds support for the `triggerFilters` and `triggerSearch` public API methods. 
  - This was motivated by the `starlight-versions` plugin having to re-implement entirely the `Search` component just to trigger this functionality.
  - With this, a component override can wrap the default `Search` component with a custom web component, `querySelector` the default web component, and call any of the public API methods on it.
  - Demos of this have been temporarilly added to the documentation.
- Adds support for the [`showEmptyFilters` option](https://pagefind.app/docs/ui/#show-empty-filters)
  - This was requested in https://github.com/withastro/starlight/discussions/2391#discussioncomment-10878644 and https://github.com/withastro/starlight/discussions/3117
- Adds support for the [`openFilters` option](https://pagefind.app/docs/ui/#open-filters)
  - This was requested in https://github.com/withastro/starlight/discussions/3117
- Adds support for the [`processTerm` option](https://pagefind.app/docs/ui/#process-term)
  - This was requested in https://github.com/withastro/starlight/discussions/1358#discussioncomment-9659960
  - As this is a non-serializable option, a new `clientOptionsModule` option was added, [similar to the DocSearch plugin](https://starlight.astro.build/guides/site-search/#additional-docsearch-options), that allows a user to define the path to a configuration file exporting the Pagefind configuration.

HiDeoo:
Updated the PR based on the feedback from T&D (14 August 2025):

- When present, the "Load more results" is now always visible in a footer of the modal. Like the always-visible query input, it's definitely not a definitive or even thought-out design, but more of an experiment.
- Refactor the API design to avoid the need to `querySelector` some web components. All pagefind related exports are now available from the `@astrojs/starlight/pagefind` module which exposes a `getPagefindApi` function that returns an object with methods to trigger filters and search.

HiDeoo:
Due to various UI changes that happened in the course of this PR, e.g. an always-visible search input or load more button (when applicable), the current UI entirely relying on the previous design feels a bit "heavy" in my opinion around these changes.

Sharing here a few screenshots of more compact versions of the same UI (not included in this PR) for reference and discussion. I am obviously not the best when it comes to design, so only sharing as these parts of the UI have been on my mind while working on this PR. As we now have entire control over the UI, there is also more room for customization depending on where we want to take this in the future.

| Current  |
| -------- |
| SCR-20251001-okga |

| Experiments |
| ----------- |
| SCR-20251002-jnsr SCR-20251001-ohle SCR-20251002-jznw |

ematipico:
Tested on Firefox Developer Tools (desktop):
- The height of the modal is smaller than before, which affects the visibility of the results shown. I preferred the previous height

  Screenshot 2025-10-14 at 11 21 05
  


One minor (nitpick) drawback on mobile. Now that input search (header) and more results (footer) are sticky, when clicking "More results", you don't have a visual change anymore that shows that new results were loaded.

HiDeoo:
Thanks for the great feedback :raised_hands:

Definitely agree, and this is also something I've started looking at in https://github.com/withastro/starlight/pull/3335#issuecomment-3360161584 (screenshots included).

To give a bit more context, this PR initially started by a rewrite to improve the accessiblity, but as we now have more control, some UI/UX changes have also started to make sense. So far, such changes have been added with minimal effort regarding the design aspect which explains such large padding around some elements. As I'm also not the best when it comes to design, this is also something where Chris inputs would be very valuable.

Regarding the last point, definitely something that is missing right now, great reminder. Such indicators have been implemented for screen-readers, but not visually yet. I'm guessing these would be added once we have a better idea of the overall design direction.

bglw:
👋 @HiDeoo 

I don't know how far you are along with this, but I thought I'd drop a ping. 

The [Pagefind v1.5.0 beta](https://github.com/Pagefind/pagefind/releases/tag/v1.5.0-beta.1) is out, which includes https://ui.pagefind.app/ — which is an order of magnitude more accessible, and an order of magnitude more customizable. Both of which might be of use... :)

HiDeoo:
> 👋 @HiDeoo
> 
> I don't know how far you are along with this, but I thought I'd drop a ping.
> 
> The [Pagefind v1.5.0 beta](https://github.com/Pagefind/pagefind/releases/tag/v1.5.0-beta.1) is out, which includes [ui.pagefind.app](https://ui.pagefind.app/) — which is an order of magnitude more accessible, and an order of magnitude more customizable. Both of which might be of use... :)

:wave: @bglw 

Wooo, thanks a lot for the ping, this sounds amazing :tada:

I would say we are definitely at a fairly advanced stage, we started doing some user testing, but we have not been able to prioritize it recently with some other work going on and Astro v6 being around the corner.

I am expecting to pick this back up in the next few weeks, so this is perfect timing! I will definitely check out the new UI, evaluate if we need to change our approach, etc. and let you know if we have any questions.

Thanks a lot for reaching out and your amazing work on Pagefind 🙌

PR #3340, opened 2025-07-25
Fix Astro Island margins in Markdown

### LINKED ISSUE #3290: No margin between Markdown content and UI framework components

0.34.4

5.11.0

npm

Mac

Chrome

### Describe the Bug

When using UI framework components, the `margin-top` between elements applied by Starlight's Markdown styling does not work on the `astro-island` element.

Image

https://stackblitz.com/edit/github-hczwebv8?file=src%2Fcontent%2Fdocs%2Findex.mdx

### Participation

HiDeoo:
#### Description

- Closes #3290

This PR attempts to fix an issue where Starlight's Markdown default margins are not applied when using interactive UI frameworks components (Astro Islands).

I don't think we can really reuse our existing rule for this and we may need to have a special case targetting the first child of an island. This PR temporarily adds a few test scenarios to the `/getting-started/` page and I'll also play around with more to see if there are other edge cases we might need to consider.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -3,6 +3,8 @@ import { defineConfig } from 'astro/config';
 import starlight from '@astrojs/starlight';
 import starlightLinksValidator from 'starlight-links-validator';
 import markdocGrammar from './grammars/markdoc.tmLanguage.json';
+// TODO(HiDeoo) Remove
+import preact from '@astrojs/preact';
 
 export const locales = {
 	root: { label: 'English', lang: 'en' },
@@ -34,6 +36,8 @@ export default defineConfig({
 	site,
 	trailingSlash: 'always',
 	integrations: [
+		// TODO(HiDeoo) Remove
+		preact(),
 		starlight({
 			title: 'Starlight',
 			logo: {
--- a/docs/package.json
+++ b/docs/package.json
@@ -17,11 +17,13 @@
   },
   "dependencies": {
     "@astro-community/astro-embed-youtube": "^0.5.6",
+    "@astrojs/preact": "^4.1.0",
     "@astrojs/starlight": "workspace:*",
     "@lunariajs/core": "^0.1.1",
     "@types/culori": "^2.1.1",
     "astro": "^5.6.1",
     "culori": "^4.0.1",
+    "preact": "^10.26.9",
     "sharp": "^0.34.2"
   },
   "devDependencies": {
--- a/docs/src/components/Debug.tsx
+++ b/docs/src/components/Debug.tsx
@@ -1,0 +1,28 @@
+import type { ComponentChildren } from 'preact';
+
+export function IslandBlock() {
+	return <p>This is an Astro Island block.</p>;
+}
+
+export function IslandInline() {
+	return <a href="#">Island inline link</a>;
+}
+
+export function IslandBlocks() {
+	return (
+		<>
+			<p>This is an Astro Island block.</p>
+			<p>And another block</p>
+		</>
+	);
+}
+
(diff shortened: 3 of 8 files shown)


delucis (commented):
Thanks for getting the ball rolling on this one @HiDeoo!

Do you have any more tests that you think we need for confidence? With more realistic scenarios?

In the abstract, this looks OK, but I find it a bit hard to predict what the real-world island usage will actually look like. Especially for slotted content.

delucis:
@KianNH shared this page and island: https://developers.cloudflare.com/ai-gateway/demos/

Source: https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/ai-gateway/demos.mdx?plain=1

HiDeoo:
> Do you have any more tests that you think we need for confidence? With more realistic scenarios?

Definitely 👍 That's what I meant in the PR body and I plan on doing that this week-end. I want to try with some random components I use in various projects and test various scenarios, and see if I can find any issues.

Thanks for linking the Cloudflare example, I'll have to test against it too as I mostly worked on relatively simple components so far. I'll undraft the PR once I have done that and feel more confident about the changes.

Regarding the refactor, totally agree, right now I just added new rules and did not even try to see if some smarter refactors could be done or not.

trueberryless:
Thanks @HiDeoo for picking up this issue and investigating on even better improvements 🙌

Also big thanks for @delucis and @KianNH for sharing knowledge too. Awesome! 🤩

HiDeoo:
Added a few more examples, but turns out after checking, I'm not a heavy island user and even less in the middle of content, most of the time it would be in layouts or just Astro components :sweat_smile:

I added one matching the Cloudflare example, a few others that came to mind, but in the end, I still feel like I may not have a full grasp of all or most common use cases of islands in the middle of content.

I'll try exploring something like [`https://github.com/search?q=language%3Amdx+%22client%3Aonly%22&type=code&p=1`](https://github.com/search?q=language%3Amdx+%22client%3Aonly%22&type=code&p=1)

HiDeoo:
As [mentioned](https://github.com/withastro/starlight/pull/3340#issuecomment-3131454150), I've been trying more examples based on searches like https://github.com/search?q=language%3Amdx+%22client%3Aonly%22&type=code&p=1 and a few others. So far, every examples behaved as expected using the new changes from this PR but I'm still not super confident about the change considering I'm not a heavy island user, even more in the middle of content.

I'm opening up this PR for review in case it can bring more awareness to the change and get more feedback from the community if some users are running into similar setups and can to test it out. I will only move forward with the suggested refactoring if we decide to ship this change.

HiDeoo:
> Do you think it should be a minor given the changes? Or do we think it’s safe enough for a patch? People will either get improved spacing OR if they already did something to remediate it nothing should change (their custom CSS wins or they have wrapper elements which are not impacted by this change).

I'm a bit afraid of the opposite case where someone was relying on the previous behavior for some reason and do not want the margin, so I'm leaning towards a minor and including in the changelog some code to get back the previous behavior.

delucis:
> I'm leaning towards a minor and including in the changelog some code to get back the previous behavior

Sure — gives us a bit more space to document the change like you say 👍

HiDeoo:
Updated and cleaned up the PR.

I introduced a new CSS custom property [as suggested](https://github.com/withastro/starlight/pull/3340#pullrequestreview-3055384098) (open on suggestions regarding the naming, e.g. what it represents vs. what it does). I also decided against merging the 2 rules into one, I think it hurts readability too much.

trueberryless:
Thanks HiDeoo for the awesome update 🙌

PR #3341, opened 2025-07-25
Prevent Cloudflare build issues due to Node.js builtins usage

### LINKED ISSUE #2875: [Astro5 + Starlight + Cloudflare] Cannot build with Starlight and Cloudflare adapter
### Astro Info

```block
Astro                    v5.0.5
Node                     v18.20.3
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/starlight
```

### Describe the Bug

If I try to build the Astro 5 website with Starlight and Cloudflare adapter, the build error will occurs.

Build error log:

```
$ npm run build

00:47:37 [build] Building static entrypoints...
00:47:38 [ERROR] [vite] x Build failed in 636ms
[commonjs--resolver] [plugin vite:resolve] Cannot bundle Node.js built-in "node:path" imported from "node_modules/@astrojs/starlight/utils/collection.ts". Consider disabling environments.ssr.noExternal or remove the built-in dependency.
  Stack trace:
    at getRollupError (file:///home/projects/astro5-starlight-cloudflare-build-error/node_modules/rollup/dist/es/shared/parseAst.js:558:41)
    at error (file:///home/projects/astro5-starlight-cloudflare-build-error/node_modules/rollup/dist/es/shared/node-entry.js:20652:33)
    at handler (file:///home/projects/astro5-starlight-cloudflare-build-error/node_modules/vite/dist/node/chunks/dep-yUJfKD1i.js:50867:15)
    at then/< (https://astro5starlightcloudflarebuild-el3s.w-corp-staticblitz.com/blitz.c8c182a3.js:40:23197)
```

### What's the expected result?

No build errors.

https://stackblitz.com/edit/astro5-starlight-cloudflare-build-error

### Participation

HiDeoo:
#### Description

- Closes #2875

The linked issue and its reproduction showcase build errors with the Cloudflare adapter due to some Node.js builtins ending up in the final bundle. Using updated dependencies seems to no longer cause this issue with some modules being automatically externalized by Vite but still logging some warnings, for example:

```
[WARN] [vite] [plugin vite:resolve] Automatically externalized node built-in module "node:url" imported from "node_modules/@astrojs/starlight/utils/collection.ts". Consider adding it to environments.ssr.external if it is intended.
```

While this may work now, this PR aims to fix such issues and prevent regressions in the future by adding a Vite plugin ([copied from Astro](https://github.com/withastro/astro/blob/8491aa56e8685677af8458ff1c5a80d6461413f8/packages/astro/test/test-plugins.js) and slightly tweaked) to one of our e2e fixtures that checks that the final bundle does not have a dependencies on Node.js builtins due to Starlight and throws an error if it does.

Regarding the issue, 3 main reasons for the issue were identified:

- In #2261, we added support for custom icons in asides and we throw an error if the icon does not exist. The code throwing the error was located in the aside remark plugin and imported in the `<Aside>` component. This caused some Node.js builtins used in the remark plugin to end up in the final bundle. To fix this, the shared error was moved to a dedicated file.
- The `collection.ts` file which contains various helpers to get paths to collections used in Starlight and used a lot throughout the codebase, contained 1 function relying on Node.js builtins which ended up leaking due to many places importing this file without using the function. To fix this, the function was moved to a dedicated file.
- The `absolutePathToLang` helper shared between many integrations was recently updated in #3298 to use a Node.js builtin (`pathToFileURL`) to convert a path to a URL and avoid issues with paths with some encoded characters, e.g. a space. This helper used by Expressive Code can end up in the final bundle due to it's usage in the preprocessor config which is also exposed in a virtual module through `packages/starlight/internal.ts` for the `<Code>` component. To fix this, the helper has been updated to no longer be in charge of figuring out the `docs` collection path, and instead the POSIX path to this collection is passed down to this helper.

An additional test for these changes is also to install and build the documentation using the Cloudflare adapter and making sure that no log warning is emitted about Node.js builtins being automatically externalized.

#### Remaining tasks

- [ ] Add changeset (will do that later, have to run for now 🏃‍➡️)

### DIFF
--- a/docs/src/content/docs/de/resources/plugins.mdx
+++ b/docs/src/content/docs/de/resources/plugins.mdx
@@ -163,6 +163,11 @@ Erweitere deine Website mit offiziellen Plugins, die vom Starlight-Team unterst�
 		title="starlight-codeblock-fullscreen"
 		description="Füge eine Funktion zum Umschalten zwischen Vollbildmodus und Normalansicht zu den Expressive Code-Blöcken in deiner Dokumentation hinzu."
 	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-auto-drafts"
+		title="starlight-auto-drafts"
+		description="Passe das Standardverhalten von Entwurfsseiten an und entferne automatisch Links zur Seitenleiste zu Entwurfsseiten im Produktionsmodus."
+	/>
 </CardGrid>
 
 ## Community-Tools und Integrationen
--- a/docs/src/content/docs/de/resources/themes.mdx
+++ b/docs/src/content/docs/de/resources/themes.mdx
@@ -74,9 +74,9 @@ Installiere ein von der Community erstelltes Theme, um das Aussehen deiner Websi
 			previews: { light: 'nova-light.png', dark: 'nova-dark.png' },
 		},
 		{
-			title: 'Starlight NextJS Theme',
-			description: 'Starlight-Theme inspiriert von der NextJS-Dokumentation.',
-			href: 'https://starlight-nextjs-theme.trueberryless.org/',
+			title: 'Starlight Next.js Theme',
+			description: 'Starlight-Theme inspiriert von der Next.js-Dokumentation.',
+			href: 'https://starlight-theme-next.trueberryless.org/',
 			previews: { light: 'nextjs-light.png', dark: 'nextjs-dark.png' },
 		},
 	]}
--- a/docs/src/content/docs/fr/resources/plugins.mdx
+++ b/docs/src/content/docs/fr/resources/plugins.mdx
@@ -163,6 +163,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-codeblock-fullscreen"
 		description="Ajouter la fonctionnalité de basculement en plein écran aux blocs Expressive Code de votre documentation."
 	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-auto-drafts"
+		title="starlight-auto-drafts"
+		description="Ajuster le comportement par défaut des ébauches et supprimer automatiquement de la barre latérale les liens vers les ébauches en mode production."
+	/>
 </CardGrid>
 
 ## Outils et intégrations communautaires
--- a/docs/src/content/docs/fr/resources/themes.mdx
+++ b/docs/src/content/docs/fr/resources/themes.mdx
@@ -74,9 +74,9 @@ Installez un thème créé par la communauté pour personnaliser rapidement l'ap
 			previews: { light: 'nova-light.png', dark: 'nova-dark.png' },
 		},
 		{
-			title: 'Starlight NextJS Theme',
-			description: 'Thème Starlight inspiré par la documentation de NextJS.',
-			href: 'https://starlight-nextjs-theme.trueberryless.org/',
+			title: 'Starlight Next.js',
+			description: 'Thème Starlight inspiré par la documentation de Next.js.',
+			href: 'https://starlight-theme-next.trueberryless.org/',
 			previews: { light: 'nextjs-light.png', dark: 'nextjs-dark.png' },
 		},
 	]}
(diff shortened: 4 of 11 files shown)


packages/starlight/__e2e__/fixtures/basics/src/noNodeModule.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @see https://github.com/withastro/astro/blob/8491aa56e8685677af8458ff1c5a80d6461413f8/packages/astro/test/test-plugins.js
+ */
+export function preventNodeBuiltinDependencyPlugin(): NonNullable<
+	ViteUserConfig['plugins']
+>[number] {
+	return {
+		name: 'verify-no-node-stuff',
+		generateBundle() {
+			const nodeModules = ['node:fs', 'node:url', 'node:worker_threads', 'node:path'];
```
delucis: Do you know why it’s specifically these four modules we test for?

packages/starlight/utils/collection-fs.ts
```diff
@@ -0,0 +1,21 @@
+import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { getCollectionUrl, type StarlightCollection } from './collection';
+
+/**
+ * @see {@link file://./collection.ts} for more context about this file.
+ *
+ * Below are various functions to easily get paths to collections used in Starlight that rely on
+ * Node.js builtins. They exist in a separate file from {@link file://./collection.ts} to avoid
+ * to potentially importing Node.js builtins in the final bundle.
```
delucis: Grammar nit 😄 

```suggestion
 * potentially importing Node.js builtins in the final bundle.
```

packages/starlight/utils/collection.ts
```diff
@@ -22,12 +19,12 @@ export type StarlightCollection = (typeof collectionNames)[number];
...
-	return new URL(`content/${collection}/`, srcDir).pathname;
+export function getCollectionUrl(collection: StarlightCollection, srcDir: URL) {
+	return new URL(`content/${collection}/`, srcDir);
 }
 
-export function resolveCollectionPath(collection: StarlightCollection, srcDir: URL) {
-	return resolve(fileURLToPath(srcDir), `content/${collection}`);
+export function getCollectionPath(collection: StarlightCollection, srcDir: URL) {
+	return getCollectionUrl(collection, srcDir).pathname;
 }
```
delucis: Wonder if anyone might mistakenly think this returns a file system “path” they can use?

Oh… and just checked, looks like this is completely unused? Was going to suggest renaming/documenting, but I guess we can just delete it! 😤 

```suggestion
```

packages/starlight/__e2e__/fixtures/basics/src/noNodeModule.ts
HiDeoo: No idea, this was added in [this PR](https://github.com/withastro/astro/pull/6537) and I assume at the time these were the only modules that were noticed.

I refactored the test to use the list of all modules provided by Node.js and also moved the test to a dedicated suite. Having 20+ unrelated tests (basics suite) failing because of this was not ideal and also way slower than just having a single dedicated test failing for this while making it easier to identify the issue.

image

packages/starlight/utils/collection.ts
HiDeoo: Great catch, totally missed that during the refactor 🙈

To explain why this helper is no longer needed after this PR:

- After splitting collection helpers into 2 files (using `fs` and not using `fs`), the `getCollectionPath` helper was only used in the `absolutePathToLang` helper.
- As this was an issue to receive an URL-encoded pathname in the `absolutePathToLang` helper (which was recently introduced in #3298), it was refactored to now accept the `docs` content collection path as an argument and compare POSIX paths.
- This new argument is now passed from the new and explicit `getCollectionPosixPath` helper making `getCollectionPath` obsolete.

delucis (approved):
Perfect — thanks for cleaning that up @HiDeoo! Great call making it a dedicated test suite too.

PR #3337, opened 2025-07-24
docs: add `starlight-auto-drafts` to plugins resources

HiDeoo:
#### Description

This PR adds the [`starlight-auto-drafts`](https://github.com/HiDeoo/starlight-auto-drafts) plugin to the plugins resources section of the documentation.

PR #3342, opened 2025-07-25
i18n(de): udpate `plugins.mdx`

trueberryless:
#### Description

- Update `plugins.mdx`
- #3337


delucis (approved):
We can’t auto-translate or auto-approve, but at last auto-drafts are now available in German — thanks @trueberryless 🚀

PR #3348, opened 2025-07-28

ArmandPhilippot:
#### Description

Adds the new plugin added in #3337 to the French translation of `resources/plugins.mdx`.


HiDeoo (approved):
Definitely a bit more words in French compared to the English one but that's perfectly translated to me 🚀

PR #3351, opened 2025-07-29
Ensure CSS layer order in custom pages

HiDeoo:
#### Description

This PR aims to fix the issue 2 in #3162 and is a follow-up to #3150.

This PR experiments (this is why this PR is a draft) with a Vite plugin that ensures in custom pages that the `<StarlightPage>` component is always imported first, which is necessary to ensure the correct CSS layer order due to how [import order](https://docs.astro.build/en/guides/styling/#import-order) for stylesheets works in Astro.

A few test pages have been added to the documentation which should all result in the same visual appearance (ignoring the last `<aside>` element). Without the plugin, some differences can be observed due to different CSS layer orders between all the pages, in the same mode (dev/build) or between dev and build modes.

With the plugin, all pages should look the same no matter the mode. Disabling the new plugin and running `pnpm test` in the `docs/` directory should also result in errors with some a11y issues due to the different layer orders.

_Note: unit tests are failing due to coverage thresholds atm._

#### Remaining tasks

If we end up moving forward with this PR, the following tasks will need to be completed:

- [x] Remove all demo content
- [x] Add changeset


HiDeoo:
Took the time to test the approach on more projects and it seems to work well.

Before continuing with this PR, I'm undrafting it to get some feedback on the approach. While this kind of approach could usually be considered a code smell, or a sign of a design flaw, in this specific case and due to how import order for stylesheets works in Astro and how it can impact layer order, I'm failing to find other ways to achieve the desired outcome consistently.

I'm hoping to get some feedback on the approach and whether it seems reasonable in this specific context or if we should continue to rely on documentation to guide users to the correct order of imports when using `<StarlightPage>` (which is easy to miss, not common, and can easily break with import auto reordering tools).

HiDeoo:
Yeah, this is definitely a tricky case. I also spent a lot of time thinking and trying out various approaches as like I mentioned, I'm also not super happy with this one but this is one of those cases where I think that the trade-offs may be needed as there's no perfect solution afaik (or even a good one providing a good DX).

> One thought I had, which you can consider, is that rather than _moving_ the user’s import, we could in theory prepend a side-effect import, i.e. make the plugin do this:

Interesting idea, I haven't thought about that. I cannot yet answer your last question as this is not something I've tried, but I'll explore the idea once I'm done catching up with Astro/Starlight and plugins. I'll report back here once I have more info.

HiDeoo:
> One thought I had, which you can consider, is that rather than _moving_ the user’s import, we could in theory prepend a side-effect import, i.e. make the plugin do this:

Great idea, works perfectly and slightly simplifies the code too.

> Not sure if that could potentially be simple enough to compute the source map for that we could avoid the need for `magic-string`.

This is tricky. As we transform source code, we [have to generate a sourcemap](https://rollupjs.org/plugin-development/#source-code-transformations). The plugin [only have to return the sourcemap of the tranformations done in that `transform` hook](https://github.com/vitejs/vite/issues/20576#issuecomment-3168681241). The change to the sourcemap is trivial as in our case it can be as simple as prepending a `;` to the sourcemap mappings (a `;` delimits lines in the sourcemap mappings). The tricky part is that if I'm not mistaken, we don't have the original sourcemap of the source code before our transformation that we could easily modify. There is a [`getCombinedSourcemap()` helper](https://rollupjs.org/plugin-development/#this-getcombinedsourcemap) altho this returns the combined sourcemaps of all previous plugins which is not what we need. So `magic-string` gives us the 2 steps for free (and there is no way we can/should compute the entire sourcemap manually :sweat_smile:).

packages/starlight/integrations/vite-layer-order.ts
```diff
@@ -0,0 +1,66 @@
...
+				const importDefaultSpecifier = node.specifiers.find(
+					(specifier) => specifier.type === 'ImportDefaultSpecifier'
+				);
+				if (!importDefaultSpecifier) continue;
+
+				isStarlightPage = true;
+				break;
+			}
+
+			if (!isStarlightPage) return;
```
delucis: Nit: `hasStarlightPageImport` might be a more accurate name here?

```suggestion
			let hasStarlightPageImport = false;

			for (const node of ast.body) {
				if (node.type !== 'ImportDeclaration') continue;
				if (node.source.value !== starlightPageImportSource) continue;

				const importDefaultSpecifier = node.specifiers.find(
					(specifier) => specifier.type === 'ImportDefaultSpecifier'
				);
				if (!importDefaultSpecifier) continue;

				hasStarlightPageImport = true;
				break;
			}

			if (!hasStarlightPageImport) return;
```

packages/starlight/integrations/vite-layer-order.ts
```diff
@@ -0,0 +1,66 @@
+import type { ViteUserConfig } from 'astro';
+import MagicString from 'magic-string';
+
+const starlightPageImportSource = '@astrojs/starlight/components/StarlightPage.astro';
```
delucis, line 4: I’m thinking to myself this will break if anyone ever aliases this, but never mind 😁

packages/starlight/index.ts
```diff
@@ -118,6 +119,7 @@ export default function StarlightIntegration(
 				updateConfig({
 					vite: {
 						plugins: [
+							vitePluginStarlightCssLayerOrder(),
 							vitePluginStarlightUserConfig(command, starlightConfig, config, pluginTranslations),
```
delucis, line 123: In our test setup utility we include `vitePluginStarlightUserConfig`:

https://github.com/withastro/starlight/blob/2db7b5e00052eeee4888d6925d2da28bb2626812/packages/starlight/__tests__/test-config.ts#L31-L32

Should we do the same for the new plugin so it’s running in unit tests?

HiDeoo: Yeah, totally slipped my mind as we only test it using e2e tests but makes sense to have the same setup in unit tests as the integration.

HiDeoo:
Thanks for the review, the PR should now be updated without any of the demo content, and I also added a changeset and reverted #3199.

.changeset/famous-grapes-develop.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/starlight': minor
+---
+
+Ensures that Starlight CSS layer order is predictable in custom pages using the `<StarlightPage>` component.
+
+Previously, due to how [import order](https://docs.astro.build/en/guides/styling/#import-order) works in Astro, the `<StarlightPage />` component had to be the first import in custom pages to set up [cascade layers](https://starlight.astro.build/guides/css-and-tailwind/#cascade-layers) used internally by Starlight to manage the order of its styles.
+
+With this change, this restriction no longer applies and Starlight’s styles will be applied correctly regardless of the import order of the `<StarlightPage>` component.
```
trueberryless, line 9: Very nitpicky, but I think `<StarlightPage />` should be equally written as the first time:

```suggestion
With this change, this restriction no longer applies and Starlight’s styles will be applied correctly regardless of the import order of the `<StarlightPage />` component.
```

Otherwise, very clear changeset, easy to understand 👍

trueberryless: I just noticed, that the first paragraph also includes a `<StarlightPage>`, so we could also adapt the second mention...

Or is there a specific reason for the different closing vs non-closing?

HiDeoo: Thanks, typo I missed. We tend to use `<Component>`, but we're not consistent over it, but we should at least be consistent in the same changeset.

packages/starlight/__e2e__/fixtures/basics/src/pages/starlight-page-css-layer-order.astro
```diff
@@ -0,0 +1,21 @@
+---
+import { LinkButton } from '@astrojs/starlight/components';
+import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
+import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
```
delucis, line 6: Bit silly, but should we separate these like this?

```suggestion
import { LinkButton } from '@astrojs/starlight/components';

import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';

import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
```

It would mean if we ever ran an automatic “Organise imports” task like we do in some repos, these would not get auto-sorted.

packages/starlight/__e2e__/basics.test.ts
```diff
@@ -467,6 +467,28 @@ test.describe('components', () => {
...
+			const [bgColor, textColor] = await linkButton.evaluate((element) => {
+				const styles = window.getComputedStyle(element);
+				return [styles.getPropertyValue('background-color'), styles.getPropertyValue('color')];
+			});
+
+			// If the background color and text color are the same, the text is not readable and the CSS
+			// layer order is incorrect.
+			expect(bgColor).not.toBe(textColor);
+		});
+	});
```
delucis, line 494: This test feels a little fragile. Do you think we should maybe test something a bit more explicit? Like a custom user style that could end up overridden or something? Or do you think the behavior here is predictable enough that the test is unlikely to keep passing “by accident”.

packages/starlight/__e2e__/fixtures/basics/src/pages/starlight-page-css-layer-order.astro
HiDeoo: Great idea, didn't think of that 🧠

packages/starlight/__e2e__/basics.test.ts
HiDeoo: I've rewritten that test to make it more about the layer order itself rather than side-effects of good or bad layer order. Let me know what you think of this version.

delucis (approved):
Nice change! Yeah, I think this feels a bit more reliable. Might still be fragile in some circumstances (major Astro/Vite updates e.g.), but feels like it would be fragile in the sense that it might _fail_ incorrectly, which is not too bad, better than _passing_ incorrectly in this case, I think.

Great work! Should be good to go for the next minor IMO 🤩

PR #3355, opened 2025-08-01
Restrict Starlight line-height to Starlight content

alvinometric:
#### Description

Starlight's `sl-line-height` value was applied even for `.not-content`. I used the same `:not` rule that you have elsewhere, but I wonder if this might not be better:

```css
body :not(.not-content *) {
  line-height: var(--sl-line-height);
}
```

Another possibility might be to set `line-height: normal` to `.not-content`, whichever you prefer 😄

### DIFF
--- a/packages/starlight/style/reset.css
+++ b/packages/starlight/style/reset.css
@@ -20,10 +20,13 @@
 
 	body {
 		font-family: var(--__sl-font);
-		line-height: var(--sl-line-height);
 		-webkit-font-smoothing: antialiased;
 		color: var(--sl-color-text);
 		background-color: var(--sl-color-bg);
+	}
+
+	body:not(:where(.not-content *)) {
+		line-height: var(--sl-line-height);
 	}
 
 	input,


packages/starlight/style/reset.css
```diff
@@ -20,12 +20,15 @@
...
 		font-family: var(--__sl-font);
-		line-height: var(--sl-line-height);
 		-webkit-font-smoothing: antialiased;
 		color: var(--sl-color-text);
 		background-color: var(--sl-color-bg);
 	}
 
+	body:not(:where(.not-content *)) {
+		line-height: var(--sl-line-height);
+	}
```
delucis, line 30: I’m not sure this CSS does what you think it does. Did you test it?

This would require the following markup for the `:not()` selector to kick in and remove the `line-height` from the entire page:

```html
<html class="not-content">
  <head>...</head>
  <body>...</body>
</html>
```

Otherwise it always matches `<body>`.

delucis (commented):
Hey @alvinometric 👋 

I’m not 100% sure it makes sense to do this. `line-height` is an [inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Inheritance), so any parent of a component which sets `line-height` also impacts the component itself.

It’s true that we could reset `line-height` inside `not-content` but I feel like any component that requires strict control of its `line-height` needs to set it explicitly rather than relying on the initial value being set for it, as it’s rarely the case that a page is not setting `line-height` _somewhere_ (many CSS resets do this, including e.g. Tailwind). In general, `not-content`’s aim is to remove styling that interferes with embedded components in ways that are not likely to happen if the component were embedded in another context. Especially to avoid needing to worry about hard to unset ones like vertical margins etc.

In the case of `line-height`, setting
```css
.my-component {
  line-height: normal;
}
```
will always work (regardless of `not-content`) for the whole component and it also guarantees that the component behaves like you want if it’s embedded somewhere other than Starlight.

delucis:
Going to close this as we haven’t heard back, and I think the proposed change would need more discussion. Thanks!

PR #3357, opened 2025-08-05
docs: add link to horizontal progress bar blog

trueberryless:
#### Description

Add new community content to a blog from @frostybee and me where we describe how to add a horizontal progress indicator to a Starlight site.

### DIFF
--- a/docs/src/content/docs/resources/community-content.mdx
+++ b/docs/src/content/docs/resources/community-content.mdx
@@ -117,7 +117,7 @@ Explore community-produced content maintained by Starlight users:
 	<LinkCard
 		href="https://blog.trueberryless.org/blog/starlight-progress-indicator/"
 		title="Add a reading progress indicator"
-		description="Learn how to create a horizontal reading progress bar for your Starlight site."
+		description="Find out how to add a horizontal reading progress bar to your Starlight site."
 	/>
 </CardGrid>
 


docs/src/content/docs/resources/community-content.mdx
```diff
@@ -114,6 +114,11 @@ Explore community-produced content maintained by Starlight users:
 		title="Starlight Plugins by Example"
 		description="A collection of notes and examples about Starlight plugins and common patterns used to build them."
 	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-progress-indicator/"
+		title="Add a horizontal progress bar"
+		description="Learn how to create a horizontal progress indicator for your Starlight site."
```
HiDeoo: I wonder if the current title makes it feels a bit unrelated to Starlight and unclear what the real purpose of the article is :thinking: Is it about using `<progress>` in Starlight?

Considering "horizontal" is a bit of a detail and already mentioned in the description, maybe dropping it in the title and replacing it with something like "reading" could better indicate what this is about, e.g. "Add a reading progress indicator". Meanwhile the description goes into more detail about the horizontal aspect, e.g "Learn how to create an horizontal reading progress bar for your Starlight site." or something like that.

Any thoughts on such a change?

trueberryless: This sounds like a good idea to me, yeah 👍

trueberryless:
Awesome! 🙌 

I also adjusted the grammar and styling a little bit so it's not always the same words in the description.

PR #3364, opened 2025-08-07
Update carbon calculator stats for v4 model

delucis:
Website Carbon Calculator [introduced a new model for estimating energy usage](https://www.wholegraindigital.com/blog/updating-website-carbon-to-v4-of-the-sustainable-web-design-model/), so this PR updates the table to use new measurements. Mostly no change in the ratings but the new model estimates lower CO₂ usage than the earlier version.

I had to update the Docus page being tested as their installation guide repeatedly errored out during testing. (They’re also notably the only entry whose CO₂ usage didn’t decrease this time round, presumably because of changes in their recent release.)


HiDeoo (approved):
Just read the link and the [model changes](https://www.wholegraindigital.com/blog/new-sustainable-web-design-model-changes-the-context-of-internet-emissions/), pretty interesting.

Thanks for updating, looks perfect to me 👍

PR #3366, opened 2025-08-07
i18n(de): update `community-content.mdx`

trueberryless:
#### Description

- Update translation
- #3357 

> [!WARNING]
> I link to the German version of my blog which is AI translated. I'm not sure, if French and German (the both languages the AI translates to) should link to their "mother" languages respectively.

Also quick note, this is just German at it's best: `Lesefortschrittsanzeige` 👨‍🍳

### DIFF
--- a/docs/src/content/docs/de/resources/community-content.mdx
+++ b/docs/src/content/docs/de/resources/community-content.mdx
@@ -114,6 +114,11 @@ Erkunde die von der Community erstellten Inhalte, die von Starlight-Benutzern ge
 		title="Starlight-Plugins anhand von Beispielen"
 		description="Eine Sammlung von Notizen und Beispielen zu Starlight-Plugins und gängigen Mustern, die zum Erstellen dieser Plugins verwendet werden."
 	/>
+  <LinkCard
+		href="https://blog.trueberryless.org/de/blog/starlight-progress-indicator/"
+		title="Lesefortschrittsanzeige hinzufügen"
+		description="Erfahre, wie du eine horizontale Lesefortschrittsanzeige zu deiner Starlight-Website hinzufügen kannst."
+	/>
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/de/resources/community-content.mdx
```diff
@@ -114,6 +114,11 @@ Erkunde die von der Community erstellten Inhalte, die von Starlight-Benutzern ge
 		title="Starlight-Plugins anhand von Beispielen"
 		description="Eine Sammlung von Notizen und Beispielen zu Starlight-Plugins und gängigen Mustern, die zum Erstellen dieser Plugins verwendet werden."
 	/>
+  <LinkCard
+		href="https://blog.trueberryless.org/de/blog/starlight-progress-indicator/"
```
HiDeoo:
```suggestion
		href="https://blog.trueberryless.org/blog/starlight-progress-indicator/"
```

> Warning
> 
> I link to the German version of my blog which is AI translated. I'm not sure, if French and German (the both languages the AI translates to) should link to their "mother" languages respectively.

Considering the user would be coming from a non-machine translated website, I think we should let the user manually decide. This also kinda set a precedent for future links, e.g. why are we not prefixing all our links with some Google Translate or w/e service URL so that users automatically get a machine translated version? (I also checked the French version, the quality is not that great, especially context-wise)

I think linking to original versions when no human translation is available is a fine default for now.

HiDeoo (changes requested):

docs/src/content/docs/de/resources/community-content.mdx
trueberryless: Perfect. I agree that linking to AI translations is not the best solution here. Thanks for sharing your thoughts 💬

PR #3359, opened 2025-08-06
Add `starlight-changelogs` to plugins resources

HiDeoo:
#### Description

This PR adds the [`starlight-changelogs`](https://github.com/HiDeoo/starlight-changelogs) to the `resources/plugins` page.

PR #3358, opened 2025-08-06
docs: add `starlight-contextual-menu` to plugins resources

reynaldichernando:
#### Description

This PR adds the [`starlight-contextual-menu`](https://github.com/corsfix/starlight-contextual-menu) plugin to the plugins resources section of the documentation.

PR #3367, opened 2025-08-07
i18n(de): update `plugins.mdx`

trueberryless:
- #3358 
- #3359
- #3369

### DIFF
--- a/docs/src/content/docs/de/resources/plugins.mdx
+++ b/docs/src/content/docs/de/resources/plugins.mdx
@@ -168,6 +168,16 @@ Erweitere deine Website mit offiziellen Plugins, die vom Starlight-Team unterst�
 		title="starlight-auto-drafts"
 		description="Passe das Standardverhalten von Entwurfsseiten an und entferne automatisch Links zur Seitenleiste zu Entwurfsseiten im Produktionsmodus."
 	/>
+	<LinkCard
+		href="https://github.com/corsfix/starlight-contextual-menu"
+		title="starlight-contextual-menu"
+		description="Füge ein Kontextmenü zu deiner Starlight-Dokumentation hinzu."
+	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-changelogs"
+		title="starlight-changelogs"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/de/resources/plugins.mdx
```diff
@@ -168,6 +168,16 @@ Erweitere deine Website mit offiziellen Plugins, die vom Starlight-Team unterst
...
 	/>
+	<LinkCard
+		href="https://github.com/corsfix/starlight-contextual-menu"
+		title="starlight-contextual-menu"
+		description="Füge ein Kontextmenü zu deiner Starlight-Dokumentation hinzu."
+	/>
+	<LinkCard
+		href="https://github.com/HiDeoo/starlight-changelogs"
+		title="starlight-changelogs"
+		description="Zeig Änderungsprotokolle neben deiner Projektdokumentation an."
```
trueberryless: Both possible

```suggestion
		description="Zeige Änderungsprotokolle neben deiner Projektdokumentation an."
```

trueberryless:
I would personally suggest waiting for another native speaker to decide which translation is better here... @delucis

docs/src/content/docs/de/resources/plugins.mdx
trueberryless: Choosing "Zeige" as the better translation because it feels more friendly and not commanding 🫶

trueberryless:
Open for review from everyone (also non-native) because the translation duality was resolved 🙌

delucis (approved):
LGTM!

Double checked “Warnungen” because it seems like a slightly odd name for e.g. a “Note” block 😅 but I guess this is true for “alert” in the English too, and I can see that [GitHub’s own German docs use “Warnungen” as well](https://docs.github.com/de/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

trueberryless:
Yeah, I took the translation from GitHub's German docs so people are familiar with it 👍

PR #3376, opened 2025-08-12
Disable Renovate `uses-with` for Node.js

HiDeoo:
#### Description

This PR is a follow-up to the `uses-with` updates in #3374 and #3375 and aims to disable Node.js version updates for `actions/setup-node` in Renovate.

As I mentioned in [this comment](https://github.com/withastro/starlight/pull/3374#issuecomment-3174710476), this is done for a few reasons:

- We want to control the Node.js version used in some workflows
- We probably don't want to deal with such PRs for every Node.js versions

As usual, I don't think there is a way to test this change without merging it, and not explicitely documented (or I missed it), but hopefully after merging this PR and waiting a bit, Renovate should close the 2 existing PRs I mentioned and should create 2 new ones only containing an update for `actions/checkout` (which was bundled in the previous PRs with the Node.js version update).

If this works as expected, I'll port this change to Astro Docs.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -27,5 +27,11 @@
 			matchDepTypes: ['packageManager'],
 			enabled: false,
 		},
+		{
+			description: 'Disable Node.js version updates with actions/setup-node',
+			matchDepNames: ['node'],
+			matchDepTypes: ['uses-with'],
+			enabled: false,
+		},
 	],
 }


delucis (commented):
I guess the thinking behind the change is to make the implicit Node versioning behaviour more explicit? i.e. currently if we specify `18`, the action will get the latest `18` release, changing across runs. After the Renovate PR, the Node version is pinned to a specific patch and will only change when we merge a Renovate PR.

In principle, I think that’s good, although I agree that a) sometimes we are intentionally using old versions and b) updating for every Node patch is a bit annoying (I can already confirm this from experience with the astro.build repo).

So might be good to accept the PR pinning Node to a specific patch and then disable subsequent updates?

HiDeoo:
> So might be good to accept the PR pinning Node to a specific patch and then disable subsequent updates?

That's also a possibility. That would mean (if everything goes well ^^)

1. Merging #3374
2. Merging this PR
3. Renovate closes #3375
4. Renovate re-opens a PR for updating `actions/checkout` only to the next major

delucis:
Sounds like a good experiment 😁

.github/renovate.json5
```diff
@@ -27,5 +27,11 @@
 			matchDepTypes: ['packageManager'],
 			enabled: false,
 		},
+		{
+			description: 'Disable Node.js version updates with actions/setup-node',
+			matchDepNames: ['node'],
+			matchDepTypes: ['uses-with'],
+			enabled: false,
+		},
```
delucis, line 35: Noting that our existing `github-actions` group uses `matchManagers: ['github-actions']` to target these specifically, but not sure that’s relevant here.

delucis (approved):
Left an observation, but happy to merge this and see what happens 👍 

(I merged #3374 already, so you can merge this whenever you like.)

.github/renovate.json5
HiDeoo: The reason I did not add it is to potentially catch similar rules that may in the future have a similar behavior for Node.js versions, e,.g. a similar rule was added for `pnpm/action-setup` but in this case the [data source is `npm`](https://github.com/renovatebot/renovate/pull/34791/files#diff-384c020b3634bb8e1fd92cb0f7f3a63b7223b6b37b72536dc41cf1e51553723dR62).

PR #3380, opened 2025-08-13
Add `head` config validation for `meta` tags with `content`

HiDeoo:
#### Description

- Closes #3044

Just had to debug an issue where a user had a `head` configuration including a `meta` tag with some `content` which is invalid HTML (it's a void element) and found out we already have a [feature request](https://github.com/withastro/starlight/discussions/3044) to better handle this case so I decided to implement it.

This PR adds a validation steps for `meta` tags in the `head` config to ensure that `content` is not provided and displays a friendly error message if it is.

The error message is different if some `attrs` are provided or not, to help users fix their config more easily:

| with `attrs` | without `attrs` |
| ------------ | --------------- |
| SCR-20250813-ohai | SCR-20250813-ogum |


.changeset/tough-weeks-speak.md
```diff
@@ -0,0 +1,20 @@
+---
+'@astrojs/starlight': minor
+---
+
+Makes head entry parsing stricter in Starlight config and content frontmatter.
+
+**⚠️ Potentially breaking change:** Previously Starlight would accept a head entry for a `meta` tag defining some `content` which generates invalid HTML as `<meta>` is a void elememt which cannot have any child nodes. Now, it is an error to define a `meta` tag including some `content`.
```
delucis: Tiny typo:
```suggestion
**⚠️ Potentially breaking change:** Previously Starlight would accept a head entry for a `meta` tag defining some `content` which generates invalid HTML as `<meta>` is a void element which cannot have any child nodes. Now, it is an error to define a `meta` tag including some `content`.
```

packages/starlight/schemas/head.ts
```diff
@@ -3,14 +3,37 @@ import { z } from 'astro/zod';
...
+				.superRefine((config, ctx) => {
+					if (config.tag !== 'meta' || config.content === undefined) return;
+					const { content, ...rest } = config;
+					ctx.addIssue({
+						code: 'custom',
+						message:
+							`The \`head\` configuration includes a \`meta\` tag with \`content\` which is invalid HTML.\n` +
+							`You should instead use a \`content\` attribute ` +
+							(Object.keys(rest.attrs ?? {}).length === 0
+								? 'with an additional attribute to identify the kind of metadata it represents '
```
delucis: Wonder if we should hint at common attributes here?

```suggestion
								? 'with an additional attribute such as `name`, `property`, or `http-equiv` to identify the kind of metadata it represents '
```

I also wondered about linking MDN’s `<meta>` docs but not sure it’s useful.

delucis (commented):
I’d forgotten about my own feature request 😄 But that’s why notes are a good idea.

Thanks for picking it up — looks solid!

(I also double checked if there‘s a better way to display the suggested config in the error messages, but Astro’s error renderer doesn’t have code block support, so we can’t easily style it any better than this 👍)

packages/starlight/schemas/head.ts
HiDeoo: Sounds like a great idea, I wanted to avoid only mentioning `name` in the sentence but multiple seems helpful :thumbsup:

Regarding the link, I initially included it, but it added a lot of content to the message why being not very helpful in contrast to the "code block" imo so I decided to not include it, no strong opinion on that though.

HiDeoo:

> (I also double checked if there‘s a better way to display the suggested config in the error messages, but Astro’s error renderer doesn’t have code block support, so we can’t easily style it any better than this 👍)

Yeah, added a todo about that to play with at some point and see if I can come up with an RFC to improve that, feels like it could be helpful in some cases (and not the first time I wanted something similar).

I also tried playing with different output format, e.g. config vs frontmatter, and while it's easy to distinguish between the two, it gets tricky with frontmatter format between JSON, YAML, and TOML, so I think the JSON format should be a good enough indicator of what the fix should look like.

delucis:
> I also tried playing with different output format

Oh, good point! I didn’t even put two and two together that this same JSON would display also when you make the error in your frontmatter.

I guess we could make the schema dynamic based on where it’s used maybe?

So, `head({ format: 'js' })` and `head({ format: 'md' })` or something like that return a schema that throws the relevant error? Then we could at least show YAML for frontmatter errors which I guess is the most common syntax (and what we show in our docs).

HiDeoo:
> So, `head({ format: 'js' })` and `head({ format: 'md' })` or something like that return a schema that throws the relevant error? Then we could at least show YAML for frontmatter errors which I guess is the most common syntax (and what we show in our docs).

We definitely could, that's even how I initially coded it, then thought about the different Markdown possible format and remove it but I can definitely quickly add it back, and decide how we feel about it 👍

delucis:
> thought about the different Markdown possible format

I _think_ Astro only supports YAML and TOML? And TOML is pretty recent and I’d guess fairly rare? So YAML seems a safe bet to me.

HiDeoo:
> And TOML is pretty recent and I’d guess fairly rare?

And I assume that one is even more rare 😅 

image

HiDeoo:
Updated the PR with the discussed changes, here is the new output:

| with `attrs` | without `attrs` |
| ------------ | --------------- |
| SCR-20250814-quhk         | SCR-20250814-qumy            |
| SCR-20250814-qvbl         | SCR-20250814-qvhx            |

While the JSON output seems fine, I wonder if for the YAML one, we should not include more context, e.g.:

```yaml
head:
  - tag: meta
    attrs:
      name: test:id
      content: '1234'
```

Instead of just:

```yaml
tag: meta
attrs:
  name: test:id
  content: '1234'
```

I also wonder if the start of the sentence should be different between the two cases, e.g.:

```
The \`head\` configuration includes…
```

```
The \`head\` frontmatter field includes…
```

Any thoughts on this?

delucis:
> While the JSON output seems fine, I wonder if for the YAML one, we should not include more context

I actually like the current one from the screenshots. If you include the `head:` line, it might look like it’s the first item always? Although maybe the `-` could be helpful?

```yml
- tag: meta
  attrs:
    name: test:id
    content: '1234'
```

Definitely would be clearer with some code block styling to distinguish it.

> I also wonder if the start of the sentence should be different between the two cases

I’m fairly neutral here as “configuration” could be interpreted OK in the frontmatter context where the lines above show more clearly that it’s about the specific entry. But yeah “frontmatter field” does _slightly_ more hint where to look.

HiDeoo:
> Although maybe the `-` could be helpful?
> 
> ```yaml
> - tag: meta
>   attrs:
>     name: test:id
>     content: '1234'
> ```
> 
> Definitely would be clearer with some code block styling to distinguish it.

Awesome idea, just pushed that change which I think is enough now.

SCR-20250816-lsbg

PR #3381, opened 2025-08-15
i18n(uk): update `index.mdx`

dolgachio:
#### Description

- Updates `uk/index.mdx` translation


docs/src/content/docs/uk/index.mdx
```diff
@@ -24,29 +30,140 @@ hero:
...
 
 <CardGrid stagger>
-	<Card title="Документація, що кидає в захват" icon="open-book">
-		Містить: навіґацію по сайту, пошук, інтернаціоналізацію, SEO, читку
-		типоґрафіку, підсвічування коду, темний режим і багато іншого.
+	<Card title="Документація, яка вражає" icon="open-book">
+		Включає в себе: навігацію сайтом, пошук, інтернаціоналізацію, SEO, чітку
+		типографіку, підсвітку коду, темний режим і багато іншого.
 	</Card>
 	<Card title="Працює на Astro" icon="rocket">
```
trueberryless, line 41: Just to confirm (I'm not a native speaker), but when translating this to English, it reads more like "Works _for_ Astro" which isn't exactly what it is supposed to mean. We would rather aim for "Powered _by_ Astro". Would you please be so kind and maybe find a better suiting translation here? (I know that you didn't change the line, but maybe you can enhance it when we are already adding and adopting translations on this page) 🫶

docs/src/content/docs/uk/index.mdx
```diff
@@ -24,29 +30,140 @@ hero:
...
+    cite="https://elk.zone/m.webtoo.ls/@loucyx@mastodon.social/111587380021362284"
+  >
+    Документація мого сайту-монорепозиторію виглядає краще, ніж будь-коли, завдяки Starlight. Її надзвичайно легко використовувати, не втрачаючи всю потужність Astro. Дякую за вашу роботу над цим!
+  </Testimonial>
+  <Testimonial
+    name="BowTiedWebReaper"
+    handle="BowTiedWebReapr"
+    cite="https://twitter.com/BowTiedWebReapr/status/1735633399501697517"
+  >
+    Starlight is my go-to tool for documentation. It made it super easy to add docs to my existing Astro product website, vs needing a subdomain to use with another tool.
```
trueberryless: We might want to translate this testimonial too (probably a small oversight) 🙌

trueberryless (changes requested):
As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://contribute.docs.astro.build/reviewers/reviewing-translations/) and have to say that the translation looks wonderful overall. Thank you so much for this awesome first contribution to the Starlight project, @dolgachio. 🥳 

I left two comments about some nitpicky suggestions.

docs/src/content/docs/uk/index.mdx
delucis: At a quick look, and again, as a non-native speaker, it seems this translation may be correct to me. For example, I see the same suggested translation for a more common combo like “Powered by electricity” (Працює на електриці). Can easily be that in another language a word is appropriate that also has a second meaning similar to the English “works” — I’m sure @dolgachio can say for sure though!

dolgachio: Thanks so much for the feedback, @delucis and @trueberryless. I agree that a word-for-word translation doesn't work well here. We just don't have an exact phrase for "Powered by Astro," and a literal translation sounds a bit odd. I think we should stick with the current version since it sounds much more natural to native speakers.

docs/src/content/docs/uk/index.mdx
dolgachio: Thanks, updated

PR #3382, opened 2025-08-15
fix: mobile toc issue with custom banners

### LINKED ISSUE #3047: Mobile ToC issue with custom banners

0.32.6

5.6.1

pnpm

Mac

Chrome

### Describe the Bug

On project having a custom banner (or theme with custom banners) that would change the height, padding-block or margin-block of the banner, when displaying the table of contents on mobile, the ToC will initially not display the current element.

This is visible on the [Starlight Flexoki](https://delucis.github.io/starlight-theme-flexoki/examples/banner/) or [Starlight Rapide](https://starlight-theme-rapide.vercel.app/examples/banner/) themes.

Image

I did not investigate yet but I would assume that this is related to how the root margin for the intersection observer is [calculated](https://github.com/withastro/starlight/blob/2768932065980d7e1b2fc615706b605a9527fb36/packages/starlight/components/TableOfContents/starlight-toc.ts#L100-L110) and does not handle for such cases.

Note that the linked reproduction on StackBlitz does not seem to reproduce the issue (didn't investigate why), so it's linked as an example that should be downloaded or just having a page with the following content will reproduce the issue:

```md
---
title: Test page with custom banner
banner:
  content: A random banner
head:
  - tag: style
    content: |
      :root .sl-banner {
        background-color: red;
        color: white;
        padding-block: 2rem;
      }
---
```

https://stackblitz.com/edit/github-zgordjca?file=src%2Fcontent%2Fdocs%2Findex.mdx

### Participation

trueberryless:
### Description

This PR fixes an issue with the Table of Contents being unable to find any heading when the page has a banner with custom styling. This issue is currently tracked in #3047.

The issue is fixed by introducing a temporary `IntersectionObserver` without any `rootMargin` (so that it definitely gets an intersection target). This new code only runs when there is no other intersecting target found by the existing [`IntersectionObserver`](https://github.com/withastro/starlight/blob/2768932065980d7e1b2fc615706b605a9527fb36/packages/starlight/components/TableOfContents/starlight-toc.ts#L83).

Please note that this issue could also be resolved in some other ways of course and I'm not sure if introducing a second `IntersectionObserver` is a good practice just for this edge case. Feel free to suggest cleaner solutions!

#### Other considered solution

Another solution that came to my mind, but where I figured that it has too many side effects (and therefore I didn't commit to it): By dynamically incorporating the actual height of the banner in the calculation of `bottom` it would be possible to always guarantee that a heading can be found (which would mean that we don't need a second observer). However, the problem with that approach would be that the observer would then permanently be too big and headings would "too early" be considered current headings. For example: If the banner is really big, headings that scroll into the middle of the screen could potentially be set as the "current" headings in the ToC, which would be misleading.

To demonstrate what I mean with this paragraph in code:

```diff
-const bottom = top + 53;
+const bottom = top + 53 + bannerHeight; // assuming we calculate the actual banner height before
```

https://github.com/withastro/starlight/blob/2768932065980d7e1b2fc615706b605a9527fb36/packages/starlight/components/TableOfContents/starlight-toc.ts#L107

### Related stuff and todo

- Closes #3047
- [ ] Remove TODOs

### DIFF
--- a/.changeset/weak-spiders-itch.md
+++ b/.changeset/weak-spiders-itch.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes and issue where the Table of Contents is unable to find any current heading when the page has a banner with custom styling regarding its height.
--- a/packages/starlight/components/TableOfContents/starlight-toc.ts
+++ b/packages/starlight/components/TableOfContents/starlight-toc.ts
@@ -60,6 +60,8 @@ export class StarlightTOC extends HTMLElement {
 
 		/** Handle intersections and set the current link to the heading for the current intersection. */
 		const setCurrent: IntersectionObserverCallback = (entries) => {
+			let found = false;
+
 			for (const { isIntersecting, target } of entries) {
 				if (!isIntersecting) continue;
 				const heading = getElementHeading(target);
@@ -67,8 +69,30 @@ export class StarlightTOC extends HTMLElement {
 				const link = links.find((link) => link.hash === '#' + encodeURIComponent(heading.id));
 				if (link) {
 					this.current = link;
+					found = true;
 					break;
 				}
+			}
+
+			// Fallback: if no heading intersects (because the banner is higher and moves the content down too much),
+			// create a temporal IntersectionObserver without rootMargin that finds the nearest heading.
+			// See https://github.com/withastro/starlight/issues/3047
+			if (!found) {
+				const tempObserver = new IntersectionObserver((tempEntries) => {
+					for (const { isIntersecting, target } of tempEntries) {
+						if (!isIntersecting) continue;
+						const heading = getElementHeading(target);
+						if (!heading) continue;
+						const link = links.find((link) => link.hash === '#' + encodeURIComponent(heading.id));
+						if (link) {
+							this.current = link;
+							break;
+						}
+					}
+					tempObserver.disconnect();
+				});
+
+				document.querySelectorAll('main [id]').forEach((el) => tempObserver.observe(el));
 			}
 		};
 
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -1,6 +1,16 @@
 ---
 title: Getting Started
 description: Learn how to start building your next documentation site with Starlight by Astro.
+banner:
+  content: 'TODO(trueberryless): Remove'
+head:
+  - tag: style
+    content: |
+      :root .sl-banner {
+        background-color: red;
+        color: white;
+        padding-block: 2rem;
+      }
 ---
 
 import { Tabs, TabItem } from '@astrojs/starlight/components';


delucis:
Ah, although re-reading your second considered solution is related to that same idea. Hmm. I’ll have to think about that.

trueberryless:
> Ah, although re-reading your second considered solution is related to that same idea. Hmm. I’ll have to think about that.

Yeah exactly, if we include the height of the banner in the `getRootMargin()` calculation, it solves this issue, but introduces a much more annoying issue - when scrolling down the page, headings that do not yet hit the top of the page are already considered "current headings" in the ToC because of the bigger root margin.

As I think that this problem would be much harder to solve with potentially more edge cases, I gave up on the idea quite early, but maybe there is potential 🤔

Also just to mention it if (maybe I forgot to say): It's impossible to change the `rootMargin` Fter we intitially set it, as mentioned on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin):

> The `rootMargin` **read-only** property of the IntersectionObserver interface is a string with syntax similar to that of the CSS margin property.

_Notice the read-only._

delucis:
> It's impossible to change the `rootMargin` Fter we intitially set it

Yeah that’s correct, that’s why we destroy and recreate the observer on window resizes currently.

trueberryless:
We could maybe use a "Package Preview" here 👀 
Or does it not make sense here? I wonder in which cases it would make the most sense... 🤔

packages/starlight/components/TableOfContents/starlight-toc.ts
```diff
@@ -60,16 +60,40 @@ export class StarlightTOC extends HTMLElement {
...
+						const link = links.find((link) => link.hash === '#' + encodeURIComponent(heading.id));
+						if (link) {
+							this.current = link;
+							break;
+						}
+					}
+					tempObserver.disconnect();
+				});
+
+				document.querySelectorAll('main [id]').forEach((el) => tempObserver.observe(el));
```
delucis: I think we concluded on our review on Talking & Doc’ing today that after we fix other issues in #3656, the remaining case we need to fix in this PR will only be for the page title, i.e. the `<h1>` element.

Given this, I think we can simplify this logic so that if no element is found, we fall back to the `<h1>` instead of needing an additional `IntersectionObserver`.

trueberryless: I adapted the logic and updated the changeset as well.
Thanks @HiDeoo for coming up with that idea and huge thanks to @delucis for explaining the bug with much effort and details 💜

trueberryless:
I will remove the red banner (TODO) in the "Getting started" guide once this PR is approved.
For now, I will move it to "Ready for review" 👍

trueberryless:
After playing around a little more and combining all changes from here and #3656, it seems that this does not at all work as expected. The `heading` is undefined between lines and elements, leading to the first heading being set to the current heading alternatively:

https://github.com/user-attachments/assets/9f2c9aea-76ef-4628-9f2a-86ad56c9dce8

I need to further investigate this. Unfortunately, I might not have that much time the next few days, but let's see...

HiDeoo:
Isn't this related to `found` being reset every time in the intersection observer callback?

trueberryless:
> Isn't this related to `found` being reset every time in the intersection observer callback?

That was it, thanks for the heads up!

trueberryless:
Now that #3656 is merged, should we also add some test cases to prevent this bug from happening again?

delucis:
> Now that #3656 is merged, should we also add some test cases to prevent this bug from happening again?

That would be great! Can you see how to do it or do you need any pointers?

trueberryless:
> That would be great! Can you see how to do it or do you need any pointers?

I small pointer in the right direction would actually be very useful to me.

Mainly:
- Would you also add them to `basics.test.ts`?
- Would you use your `testTOCHighlighting` function as well or shall I create a new one to emulate a bigger banner?

delucis:
> * Would you also add them to `basics.test.ts`?

I would. But you may want to add a new _page_ to the basics fixture that includes the custom banner.

> * Would you use your `testTOCHighlighting` function as well or shall I create a new one to emulate a bigger banner?

I think that function should work, yeah! Just pass it the `path` to your page with the custom banner.

trueberryless:
For some reason, when I run `pnpm test:e2e` locally, it get's stuck. So I pushed to see the CI run here

image

I use Zed on MacOS (still on v18, not v26)... But let's discuss this bug somewhere else 👍 

Edit: This is probably because I run Node in a Nix shell where playwright struggles to start either Chrome or Firefox... This should be fixable on my end by modifying some env variables in the Nix shell 🙌

trueberryless:
Sorry for the force push, but I found some completely unrelated commits in this PR (I don't know how I managed to get them in here), which messed up the labels, so I cleaned up the history a little bit (even if everything get's squashed nonetheless). Now at least the [commits](https://github.com/withastro/starlight/pull/3382/commits) look clean here 🥳

.changeset/weak-spiders-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue where the mobile table of contents is unable to find the first heading when the banner of the page is higher than expected.
```
delucis: Small tweak here for clarity:

```suggestion
Fixes an issue where the mobile table of contents is unable to find the first heading when a page has a tall banner.
```

Most specifically: “taller” rather than “higher” because “higher” would normally just indicate a position (“The sixth-floor window is **higher** than the tree tops”)  vs dimensions of an item (“The tree is **taller** than the building”)

packages/starlight/components/TableOfContents/starlight-toc.ts
```diff
@@ -71,9 +73,23 @@ export class StarlightTOC extends HTMLElement {
...
+					hasIntersectingHeading = true;
 					break;
 				}
 			}
+
+			// Fallback: if no heading intersects (because the banner is higher and
+			// moves the content down too much), set the page title to the current link.
+			// See https://github.com/withastro/starlight/issues/3047
+			if (!hasIntersectingHeading) {
+				const firstHeading = document.querySelector('main h1');
```
delucis: I think we can use the same approach as used on L46 above for finding the title:

```suggestion
				const firstHeading = document.getElementById(PAGE_TITLE_ID) as HTMLHeadingElement;
```

(IIRC this is slightly more performant because the browser already has a map of elements with IDs vs a query which needs parsing and then searching the DOM.)

packages/starlight/components/TableOfContents/starlight-toc.ts
```diff
@@ -71,9 +73,23 @@ export class StarlightTOC extends HTMLElement {
...
+				if (firstHeading) {
+					const link = links.find((l) => l.hash === '#' + encodeURIComponent(firstHeading.id));
+					if (link) {
+						this.current = link;
+					}
+				}
```
delucis: Hmm, bugs me a bit that we’re repeating the code from immediately above. I think we can refactor things to avoid that. Would you mind if I push up a change? I played around with the code a bit, but it’s a bit tricky to change in review comments.

trueberryless: Sure, is it easier for you if I apply the other suggestions first or later?

delucis: You can apply them! I have the refactor locally, but now I’m thinking about another option so won’t push just yet.

.changeset/weak-spiders-itch.md
trueberryless: Thanks for clarification, I wasn't sure which was the correct adjective here 😅 
I'll adapt this change consistently tho, as there are also some comments in the logic where this is wrong then + the test fixture heading is also misleading then 👍

packages/starlight/components/TableOfContents/starlight-toc.ts
trueberryless: Sorry, I got distracted creating an git alias that automates adding Co-authors with just the GitHub username 😅 Seems to work now 🙌

packages/starlight/components/TableOfContents/starlight-toc.ts
```diff
@@ -41,8 +41,8 @@ export class StarlightTOC extends HTMLElement {
 			if (!el) return null;
 			const origin = el;
 			while (el) {
-				// Short circuit if we reach the top-level content container.
-				if (el.classList.contains('sl-markdown-content')) {
+				// Short circuit if we reach the top-level content container or one of the other containers in main.
+				if (el.matches('.sl-markdown-content, main > *')) {
```
trueberryless, line 45: I don't fully understand why the addition of `main > *` is necessary just for the banner? Couldn't we restrict more to avoid unexpected matches or is this intentional?

delucis: This is what I was trying to explain in this part of my comment:

> any direct child of `<main>` just in case someone adds e.g. a banner that doesn’t use the `sl-banner` class name

In other words, we _could_ check for `.sl-banner` specifically, but we support overriding the banner component and people could use a different component without that class name, OR add some other additional above the page title. And we want to cover both those cases, so that’s why I preferred a more generic selector.

In practice this cannot match most elements in `<main>` because we exclude most of them in the selector that determines what to watch.

PR #3387, opened 2025-08-20
Adds OmniPrint documentation to the showcase

omni-byte:
Add [Omni Print Documentation](https://omni-byte.com/docs/) to showcase.


HiDeoo (approved):
Thanks for your contribution and this great addition to the showcase :raised_hands:

PR #3393, opened 2025-08-23
docs: add more community plugins resources

chilensis:
#### Description
Add a new [`page`](https://github.com/pelagornis/starlight-theme-page) theme.

### DIFF
--- a/docs/src/content/docs/resources/themes.mdx
+++ b/docs/src/content/docs/resources/themes.mdx
@@ -88,7 +88,7 @@ Install a theme built by the community to quickly customize the look and feel of
 		},
 		{
 			title: 'Starlight Page',
-			description: 'An intuitive, modern Starlight theme with smooth, floating animations.',
+			description: 'An intuitive, modern Starlight theme with smooth, float-in animations.',
 			href: 'https://pelagornis-page.netlify.app/',
 			previews: { light: 'page-light.png', dark: 'page-dark.png' },
 		}


trueberryless:
Wow, this theme looks really impressive, I would not have believed that it is built with Starlight, great job 🙌

I think the name of the plugin might be pretty vague, but it's fine 👍 

Thank you very much for such an awesome first contribution to the project 🚀

docs/src/content/docs/resources/themes.mdx
```diff
@@ -86,6 +86,12 @@ Install a theme built by the community to quickly customize the look and feel of
 			href: 'https://frostybee.github.io/starlight-theme-galaxy/',
 			previews: { light: 'galaxy-light.png', dark: 'galaxy-dark.png' },
 		},
+		{
+			title: 'Starlight Page',
+			description: 'Starlight theme for Pelagornis Docs',
```
delucis: Thanks @chilensis!

Would you be able to provide a slightly more detailed `description`? “Pelagornis Docs” makes it sound like this theme is just for a specific product or company, but the description should say something about the theme style or inspiration ideally.

chilensis: Thank you for your review @delucis!

There is a design system project currently being carried out by our company, so can I say I was inspired by it?

chilensis:
@delucis Thank you for your review.
As you mentioned, I revised the description.

docs/src/content/docs/resources/themes.mdx
```diff
@@ -86,6 +86,12 @@ Install a theme built by the community to quickly customize the look and feel of
 			href: 'https://frostybee.github.io/starlight-theme-galaxy/',
 			previews: { light: 'galaxy-light.png', dark: 'galaxy-dark.png' },
 		},
+		{
+			title: 'Starlight Page',
+			description: 'A sophisticated and intuitive Starlight theme.',
```
trueberryless: If I may suggest something more "fitting" / descriptive:

```suggestion
			description: 'A sleek, modern Starlight theme with smooth float-in animations.',
```

delucis: That’s kind of you @trueberryless! But I would still like to give @chilensis a chance to frame how they see the plugin, so we get a bit of insight into how they think about their work.

chilensis: Thank you for your suggestion, @trueberryless! 👏
"An intuitive, modern Starlight theme with smooth, float-ing animations."

We'll decide that when we produce the project, we design it for the purpose of being sophisticated and intuitive in design.

And I accepted the feedback from @trueberryless and attached it to the explanation, thank you for the good feedback.

trueberryless: Alright, sorry Chris. That makes sense!

Feel free to adapt the description then to better fit what you have in mind, @chilensis

delucis:
D’oh and I still forgot to add @trueberryless as a co-author 😅 
Sorry, Felix!

trueberryless:
All good 😆

PR #3390, opened 2025-08-20
Add link to new Starlight theme

frostybee:
#### Description
This PR adds a link to the newly released Starlight Galaxy theme.

PR #3399, opened 2025-08-26
i18n(fr): update `resources/themes.mdx`

ArmandPhilippot:
#### Description

Adds changes from #3390 and #3393 to the French translation of `resources/themes.mdx`.

> [!NOTE]
> I'm not quite sure what `float-in`  means (floating?) so the translation could be wrong.

### DIFF
--- a/docs/src/content/docs/fr/resources/themes.mdx
+++ b/docs/src/content/docs/fr/resources/themes.mdx
@@ -79,6 +79,20 @@ Installez un thème créé par la communauté pour personnaliser rapidement l'ap
 			href: 'https://starlight-theme-next.trueberryless.org/',
 			previews: { light: 'nextjs-light.png', dark: 'nextjs-dark.png' },
 		},
+		{
+			title: 'Starlight Galaxy',
+			description:
+				'Un thème Astro Starlight conçu pour la publication de documentation et de contenu éducatif.',
+			href: 'https://frostybee.github.io/starlight-theme-galaxy/',
+			previews: { light: 'galaxy-light.png', dark: 'galaxy-dark.png' },
+		},
+		{
+			title: 'Starlight Page',
+			description:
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/resources/themes.mdx
```diff
@@ -79,6 +79,20 @@ Installez un thème créé par la communauté pour personnaliser rapidement l'ap
...
+			title: 'Starlight Galaxy',
+			description:
+				'Un thème Astro Starlight conçu pour la publication de documentation et de contenu éducatif.',
+			href: 'https://frostybee.github.io/starlight-theme-galaxy/',
+			previews: { light: 'galaxy-light.png', dark: 'galaxy-dark.png' },
+		},
+		{
+			title: 'Starlight Page',
+			description:
+				'Un thème Starlight intuitif et moderne avec des animations fluides et flottantes.',
```
ArmandPhilippot: Thanks Chris, this makes more sense!
I'm not sure we have an appropriate term to describe this in French though... I guess the idea would be `des éléments émergeant progressivement tout en donnant l'impression qu'ils flottent`. But I can't find a shorter version. 😆 `flottantes` could be a bit vague to describe that, maybe we can just say `progressives`:
```suggestion
				'Un thème Starlight intuitif et moderne avec des animations fluides et progressives.',
```
Not sure if this is better.

HiDeoo: Not literal at all, but I think I would personally go with "Un thème Starlight intuitif et moderne avec des animations fluides et de subtiles transitions."

ArmandPhilippot: This sounds even better than `progressives`. This works for me, thanks HiDeoo! 🙌🏽 
```suggestion
				'Un thème Starlight intuitif et moderne avec des animations fluides et de subtiles transitions.',
```

PR #3409, opened 2025-09-04
[i18nIgnore] Fix duplicate i18n collection definition

salvaKaraka:
From:

image

To:

image


HiDeoo (approved):
Thanks for the fix :raised_hands:

I've added the `[i18nIgnore]` keyword to the PR title so that this change does not mark this file as updated in our [translation tracker](https://i18n.starlight.astro.build/).

I'll also add that this current page of the Spanish translation is quite outdated, specially for anything related to content collections and it may be better to refer to the [English version](https://starlight.astro.build/guides/i18n/) until it gets updated. If you want to help with that, our contribution guide is available [here](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#translating-starlights-docs) for reference.

PR #3412, opened 2025-09-05
feat(starlight): re-export Sidebar types and add consumer-side test

KerimCETINBAS:
#### Description

- Closes #3411 
- Re-exports `SidebarLink`, `SidebarGroup`, and `SidebarEntry` types from the root `types.ts`, making them directly importable for consumers.
- Adds a consumer-side test to ensure these types are available and assignable correctly without polluting existing test files.
- This change supports building custom navigation bars and using the sidebar types safely in TypeScript projects.

#### Visual Changes
- None. This PR only affects TypeScript types and tests.

### DIFF
--- a/packages/starlight/__tests__/basics/navigation.test.ts
+++ b/packages/starlight/__tests__/basics/navigation.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test, vi } from 'vitest';
 import { flattenSidebar, getPrevNextLinks, getSidebar } from '../../utils/navigation';
+import type { SidebarEntry, SidebarGroup, SidebarLink } from '../../types';
 
 vi.mock('astro:content', async () =>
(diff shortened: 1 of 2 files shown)


HiDeoo:

I'd like to spend more time discussing the overall issue and potential solutions in the [discussion](https://github.com/withastro/starlight/discussions/3411) you just created before moving forward with a specific implementation.

I commented on the discussion as it's already possible to access these types but it may not be obvious enough. Although, deciding to expose more types directly requires more thought as we need to decide if all types should be exposed or only a subset, how to decide which ones, and we also need to document each of them as they become part of the public API. There are also potential improvements to the existing documentation and examples which could be explored first to see if they help with the DX.

Considering the above, I'll close this PR for now so we can take the time to properly discuss and evaluate the best approach to improve the situation with everyone's input.

PR #3414, opened 2025-09-08
Correct directory name for Chinese localization

IEvangelist:
#### Description

If the folder name doesn't match the configured `lang` it will not translate astro components.

I discovered that localization doesn't work unless the folder names match the casing of the configured language.

### DIFF
--- a/docs/src/content/docs/guides/i18n.mdx
+++ b/docs/src/content/docs/guides/i18n.mdx
@@ -57,7 +57,7 @@ Starlight provides built-in support for multilingual sites, including routing, f
        - docs/
          - ar/
          - en/
-         - zh-cn/
+         - zh-CN/
 
    </FileTree>
 
@@ -108,7 +108,7 @@ When using a `root` locale, keep pages for that language directly in `src/conten
   - content/
     - docs/
       - **index.md**
-      - zh-cn/
+      - zh-CN/
         - **index.md**
 
 </FileTree>


delucis (commented):
Hi @IEvangelist, could you share the Starlight config you tried? This example code should be correct — it matches the key in the `locales` configuration.

IEvangelist:
> Hi @IEvangelist, could you share the Starlight config you tried? This example code should be correct — it matches the key in the `locales` configuration.

Hi @delucis, thanks for the reply.

Here's the locales that I've defined in the config:

```js
export const locales = {
  root: { label: 'English', lang: 'en' },
  de: { label: 'Deutsch', lang: 'de' },
  es: { label: 'Español', lang: 'es' },
  ja: { label: '日本語', lang: 'ja' },
  id: { label: 'Bahasa Indonesia', lang: 'id' },
  'zh-cn': { label: '简体中文', lang: 'zh-CN' },
  'pt-br': { label: 'Português do Brasil', lang: 'pt-BR' },
  'pt-pt': { label: 'Português', lang: 'pt-PT' },
  ko: { label: '한국어', lang: 'ko' },
  tr: { label: 'Türkçe', lang: 'tr' },
  ru: { label: 'Русский', lang: 'ru' },
  hi: { label: 'हिंदी', lang: 'hi' },
  da: { label: 'Dansk', lang: 'da' },
  uk: { label: 'Українська', lang: 'uk' },
};
```

I've noticed that all of my astro components fallback when attempting to use any of the hyphenated locales. The only difference I observed was a different casing. Maybe the issue is instead that the casing of the `*.json` files needs to match? I really wish it was case insensitive altogether.

delucis:
> Maybe the issue is instead that the casing of the `*.json` files needs to match?

Ah yeah, I think currently the JSON (or YAML) translation files _do_ need the `zh-CN.json` style, while the documentation page directories _don’t_. That’s definitely confusing.

For a bit of context: we went with recommending lowercase for the directories/docs paths, because these get translated to URLs (e.g. `src/content/docs/zh-cn/example.md` becomes the `/zh-cn/example/` page) and we found all-lowercase URLs were a good practice for various reasons — it helps avoid accidental SEO issues (e.g. canonical URLs not matching due to casing differences), some common hosts lowercase everything _anyway_, plus some filesystems are case insensitive too, so uppercase filenames can cause issues for teams working across different OSes.

There’s no real reason for us to care about the casing of the translation files though — we should be able to match `zh-cn.json` and `zh-CN.json` just as easily, so we should fix that to be more flexible.

IEvangelist:
> There’s no real reason for us to care about the casing of the translation files though — we should be able to match `zh-cn.json` and `zh-CN.json` just as easily, so we should fix that to be more flexible.

Yes, please!

When we map out a translation for a sidebar entry, something like this:

```ts
translations: {
  da: 'Arkitektur',
  de: 'Architektur',
  en: 'Architecture',
  es: 'Arquitectura',
  fr: 'Architecture',
  hi: 'आर्किटेक्चर',
  id: 'Arsitektur',
  it: 'Architettura',
  ja: 'アーキテクチャ',
  ko: '아키텍처',
  'pt-br': 'Arquitetura',
  'pt-pt': 'Arquitetura',
  ru: 'Архитектура',
  tr: 'Mimari',
  uk: 'Архітектура',
  'zh-cn': '架构',
}
```

Should these keys also be all lowercase? Or should they be something like `zh-CN`?

delucis:
> Should these keys also be all lowercase? Or should they be something like `zh-CN`?

Ah, good point — I think they should also match the `lang` tag, i.e. `zh-CN`. Basically: correct [BCP-47 tag](https://en.wikipedia.org/wiki/IETF_language_tag) everywhere _except_ object keys in `locales` and directories in `src/content/docs/`.

I’ll close this PR as I guess it’s not a change we want, but we’ll look into improving the experience here and feel free to keep asking questions!

IEvangelist:
I'm still a bit confused. See here: https://github.com/withastro/starlight/blob/e7fd6de4f4725f6e3d976f8cf8ee6a49d79fa97e/docs/lunaria.config.json#L70 These `lang` values aren't BCP-47 tags, they're lowercase.

delucis:
Ah, that’s our [Lunaria](https://lunaria.dev/) config, rather than Starlight. I’d need to double check why that’s the way it is.

**Edit:** The [Lunaria docs](https://lunaria.dev/reference/configuration/#defaultlocale-required) also say “BCP-47” for those config entries, but not sure why our config is using a lowercase version.

IEvangelist:
If you had five mins you could probably figure out the issue I'm having. Everything looks right, in fact, it works locally... but when I deploy the site my astro components that try rendering localized strings for their locales that include a `-` do not work.

delucis:
Ohhh, I'm guessing you renamed the files to the different casing and git didn't pick that up maybe? I've seen that happen before where a local casing change doesn't make it to the git remote. Sometimes you need to `git mv` the files with the casing change to ensure they end up in the repo. 

I'm logging off for the night but if it's not that, happy to take a look at a repo if you can share one!

IEvangelist:
@delucis — To recap, what values need to have what casing?

Here's what I have thus far:

| Location | Example | Casing |
|----------|---------|--------|
| Folder names  | `content/docs/zh-cn` | lowercase |
| JSON file names | `content/i18n/zh-CN.json` | uppercase |
| JSON keys | astro.config.mjs locales: `{ 'pt-br': { label: 'Português do Brasil', lang: 'pt-BR' }` | lowercase |
| JSON `lang` values | astro.config.mjs locales: `{ 'pt-br': { label: 'Português do Brasil', lang: 'pt-BR' }` | uppercase |
| JSON `label` or `translations` keys | `{ label: { 'pt-BR': 'Documentos' }` <br> `{ label: 'example', slug: 'exmaple', translations: { 'pt-BR': 'Example' } }` | uppercase |

And by uppercase, I mean BCP-47 proper.

We have a new docs UX launch on the 23rd, and I'd like to keep it a surprise. I could share a deployment of it with you, but privately as to avoid it being public. The repo is private now too, but I'd love for you to see what I'm experiencing and if you can make sense of it. Are you open to email, if so my email is my first name followed by a dot, and then my last name `@microsoft.com` (trying to avoid putting it here 😄).

delucis:
That table looks correct! I’ll drop you an e-mail.

IEvangelist:
> That table looks correct! I’ll drop you an e-mail.

Awesome, thank you. Replied... Thanks again.

PR #3420, opened 2025-09-11
Don’t generate invalid `<link rel="canonical"/>` with missing `href`

andersk:
Remove these invalid HTML tags from the output when `context.site` is unset:

- `<link rel="canonical"/>` with missing `href`
- `<meta property="og:url"/>` with missing `content`


.changeset/quick-apes-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Removes invalid HTML tags `<link rel="canonical"/>` and `<meta property="og:url"/>` from output when `context.site` is unset
```
trueberryless: I think we could make the changeset clearer (include the description changes as well):

```suggestion
Fixes invalid `<head>` output when configuration is missing:
- Omits `<meta property="og:description" />` if Starlight’s `description` option is unset
- Omits `<link rel="canonical" />` and `<meta property="og:url" />` if Astro’s `site` option is unset
```

packages/starlight/utils/head.ts
```diff
@@ -51,9 +53,13 @@ export function getHead(
...
 		{ tag: 'meta', attrs: { property: 'og:title', content: data.title } },
 		{ tag: 'meta', attrs: { property: 'og:type', content: 'article' } },
-		{ tag: 'meta', attrs: { property: 'og:url', content: canonicalHref } },
+		...(canonicalHref !== undefined
+			? ([{ tag: 'meta', attrs: { property: 'og:url', content: canonicalHref } }] as const)
+			: []),
 		{ tag: 'meta', attrs: { property: 'og:locale', content: lang } },
-		{ tag: 'meta', attrs: { property: 'og:description', content: description } },
+		...(description !== undefined
+			? [{ tag: 'meta', attrs: { property: 'og:description', content: description } } as const]
```
trueberryless: I think the brackets here should be the same as in the other changes, but maybe it's better to adapt the other two instead of this:

```suggestion
			? ([{ tag: 'meta', attrs: { property: 'og:description', content: description } }] as const)
```

andersk:
Adjusted as suggested, thanks.

The [specification](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-href) for `<link>` `href` says:

> One or both of the [`href`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-href) or [`imagesrcset`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-imagesrcset) attributes must be present.

For `<meta>` `content`, since OpenGraph is [based on RDFa](https://ogp.me/#metadata), here’s the [relevant specification](https://www.w3.org/TR/html-rdfa/#h-extensions-to-the-html5-syntax):

> If the RDFa `@property` attribute is present on the `meta` element, neither the `@name`, `@http-equiv`, nor `@charset` attributes are required and the `@content` attribute MUST be specified.

delucis:
Thanks again for the PR @andersk!

@erbierc has kindly offered to take over writing tests for this, so I’m going to merge this into an intermediate branch — https://github.com/withastro/starlight/tree/stack/invalid-canonical — so we can collaborate there before merging into `main`.

PR #3422, opened 2025-09-11
Convert invalid `<div>` child of `<summary>` to `<span>`

andersk:
The [`<summary> element`](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element) can only contain phrasing content elements like `<span>`, not flow content elements like `<div>`.

This makes no styling difference in these cases, as these are flexbox children.


delucis (approved):
Thanks for this one too 🙌

PR #3423, opened 2025-09-11
Move `<summary>` to top of `<details>` for validity

andersk:
The [`<summary>` element](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element) must be the _first_ child of `<details>`.

### DIFF
--- a/.changeset/small-singers-smell.md
+++ b/.changeset/small-singers-smell.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Makes `<summary>` the first child of `<details>` in sidebar groups
+Fixes HTML validity in sidebar groups by ensuring `<summary>` is the first child of `<details>`


andersk:
I’m just trying to get my Starlight output to pass an HTML validator. ([Example](https://validator.w3.org/nu/?doc=https%3A%2F%2Fstarlight.astro.build%2Fgetting-started%2F).) Of course I’m aware that not all validation errors are problems in the real world, so I’m filtering out a mix of validator bugs, legitimate nonstandard extensions, and browser bug workarounds to find the real issues. This case clearly does violate the standard, so even if it doesn’t cause problems in practice with today’s common browsers, we might as well fix it to make it easier for real issues to stand out.

packages/starlight/components/SidebarSublist.astro
```diff
@@ -37,7 +37,6 @@ const { sublist, nested } = Astro.props;
 					<details
 						open={flattenSidebar(entry.entries).some((i) => i.isCurrent) || !entry.collapsed}
 					>
-						<SidebarRestorePoint />
 						<summary>
 							<div class="group-label">
```
delucis: I wonder if another option here would be to move the restore point inside the `<summary>` so it’s still as early as possible in the tree:

```suggestion
							<SidebarRestorePoint />
							<div class="group-label">
```

packages/starlight/components/SidebarSublist.astro
```diff
@@ -51,6 +50,7 @@ const { sublist, nested } = Astro.props;
 							</div>
 							<Icon name="right-caret" class="caret" size="1.25rem" />
 						</summary>
+						<SidebarRestorePoint />
```
delucis, line 53: And then remove it here:
```suggestion
```

andersk:
Yeah that seems fine to me: `this.closest('details')` in `SidebarPersister` will still be able to find `<details>`. Updated.

delucis:
Oops, I should have tested this before suggesting it 😅 
image

Might be best to go with your original suggestion after all 👍 (Or introduce styles, but that seems overkill?)

PR #3430, opened 2025-09-16
i18n(fr): small rewording in `components`

ArmandPhilippot:
#### Description

Updates the French translations in `components` to:
* replace `import` with `importation` (`import` is an anglicism, we choose to use `importation` according to the [French i18n guide in Astro Docs](https://github.com/withastro/docs/blob/main/i18n-guides/fran%C3%A7ais.md#langage-courant))
* translate `label`, I don't have a strong opinion between `étiquette` and `libellé`, but it seems `étiquette` is [the right translation](https://www.larousse.fr/dictionnaires/francais/label/45761)
* fix some typo/missing translation
* slightly reword a sentence to use the right pronouns (`le lecteur` > `son` / `les lecteurs` > `leur`)

### DIFF
--- a/docs/src/content/docs/fr/components/file-tree.mdx
+++ b/docs/src/content/docs/fr/components/file-tree.mdx
@@ -26,7 +26,7 @@ import Preview from '~/components/component-preview.astro';
 
 </Preview>
 
-## Import
+## Importation
 
 ```tsx
 import { FileTree } from '@astrojs/starlight/components';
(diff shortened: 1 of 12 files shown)


docs/src/content/docs/fr/components/file-tree.mdx
```diff
@@ -133,7 +133,7 @@ import { FileTree } from '@astrojs/starlight/components';
 ### Ajouter des commentaires
 
 Ajoutez un commentaire à un fichier ou à un répertoire en ajoutant du texte après le nom.
-Le formatage Markdown en ligne, tel que le gras et l'italique, est pris en charge dans les commentaires.
+La syntaxe de mise en forme Markdown, telle que le gras et l'italique, est prise en charge dans les commentaires.
```
ArmandPhilippot, line 136: The goal here was mostly to replace `en ligne` because we agreed in the [French i18n guide in Astro Docs](https://github.com/withastro/docs/blob/main/i18n-guides/fran%C3%A7ais.md#langage-courant) that `en ligne` is not the right translation. I think in this case it's preferable to omit `inlined`...

I take advantage of this to reword the sentence because `le formatage Markdown` bothered me. But maybe it was just me. I can revert if you preferred the previous translation! 😄

PR #3431, opened 2025-09-16
i18n(fr): small rewording and fixes in `guides`

ArmandPhilippot:
#### Description

Updates the French translations in `guides` to:
* revert some unwanted translations (`syntaxe frontale`, `matière première` for `frontmatter`)
* translate `docs`
* replace the translation for `inline` (`en ligne` is usually the wrong translation)
* fix some bad translations (`backticks` > `accents graves`, `blockquotes` > `bloc de citation`)
* replace `supporte` with `prend en charge` (`support` is an anglicism)
* use French quotes
* fix some Astro docs links because of updates on the French version
* translate `plugin` to be consistent with the other docs pages
* respect verb agreement (`propriété` > `utilisée` instead of `utilisé`)
* translate `package`
* replace `actifs` (financial/crypto word) with `ressources` for `assets`
* use non breaking space before punctuation in some places to avoid:

starlight-without-non-breaking-spaces-example

(note the colon on a new line)

---

One thing I haven't updated for consistency: `en-tête`/`titre`. In Astro Docs we use `titre` for `headings` while in most places, Starlight uses `en-tête`. I could have replaced the `titre` occurrences in `authoring-content` but I'm a bit torn.
While `titre de niveau 2` sounds correct to me, `en-tête de niveau 2` sounds unusual to me. So I wasn't sure what to do and so I preferred to leave them as they are. For reference, LibreOffice uses `titre` in French when the English version uses `heading`:
|French|English|
|---|---|
|libre-office-french-heading|libre-office-english-heading|

**Note:** This is not necessarily blocking for this PR since I haven't updated that! But we might want to think about it to make things consistent in a future PR!

### DIFF
--- a/docs/src/content/docs/fr/guides/authoring-content.mdx
+++ b/docs/src/content/docs/fr/guides/authoring-content.mdx
@@ -3,9 +3,9 @@ title: Création de contenu en Markdown
 description: Un aperçu de la syntaxe Markdown prise en charge par Starlight.
 ---
 
-Starlight prend en charge l'ensemble de la syntaxe [Markdown](https://daringfireball.net/projects/markdown/) dans les fichiers `.md` ainsi que la syntaxe frontale [YAML](https://dev.to/paulasantamaria/introduction-to-yaml-125f) pour définir des métadonnées telles qu'un titre et une description.
+Starlight prend en charge l'ensemble de la syntaxe [Markdown](https://daringfireball.net/projects/markdown/) dans les fichiers `.md` ainsi que le frontmatter en [YAML](https://dev.to/paulasantamaria/introduction-to-yaml-125f) pour définir des métadonnées telles qu'un titre et une description.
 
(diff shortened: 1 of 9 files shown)


HiDeoo (approved):
Amazing, definitely some old translations that probably predate both of us in there, really neat update :raised_hands:

Regarding "titre" vs "en-tête", while some websites like MDN or even the W3C use both of them, I think I personally like the explicit different word for headings in the Starlight context. In Starlight, we have the project title in the Astro config that impacts the document `<title>`, we have the page title in content files that also impacts the document `<title>` and what is displayed in the `<PageTitle/>` component. I think having a dedicated word for headings, like in English, can avoid some potential confusions (but maybe I'm wrong).

ArmandPhilippot:
Thanks for the feedback, I know you've explained this in the past but reading "en-tête de niveau 2" made me doubt. 😅 

So, before merging this one, I'll add a new update to make the use of `en-tête`/`title` consistent, no need for a separate PR for this!

ArmandPhilippot:
Done! I spotted another one in `reference/configuration` so I included it here.

HiDeoo (approved):
Thanks for the follow-up 🌟

PR #3432, opened 2025-09-16
i18n(fr): small rewording and fixes in top-level pages

ArmandPhilippot:
#### Description

Updates the French translations of the top-level pages to:
* replace the translation of `analytics` (for the same reason explained in #3429, `analyses` by itself sounds wrong)
* add non breaking spaces in some places to avoid the colon to be on a new line alone
* translate `package`
* use French quotes
* slightly reword some sentences
* replace the translation for `type-safety` (`sûreté du typage` is what we used both on Astro Docs and on the other pages)
* replace `supporte` with `est compatible avec`/`prend en charge` because `supporte` is an anglicism

In `index.md`, I'm a bit torn regarding `Apportez votre...`, this sounds like a literal translation and I think in French we would rather use `Utilisez votre` but in doubt I left it untouched.


ArmandPhilippot:
Thanks, I replaced "Apportez" and "mise en évidence du code"!

Regarding the last card, it's the translation for `Starlight ships as a framework-agnostic, complete docs solution.`, but I believe I already reworded that unless I missed a place?
> Starlight est une solution de documentation complète et indépendante du framework utilisé.

Now, I'm not sure my suggestion is the perfect translation so if you have another idea it will be much appreciated. 😄

PR #3433, opened 2025-09-17
Run all tests on CI workflow changes

HiDeoo:
#### Description

I noticed a while ago in https://github.com/withastro/starlight/pull/3374 that if the `.github/workflows/ci.yml` is updated but nothing in the `packages/**` directories is, our unit tests and e2e tests do not run. In the case of the linked PR, this was not really a problem, altho, if we were to update an environment variable like `NODE_VERSION` in this file and nothing else, we want to make sure that all tests run.

This PR updates the [`dorny/paths-filter`](https://github.com/dorny/paths-filter) configuration to include a new `ci` filter matching this file used to trigger some jobs that were previously skipped when only this file was changed (I considered adding this file to the existing filters, which would have avoided the need to edit the `if` conditions, but I felt that it was cleaner to have a dedicated filter for this purpose).

Here are 2 runs for this PR only modifying the CI workflow file:

- [Before this changes is fully implemented, tests are skipped](https://github.com/withastro/starlight/actions/runs/17798652381?pr=3433)
- [After this change, all tests run as expected](https://github.com/withastro/starlight/actions/runs/17798763172?pr=3433)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
     outputs:
       docs: ${{ steps.filter.outputs.docs }}
       packages: ${{ steps.filter.outputs.packages }}
+      ci: ${{ steps.filter.outputs.ci }}
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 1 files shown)


delucis (approved):
Oh, very well spotted! I’d never noticed this.

Thanks for the great catch 🙌

PR #3443, opened 2025-09-25
Set packages author to "withastro"

trueberryless:
#### Description

Recently I noticed that most packages in this mono-repo are authored under "Chris Swithinbank".
I know that he is the founder of Star(book)light, but at the same time I noticed that most other Astro repos either have no author or just `"withastro"`. So I adapted it here as well.

If I am not aware that this project is still authored by Chris legally, then I am sorry for creating this PR.
Just let me know if I should adapt or close this PR!

### DIFF
--- a/packages/docsearch/package.json
+++ b/packages/docsearch/package.json
@@ -2,7 +2,7 @@
   "name": "@astrojs/starlight-docsearch",
   "version": "0.6.0",
   "description": "Algolia DocSearch plugin for the Starlight documentation theme for Astro",
-  "author": "Chris Swithinbank <swithinbank@gmail.com>",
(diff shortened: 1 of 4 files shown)


HiDeoo:
Thanks for the PR.

As you mentioned, across the Astro organization, we do not really have a consistent approach, e.g. random authors, `withastro`, `astro`, or no author at all. If we were to standardize at some point, we would probably do that org-wide and maybe use the object form with a name and URL (rather than the string) as this is probably the most informative option for Astro projects but this is not something we have planned or want to prioritize right now.

Considering the above, and that this field is not really important or have any practical impact, I'm going to close this PR. Thanks for your effort nonetheless.

PR #3451, opened 2025-10-01
feat: disable DarkReader

trueberryless:
## Description

This PR adds a new meta tag to Starlight which prevents the extension [DarkReader](https://darkreader.org/) from manipulating the dark mode Starlight provides (https://northstarthemes.com/blog/web-dev/disable-darkreader-on-site/#how-to-disable-darkreader-on-your-website).

I set the changeset to make this a minor change, as it might be a breaking change for some websites - I think. Open for suggestions.

## Some thoughts about these changes:
- Could there ever be the possibility that Starlight wants to show the light mode, and the user wants the DarkReader to enforce the dark mode? (drafting for this reason)
- Should it even be Starlight's responsibility to add such a meta tag to the head as this would probably mean that many other edge-cases should/could be included. I am leaning towards not including such "interactions with other tools/extensions" in Starlight as this would make everything bloat.
- On the other hand: As a user of DarkReader, I find it pretty annoying to have to disable DarkReader for all Starlight pages, as the are in dark mode by default and the DarkReader version makes them pretty ugly.
- Maybe we could make DarkReader detect the dark mode with some other cleaner changes.

Maybe we should turn this into an issue or discussion, just wanted to start with actual changes, as they are pretty minimal and explain better what's going on.

### DIFF
--- a/.changeset/sharp-melons-add.md
+++ b/.changeset/sharp-melons-add.md
@@ -1,0 +1,17 @@
+---
+'@astrojs/starlight': minor
+---
+
+Disables the [DarkReader](https://darkreader.org/) extension for Starlight sites by adding a new meta tag: `<meta name="darkreader-lock" />`.
+
+⚠️ **Potentially breaking change:** If you were relying on this behaviour or you want the DarkReader extension to modify Starlight's built in dark mode, you can [add this middleware](https://starlight.astro.build/guides/route-data/#how-to-customize-route-data) to remove the new meta tag:
+
+```ts
+import { defineRouteMiddleware } from "@astrojs/starlight/route-data";
+
+export const onRequest = defineRouteMiddleware((context) => {
+  const { head } = context.locals.starlightRoute;
+  const filteredHead = head.filter((entry) => !entry.attrs || entry.attrs["name"] !== "darkreader-lock");
+  context.locals.starlightRoute.head = filteredHead;
+});
+```
--- a/packages/starlight/utils/head.ts
+++ b/packages/starlight/utils/head.ts
@@ -60,6 +60,8 @@ export function getHead(
 			tag: 'meta',
 			attrs: { name: 'twitter:card', content: 'summary_large_image' },
 		},
+		// Disable DarkReader
+		{ tag: 'meta', attrs: { name: 'darkreader-lock' } },
 	];
 
 	if (description)


delucis (commented):
Hmm, I’m not sure we should add metadata for a very specific non-standard tool like this.

Also, from my experience, Dark Reader should already detect dark mode automatically, I think? I don’t see it applying to sites that are already in dark mode usually, including e.g. Starlight docs.

trueberryless:
> I don’t see it applying to sites that are already in dark mode usually, including e.g. Starlight docs.

That's interesting, for me many Starlight sites are not getting detected, but maybe this is also some weird and inconsistent behaviour of the Zen browser.

Nonetheless, as already stated: I agree that Starlight should not have such meta links, writing a blog teaching how to achieve something like forcing DarkReader disabled would suffice. From my side, this could be closed, but @HiDeoo opinion would be interesting I have to admit.

HiDeoo:
I'm personally against such additions for multiple reasons:

- As Chris said, this is definitely non-standard and also tool specific, so what about other similar tools? I don't think we should add such markup for all similar tools.
- I tested on Starlight Docs, Astro Docs, and many other Starlight websites and even themes, their "Detect dark theme" feature (enabled by default) which does not override a website's dark theme works perfectly fine for me.
   SCR-20251002-jbpn
   SCR-20251002-jbgs
   SCR-20251002-jbio
- Using such markup just stop any changes the extension would have made without any feedback and no way to override it. This means that if someone use specific settings for any reason, e.g. a specific contrast or sepia tone, or even a specific font (as DarkReader support such feature) due to a disability, none of their settings would apply.

ArmandPhilippot:
I haven't been using this extension for a long time, but from what I've seen, the extension handles websites that have a properly configured dark theme quite well.
So, I don't think this should be built-in into Starlight. Either a website don't care about dark themes and should include the meta on its own, or if there is a styling issue affecting many websites, maybe this should be fixed in the extension itself. And so, I agree with HiDeoo.

trueberryless:
Thanks for all the awesome feedback! As mentioned before, I am also opposed to the idea of including third-party tools like this.

Just to reason why I came up with this discussion/PR, here are some of the sites where auto-detection did not work for me (but I think this might be a "doesn't work on my machine"-problem, maybe because I use the Zen Browser?):

- [Starlight Sidebar Topics](https://starlight-sidebar-topics.netlify.app/)
- [Starlight Sidebar Topics Dropdown](https://starlight-sidebar-topics-dropdown.trueberryless.org/)
- [Starlight View Modes](https://starlight-view-modes.trueberryless.org/)
- https://pls.cli.rs/, https://xs-dev.js.org/, https://docs.mrrobot.app/, and many more from [showcase](https://starlight.astro.build/resources/showcase/)...

In all those examples, the manipulated dark mode looks like this:

image

---

Even Starlight docs themselves are inconsistently themed (sometimes auto-detection works, sometimes not): 

image

When navigating around a bit, DarkReader eventually will detect the dark-mode, but weirdly enough not on initial load 🤔 

---

I'm closing this as this seems to be more a DarkReader issue, and I agree to all of you that this should not be part of Starlight 👍

PR #3457, opened 2025-10-06
Dedupe head sitemap link tags

### LINKED ISSUE #3456: Impossible to cleanly opt out of default sitemap filename

0.36.0

5.14.1

npm 11.2.0

macos 14.7

not relevant to the bug

### Describe the Bug

I need to customize the filename of the sitemap generated for my Starlight project. To do so, I am adding `@astrojs/sitemap` to `astro.config.ts` myself:
```js
export default defineConfig({
  adapter: vercelAdapter(),
  integrations: [
    sitemap({
      filenameBase: 'sitemap-foobar',
   }),
  ]
})
```

However, I still want canonical links on my pages, so I do want to include `site: 'https://example.com'` in the config as well.

Starlight has code that detects that I have already added the sitemap plugin, so it doesn't add the plugin again and so it won't generate a second sitemap. That's good. However, it still adds a link to my page's head that links to the default `sitemap-index.xml` file (that doesn't exist in my project) because I set `site` in my config.

I would expect Starlight not to add a link to `sitemap-index.xml` the page's head if I am explicitly adding the sitemap plugin to the Astro config, especially if I'm using a different filename.

https://idx.google.com/astro-starlight-75283240

### Participation

HiDeoo:
#### Description

- Closes #3456

As discussed in the linked issue, this PR dedupes `<link rel="sitemap" href="" />` tags in the head, similar to how `<link rel="canonical" href="" />` tags are deduped.

I think the changeset is a bit verbose, not quite sure how to make it more concise while still being clear and explaining the use-case. Happy to adjust as needed.


packages/starlight/utils/head.ts
trueberryless: We might want to update the comment here as well, alto this leads to
1. a long line (could be splitted into two lines or even a list (ordered or unordered)).
2. too many "or"s for my taste.
So I think it's also okay to not change here. However, I think it should be mentioned somewhere in the comment if all other cases are covered as well.

```suggestion
 * Test if a head config object contains a matching `<title>`, `<meta>`, `<link rel="canonical">`
 * or `<link rel="sitemap">` tag.
```

.changeset/mean-icons-sneeze.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/starlight': patch
+---
+
+Dedupes sitemap link tags in the head.
```
trueberryless: This is not a fix suggestion, just an idea to simplify the meaning: Maybe the term "merge" (or "overrides" — like you write it below —) is more intuitive here for the user to image what this change does 🤔 

```suggestion
Merges sitemap link tags in the head.
```

Personally I think the rest of the changeset is really well-written and understandable with clear links to documentation 👍

delucis: “Merge” suggests they would be somehow _combined_, which isn’t what happens. Might help to spell out the full word here though for people less familiar with the term?
```suggestion
Deduplicates sitemap link tags in the head.
```

packages/starlight/utils/head.ts
delucis: Thanks @trueberryless — I updated your suggestion to break over two lines and avoid the repeated “or”.

PR #3471, opened 2025-10-13
Add `starlight-links` Visual Studio Code extension to showcase

HiDeoo:
#### Description

While initially only supporting basic Markdown links, I [just updated](https://github.com/HiDeoo/starlight-links/releases/tag/starlight-links%400.2.0) the `starlight-links` Visual Studio Code extension to provide IntelliSense for most type of links found in Starlight projects, even custom components, so this is probably a good time to add it to the list of community tools.

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -247,4 +247,9 @@ These community tools and integrations can be used to add features to your Starl
 		title="astro-d2"
 		description="Transform D2 Markdown code blocks into diagrams."
 	/>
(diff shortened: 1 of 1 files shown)


HiDeoo:
> Let us know when you think it could make sense to add as a recommended extension to our templates (and this repo for that matter).

Sure, my basic plan before doing so is:

- Address new feedback as usage grows slightly by sharing it a bit more
- Make sure it works with Astro Docs (it does already but without taking i18n into account though)
  - I'll make a PR to Docs first refactoring the i18n config so that it's statically analyzable

PR #3472, opened 2025-10-14
Remove shortcut from favicon

ghost:
#### Description

- Closes # 

Removes `shortcut` from `rel="icon"` because its not a valid link relation.
[How to Favicon in 2025: Three files that fit most needs](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs)


delucis:
Hi @jacobdalamb — thanks for the PR!

As you can see from the failing tests, we currently depend on the presence of `shortcut`, so it’s not so straightforward to remove. You can see more details in #2646.

Given there’s little benefit to removing it (it’s harmless), and the effort involved in removing it is nontrivial, I’ll close this PR.

PR #3491, opened 2025-10-17
Fix: Update overflow property in reset.css

### LINKED ISSUE #1321: "overflow-wrap: anywhere" is problematic in table cell code blocks

0.15.1

4.0.8

npm

mac

chromium-based

### Describe the Bug

There's a style somewhere that sets `overflow-wrap: anywhere` -- probably across the markdown content.  But this is problematic when the code block lives in a table cell and the code block is particularly short:

image

Source of the above:

```md
| Name      | Type      | Default | Description                                                                                  |
| --------- | --------- | ------- | -------------------------------------------------------------------------------------------- |
| `types`   | `boolean` | `true`  | Check the file referenced by the conditional `types` export has a `.d.ts` extension          |
| `require` | `boolean` | `true`  | Check the file referenced by the conditional `require` export references a CommonJS script   |
| `import`  | `boolean` | `true`  | Check the file referenced by the conditional `import` export references an ECMAScript module |
| `order`   | `boolean` | `true`  | Check the conditional `default` export is the _last_ field in its object                     |
| `glob`    | `boolean` | `true`  | If `false`, disallow glob patterns in the `exports` field                                    |
```

Suggestion:

Something like

```css
.sl-markdown-content :is(th, td):not(:where(.not-content *)) > code {
  overflow-wrap: normal;
}
```

...and whatever CSS is needed to except `pre > code` from the rule.

n/a

### Participation

JusticeMatthew:
#### Description

Implements the change discussed in https://github.com/withastro/starlight/issues/1321
Closes #1321


.changeset/blue-trees-shave.md
```diff
@@ -0,0 +1,15 @@
+---
+"@astrojs/starlight": minor
+---
+
+Changes text overflow styling in Markdown content
+
+⚠️ **Potentially breaking change:** This release switches the [`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) CSS style for common elements to `break-word`. In most cases, there should be little visual impact, but this change can impact how layouts with implicit sizing (such as tables) look, improving legibility in how words wrap.
+
+If you want to preserve the previous styling, you can add the following custom CSS to your site:
```
trueberryless: Might be helpful if we include a link to how one can set custom CSS here:

```suggestion
If you want to preserve the previous styling, you can add the following [custom CSS](https://starlight.astro.build/guides/css-and-tailwind/#custom-css-styles) to your site:
```

delucis:
> Does this affect #1885 in any way? Could we close it or should we still take a look at that?

Good question. The intention of that PR still stands: that these styles maybe belong in the Markdown styles rather than the reset. But I honestly lost the context for it, so can’t remember whether we should do that or not.

PR #3495, opened 2025-10-22
docs: update CONTRIBUTING.md to clarify development environment setup

erbierc:
#### Description

Added additional information to the contributing guidelines to clarify that a dev server should be ran on project setup for type generation.

**NOTE:** I think we should also change/remove the Gitpod info, as it's a completely different, AI-centric platform called "Ona" now. Open to feedback!


HiDeoo:
Just linking to the [Discord conversation](https://discord.com/channels/830184174198718474/845430950191038464/1430598380899532840) about this PR for additional context for now.

The scenario could basically be someone trying to contribute to unit tests, who may not have run `pnpm dev` in the `docs/` directory (which is mostly documented in the "Testing visual changes while you work" section which may not be obvious to someone just trying to run unit tests), and hitting type errors in some files.

CONTRIBUTING.md
```diff
@@ -64,6 +64,13 @@ You can [develop locally](#developing-locally) or use an online coding developme
    pnpm i
    ```
 
+5. **Run the dev server** to generate types:
+
+   ```sh
+   cd docs
+   pnpm dev
+   ```
```
HiDeoo: Considering we already have a ["Testing visual changes while you work" section](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#testing-visual-changes-while-you-work) which explains how to run the dev server, and it may not be obvious why a dev server is needed to generate types, I think it may be easier to instead use the [`astro sync`](https://docs.astro.build/en/reference/cli-reference/#astro-sync) command to only do the type generation step.

````suggestion
5. **Generates TypeScript types** for all Astro modules:

   ```sh
   cd docs
   pnpm astro sync
   ```
````

I guess such step should be also added to the ["Developing using GitHub Codespaces" section](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#developing-using-github-codespaces) in second position.

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -80,7 +87,7 @@ You can [develop locally](#developing-locally) or use an online coding developme
 
 1. **Create a new codespace** via https://codespaces.new/withastro/starlight
 
-2. If running the docs site, pass the `--host` flag to avoid “502 Bad Gateway” errors:
+2. **Run the dev server** to generate types. Pass the `--host` flag to avoid “502 Bad Gateway” errors:
```
HiDeoo: If we add an extra step to this section dedicated to types before this step, we should be able to revert these changes and keep this step focused on the `--host` requirement when and if running the docs site.

```suggestion
2. If running the docs site, pass the `--host` flag to avoid “502 Bad Gateway” errors:
```

HiDeoo (commented):
Thanks for spinning this pull request :raised_hands:

I left a few suggestions that came to my mind while reading the changes, curious to hear everyone's thoughts or alternative ideas!

Regarding Gitpod, I went through the steps. It looks like the product now goes by different names: Ona or GitPod Flex. While the URL we provide still works, the process now requires to give more GitHub permissions than before (e.g. changing workflows). After granting all these permissions in a multiple step flow, I was finally asked for my credit card details to "verify" something. 

image

Clicking "Not now" just halts the process.

Considering all this, I think we should just remove this entire section about Gitpod (note that [this line](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md?plain=1#L41) should be edited to remove the Gitpod reference too).

CONTRIBUTING.md
erbierc: Both changes were added!

erbierc:
I have removed the GitPod info and added the requested changes.

HiDeoo (approved):
Thanks for updating the PR, looks really great to me now :star2: Ready for new contributors to follow the steps more easily!

PR #3496, opened 2025-10-22
Don’t generate invalid meta tags in `<head>`

delucis:
This is a follow-up PR to #3420 which was merged into this intermediate branch to allow @erbierc to pick up work on adding tests.

Opening as a draft for now until we have those tests!

### DIFF
--- a/packages/starlight/__tests__/test-utils.ts
+++ b/packages/starlight/__tests__/test-utils.ts
@@ -103,7 +103,15 @@ export async function mockedCollectionConfig(docsUserSchema?: Parameters<typeof 
 }
 
 type RouteDataTestContextOptions = {
+	/**
+	 * The pathname to get route data context for, e.g. `"/getting-started/"`.
+	 * @default "/"
+	 */
 	pathname?: string,
+	/**
+	 * Whether or not the context should include a value for `site`. Set to `false` to test without a `site` value.
+	 * @default true
+	 */
 	setSite?: boolean
 }
 


packages/starlight/__tests__/head/head.test.ts
```diff
@@ -188,7 +189,22 @@ test('places the default favicon below any user provided icons', () => {
...
+	const head = getTestHead({setSite: false});
+
+	const ogUrlExists = head.some((tag) => tag.tag === 'meta' && tag.attrs?.property === 'og:url');
+
+	expect(ogUrlExists).toBe(false);
+});
+
+type GetTestHeadOptions = {
+	heads?: HeadConfig,
+	route?: Route | undefined,
```
HiDeoo:
```suggestion
	route?: Route,
```

We always expect a route so we might as well properly reflect that in the type.

packages/starlight/__tests__/head/head.test.ts
```diff
@@ -44,7 +45,7 @@ test('includes description based on Starlight `description` configuration', () =
 });
 
 test('includes description based on page `description` frontmatter field if provided', () => {
-	const head = getTestHead([], routes[1]);
+	const head = getTestHead({ heads: [], route: routes[1] });
```
HiDeoo:
```suggestion
	const head = getTestHead({ heads: [], route: routes[1]! });
```

And to go alongside the type change, we can use the non-null assertion operator here like we do in other places or for the default value, as we know this route is mocked.

packages/starlight/__tests__/test-utils.ts
```diff
@@ -102,11 +102,17 @@ export async function mockedCollectionConfig(docsUserSchema?: Parameters<typeof
 	};
 }
 
-export function getRouteDataTestContext(pathname?: string): RouteDataContext {
+type RouteDataTestContextOptions = {
+	pathname?: string,
+	setSite?: boolean
```
HiDeoo, line 115: Nit: not sure if it's just me or not, but with a parameter name `setSomething` I kinda expect to pass down a function. Could it be worth accepting instead an optional `site` property that can be `URL | false`? Or maybe just a different name?

delucis: Ah interesting, it didn’t suggest that to me, although we could also add a JSDoc comment to communicate its purpose potentially.

A `site` property is definitely also an option, although I guess a bit weird given we wouldn’t actually use the URL form anywhere currently?

HiDeoo: Ah yeah, a comment explaining the use-case would also do the trick for tests as I was initially confused about it when encountering it during the review 👍

delucis, line 115:
```suggestion
	/**
	 * The pathname to get route data context for, e.g. `"/getting-started/"`.
	 * @default "/"
	 */
	pathname?: string,
	/**
	 * Whether or not the context should include a value for `site`. Set to `false` to test without a `site` value.
	 * @default true
	 */
	setSite?: boolean
```

delucis: Done in fb12d2b5b679729e5a8659c905b20c1032811730

Thanks for pointing this out.

PR #3500, opened 2025-10-23
Update release workflow for OIDC

delucis:
#### Description

- This PR updates our workflow for releasing packages to npm to use OIDC trusted publishing
- While there I also locked down some security by removing some permissions that weren’t required
- The weird `NPM_TOKEN: ""` line is a workaround for changesets to support OIDC publishing


HiDeoo (approved):
Looks perfect to me, and love the global patch bump to verify/notify about the change 👍 

Can be merged whenever the npmjs side of the changes are done.

PR #3501, opened 2025-10-23
Add testing to the "don’t generate invalid <link rel="canonical"/> with missing href" solution

erbierc:
#### Description

Continuation of #3420 and #3496
I have modified the `getRouteDataTestContext()` function to make sure I can get test context without site set, same with `getTestHead()`. The new parameter is optional, so it does not break any other usage of the functions.
I also wrote a test checking that `<meta property="og:url"/>` with missing content is not generated. The test passes.

I wanted to write one more test checking `<link rel="canonical"/>` in the same way (if site is not set, this element should not appear). However, `getRouteDataTestContext()` has a return type of

```typescript
type RouteDataContext = {
    generator: string;
    site: URL | undefined;
    url: AstroSharedContext["url"];
}
```
or, more specifically:
```typescript
type RouteDataContext = Pick<APIContext, 'generator' | 'site' | 'url'>;
```
`url` is required, and that url actually sets `<link rel="canonical"/>` to it, from what I logged in the console.

I would kindly ask you to check if my test logic is alright, as I have not done a lot of testing before and would appreciate any help with figuring out what to do to be able to write the next test. I don't really want to mess with existing types and I don't know the codebase well enough yet to come up with something new.

### DIFF
--- a/packages/starlight/__tests__/basics/omit-canonical.test.ts
+++ b/packages/starlight/__tests__/basics/omit-canonical.test.ts
@@ -1,0 +1,32 @@
+import { expect, test, vi } from 'vitest';
+import { getRouteDataTestContext } from '../test-utils';
+import { generateRouteData } from '../../utils/routing/data';
+import { routes } from '../../utils/routing';
+
+vi.mock('astro:content', async () =>
+	(await import('../test-utils')).mockedAstroContent({
+		docs: [
+			['index.mdx', { title: 'Home Page' }],
+			[
+				'environmental-impact.md',
+				{
+					title: 'Eco-friendly docs',
+					description:
+						'Learn how Starlight can help you build greener documentation sites and reduce your carbon footprint.',
+				},
+			],
+		],
+	})
+);
+
+test('omits link canonical tag when site is not set', () => {
+	const route = routes[0]!;
+	const { head } = generateRouteData({
+		props: { ...route, headings: [] },
+		context: getRouteDataTestContext(undefined, false),
+	});
+
+	const canonicalExists = head.some((tag) => tag.tag === 'link' && tag.attrs?.rel === 'canonical');
+
+	expect(canonicalExists).toBe(false);
+});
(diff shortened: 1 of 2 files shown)


erbierc:
Ok, I had some git confusion there. Added the failing test!

erbierc:
Btw I just wanted to add that I tested this manually by packing my local astro version (with the canonical changes) and comparing it to the current release, the invalid canonical and og:url did, in fact, disappear

erbierc:
@delucis Could you please take a look at [vitest.config.ts](https://github.com/withastro/starlight/blob/stack/invalid-canonical/packages/starlight/__tests__/head/vitest.config.ts) for head tests? I think this is the reason why the test fails. We keep getting
```
{
    tag: 'link',
    attrs: { rel: 'canonical', href: 'https://example.com/test' }
  },
  ```
  in the head, no matter the test setup, because config has it by default. 

I noticed that because `getRouteDataTestContext()` sets the URL to `https://example.com`, not `https://example.com/test`, and yet the `test` part kept appearing in logs. That might be it! :D

delucis:
> @delucis Could you please take a look at [vitest.config.ts](https://github.com/withastro/starlight/blob/stack/invalid-canonical/packages/starlight/__tests__/head/vitest.config.ts?rgh-link-date=2025-10-31T21%3A36%3A55Z) for head tests? I think this is the reason why the test fails

Ah yes, that makes sense! We were already testing that canonicals get deduplicated, so we set that in the test config to make sure the user version is used.

I guess in this case we can instead add the new test to our “Basics” test folder. Those tests run with a minimal config, so should work in this case, I think?

Thanks for figuring that out!

erbierc:
@delucis I figured it out, the test is passing! If everything is okay we could merge it and then I would like to work on making the helper functions look nicer (add the options object) in another PR.

delucis (approved):
Wonderful — this looks great! Thanks for the determination seeing this through @erbierc 👏

PR #3504, opened 2025-10-24
docs(plugin): starlight-docsearch-typesense

haydenhoang:
#### Description

This PR adds the [Starlight DocSearch Typesense plugin](https://github.com/typesense/starlight-docsearch-typesense) to the "Site search - Alternative search providers" documentation and plugin showcase.

The plugin lets you use Typesense as the backend for the DocSearch interface in Starlight, it works just like Algolia DocSearch.


docs/src/content/docs/guides/site-search.mdx
```diff
@@ -232,3 +232,34 @@ Add translations of the modal UI for your language using Starlight’s built-in
...
+           typesenseServerConfig: {
+             nodes: [{ url: 'TYPESENSE_SERVER_URL' }],
+             apiKey: 'TYPESENSE_SEARCH_ONLY_API_KEY',
+           },
+         }),
+       ],
+    }),
+  ],
+})
+```
```
HiDeoo: Thanks for the great new addition to the plugin showcase :star2:

The current [Site Search guide](https://starlight.astro.build/guides/site-search/#algolia-docsearch), and specifically the "Alternative search providers" section is right now mostly structured around 1 plugin, which we officially maintain, document, and support. While this is the obvious place to mention other search providers, I think there are a few points worth considering:

- Just like we mention for the DocSearch plugin that it's official, we should mention that this is a community plugin.
- I'm not sure we should include any code/configuration examples for community plugins:
  - We do it for the DocSearch plugin because it's where we document it, but community plugins would already have their own documentation for that.
  - We would need to keep those examples (and any translations) up to date with any changes to the plugin.
  - We could get technical questions about such configuration code example that we might not be able to answer.

Maybe something like this:

```suggestion
The [Starlight DocSearch Typesense](https://starlight-docsearch-typesense-docs.vercel.app/) community plugin provides a self-hosted and open-source alternative which integrates the [DocSearch](https://docsearch.algolia.com/) interface with a [Typesense](https://typesense.org/) backend.

Check the [“Getting Started”](https://starlight-docsearch-typesense-docs.vercel.app/getting-started/) guide in the Starlight DocSearch Typesense plugin documentation to use it instead of Pagefind.
```

Curious to hear everyone's thoughts.

haydenhoang:
Good points! I updated the Site search guide.

docs/src/content/docs/guides/site-search.mdx
```diff
@@ -232,3 +232,9 @@ Add translations of the modal UI for your language using Starlight’s built-in
    ```
 
 </Steps>
+
+### Typesense DocSearch
```
HiDeoo: We discussed the change during our public Talking & Doc'ing session on [Discord](https://astro.build/chat/) yesterday and we decided to go with an extra heading to highlight community plugins for search providers. Allows for a bit more detail and also the ability to link to an entire section containing all of them.

```suggestion
### Community search providers

[Community plugins](/resources/plugins/#community-plugins) can also provide alternatives to the built-in search provider.
Replace Starlight’s default Pagefind search with one of the following community-developed options.

#### Typesense DocSearch
```

PR #3508, opened 2025-10-31
[i18nIgnore] Match title case convention for components/using-components

hudlow:
All other page titles use title case:

image


delucis (approved):
Well spotted @hudlow! Thanks for the update 🌟

PR #3519, opened 2025-11-03
i18n(de): update translation resources/plugins.mdx

trueberryless:
#### Description
Updates German translation of `resources/plugins.mdx` with changes from #3515 and #3504


delucis (approved):
Looks great! Thanks for the speedy update @trueberryless 💨

PR #3507, opened 2025-10-30
Tweak a11y e2e tests

HiDeoo:
#### Description

As we discussed, this PR tweaks the a11y e2e tests. At the moment, we process the sitemap after building the docs site, filter out some slugs, and run the a11y tests against the remaining URLs for 2 locales (English and Japanese).

We don't get a lot of value from testing all the same pages in both English and Japanese, even more so many of them. This PR refactors the test utils to be more i18n-aware:

- For the default locale, we test all pages except for some excluded slugs.
- For non-default locales, they are excluded by default and we can specify which slugs to include for testing for a specific locale (this would allow us to include more pages if we ever get an RTL locale, for example).

Here are a comparison of the test runs before and after this change:

|                                          | Before                        | After                        |
| ---------------------------------------- | ----------------------------- | ---------------------------- |
| Number of page tested                    | 48 (24 English - 24 Japanase) | 24 (22 English - 2 Japanase) |
| Hyperfine benchmark (min … max) (3 runs) | 94.062 s … 96.962 s           | 64.291 s … 66.667 s          |

<details>
<summary>List of pages tested before the change</summary>

- `/`
- `/components/asides/`
- `/components/badges/`
- `/components/card-grids/`
- `/components/cards/`
- `/components/code/`
- `/components/file-tree/`
- `/components/icons/`
- `/components/link-buttons/`
- `/components/link-cards/`
- `/components/steps/`
- `/components/tabs/`
- `/environmental-impact/`
- `/guides/authoring-content/`
- `/guides/css-and-tailwind/`
- `/guides/route-data/`
- `/guides/sidebar/`
- `/ja/`
- `/ja/components/asides/`
- `/ja/components/badges/`
- `/ja/components/card-grids/`
- `/ja/components/cards/`
- `/ja/components/code/`
- `/ja/components/file-tree/`
- `/ja/components/icons/`
- `/ja/components/link-buttons/`
- `/ja/components/link-cards/`
- `/ja/components/steps/`
- `/ja/components/tabs/`
- `/ja/environmental-impact/`
- `/ja/guides/authoring-content/`
- `/ja/guides/css-and-tailwind/`
- `/ja/guides/route-data/`
- `/ja/guides/sidebar/`
- `/ja/reference/configuration/`
- `/ja/reference/icons/`
- `/ja/reference/route-data/`
- `/ja/resources/community-content/`
- `/ja/resources/plugins/`
- `/ja/resources/showcase/`
- `/ja/resources/themes/`
[...]

PR #3520, opened 2025-11-03
Skip a11y CI checks for docs changes that aren’t tested anyway

delucis:
#### Description

- Follow-up #3507
- I realised we could just skip the a11y CI job early for docs changes where content changes in languages we’re not running the a11y checks for. This PR updates the workflow file to check if Japanese and English pages changed (or if any non-content files in the docs package changed) and only runs the a11y job if they have.
- The filter for English content excludes all content files with a `xx` or `xx-xx` initial path segment. This could in theory break if we ever had an English `to-do` folder or something, but I think that’s not super likely and we can deal with it if it does come up.

### DIFF
--- a/docs/__a11y__/test-utils.ts
+++ b/docs/__a11y__/test-utils.ts
@@ -43,6 +43,8 @@ const config: Config = {
 		],
 		// Locale-specific included slugs (non-default locale slugs are excluded by default).
 		locales: {
+			// N.B. If adding more locales here, also update the changed files filters in
+			// `.github/workflows/ci.yml` to ensure tests run when files for those locales change.
 			ja: ['guides/route-data', 'reference/frontmatter'],
 		},
 	},


.github/workflows/ci.yml
```diff
@@ -113,6 +121,8 @@ jobs:
 
   a11y:
     name: Check for accessibility issues
+    needs: changes
+    if: ${{ needs.changes.outputs.docs_not_content == 'true' || needs.changes.outputs.docs_ja == 'true' || needs.changes.outputs.docs_en == 'true' || needs.changes.outputs.ci == 'true' || needs.changes.outputs.packages == 'true' }}
```
HiDeoo, line 128: Maybe it could be worth to add a comment after this one https://github.com/withastro/starlight/blob/main/docs/__a11y__/test-utils.ts#L44 in `docs/__a11y__/test-utils.ts` mentioning that if we ever add new locales, we should also update the logic in `.github/workflows/ci.yml`?

delucis: Good idea — done in fb005a084802f58d2b2e5e06f67529d653f37349 👍

PR #3521, opened 2025-11-03
Tabs: Remove margin-bottom being used to hide border.

shubham-padia:
#### Description

We were setting margin-bottom as -2px when at the same time we had a bottom border of 2px. Using this indirect margin was causing the tab container to be greater in height than it's wrapper when we zoomed out from the default zoom level. This introduced an unnecessary scroll bar.
| Zoom Level | Before | After |
|-------------|---------|--------|
| **90%** | Screenshot 2025-11-03 at 4 01 35 PM | Screenshot 2025-11-03 at 4 01 43 PM |
| **100%** | Screenshot 2025-11-03 at 3 53 30 PM | Screenshot 2025-11-03 at 3 53 57 PM |

I'm not sure if this PR warrants a minor version bump or not, let me know if it does.


shubham-padia:
Updated the PR @delucis ! Let me know if we want this to be part of a patch release instead of minor.

.changeset/fresh-horses-burn.md
```diff
@@ -0,0 +1,24 @@
...
+
+starlight-tabs .tab > [role='tab'] {
+  border-bottom: 2px solid var(--sl-color-gray-5);
+}
+
+starlight-tabs .tab [role='tab'][aria-selected='true'] {
+  border-color: var(--sl-color-text-accent);
+  box-shadow: none;
+}
+```
```
delucis: This looks great @shubham-padia — just suggesting a few tweaks for the changelog:

````suggestion
Fixes an issue where a vertical scrollbar could be displayed on the Starlight `<Tabs>` component when zooming the page

⚠️ **Potentially breaking change:** The `<Tabs>` component no longer uses `margin-bottom` and `border-bottom` to highlight the current tab. This is now done with a `box-shadow`. If you have custom styling for your tabs, you may need to update it.

If you want to preserve the previous styling, you can add the following custom CSS to your site:

```
starlight-tabs .tab {
  margin-bottom: -2px;
}

starlight-tabs .tab > [role='tab'] {
  border-bottom: 2px solid var(--sl-color-gray-5);
}

starlight-tabs .tab [role='tab'][aria-selected='true'] {
  border-color: var(--sl-color-text-accent);
  box-shadow: none;
}
```
````

**Review by HiDeoo — COMMENTED** [review]
Removing the border on `.tab > [role='tab']` would end up with a different behavior when the tab list is scrollable.

| Before | After |
| -------- | -------- |
| SCR-20251104-joog | SCR-20251104-jotp |

I guess this may be fixable with a box-shadow instead of removing the border, but didn't investigate further yet.

**Comment by delucis** [comment]
> I guess this may be fixable with a box-shadow instead of removing the border, but didn't investigate further yet.

Ah, good catch. We probably need the `box-shadow` on `[role=tab]` as well to avoid this.

File: packages/starlight/user-components/Tabs.astro [hunk 2]
```diff
@@ -125,14 +124,13 @@ if (isSynced) {
 			line-height: var(--sl-line-height-headings);
 			padding: 0.275rem 1.25rem;
 			text-decoration: none;
```
delucis, line 127:
```suggestion
--sl-tab-color-border: var(--sl-color-gray-5);
display: flex;
align-items: center;
gap: 0.5rem;
line-height: var(--sl-line-height-headings);
padding: 0.275rem 1.25rem;
text-decoration: none;
box-shadow: 0 2px 0 var(--sl-tab-color-border);
```

packages/starlight/user-components/Tabs.astro
```diff
@@ -125,14 +124,13 @@ if (isSynced) {
...
-			border-bottom: 2px solid var(--sl-color-gray-5);
 			color: var(--sl-color-gray-3);
 			outline-offset: var(--sl-outline-offset-inside);
 			overflow-wrap: initial;
 		}
 		.tab [role='tab'][aria-selected='true'] {
 			color: var(--sl-color-white);
```
delucis, line 135:
```suggestion
			--sl-tab-color-border: var(--sl-color-text-accent);
			color: var(--sl-color-white);
```

packages/starlight/user-components/Tabs.astro
```diff
@@ -125,14 +124,13 @@ if (isSynced) {
...
-			border-color: var(--sl-color-text-accent);
+			box-shadow: 0 2px 0 var(--sl-color-text-accent);
```
delucis:
```suggestion
```

packages/starlight/user-components/Tabs.astro
HiDeoo: Yeah, that is pretty much what I had in mind. Just tested locally and this works as expected (just took the liberty of editing this suggestion as it was suggesting to replace the `color: var(--sl-color-white);` line instead of appending after it.

shubham-padia:
Sorry for the delay, made the requested changes!

Issue #1223, opened 2023-12-04
Sidebar folders order is 


cbontems:
### What version of `starlight` are you using?

0.14.0

### What version of `astro` are you using?

3.6.4

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

As discussed on Discord,
```
Structure:
 - docs/
    - category/
      - folder-1/
        - page-1.md
        - page-2.md
      - folder-2/
        - page-1.md
        - page-2.md
```

If I give my folder-1/page-1.md an order property of 10, and my folder-2/page-1.md an order property of 5, I end-up with my folder-2 being ranked above folder-1 in my sidebar.

Said differently, autogenerated folders are sorted according to the order of the first page in the directory.

In the stackblitz example linked below, in the fruits directory, the banana subdirectory comes before the apples subdirectory because it has a page with order 1, when apples has a page with order 5.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-dssfwv?file=src%2Fcontent%2Fdocs%2Ffruits%2Fbananas%2Flatundan.md

### Participation

- [X] I am willing to submit a pull request for this issue.

cbontems:
I just started looking at a potential solution, but found this comment:

```astro
/**
 * Get the sort weight for a given route or directory. Lower numbers rank higher.
 * Directories have the weight of the lowest weighted route they contain.
 */
 function getOrder(routeOrDir: Route | Dir): number {
```
for the `getOrder`function in `utils/navigation.ts` so it looks like it is the expected behaviour for now.

This still look like a problem for me, as I don't want to check potential impact on other folders when I define order between pages inside a given folder.

stavros-k:
In addition, nested directories cannot have custom labels.
eg a dir named `my-folder` cannot be shown as `My Folder` unless you manually add the links.

My suggestion would be to have something simillar on how docusaurus do.
- index.md(x) with frontmatter. which will also make the directory, a page 
  so clicking on it on the sidebar willnot just collapse or un-collapse the tree, but show an actual page.
- A `_config.yaml` file (docusaurus has it as `_category_.yml`) which you can just add front-matter data, like title,order etc

apfelbox:
I am also highlight interested in both issues:

- Giving proper names to directories
- defining the order of folders and files in a directory, without impacting files inside these directories / other directories.

Is there already a solution for that?
[...]

PR #3525, opened 2025-11-04
feat: add some of trueberryless' new Starlight blogs to community content

trueberryless:
#### Description

I have added some more or less interesting blog posts to my blog and noticed that some of them are not yet listed on the awesome community content site. So I decided to create this PR adding them in chronological order 🙌

### DIFF
--- a/docs/src/content/docs/fr/resources/themes.mdx
+++ b/docs/src/content/docs/fr/resources/themes.mdx
@@ -18,6 +18,8 @@ Un thème est un module d'extension Starlight qui modifie l'apparence visuelle d
 
 Installez un thème créé par la communauté pour personnaliser rapidement l'apparence de votre site.
 
+Découvrez ci-dessous un aperçu de tous les thèmes ou essayez-les de manière interactive sur le [site de démonstration de Starlight Themes](https://starlight-themes.netlify.app/).
+
 <ThemeGrid
 	labels={{
 		/** Accessible label for the theme toggle. */


docs/src/content/docs/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Explore community-produced content maintained by Starlight users:
 		title="Add a reading progress indicator"
 		description="Find out how to add a horizontal reading progress bar to your Starlight site."
 	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-sidebar-whitespace/"
+		title="Make your Starlight sidebar items look better"
```
delucis: I wonder if there’s a more helpful way to describe this? “better” is quite vague and subjective and I’m not sure this title helps people know whether this would be useful for them to read. (i.e. of course, everyone would prefer something that looks better, but what actually changes? why would I want this? why isn’t this the Starlight default? etc.)

Maybe something like “How to apply custom styles to your Starlight sidebar” would be another way to frame it?

```suggestion
		title="Apply custom styles to your Starlight sidebar"
```

docs/src/content/docs/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Explore community-produced content maintained by Starlight users:
...
+		description="Using a little bit of CSS to adjust fonts, colours and more of the Starlight sidebar."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-dropdown-and-list-together/"
+		title="Starlight Topics Dropdown on mobile, List on desktop"
+		description="Combine two plugins to have the best of both on desktop and mobile devices."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-autogenerate-whole-sidebar/"
+		title="Convention for better Starlight sidebars with middleware"
```
delucis: Same thought here! What’s “better”? I also notice it’s not the title you went for on the site itself.

Maybe it could be something like this?

```suggestion
		title="Processing Starlight sidebars with middleware"
```

docs/src/content/docs/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Explore community-produced content maintained by Starlight users:
...
+		description="Use middleware and your own convention to have better control of the order of sidebar items and group labels."
```
delucis: I’m not very clear what “convention” means in this context, so maybe it’s not helping get the message across. Could just be edited down to something like this?
```suggestion
		description="Use middleware to take control of the order of sidebar items and group labels."
```

trueberryless:
Wonderful. I myself had a hard time finding the right words, your suggestions all sound much _better_. I'm gonna apply them later.

docs/src/content/docs/resources/community-content.mdx
trueberryless: Yeah, I didn't want to copy and paste the longer titles from the actual blog to this title, as I think they should not be as "catchy" as on my blog. So I decided to go for a little more "summary" and compact version. Processing feels good, I see however that "convention" was also removed in the description, it was supposed to convey the "opinionated" version, that's what I meant with convention. So I think I'll rewrite your description suggestion to include "opinionated" if that's more clear 👍

docs/src/content/docs/resources/community-content.mdx
trueberryless: What I meant with "convention" was like that the blog introduces a very opinionated folder structure, copied from [Nuxt Content](https://v2.content.nuxt.com/usage/content-directory#ordering). Maybe it's worth also writing this inspiration into the description (Nuxt Content), so people looking for it can find it 🤔 

Something like this came to my mind here:

```suggestion
		description="An opinionated middleware approach to ordering sidebar items and grouping labels, inspired by Nuxt Content’s conventions."
```

Maybe we could rewrite it, that this is Nuxt's "ordering" feature, but I didn't want to repeat the word... 🤔 Very open for other suggestions here 🙌

trueberryless: For even more context, the idea for this blog came from this comment in the Starlight issue: https://github.com/withastro/starlight/issues/1223#issuecomment-2371621344

delucis: Ahh, thanks for the explanation! I think maybe this would work well:

```suggestion
		description="Use route middleware to control sidebar ordering and group labels, inspired by Nuxt Content’s conventions."
```

trueberryless:
Awesome wording! Ready for release 🎸

PR #3528, opened 2025-11-04
Fixing typos

deining:
This PR fixes various typos I spotted in the project.


delucis (approved):
Thanks for the fixes @deining! I double checked the syntax highlighting themes fix didn’t have any unwanted consequences, but looks safe to me.

Normally we add a changelog entry when modifying code in `packages/`, but given this should all have no end-user impact for the most part, I think we can go ahead and merge this without.

PR #3529, opened 2025-11-04
i18n(de): update community-content.mdx

trueberryless:
#### Description

Update German translation of `resources/community-content.mdx` with changes from #3525

### DIFF
--- a/docs/src/content/docs/fr/resources/community-content.mdx
+++ b/docs/src/content/docs/fr/resources/community-content.mdx
@@ -119,6 +119,21 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 		title="Ajouter un indicateur de progression de lecture"
 		description="Découvrez comment ajouter une barre de progression de lecture horizontale à votre site Starlight."
 	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-sidebar-whitespace/"
+		title="Appliquer des styles personnalisés à votre barre latérale Starlight"
+		description="En utilisant un peu de CSS pour ajuster les polices, les couleurs et d'autres éléments de la barre latérale de Starlight."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-dropdown-and-list-together/"
+		title="Starlight Topics : menu déroulant de rubriques sur mobile, liste sur ordinateur"
+		description="Combinez deux modules d'extension pour profiter du meilleur des deux mondes sur ordinateur et appareils mobiles."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-autogenerate-whole-sidebar/"
+		title="Agencement de la barre latérale de Starlight avec un middleware"
+		description="Utilisez un middleware de routage pour contrôler l'agencement de la barre latérale et les étiquettes de groupe, inspiré des conventions de Nuxt Content."
+	/>
 </CardGrid>
 
 ## Contenu vidéo
--- a/docs/src/content/docs/ko/guides/site-search.mdx
+++ b/docs/src/content/docs/ko/guides/site-search.mdx
@@ -232,3 +232,13 @@ Starlight에 내장된 [국제화 시스템](/ko/guides/i18n/#starlight-ui-번�
    ```
 
 </Steps>
+
+### 커뮤니티 검색 공급자
+
+[커뮤니티 플러그인](/ko/resources/plugins/#커뮤니티-플러그인)은 Starlight의 기본 Pagefind 검색 공급자에 대한 대안을 제공할 수도 있습니다.
+
+#### Typesense DocSearch
+
+[Starlight DocSearch Typesense](https://starlight-docsearch.typesense.org/) 커뮤니티 플러그인은 [DocSearch](https://github.com/typesense/typesense-docsearch.js) 인터페이스와 [Typesense](https://typesense.org/) 백엔드를 통합하는 자체 호스팅 가능한 오픈 소스 대안을 제공합니다.
+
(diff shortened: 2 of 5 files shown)


docs/src/content/docs/de/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Erkunde die von der Community erstellten Inhalte, die von Starlight-Benutzern ge
 		title="Lesefortschrittsanzeige hinzufügen"
 		description="Erfahre, wie du eine horizontale Lesefortschrittsanzeige zu deiner Starlight-Website hinzufügen kannst."
 	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-sidebar-whitespace/"
+		title="Verpass deiner Starlight-Seitenleiste eigene Stile"
```
delucis: Is this translation correct? I’ve never seen this use of “verpassen” before.

Maybe? “Verwende benutzerdefinierte Stile in deiner Starlight-Seitenleiste”? Or “Deine Starlight-Seitenleiste anpassen mit benutzerdefinierte Stile”?

deining: May I step in, I'm native German speaker:

> Is this translation correct? I’ve never seen this use of “verpassen” before.

Yes it is.

See https://de.wiktionary.org/wiki/verpassen:
[5] eine Eigenschaft (meist optisches Erscheinungsbild) ändern

I don't really like this translation, though.

**Original:** `Apply custom styles to your Starlight sidebar`
**My proposal:** `Passe das Erscheinungsbild deiner Starlight-Seitenleiste gemäß Deinen Vorstellungen an` 

If we want to stick close to the proposed translation, I would better use:
`Spendiere deiner Starlight-Seitenleiste ein eigenes Erscheinungsbild`.

Just my 2 cents, ...

> Maybe? “Verwende benutzerdefinierte Stile in deiner Starlight-Seitenleiste”? Or “Deine Starlight-Seitenleiste anpassen mit benutzerdefinierte Stile”?

If, then use `Deine Starlight-Seitenleiste mit benutzerdefinierten Stilen anpassen`, otherwise is grammatically incorrect. While not wrong, I don't really like any of the two proposals, though.

docs/src/content/docs/de/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Erkunde die von der Community erstellten Inhalte, die von Starlight-Benutzern ge
...
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-dropdown-and-list-together/"
+		title="Starlight-Themen Dropdown auf Mobilgeräten, Liste auf Desktop-Computern"
+		description="Kombiniere zwei Plugins, um das Beste aus beiden auf Desktop- und Mobilgeräten zu haben."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-autogenerate-whole-sidebar/"
+		title="Verarbeitung von Starlight-Seitenleisten mit Middleware"
+		description="Benutz Route-Middleware, um die Reihenfolge der Seitenleisten und Gruppenbeschriftungen zu steuern, inspiriert von den Konventionen von Nuxt Content."
```
deining: better: `Benutze ..`

docs/src/content/docs/de/resources/community-content.mdx
```diff
@@ -119,6 +119,21 @@ Erkunde die von der Community erstellten Inhalte, die von Starlight-Benutzern ge
...
 	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-sidebar-whitespace/"
+		title="Verpass deiner Starlight-Seitenleiste eigene Stile"
+		description="Mit ein bisschen CSS kannst du Schriftarten, Farben und mehr in der Starlight-Seitenleiste anpassen."
+	/>
+	<LinkCard
+		href="https://blog.trueberryless.org/blog/starlight-dropdown-and-list-together/"
+		title="Starlight-Themen Dropdown auf Mobilgeräten, Liste auf Desktop-Computern"
+		description="Kombiniere zwei Plugins, um das Beste aus beiden auf Desktop- und Mobilgeräten zu haben."
```
deining: **Proposal:** maybe better use `zu erhalten` instead of `zu haben` here.

docs/src/content/docs/de/resources/community-content.mdx
trueberryless: Thanks for all those detailed reviews with suggestions!

> While not wrong, I don't really like any of the two proposals, though.

I'm personally curious why you dislike Chris' (delucis) first suggestion, I mean I would prefer I slight use of other words in this case, something like: `Verwende eigene Stile für deine Starlight-Seitenleiste`, but I think that would be fine.

But at the same time, I am also heavily leaning toward your suggestion, but the `"Deinen"` must be small since we use the informal form in the German translation consistently throughout Astro Docs and Starlight Docs (you already did it right with the first `"deiner"`, so: `Passe das Erscheinungsbild deiner Starlight-Seitenleiste gemäß deinen Vorstellungen an`.

I'll not directly adapt any suggestions as I am curious what @delucis would prefer in this case, and @deining feel free to share your thoughts as well 🙌

docs/src/content/docs/de/resources/community-content.mdx
trueberryless: Done ✅

docs/src/content/docs/de/resources/community-content.mdx
trueberryless: Done ✅

docs/src/content/docs/de/resources/community-content.mdx
deining: > something like: `Verwende eigene Stile für deine Starlight-Seitenleiste`, but I think that would be fine.

Yes, I like this translation, I'm fine with that suggestion.

trueberryless: Done ✅

delucis (approved):
Super! Thanks for the updates and reviews both of you 💖

PR #3534, opened 2025-11-05
Pagefind CLI → Pagefind API

### LINKED ISSUE #3503: Pagefind forces the use of npx to run which may not be available

0.36.1

5.14.6

pnpm

Linux

Chrome

### Describe the Bug

Hello,

When using pagefind (which is the default) the hook forces the use of `npx`. See https://github.com/withastro/starlight/blob/5018793b5c03d6d22eaec4c7174b5fe1c0aeace5/packages/starlight/index.ts#L162C6-L162C17 

I'm in a constrained environment (bazel) where I only have access to pnpm so it would be great to be able to choose the specific env package manager.

none

### Participation

HiDeoo:
#### Description

- Closes #3503

This PR refactors how Pagefind is run after the build process. Instead of using the CLI via `npx`, it now uses the [Pagefind API](https://pagefind.app/docs/node-api/) directly.

Feature-wise, there should be no changes but one aspect which will differ is logging.

| Before   | After    |
| -------- | -------- |
| SCR-20251105-oazh | SCR-20251105-ocrn |

We are now directly using a fork of the Starlight ingration logger. I didn't spend too much time on tweaking the logging output so far, but one huge difference is that we don't have access to the same informations when using the API vs CLI which access them using some internal Rust code and that data is not exposed via the API.

Personally, I never found the logging output of the CLI very useful, so I think this is an overall improvement but definitely something worth mentioning and discussing if needed.

#### Todo

- [x] Add a changeset. (Chris added a temporary one to test preview releases, but we may want to revisit the wording.)


Mikescops:
Hello, thanks a lot for the PR, is there a chance this goes into the next version? 😄

delucis:
> is there a chance this goes into the next version?

Yes, planned for v0.38!

Mikescops:
Nice, thanks a lot!

packages/starlight/integrations/pagefind.ts
```diff
@@ -0,0 +1,60 @@
...
+export async function starlightPagefind({
+	dir,
+	logger: starlightLogger,
+}: PagefindIntegrationOptions) {
+	const logger = starlightLogger.fork('starlight:pagefind');
+	const options = { dir, logger };
+
+	try {
+		const now = performance.now();
+		logger.info('Running Pagefind…');
```
delucis: Was just looking at the example logging and thought maybe we should be more explicit for users who aren’t yet familiar with “Pagefind”:

```suggestion
		logger.info('Building search index with Pagefind...');
```

(And switched `…` to `...`, which is a bit of a nit, but I find the ellipsis character tends to be a bit squashed in monospace fonts, so it’s better to spell it out, I think?)

packages/starlight/integrations/pagefind.ts
```diff
@@ -0,0 +1,60 @@
...
+		logger.info(`Found ${page_count} HTML files.`);
+
+		const writeFilesResponse = await index.writeFiles({
+			outputPath: fileURLToPath(new URL('./pagefind/', dir)),
+		});
+		assertPagefindResponse<pagefind.WriteFilesResponse>(writeFilesResponse, options);
+
+		const pagefindTime = performance.now() - now;
+		logger.info(
+			`Finished running Pagefind in ${pagefindTime < 750 ? `${Math.round(pagefindTime)}ms` : `${(pagefindTime / 1000).toFixed(2)}s`}.`
```
delucis: And similar here to convey the meaning more than the specific tool:

```suggestion
			`Finished building search index in ${pagefindTime < 750 ? `${Math.round(pagefindTime)}ms` : `${(pagefindTime / 1000).toFixed(2)}s`}.`
```

packages/starlight/integrations/pagefind.ts
```diff
@@ -0,0 +1,60 @@
...
+		);
+	} catch {
+		throw new Error('Failed to run Pagefind.');
```
delucis: Should we include the original error as a cause here? I noticed the error thrown in `assertPagefindResponse()` would always be caught here. That one probably doesn’t matter too much, because we already log each of the `response.errors`, but maybe for some other unexpected error, this could help with debugging?

```suggestion
	} catch (cause: unknown) {
		throw new Error('Failed to build search index.', { cause });
```

(Also updated the error message similar to the logging changes proposed above.)

.changeset/great-squids-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fixes support for running builds when `npx` is unavailable.
```
delucis, line 5: Not sure how much detail is helpful here, but including a proposal for some more for discussion. The main user-facing change is the logging, but I’m not sure it’s super important.

```suggestion
Fixes support for running builds when `npx` is unavailable.

Previously, Starlight would spawn a process to run the Pagefind search indexing binary using `npx`. On platforms where `npx` isn’t available, this could cause issues. Starlight now runs Pagefind using its Node.js API to avoid a separate process. As a side effect, you may notice that logging during builds is now less verbose.
```

delucis (approved):
Thanks for adding the tests @HiDeoo! I think this looks good to go.

Also, realised this is technically only a patch, so no need to wait for the next minor either 🎉

PR #3538, opened 2025-11-06
ci: use Firefox for Windows E2E tests

HiDeoo:
#### Description

Continuing on the effort to speed up CI, I [noticed](https://github.com/withastro/starlight/actions/runs/19137426946/job/54692987354?pr=3516#step:7:14) that on Windows, the E2E tests took 3 minutes to install some dependendcies before even starting to download Chrome.

After checking the Playwright core codebase, it turns out that this is due to the [installation of the Server Media Foundation windows feature](https://github.com/microsoft/playwright/blob/e7bff526433b6dcb02801763ab5b1c6407902d47/packages/playwright-core/bin/install_media_pack.ps1#L4) on Windows Server.

Altho, something I noticed is that this is [only required for Chromium-based browsers](https://github.com/microsoft/playwright/blob/e7bff526433b6dcb02801763ab5b1c6407902d47/packages/playwright-core/src/server/registry/dependencies.ts#L79-L89).

Therefore, this PR changes the Windows E2E tests on CI to use Firefox instead of Chrome, which skips this step and immediately starts downloading Firefox before running the tests.

| Before   | After    |
| -------- | -------- |
| SCR-20251106-pldu | SCR-20251106-pmaz |

This also has the added benefit of running our E2E tests on a different browser.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,9 @@ jobs:
       - name: Test packages (Linux)
         if: matrix.os == 'ubuntu-latest'
         run: pnpm -r test:e2e:chrome
+      # Firefox is used for Windows E2E tests as Chrome on Windows requires the installation of the
+      # Server Media Foundation windows feature, which can take up to 3 minutes to install on CI.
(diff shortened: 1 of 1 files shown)


delucis (approved):
Incredible investigation and discovery 🔥 

I think we’ll need a blog post soon about all these tricks!

Thanks @HiDeoo 🙌

PR #3537, opened 2025-11-06
just a little adjustment

kakrusliandika:
just a little adjustment
- Aligns `docs/src/content/docs/id/getting-started.mdx` with the latest EN version.
- Updates upgrade instructions to use `@astrojs/upgrade` (npm/pnpm/yarn).
- Adds an **Extend** bullet (Plugins & Themes) under “Next steps”.

PR #3540, opened 2025-11-08
updated

kakrusliandika:
updated

### DIFF
--- a/docs/src/content/docs/id/getting-started.mdx
+++ b/docs/src/content/docs/id/getting-started.mdx
@@ -102,7 +102,7 @@ Karena Starlight adalah perangkat lunak beta, akan ada pembaruan dan peningkatan
 Pastikan untuk memperbarui Starlight secara teratur!
 :::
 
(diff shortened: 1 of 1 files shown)


HiDeoo:
As I [pointed out](https://github.com/withastro/starlight/pull/3537#issuecomment-3498419926) last week in #3537, PRs that can conditionally use AI for translations are a problem for us, as it makes it very difficult to moderate, verify if AI was used or not, and add a lot of extra burden on maintainers. This PR seems to be identical to #3537, and just fix one of multiple examples I pointed out during my previous review but does not clarify if AI was initially used or not. I'll also add that https://github.com/withastro/docs/pull/12679 also seems to contain a few signs of AI usage, as [pointed out](https://github.com/withastro/docs/pull/12679#issuecomment-3496673854) by other maintainers.

With such uncertainty, I personally didn't feel confident merging the older PR, and same applies to this one.

Thanks for your understanding and time nonetheless!

PR #3510, opened 2025-10-31
docs: add starlight-themes link

HiDeoo:
#### Description

As discussed, this PR adds a link to the [Starlight Themes](https://starlight-themes.netlify.app/) website in the themes showcase as it provides an different way to explore all themes.

I tried to make it concise but happy to adjust the wording if needed.

PR #3541, opened 2025-11-08
i18n(de): update themes.mdx

trueberryless:
#### Description
Updates German translation of `resources/themes.mdx` with changes from #3510

> [!note]
> I have chosen to once translate "theme" to "Design" to not constantly repeat it, but we can also keep it consistent if desired.


delucis (approved):
LGTM! Thanks @trueberryless ✨

PR #3542, opened 2025-11-09
refactor(test utils): Add options object to test utility functions (getRouteDataTestContext, getTestHead) 

erbierc:
#### Description

- Closes #N/A - follow up to #3496 
In a previous PR, I have added tests to the invalid canonical changes. This required me to include additional arguments to getRouteDataTestContext and getTestHead functions. This refactor is meant to make these changes more readable and developer friendly by adding an options object to both of them. This PR also makes sure all usage of these functions are updated as well. As of now, all tests pass.

```typescript
type GetTestHeadOptions = {
	heads?: HeadConfig,
	route?: Route | undefined,
	setSite?: boolean
}
```
Here, route has an additonal `| undefined` because of the 'includes description based on page `description` frontmatter field if provided' test in the same file (TypeScript doesn't like it without that).
Other options object type added:
```typescript
type RouteDataTestContextOptions = {
	pathname?: string,
	setSite?: boolean
}
```

Note: I have placed the new types above these functions. If I need to move them, or if you would prefer interfaces instead, I'd really like to know!


erbierc:
Okay, maybe it was a bad idea to sync the branches before the PR! Here are my commits:
https://github.com/withastro/starlight/pull/3542/commits/3763950091fe5ce3c4792ee383c1f37bca3641dd
https://github.com/withastro/starlight/pull/3542/commits/7d5a003cb26641ac6a4afc58ed487ccfde7d4e42
https://github.com/withastro/starlight/pull/3542/commits/b4bee2390d9da7a04419aca84d2db61729f25c8c

PR #3544, opened 2025-11-09
 refactor(test utils): Add options object to test utility functions (getRouteDataTestContext, getTestHead)

erbierc:
Created a new PR, as the previous one (https://github.com/withastro/starlight/pull/3542) was synced with main, thus making reviewing it a pain. And the one after that, #3543 was set to main, not stack/invalid-canonical 🤡

#### Description

- Closes #N/A - follow up to #3496 
In a previous PR, I have added tests to the invalid canonical changes. This required me to include additional arguments to getRouteDataTestContext and getTestHead functions. This refactor is meant to make these changes more readable and developer friendly by adding an options object to both of them. This PR also makes sure all usage of these functions are updated as well. As of now, all tests pass.
```typescript
type GetTestHeadOptions = {
	heads?: HeadConfig,
	route?: Route | undefined,
	setSite?: boolean
}
```
Here, route has an additonal `| undefined` because of the 'includes description based on page description frontmatter field if provided' test in the same file (TypeScript doesn't like it without that).
Other options object type added:
```typescript
type RouteDataTestContextOptions = {
	pathname?: string,
	setSite?: boolean
}
```
Note: I have placed the new types above these functions. If I need to move them, or if you would prefer interfaces instead, I'd really like to know!

### DIFF
--- a/packages/starlight/__tests__/basics/omit-canonical.test.ts
+++ b/packages/starlight/__tests__/basics/omit-canonical.test.ts
@@ -23,7 +23,7 @@ test('omits link canonical tag when site is not set', () => {
 	const route = routes[0]!;
 	const { head } = generateRouteData({
 		props: { ...route, headings: [] },
-		context: getRouteDataTestContext({setSite: false}),
+		context: getRouteDataTestContext({ setSite: false }),
 	});
 
 	const canonicalExists = head.some((tag) => tag.tag === 'link' && tag.attrs?.rel === 'canonical');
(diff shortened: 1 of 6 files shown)


delucis (approved):
Lovely work @erbierc 👏 Appreciate how you laid out the idea and the work in the description — really helpful.

Btw, re: the second false-start PR — you can actually switch the target after opening a PR using the <kbd>Edit</kbd> button in the top right. Clicking that will make the PR title editable, but also makes it possible to switch the branch target just below there.

Anyway, I think this means we can merge this and then get that fix merged too 🎉

erbierc:
@delucis Good to know, thanks so much!

PR #3546, opened 2025-11-11
feat: clean up and add resources to community-content, plugins and themes

trueberryless:
#### Description

This PR adds some resources that have not been added yet to one of those three files:
- `community-content.mdx`
- `plugins.mdx`
- `themes.mdx`

Moreover, the PR moves around some existing resources to enhance the structure (this might be opinionated, so please ask me for reverts if necessary). In general: The articles and guides have been sorted by publishing date, so we can consistently continue to add new entries to the bottom as recommended.

Other cleanups:
- Make Theme names consistent
- Group videos into "official" and "community", like plugins and therefore move the "1 MIN Speedrun of creating new Starlight project" down to community.

This is the list of all added resources:
- YouTube Video of [sarah11918](https://github.com/sarah11918/) talking about Starlight (One year celebration): https://www.youtube.com/watch?v=j8K-ESJF814
- 2 new themes: [Starlight Gruvbox](https://starlight-theme-gruvbox.otterlord.dev/) by [TheOtterlord](https://github.com/TheOtterlord/) and [Starlight Six](https://six-tech.github.io/Six.StarlightTheme/) by [six-tech](https://github.com/six-tech)
- First article from [HiDeoo](https://github.com/HiDeoo/): [Contribute to Starlight](https://hideoo.dev/notes/contributing-to-starlight) (I think this article is valuable, teaching and in general an awesome read, so it's fits pretty well in that section)
- [All blog posts](https://hideoo.dev/notes) from [HiDeoo](https://github.com/HiDeoo/) related to Starlight cronologically that have not been added before (@HiDeoo please approve if you want these (and also your first note ⬆️) shared or not)
- Add new "Component" section to `plugins.mdx` for packages that simply provide Astro components that can be used specifically in Starlight projects MDX and Marcdoc. I only know of my components, if anyone knows other Starlight-specific Astro components, feel free to suggest them or add them later, we'll see.

I also moved the `starlight-sidebar-topics-dropdown` packages down to the new "Component" section as it's technically a component. I'm not sure if this results in any better discoverability of the package, personally I think it also perfectly fits besides its `peerDependency` `starlight-sidebar-topic` plugin where it was before. Open for any suggestions!


docs/src/content/docs/resources/themes.mdx
```diff
@@ -51,7 +51,7 @@ Preview a list of all themes below or try them out interactively on the [Starlig
 			previews: { light: 'catppuccin-light.png', dark: 'catppuccin-dark.png' },
 		},
 		{
-			title: 'Ion',
+			title: 'Starlight Ion',
```
delucis: This was named as it is on the demo site itself: https://louisescher.github.io/starlight-ion-theme/

I think “Ion” is fine here — there’s no need to add “Starlight” in each case.
```suggestion
			title: 'Ion',
```

docs/src/content/docs/resources/community-content.mdx
delucis, line 1: Could we just add the new cards to the bottom instead of re-ordering things here? I don’t think chronological by publish date is essential — we’re already able to add new links at the bottom even if technically some page was published before another.

docs/src/content/docs/resources/plugins.mdx
```diff
@@ -193,13 +188,46 @@ Extend your site with official plugins supported by the Starlight team and commu
...
+		title="starlight-save-file-component"
+		description="Quickly display a link to some download asset on your Starlight site."
+	/>
+	<LinkCard
+		href="https://github.com/trueberryless-org/starlight-contributor-list"
+		title="starlight-contributor-list"
+		description="Display a list of all contributors to your project."
+	/>
+</CardGrid>
+
```
delucis: Hah, a _bit_ cheeky to create a whole new section just for your own projects 😄 

I think these could just go under the existing “tools” section maybe? That’s already intended as a general bucket for non-plugin packages and some of those are also components like `starlight-package-managers`.

docs/src/content/docs/resources/themes.mdx
trueberryless: Hmm, I strive for consistency tbh, and it triggers the perfectionist inside me that this is the only theme not listed with Starlight. May I ping @louisescher for an opinion on this one? Maybe we could update the docs of Starlight Ion for a cleaner list 😉

docs/src/content/docs/resources/plugins.mdx
trueberryless: Ah, the tools section is a nice place to put component packages, true. I didn't even spot the manager package there. And yeah, it really felt weird having such a lonely section 😅

docs/src/content/docs/resources/themes.mdx
delucis: No, let’s just stick to “Ion”. It’s already a _perfectly_ beautiful name ;-)

louisescher: Ion is fine!

trueberryless: Thanks for the feedback!

docs/src/content/docs/resources/community-content.mdx
trueberryless: Not the biggest fan here as well, as it feels like a missed opportunity that would also enhance the "randomness" between blog author order, but okay...

trueberryless:
Closed by accident.

Now the question is: Do we wanna separate Sidebar Topics and Topics Dropdown just because the later is a component having a `peerDep` to the former one, or should I also revert so they stick together since there is no component section?

trueberryless:
> I think the current change works! It might actually be clearer, because I was confused when first checking out the topics dropdown expecting it to be a plugin — it wasn’t clear to me immediately that it wasn’t one. So maybe placing it in tools helps that?

Yeah, I agree. I was just pointing out due to discoverability. But since I also wrote a blog about combining them, I think there are enough places to get the info from 🙌

PR #3549, opened 2025-11-14
fix: Update fonts

ghost:
#### Description

Remove redundant vendor-specific fonts (ie. `-apple-system` and `BlinkMacSystemFont`)


delucis:
Could you explain why these fonts are “unnecessary”?

ghost:
The vendor-specific fonts (ie. `-apple-system` and `BlinkMacSystemFont`)  were necessary before `system-ui` had broad browser support, but are now redundant.

The `system-ui` generic font family is now supported in all modern browsers and automatically loads the OS default font. Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#system-ui):

> Glyphs are taken from the default user interface font on a given platform.

Removing them simplifies the font stack with zero functional impact, as `system-ui` handles all platforms uniformly.

delucis:
> The vendor-specific fonts (ie. `-apple-system` and `BlinkMacSystemFont`) were necessary before `system-ui` had broad browser support, but are now redundant.

That seems reasonable enough. I checked [Can I Use](https://caniuse.com/?search=system-ui) with Starlight’s [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+105%2C+Edge+%3E%3D+105%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+15.4%2C+iOS+%3E%3D+15.4%2C+not+op_mini+all) and it does looks like `system-ui` covers our browser support range.

However, this PR also seems to change other stuff. The named fonts are different from the current ones. Is that intentional?

ghost:
Yes, that's intentional. I replaced some of the named fonts with the "Neo-Grotesque" and "Monospace Code" stacks from [modernfontstacks](https://modernfontstacks.com/) for a cleaner cross-platform look. Also, I removed the Emoji fonts though I couldn't find out if `system-ui` covers those as well so might add them back just in case.

delucis:
I think it’s probably best to _only_ remove the vendor-prefixed ones for now. That seems safe, but changing the rest of the font stack risks changing how people’s sites look which could be surprising or disruptive to them, so in the absence of a strong reason to do that, I think I’d rather stick with what we have. (We have to keep in mind that there are 10,000+ Starlight sites out there, so prefer to avoid arbitrary changes without having a clear idea of the impact.)

delucis (approved):

PR #3550, opened 2025-11-14
feat: Export SidebarRestorePoint

fabien0102:
#### Description

This PR exposes `SidebarRestorePoint`, used by `SidebarSublist`, to make custom overrides easier.

### DIFF
--- a/packages/starlight/package.json
+++ b/packages/starlight/package.json
@@ -59,6 +59,10 @@
       "types": "./components/SidebarPersister.astro.tsx",
       "import": "./components/SidebarPersister.astro"
     },
+    "./components/SidebarRestorePoint.astro": {
+      "types": "./components/SidebarRestorePoint.astro.tsx",
+      "import": "./components/SidebarRestorePoint.astro"
+    },
     "./components/SidebarSublist.astro": {
       "types": "./components/SidebarSublist.astro.tsx",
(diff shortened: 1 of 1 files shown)


delucis:
In the case of this one, I’m really not confident about exporting it as it’s very specific to how Starlight’s own internals are wired up: the custom element logic is not even defined here, but inside the `<StarlightPersister>` component which is included in our `<Sidebar>` component.

I’m noticing now that we _did_ export `StarlightPersister` last year, but that kind of feels like a mistake to me too 😅 (I made this mistake myself, so I feel comfortable saying that.)

The issue we face is that we author and maintain these “internal” sidebar things with assumptions about what a sidebar structure looks like. Once people start using them with their own custom `<Sidebar>`, it’s very easy for our assumptions to no longer match their markup and for things to break in subtle or not-so-subtle ways. If we export and document them officially, that suggests we’re responsible for fixing and supporting all these possible user customisations, and I’m not sure we can guarantee that. It also means if we _change_ the internals, we need to be much more careful because these changes now could have an impact beyond what we’re able to easily test, which is not great for being able to maintain things confidently.

delucis:
Going to close this and the related PRs pending the active discussion in https://github.com/withastro/starlight/discussions/3411! We can always reopen when the time comes — thanks @fabien0102.

PR #3551, opened 2025-11-14
feat: Export `SidebarEntry` types

fabien0102:
#### Description

Export `SidebarEntry` and subtypes to allow easier overrides.

#### Notes

Currently, we can access those types like this:

```ts
import DefaultSidebarSublist from "@astrojs/starlight/components/SidebarSublist.astro";

type Props = Parameters<typeof DefaultSidebarSublist>[0];
type SidebarEntry = Props["sublist"][number];
type SidebarLink = Extract<SidebarEntry, { type: "link" }>;
```

after this being merge, this will be:

```ts
import type { SidebarEntry, SidebarLink } from "@astrojs/starlight/types"
```

### DIFF
--- a/packages/starlight/types.ts
+++ b/packages/starlight/types.ts
@@ -1,3 +1,4 @@
+export type { SidebarEntry, SidebarGroup, SidebarLink } from './utils/routing/types';
 export type { StarlightConfig } from './utils/user-config';
 export type {
 	StarlightPlugin,


HiDeoo:

This is definitely a topic that needs some attention as it has already come up a few times.

It's also something being discussed in this [discussion](https://github.com/withastro/starlight/discussions/3411), and considering such types are already accessible (even though it may not be obvious or the best experience), I think I'd like us to spend some time thinking about the best way to improve this use-case before jumping into implementation.

When it comes to exposing new types like this, something I [noted](https://github.com/withastro/starlight/discussions/3411#discussioncomment-14315751) and want to be careful about is which types should be exposed. Why such types should be exposed but not others. Should we expose all possible types? Should we only expose types that are commonly used (how do we define that?)? Or maybe only some "advanced" types? Something to also consider is that any type we expose becomes part of our public API, and also needs to be documented.

Maybe there are also other ways to improve the DX around this without exposing new types, maybe improving the documentation or providing examples.

Considering all this, I'll close this PR for now, but please let's continue the conversation in the discussion linked above. It's definitely not me saying "no" to this, just that I want us to think through the entire picture before moving forward, and making sure we do it in the best possible way.

fabien0102:
I didn’t see the discussion, thanks for pointing it out :) I totally understand!

PR #3552, opened 2025-11-14
feat: Export `flattenSidebar` function

fabien0102:
#### Description

Export `flattenSidebar` function, used by `<SidebarSublist>` for easier custom overrides.

### DIFF
--- a/packages/starlight/internal.ts
+++ b/packages/starlight/internal.ts
@@ -1,6 +1,10 @@
 /**
- * @file This file contains utility functions imported by the `<Code>` component.
+ * Utility functions imported by the `<Code>` component.
  */
-
 export { useTranslations } from './utils/translations';
 export { getStarlightEcConfigPreprocessor } from './integrations/expressive-code/preprocessor';
+
+/**
+ * Utility functions imported by the `<SidebarSublist>` component.
+ */
+export { flattenSidebar } from './utils/navigation';


HiDeoo:
Thanks for the contribution.

> Export `flattenSidebar` function, used by `<SidebarSublist>` for easier custom overrides.

Question about this PR and also #3550: `<SidebarSublist>` is not a component that can be overridden, and this function is definitely part of Starlight's internals as it's only meant to be used by that component right now. Would you be able to explain a bit more about your use case and the desired outcome?

fabien0102:
Sure, let me explain my problem. My main goal was to add a link at the directory level of the sidebar. So tweak this line https://github.com/withastro/starlight/blob/c8bbae8977a2e4b7be64d90d19f7f8ac21dd627b/packages/starlight/components/SidebarSublist.astro#L42 but keep as much as possible the existing behaviours.

To achieve this, since `SidebarSublist` is used by `Sidebar`, I had to copy and paste several elements to modify this behavior.

Multiple friction points:
* `MobileMenuFooter`, used in `Sidebar` is internal
* `SidebarRestorePoint`, used in `SidebarSublist` is internal
* `flattenSidebar`, used in `SidebarSublist` is internal

If I’m overriding the sidebar, it’s very likely that I need to use some of the existing logic part of `SidebarSubList`.

Another point not addressed yet, I would have loved to copy the style from `@astrojs/starlight/components/SidebarSublist.astro` to avoid duplicating.

I hope this helps understanding the usecase

delucis:
Taking a look at this, I think I’d prefer a different approach rather than exposing internals.

In general we’ve found it’s better for our components to contain as little business logic as possible and instead be pretty simple representations of the `locals.starlightRoute` data object we use for each page.

We haven’t really applied this principle as thoroughly to components like this because we consider them “internal” like @HiDeoo mentioned — the intention is for anyone who is reimplementing these to do so from scratch. But we probably should consider it, because like this PR shows, people reimplementing things are likely to have similar requirements.

So in this case I look at it and instead of thinking “`flattenSidebar` should be exported”, I think “there should be data equivalent to what `flattenSidebar()` achieves in the route data object”. That way neither our component, nor a user component needs to worry about computing it.

So in this case, I’d propose we add a `hasCurrent` or `containsCurrent` property to sidebar groups and then use those here if possible.

delucis:
Going to close this and the related PRs pending the active discussion in https://github.com/withastro/starlight/discussions/3411! We can always reopen when the time comes — thanks @fabien0102.

PR #3553, opened 2025-11-14
chore(docs): add modestbench.dev to showcase

boneskull:
#### Description

This PR adds [modestbench](https://modestbench.dev) to the site showcase.

Before:

image

After:

image


boneskull:
I've added a second commit here which adds [LavaMoat](https://lavamoat.github.io) to the showcase. I can open a second PR for this, but just didn't want to make conflicts with myself!

image

PR #3555, opened 2025-11-15
fix: customCSS catch and throw error

Its-Just-Nans:
Related to https://github.com/withastro/starlight/discussions/3554


delucis (approved):
Thanks again for the PR @Its-Just-Nans!

I just took the time to update it to run the validation alongside all our other user config validation logic and throw a clear error explaining how to resolve the issue.

Its-Just-Nans:
> Thanks again for the PR @Its-Just-Nans!
> 
> I just took the time to update it to run the validation alongside all our other user config validation logic and throw a clear error explaining how to resolve the issue.


Thanks to you for all your work.

Just a little question, the goal was also to help the user by saying "you can add css using the head configuration of starlight". Why not adding that part?

Thanks you

delucis:
> Why not adding that part?

That’s not the best approach: you really should use `src/` and let Starlight bundle CSS for you as it improves the site performance in various ways.

Of course, using `head` is still _possible_ in the few circumstances where `public/` is really needed, but in an error message like this we need to provide clear advice that’s as simple as possible to point users to what is the right choice in most cases so that they can resolve the error quickly.

PR #3556, opened 2025-11-15
feat: add isIndex fields

Its-Just-Nans:
This PR should allow the user to add a frontmatter filed `isIndex`.

Like this a filename test.md can be the index of the folder "test"
The URL will be example.com/guides/test

instead of
example.com/guides/test/test

### DIFF
--- a/.changeset/three-otters-rhyme.md
+++ b/.changeset/three-otters-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Add a isIndex field to show that this is the index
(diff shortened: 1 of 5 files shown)


HiDeoo:
Thanks for the contribution.

We usually recommend opening [a discussion](https://github.com/withastro/starlight/discussions) first before making a pull request. This helps us understand the context of the issue you are trying to solve, discuss possible solutions, and ensure that your contribution aligns with the project's goals. This also helps us to prioritize various contributions effectively as we can receive feedback from the community, see if other people are also running into the same issue, and gauge the overall interest in the proposed change. This also ensures that your efforts are not spent on a solution that may not be accepted.

With the current context, I'm failing to totally understand the problem you are facing and the proposed solution vs just using an `index.mdx` file.

Considering the above, I am going to close this PR for now, and I encourage you to open a discussion first so we can better understand the context and determine the best path forward. Thank you for your understanding!

Its-Just-Nans:
In here discussion https://github.com/withastro/starlight/discussions/3557

PR #3558, opened 2025-11-16
Add starlight-ui-tweaks

dlcastillop:
#### Description

Add the plugin starlight-ui-tweaks to the plugin list

PR #3562, opened 2025-11-17

ArmandPhilippot:
#### Description

Adds changes from #3558 to the French translation of `resources/plugins`.

### DIFF
--- a/docs/src/content/docs/fr/resources/plugins.mdx
+++ b/docs/src/content/docs/fr/resources/plugins.mdx
@@ -196,7 +196,7 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 	<LinkCard
 		href="https://github.com/dlcastillop/starlight-ui-tweaks"
 		title="starlight-ui-tweaks"
-		description="Étend les composants d'interface utilisateur de base avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré."
+		description="Étend les composants natifs avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré."
 	/>
 </CardGrid>
 


docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -193,6 +193,11 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		title="starlight-docsearch-typesense"
 		description="Une alternative au module d'extension officiel Algolia DocSearch reposant sur Typesense."
 	/>
+	<LinkCard
+		href="https://github.com/dlcastillop/starlight-ui-tweaks"
+		title="starlight-ui-tweaks"
+		description="Étend les composants d'interface utilisateur de base avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré."
```
HiDeoo:
```suggestion
		description="Étend les composants d'interface utilisateur avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré."
```

Any thoughts about slightly diverging from the original as I don't think "de base" adds a lot of value and also brings us down to 5 lines?

HiDeoo (dismissed):
Left a tiny suggestion/proposal, let me know what you think about it, but otherwise, looks perfect to me 🌟

docs/src/content/docs/fr/resources/plugins.mdx
ArmandPhilippot: Yeah, this one is a bit verbose. I don't mind removing `de base`, but it doesn't seem to save space at least with my screen size:
Screenshot of the translated plugin without translating "core"

I also thought about using "UI" to reduce the size, but I think we always translate it in Starlight:
Screenshot of the translated plugin keeping "UI"
(but I don't know, while this reduces the number of lines, I prefer it translated... something that bothers me in Astro Docs 😅 )

Another possibility would be to not include "UI" at all in the translation:
> Étend les composants natifs avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré.

I don't think that adds much to the meaning and this is less verbose. What do you think?

HiDeoo: Oh yeah, your last suggestion sounds really nice, does it reduce the number of lines on your end?

ArmandPhilippot: Yep! I didn't take the screenshot, but I checked before posting. 😄 

So, let's go with this:
```suggestion
		description="Étend les composants natifs avec des liens de navigation personnalisés, des publicités, des pieds de page axés marketing et un bouton de thème plus épuré."
```

PR #3572, opened 2025-11-26
[i18nIgnore] Transpile Starlight packages

### LINKED ISSUE #2644: Publish Modern Javascript instead of Typescript

0.29.2

4.16.16

npm

Windows 11 + WSL2 (Ubuntu 20.04)

Chrome

### Describe the Bug

Whenever I'm upgrading Starlight, I always seem to be running into type issues:

image

These cannot be ignored by using `skipLibCheck` because that flag only disables the type validation of `.d.ts` files in node_modules.

Typescript doesn't care that a `.ts` file is coming from node_modules, if you're using it, it needs to compile it and verify the types.

I think the fundamental problem of this library is the fact that it's publishing `.ts` files to NPM rather than compiling it to `.js` + `.d.ts` and publishing that. This renders `skipLibCheck` useless and means that in order to prevent type issues, I have to be on the Typescript version that this library was published with. If this library was built in typescript 5.5.5 and 5.7.2 (current) is more restrictive somewhere, it means that I have to deal with the type issues of this library in my project. What's worse is, not just different typescript versions can cause these issues, if I have a more restrictive TSConfig than you guys do e.g. I'm using `noImplicitAny` but you guys are not, then I will also get type issues as a consumer about implicit any in my installation of your library.

Furthermore, things such as virtual import paths defined by module declarations inside handwritten .d.ts files only work for this library locally, but do not work in the context of this library being installed in some node_modules folder for the consumer.

So while I would appreciate it if the type issues were fixed, it would be a lot smarter to publish JS/D.TS files in the future to prevent these issues from arising to begin with.

To reproduce some of the type issues as an example, go to the stackblitz and run `npm run lint:types`. This also highlights that for example my "strict" and "noImplicitAny" settings are causing extra issues in starlight dependency, because it seems that the tsconfig that starlight uses is also more lenient than mine. This again emphasizes the fundamental problem of publishing TypeScript files: you force your consumer to adhere to your TS Config settings, because TypeScript will compile and verify types of Starlight using the settings of the consumer.

https://stackblitz.com/edit/withastro-astro-jtj49r

### Participation

HiDeoo:
#### Description

- Closes #983
- Closes #1891
- Closes #2644

This PR is an early draft (not yet ready for review) that transpiles Starlight packages from TypeScript to JavaScript to avoid publishing TypeScript which is [not a recommended or even supported practice](https://github.com/microsoft/TypeScript/issues/40426#issuecomment-1373167253).

#### Approach

[Multiples approaches](https://colinhacks.com/essays/live-types-typescript-monorepo) are usually available when it comes to transpiling in monorepos and keeping the flexibility of working with TypeScript source code in development.

[Custom export condition](https://nodejs.org/api/packages.html#conditional-exports) was one of the first options considered, and would probably be the most long-term solution, but we cannot currently use them for Starlight. For the `docs/` site to properly consume custom conditions in development, in Astro, such conditions would need to be defined in the Astro configuration using the `vite` property, altho Starlight itself is imported in such configuration so it's a chicken-and-egg problem right now. There is a world in the future where we would be able to run Astro using `node --conditions` directly, and Vite automatically picks up those conditions, but this requires [some changes in Node](https://redirect.github.com/nodejs/node/pull/57688) that are not yet available.

Considering the above, this PR relies on `pnpm` and specifically on the fact that `pnpm` automatically overrides some fields in `package.json` files when packages are published with `pnpm publish` by the one defined in the `package.json` `publishConfig` property. In the case of Starlight, we maintain 2 `exports` fields: one for development (pointing to the TypeScript source files like it is today) and one for production (pointing to the transpiled JavaScript files).

#### How to test

When getting closer to a reviewable state, some pre-release packages will probably be published to test the changes in a real-world scenario.

In the meantime, you can test the changes locally using a different Starlight project outside of the monorepo, e.g. one setup using `pnpm create astro --template starlight`.

```
# Move to the Starlight package
$ cd packages/starlight
# Build the package
$ pnpm build
# Pack the package to a temporary location like it would be published
$ pnpm pack --pack-destination ~

# Move to your test Starlight project
$ cd ~/path/to/your/test/starlight/project/that/is/outside/of/the/monorepo
# Remove existing Starlight package
$ pnpm remove @astrojs/starlight && pnpm add ~/astrojs-starlight-0.41.7.tgz
```

#### Remaining tasks

- [x] Transpile Starlight packages
  - [x] Starlight
  - [x] DocSearch plugin
  - [x] Markdoc preset (nothing to do)
  - [x] Tailwind CSS (nothing to do)
- [x] Male sure our release workflow, specifically [`pnpm changeset publish`](https://github.com/withastro/starlight/blob/main/.github/workflows/release.yml#L42) uses `pnpm publish` under the hood (it probably does, as [`package-manager-detector`](https://github.com/antfu-collective/package-manager-detector) is [used](https://github.com/changesets/changesets/blob/main/packages/cli/src/commands/publish/npm-utils.ts#L194) when actually publishing to figure out the package manager but need to double check) (explained [here](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish): `If you are using pnpm as a package manager, this automatically detects it and uses pnpm publish instead.`)
- [x] Changesets


HiDeoo:
~~Small note (mostly for me to not forget) that [tsdown 0.17.3](https://github.com/rolldown/tsdown/releases/tag/v0.17.3) released last week added glob support to the `copy` option, which means we can probably use it instead of our own `globCopyPlugin()` currently included in this PR that I initially created due to the lack of glob support for this option.~~

Done in https://github.com/withastro/starlight/pull/3572/changes/b2f932c2c34e9ab2d174fda54e333514b3f67d37

packages/starlight/src/components-internals/Icons.ts
```diff
@@ -1,4 +1,4 @@
-import { FileIcons } from '../user-components/file-tree-icons';
+import { FileIcons, type FileIcon } from '../user-components/file-tree-icons';
 
 export const BuiltInIcons = {
```
delucis, line 3: Note/thought: is it safe for us to do this?
```suggestion
export const BuiltInIcons: Record<BuiltInIcon, string> = {
```

The current test that checks the types align did not provide very readable output while I was resolving the merge — it’s almost impossible to tell from it which keys are missing:

```
packages/starlight/__tests__/basics/icon.test-d.ts:6:58 - error ts(2344): Type 'BuiltInIcon' does not satisfy the constraint '"Expected: literal string: download, Actual: literal string: firefox" | "Expected: literal string: download, Actual: literal string: chrome" | "Expected: literal string: download, Actual: literal string: edge" | ... 420 more ... | "Expected: literal string: substack, Actual: literal string: safari"'.
  Type '"error"' is not assignable to type '"Expected: literal string: download, Actual: literal string: firefox" | "Expected: literal string: download, Actual: literal string: chrome" | "Expected: literal string: download, Actual: literal string: edge" | ... 420 more ... | "Expected: literal string: substack, Actual: literal string: safari"'.

6  expectTypeOf<keyof typeof BuiltInIcons>().toEqualTypeOf<BuiltInIcon>();
                                                           ~~~~~~~~~~~
```
Whereas with an explicit `Record` it’s easy to identify missing keys.

I would need to test if this causes issues with inlining the union type as noted later in this file.

packages/starlight/src/schemas/locale.ts
```diff
@@ -0,0 +1,31 @@
...
+			.describe('The BCP-47 tag for this language, e.g. `"en"`, `"ar"`, or `"zh-CN"`.'),
+		dir: z
+			.enum(['rtl', 'ltr'])
+			.optional()
+			.default('ltr')
+			.describe(
+				'The writing direction of this language; `"ltr"` for left-to-right (the default) or `"rtl"` for right-to-left.'
+			),
+	});
+}
```
HiDeoo, line 18: Only adding a note for later as I'm not on my computer right now and to not forget, but it just hit me that these `describe()` have probably been removed when updating to Zod 4 and re-introduced when updating the PR.

HiDeoo: Done in https://github.com/withastro/starlight/pull/3572/changes/4448ad0f4917b205ddf4bda303e40682403b495c

packages/starlight/src/components-internals/Icons.ts
HiDeoo: Played quite a lot with this and finally landed on a better solution.

For example, let's imagine some random icons are missing from the union type, e.g. `twitter`, `slack`, and `storybook`, the error will now mention exactly which icons are missing:

```
Type 'ExpectNever<"twitter" | "slack" | "storybook">' has no call signatures.
```

CONTRIBUTING.md
delucis, line 1: Adding a note to self that assuming we keep these changes we should probably document something about this in `CONTRIBUTING.md`. Here’s @HiDeoo’s description of the new scripts for working with the `dist` output:
> - A new `pnpm test:dist` command that runs all our unit tests against the build version. You can use it locally and it's run on CI too as an extra step so we don't lose coverage and our unit tests are relatively fast so it's not a huge overhead. Zero tests were modified for this.
> - A new `pnpm build:dist` command in the docs that can be used to build the docs against the built version of Starlight. Right now, it can be used locally but once merged, we could update our Netlify build command from `cd docs && pnpm build` to `cd docs && pnpm build:dist`.

packages/starlight/vitest.config.ts
```diff
@@ -5,22 +5,23 @@ export default defineConfig({
 		projects: ['__tests__/*/vitest.config.ts'],
 		coverage: {
 			reportsDirectory: './__coverage__',
-			include: ['**.ts'],
+			include: ['**.ts', 'src/*.ts'],
```
delucis, line 8: Why exactly are both these needed? Doesn’t this config mean “all `.ts` files” and “`.ts` files directly inside `src/`” (which is already included in the first pattern)?

I guess also we could scope to only the `src/` directory now?
```suggestion
			include: ['src/**/*.ts'],
```

Which could also allow us to clean up the `exclude` config?

packages/starlight/package.json
```diff
@@ -96,7 +93,55 @@
...
+        "types": "./dist/expressive-code.d.mts",
+        "import": "./dist/expressive-code.mjs"
+      },
+      "./expressive-code/hast": {
+        "types": "./dist/integrations/expressive-code/hast.d.mts",
+        "import": "./dist/integrations/expressive-code/hast.mjs"
+      },
+      "./routes/*": "./dist/routes/*",
+      "./style/markdown.css": "./dist/style/markdown.css"
+    }
```
delucis, line 148: Given it‘s a lot to manage, I wonder if we’re able to use tsdown’s [automatic exports management](https://tsdown.dev/options/package-exports)? 🤔 

It has a [`devExports`](https://tsdown.dev/options/package-exports#dev-exports) option for `publishConfig`.

packages/starlight/tsdown.config.ts
```diff
@@ -0,0 +1,44 @@
...
+	entry: [
+		'src/**/*.ts',
+		// The user components barrel file should not be transpiled as it's consumed as-is by Astro.
+		// https://github.com/withastro/astro/blob/c4c99aa7a5e8e45ada0efa4fda6e6fb96f334663/packages/astro/package.json#L55
+		'!src/components.ts',
+	],
+	plugins: [rawImportPlugin()],
+	publint: { strict: true },
+	unbundle: true,
+});
```
delucis, line 21: Not super essential, but any thoughts on enabling the [`fixedExtension`](https://tsdown.dev/reference/api/Interface.UserConfig#fixedextension) option so we get `.js`/`.d.ts` extensions instead of `.mjs`/`d.mts`? It’s something I usually do in personal projects but it’s mostly an aesthetic thing.

(If we do do this, we should do it consistently also in the other tsdown config files.)

packages/docsearch/__tests__/config.test-d.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'astro/zod';
+import { expectTypeOf, test } from 'vitest';
+import type { DocSearchConfigSchema } from '../src/config';
+import type { DocSearchUserConfig } from '../src/index';
+
+test('has matching `DocSearchUserConfig` input type for `DocSearchConfigSchema`', () => {
+	expectTypeOf<z.input<typeof DocSearchConfigSchema>>().toEqualTypeOf<DocSearchUserConfig>();
+});
```
delucis: I know we’ve discussed this in the past, but just so we can document this in this PR: can we explain **why** we can’t lean on Zod’s automatically generated types and need to duplicate them manually and keep them in sync?

It’s by far one of my least favourite end results of transpiling our TypeScript, as despite the tests which will ensure we keep things in sync, we will still likely hit things in CI and then fix, so it’ll be a less pleasant DX. It makes the use of Zod quite a bit less helpful/attractive. Obviously we shouldn’t do anything here, but I’d be curious if there are other validators that wouldn’t need handling like this.

packages/docsearch/package.json
```diff
@@ -2,39 +2,50 @@
...
-    "./DocSearch.astro": "./DocSearch.astro",
-    "./schema": "./schema.ts"
+    ".": "./src/index.ts",
+    "./DocSearch.astro": "./src/DocSearch.astro",
+    "./schema": "./src/schema.ts"
   },
   "peerDependencies": {
-    "@astrojs/starlight": ">=0.38.0"
+    "@astrojs/starlight": ">=0.38.0",
+    "astro": "^6.0.0 || ^7.0.0"
```
delucis, line 28: What was the reason for adding `astro` here? In general I’d guess we can leave the peer dep check to Starlight unless there’s a difference?

packages/starlight/src/utils/user-config.ts
```diff
@@ -1,43 +1,54 @@
...
+//
+// @see https://github.com/microsoft/TypeScript/issues/62309
+//
+// As such approach uses 2 sources of truth (a schema and an hand-written type), we use type tests
+// to ensure that the hand-written type always matches the inferred input type from the schema.
+// Any properties that use a sub-schema also follow the same approach, with type tests ensuring
+// the same consistency there as well.
+// Any new sub-schema used in the configuration should also have corresponding type tests added.
+//
+// @see {@link file://./../../__tests__/basics/schema.test-d.ts}
```
delucis, line 42: (non-actionable grumble 😅)

This is by far one of my least favourite end results of transpiling our TypeScript, as despite the tests which will ensure we keep things in sync, we will still likely end up only noticing things once they reach CI and then fixing, so it’ll be a less pleasant DX. It makes the use of Zod quite a bit less helpful/attractive. Obviously we shouldn’t do anything here, but I’d be curious if there are other validators that wouldn’t need handling like this.

packages/starlight/virtual-internal.d.ts
delucis, line 1: Are there any issues with moving things from this internal declaration to `virtual.d.ts`? Will the types for the internal modules now be visible to users?

packages/starlight/vitest.dist.config.ts
```diff
@@ -0,0 +1,5 @@
+import config from './vitest.config';
+
+process.env.STARLIGHT_TEST_DIST = 'true';
+
+export default config;
```
delucis: Given this config does so little, would it make sense to just move the `STARLIGHT_TEST_DIST` variable to the package script instead of having a separate config file?

delucis (commented):
I read all 336 files and all you get is this lousy review 😅 

I don’t think there’s all that much substantive to say given we discussed many decisions already. I’ve left some queries to clarify things, but nothing major I don’t think.

Thanks again for the huge lift in getting this off the ground!

CONTRIBUTING.md
HiDeoo: Good call, added 2 small notes for now in https://github.com/withastro/starlight/pull/3572/changes/f52c51eb916da9c59ede30d614b10211568c0beb.

packages/starlight/vitest.config.ts
HiDeoo: Oh, I wanted to add a note about this but forgot with the amount of changes.

- While working on this PR, I discovered our previous `'**.ts'` pattern includes TS files at the root and only other TS files loaded as part of the tests, but not all TS files.
- If we were to switch to `'src/**/*.ts'`, this would mean all TS files in `src/` and nested directories. This basically reduces greatly our coverage as before, not all TS files were considered, only the ones at the root and loaded as part of the tests.
- The pattern in this branch perserves the previous pattern so we don't have any regression in coverage.

Ultimately, I want to clean up our pattern to include any TS file in `src/` and nested directories like you suggested, but it felt out of scope for this PR.

packages/starlight/package.json
HiDeoo: I did tried it but did not find it super helpful in our case:

- This would add an entry for each TS files so we would need to filter them.
- There is some manual remapping to do, e.g. `integrations/expressive-code/hast.ts` → `expressive-code/hast`.
- We still need to manually handle components, routes, CSS, etc.

In the end, it felt like a lot of custom logic to maintain in the tsdown config to handle this rather than just doing it manually in the `package.json` which is more explicit imo but maybe I'm missing something? Happy to revisit if I did not understand how this is supposed to work.

packages/starlight/tsdown.config.ts
HiDeoo: > (If we do do this, we should do it consistently also in the other tsdown config files.)

Done in https://github.com/withastro/starlight/pull/3572/commits/e8c1ac6e2e1874abaf20fd46373c805466c39d43 and https://github.com/withastro/starlight/pull/3572/changes/46c968eea7bba108c83441e4b19004b4689f58a5 assuming you meant disabling it but I did not introduce a shared tsdown config at this point.

packages/docsearch/package.json
HiDeoo: I noticed this when working on the plugin as it imports from Astro when doing `import { z } from 'astro/zod';`.

Not adding `astro` as a peer dependency would mean that tsdown would by default try to incorrectly bundle Zod from `node_modules` (e.g. `dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/schemas.mjs`) and emits many warnings like the following:

```
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/az.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ar.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/de.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/bg.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ca.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/cs.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/da.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/fr.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/el.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/en.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/eo.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/es.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/fa.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/fi.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/be.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/it.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/fr-CA.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/he.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/hy.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/hu.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/id.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/is.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ms.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/hr.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/kh.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ka.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ja.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ko.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/lt.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/mk.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/sl.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/nl.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/no.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ota.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/yo.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ps.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/pl.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/sv.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/pt.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ta.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ro.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ru.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ur.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/km.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/uz.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/th.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/ua.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/vi.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/tr.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/zh-CN.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/zh-TW.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
(rolldown-plugin-dts:fake-js plugin) /Users/hideo/Projects/starlight/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/locales/uk.d.cts uses CommonJS dts syntax. CommonJS dts modules cannot be bundled by rolldown-plugin-dts. Please mark this module as external in your Rolldown config.
ℹ Hint: consider adding deps.onlyBundle option to avoid unintended bundling of dependencies, or set deps.onlyBundle: false to disable this hint.
See more at https://tsdown.dev/options/dependencies#deps-onlybundle
Detected dependencies in bundle:
- zod
ℹ dist/schema.mjs                                                                         3.73 kB │ gzip: 1.14 kB
ℹ dist/index.mjs                                                                          2.10 kB │ gzip: 0.86 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/schemas.mjs                 37.52 kB │ gzip: 7.75 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/schemas.mjs              26.15 kB │ gzip: 5.29 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/api.mjs                     11.61 kB │ gzip: 1.77 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/checks.mjs                  11.23 kB │ gzip: 2.09 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/to-json-schema.mjs          10.47 kB │ gzip: 3.04 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/util.mjs                    10.38 kB │ gzip: 3.09 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/json-schema-processors.mjs   7.88 kB │ gzip: 1.96 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/regexes.mjs                  4.25 kB │ gzip: 1.72 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/parse.mjs                    3.67 kB │ gzip: 0.73 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/errors.mjs                   2.21 kB │ gzip: 0.78 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/core.mjs                     1.92 kB │ gzip: 0.83 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/iso.mjs                   1.35 kB │ gzip: 0.39 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/registries.mjs               1.29 kB │ gzip: 0.51 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/parse.mjs                 1.24 kB │ gzip: 0.34 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/errors.mjs                1.01 kB │ gzip: 0.45 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/doc.mjs                      0.91 kB │ gzip: 0.48 kB
ℹ dist/config.mjs                                                                         0.57 kB │ gzip: 0.33 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/versions.mjs                 0.17 kB │ gzip: 0.14 kB
ℹ dist/schema.d.mts                                                                       3.39 kB │ gzip: 0.96 kB
ℹ dist/index.d.mts                                                                        2.75 kB │ gzip: 1.12 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/schemas.d.mts               44.90 kB │ gzip: 8.61 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/schemas.d.mts            18.70 kB │ gzip: 4.06 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/api.d.mts                    7.15 kB │ gzip: 1.79 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/checks.d.mts                 6.42 kB │ gzip: 1.38 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/standard-schema.d.mts        5.97 kB │ gzip: 1.40 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/errors.d.mts                 5.84 kB │ gzip: 1.50 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/to-json-schema.d.mts         3.94 kB │ gzip: 1.37 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/util.d.mts                   3.52 kB │ gzip: 1.36 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/json-schema.d.mts            2.13 kB │ gzip: 0.70 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/registries.d.mts             1.52 kB │ gzip: 0.61 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/core.d.mts                   1.37 kB │ gzip: 0.56 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/errors.d.mts              1.08 kB │ gzip: 0.46 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/classic/parse.d.mts               0.45 kB │ gzip: 0.24 kB
ℹ dist/node_modules/.pnpm/zod@4.4.3/node_modules/zod/v4/core/versions.d.mts               0.21 kB │ gzip: 0.16 kB
ℹ 36 files, total: 249.01 kB
```

This also increases the bundle size from 12.36 kB to 249.01 kB.

We could workaround this with some tsdown option, e.g. `deps.neverBundle`, but this would not be the proper fix afaik as not having it as a peer dependency would also mean that with some package managers / package manager configurations combo, this could become an error.

packages/starlight/src/utils/user-config.ts
HiDeoo: I did evaluate most popular validators a long time ago (during the Zod 4 beta) when I was specifically working on this issue and sadly, none of them would avoid this for our case (as this was mostly a TS issue), but it would be nice to re-evaluate them again after this PR is merged and see if somehow time, new versions, or TS 7 would have fixed this issue (I would be surprised if it did tho).

packages/starlight/virtual-internal.d.ts
HiDeoo: It shouldn't, this file is not in `src/` so it does not end up in `dist/` as it's never directly referenced by any other file in `src/` or copied to it. Switching to a transpiled versions means we no longer have to maintain some public and private ambient files.

The ones that would end up being in `dist/` are the ones that have been transformed to TS files and are referenced by other files in `src/`, e.g. `packages/starlight/src/global.ts` which is imported in `packages/starlight/src/index.ts`.

packages/starlight/vitest.dist.config.ts
HiDeoo: Good call, done in https://github.com/withastro/starlight/pull/3572/commits/cf0b24ae64ea0990b5c770fad659a0541e2c7e04.

packages/docsearch/package.json
delucis: Oh, I see, so actually it was always technically a peer dependency, we just didn’t specify it before. Thanks for clarifying!

delucis (dismissed):
I think this is ready when the time comes! 🚀 

Thanks again @HiDeoo 💖

HiDeoo:
Thanks for thorough review considering the amount of changes :raised_hands:

Added a changeset for the main change, altho not sure I'm super happy with the wording but didn't find a better way to phrase it for now.

delucis (dismissed):

delucis (approved):

michaelfaith:
Woohoo! Thanks for all the work on this 🙌

PR #3573, opened 2025-11-26
fix: use connectedCallback to handle site search

Grafikart:
#### Description

Change how site search logic is appended to the page. Use web component `connectedCallback` instead of listening to `window.addEventListener('DOMContentLoaded'`.

This change make Search work when using `<ClientRouter/>`.

### DIFF
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -76,89 +76,117 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 <script>
 	import { pagefindUserConfig } from 'virtual:starlight/pagefind-config';
 
-	class SiteSearch extends HTMLElement {
-		constructor() {
-			super();
-			const openBtn = this.querySelector<HTMLButtonElement>('button[data-open-modal]')!;
-			const closeBtn = this.querySelector<HTMLButtonElement>('button[data-close-modal]')!;
-			const dialog = this.querySelector('dialog')!;
-			const dialogFrame = this.querySelector('.dialog-frame')!;
+  class SiteSearch extends HTMLElement {
+    openBtn: HTMLButtonElement | null = null;
+    closeBtn: HTMLButtonElement | null = null;
+    dialog: HTMLDialogElement | null = null;
+    dialogFrame: HTMLDivElement | null = null;
+    finder: null | {destroy: () => void} = null;
 
-			/** Close the modal if a user clicks on a link or outside of the modal. */
-			const onClick = (event: MouseEvent) => {
-				const isLink = 'href' in (event.target || {});
-				if (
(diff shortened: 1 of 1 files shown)


HiDeoo:

Starlight does not support Astro's client router out of the box. While this PR is a possible fix for one of the issues you encountered when using the client router, many other issues would still exist. For example, some state that Starlight persists between navigations, e.g. the sidebar state, would not work as exepcted with the client router.

Adding support for the client router is something we have decided to not bake into Starlight core, as it turns out to not be a common use case for documentation sites using Starlight, and this is something that can potentially add unnecessary code for users not using the client router. Instead, we recommend and rely on [a dedicated package](https://events-3bg.pages.dev/jotter/starlight/guide/) that provides such support for users who want to use Starlight with the client router. This means that users can choose to opt-in to this functionality if they need it, and such package can handle all the necessary adjustments to make Starlight work correctly with the client router. Relying on a package also allows for better and more targeted changes, for example in the case of the search feature, instead of re-instantiating the search on every navigation, the search UI can actually be persisted entirely between navigations, which is a better user experience in the case of client-side routing.

Right now, I assume you are currently using an [override](https://starlight.astro.build/guides/overriding-components/) for the [`<Head>` component](https://starlight.astro.build/reference/overrides/#head-1) that may look something like this to add support for the client router:

```astro
---
import Default from '@astrojs/starlight/components/Head.astro';
import { ClientRouter } from "astro:transitions";
---

<!-- Render Starlight's default Head content -->
<Default><slot /></Default>

<!-- Add the client router -->
<ClientRouter />
```

After [installing the `astro-vtbot` package](https://events-3bg.pages.dev/jotter/starlight/guide/#step-1-install-astro-vtbot), such override can be modified to use it:

```astro
---
import Default from '@astrojs/starlight/components/Head.astro';
import VtbotStarlight from 'astro-vtbot/components/starlight/Base.astro';
---

<VtbotStarlight>
  <Default><slot /></Default>
</VtbotStarlight>
```

At this point, the `astro-vtbot` package will take care of making all features of Starlight work correctly with the client router, including the search component which will now persist the search UI between navigations. If animations are not desired, they can of course be disabled using `<VtbotStarlight transition:animate="none">` instead and the packages also offers many other [optional features](https://events-3bg.pages.dev/jotter/starlight/guide/#optional-features) that can be configured as needed but none are required.

On the other hand, after doing some review of our own documentation, I think we could definitely improve discoverability of this package, e.g. we never explicitly mention that it's our recommended approach when using the client router so searching for "client router" would not lead to any relevant documentation page. I'll be working on improving this and I'll also discuss and explore some ways to potentially simplify the onboarding experience for this package for such basic use cases.

Considering all of the above, I'll close this PR. Thanks for the effort you put into it and for bringing this up.

PR #3575, opened 2025-11-27
Add footnotes section to markdown guide

johan456789:
#### Description

- Documents #2622. 

Same as the [comment](https://github.com/withastro/starlight/discussions/2622#discussioncomment-11347726), I tried to search the doc for footnote and couldn't find it. I think it deserves to be mentioned within the guide itself instead of leaving it to a link in the "Other common Markdown features" section.


docs/src/content/docs/guides/authoring-content.mdx
```diff
@@ -543,6 +543,16 @@ The [Andromeda constellation](<https://en.wikipedia.org/wiki/Andromeda_(constell
 </details>
 ```
 
+## Footnotes
+
+Footnotes[^1] are useful for adding brief references or clarifications without interrupting the flow of your content. They appear as superscript links that jump to referenced text collected at the bottom of the page.
```
delucis: I think we probably shouldn’t actually _use_ a footnote in this case is we’d normally link to a resource like this. (Also noting that Starlight includes the “Footnotes” subheading in the table of contents, which I didn’t know 😁)

```suggestion
[Footnotes](https://www.markdownguide.org/extended-syntax/#footnotes) are useful for adding brief references or clarifications without interrupting the flow of your content. They appear as numbered superscript links that jump to referenced text collected at the bottom of the page.

Define a footnote using square brackets with a caret and numbered reference:
```

docs/src/content/docs/guides/authoring-content.mdx
```diff
@@ -684,3 +694,5 @@ export default defineMarkdocConfig({
 	extends: [starlightMarkdoc({ headingLinks: false })],
 });
 ```
+
+[^1]: https://www.markdownguide.org/extended-syntax/#footnotes
```
delucis: So then we’d remove the reference too:

```suggestion
```

PR #3590, opened 2025-11-29
i18n(fr): update `reference/configuration.mdx`

ArmandPhilippot:
#### Description

Adds changes from #3332 to the French translation of `reference/configuration`.

### DIFF
--- a/docs/src/content/docs/fr/guides/pages.mdx
+++ b/docs/src/content/docs/fr/guides/pages.mdx
@@ -77,19 +77,13 @@ Pour en savoir plus, consultez le guide [« Pages » dans la documentation d'Ast
 
 Pour utiliser la mise en page de Starlight dans des pages personnalisées, englobez le contenu de votre page avec le [composant `<StarlightPage />`](#composant-starlightpage).
 Cela peut s'avérer utile si vous générez du contenu de manière dynamique, mais que vous souhaitez tout de même utiliser le design de Starlight.
-Ce composant doit être la première importation dans votre fichier pour configurer les [couches de cascade](/fr/guides/css-and-tailwind/#couche-de-cascade) et garantir un ordre prévisible pour le CSS.
 
 Pour ajouter des liens d'ancrage vers des en-têtes avec une apparence équivalente aux liens d'ancrage Markdown de Starlight, vous pouvez utiliser le [composant `<AnchorHeading>`](#composant-anchorheading) dans vos pages personnalisées.
 
 ```astro
 ---
 // src/pages/page-perso/exemple.astro
-// Importer le composant `<StarlightPage>` en premier pour configurer les
-// couches de cascade.
 import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
-
-// Importer tous les autres composants que vous souhaitez utiliser dans votre
-// page personnalisée.
 import AnchorHeading from '@astrojs/starlight/components/AnchorHeading.astro';
 import CustomComponent from './CustomComponent.astro';
 ---
@@ -113,8 +107,6 @@ Le composant `<StarlightPage />` affiche une page complète de contenu en utilis
 
 ```astro
 ---
-// Importer le composant `<StarlightPage>` en premier pour configurer les
-// couches de cascade.
 import StarlightPage from '@astrojs/starlight/components/StarlightPage.astro';
 ---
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/reference/configuration.mdx
```diff
@@ -399,6 +399,28 @@ starlight({
 }),
 ```
 
+#### `processedDirs`
+
+**Type :** `string[]`  
+**Par défaut :** `[]`
+
+Définit les répertoires supplémentaires dans lesquels les fichiers doivent être traités par le pipeline Markdown de Starlight.
```
HiDeoo: Tried to think about something else rather than "pipeline" but the only one that came to mind was ["chaîne de traitement"](https://fr.wikipedia.org/wiki/Pipeline_(architecture_des_processeurs)) like in processors but didn't feel any better.

docs/src/content/docs/fr/reference/configuration.mdx
```diff
@@ -399,6 +399,28 @@ starlight({
...
+Par défaut, seuls les contenus Markdown et MDX chargés à l'aide de [`docsLoader()`](/fr/reference/configuration/#docsloader) de Starlight sont traités.
+Prend en charge les répertoires locaux relatifs à la racine de votre projet, par exemple `'./src/data/comments/'`.
+
+Le traitement de Starlight inclut la prise en charge les [liens d'ancrage cliquables des en-têtes](#headinglinks), la [syntaxe de directive Markdown pour les encarts](/fr/guides/authoring-content/#encarts) et la prise en charge RTL des blocs de code.
```
HiDeoo:
```suggestion
Le traitement de Starlight inclut la prise en charge des [liens d'ancrage cliquables des en-têtes](#headinglinks), la [syntaxe de directive Markdown pour les encarts](/fr/guides/authoring-content/#encarts) et la prise en charge RTL des blocs de code.
```

HiDeoo (changes requested):
Left a suggestion but other than that, looks perfect to me 👍 

Thanks for the update.

docs/src/content/docs/fr/reference/configuration.mdx
ArmandPhilippot: Nice catch! But, we also need to add `de` before the other ones. I was looking for another solution, couldn't find one and I forgot to fix that afterwards. 😅 
And now, I notice we have two "prise en charge"... that's probably why I was bothered by this sentence!

Maybe this:
```suggestion
Le traitement de Starlight inclut la prise en charge des [liens d'ancrage cliquables des en-têtes](#headinglinks), de la [syntaxe de directive Markdown pour les encarts](/fr/guides/authoring-content/#encarts) et de RTL dans les blocs de code.
```

docs/src/content/docs/fr/reference/configuration.mdx
ArmandPhilippot: Ah yes, that bothered me a bit too. I wasn't sure if this was still considered as an anglicism or not, and couldn't find something better. I like your suggestion! But "traités... chaîne de traitement" is a bit repetitive. Maybe we could reword a bit the sentence:
> Définit les répertoires supplémentaires contenant les fichiers qui doivent être ajoutés à la chaîne de traitement Markdown de Starlight.

docs/src/content/docs/fr/reference/configuration.mdx
HiDeoo: Interesting, the lack of "de" didn't bother me that much initially but it definitely feels right with your new changes.

Unrelated, I wonder if RTL is something well-known or used as-is in French. Maybe it should be "et de la direction RTL" 🤔 Not quite sure and not easy to find references.

ArmandPhilippot: Good call, it seemed obvious to me but I think you right it might be better to clarify this. Instead of "direction", I can think of ["système d'écriture"](https://fr.wikipedia.org/wiki/Syst%C3%A8me_d%27%C3%A9criture) or ["sens de l'écriture"](https://ecole-publique-ploeren.ac-rennes.fr/spip.php?article2927) because "direction" alone might not be enough?
* "du système d'écriture RTL"
* "du sens d'écriture RTL"
* Or more verbose: "du système d'écriture de droite à gauche (RTL)"/"du sens d'écriture de droite à gauche (RTL)"?

HiDeoo: Just learned this is actually called "sens horizontal sinistroverse" 😅 So, let's ignore that and go with "du sens d'écriture RTL" for now, and if ever needed, we can revisit.

ArmandPhilippot: Oh right, now you say that this is obvious. 😆 Well, I encountered that word in the past but yeah I don't think this would be more meaningful for French readers... I didn't even remember anymore (guilty, I haven't read everything in the Wikipedia link).

```suggestion
Le traitement de Starlight inclut la prise en charge des [liens d'ancrage cliquables des en-têtes](#headinglinks), de la [syntaxe de directive Markdown pour les encarts](/fr/guides/authoring-content/#encarts) et du sens d'écriture RTL dans les blocs de code.
```

docs/src/content/docs/fr/reference/configuration.mdx
ArmandPhilippot: I think this changes sounds good too, so:
```suggestion
Définit les répertoires supplémentaires contenant les fichiers qui doivent être ajoutés à la chaîne de traitement Markdown de Starlight.
```

PR #3592, opened 2025-11-29
feat: support for a different icon for the `FileTree` by specifying the extension to use

maelgangloff:
#### Description

- Closes #3580

This change allows developers to specify a different extension than the one deduced from the filename extension to display the correct icon.

> For example, in some projects, local environment variables are written to a `.env.local` file, which is copied from `.env`.
> 
> When I explain this in my documentation, the first file has a default icon, while the second has the correct icon ( ⚙️ ) to clearly indicate that it's a configuration file.

image

### DIFF
--- a/.changeset/chatty-badgers-mate.md
+++ b/.changeset/chatty-badgers-mate.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Support for a different icon for the FileTree by specifying the extension to use


HiDeoo:
Thanks for the contribution.

Before moving to implementation, we usually spend some time discussing feature requests to make sure it aligns with the project goals, measure interest and gather feedback from the community, and also explore and discuss possible approaches to implement a feature to make sure we pick the best one.

In this case, I shared some initial thoughts in the [feature request](https://github.com/withastro/starlight/discussions/3580) you opened, and I think this may be one of those cases where we should definitely spend some time discussing this further before moving forward with implementation. Lots of interesting questions comes to mind when thinking about such feature.

I'm going to close this PR for now, and we can reopen/revisit it later if we decide to move forward and agree on the approach to implement this new feature.

PR #3595, opened 2025-12-01
Update PR template about new features

HiDeoo:
#### Description

This PR updates the pull request template to include some guidelines regarding new features.

The goal is to reduce the number of PRs adding new features that have not been previously discussed which can lead to such PRs being closed without merging.


HiDeoo:
Oh, I think I merged this a few seconds before your review 🙈 Totally agree on avoiding as much as possible overwhelming templates, I'd say what we currently have with this change is probably the limit we want to never go over.

PR #3597, opened 2025-12-02
docs: remove duplicated plugin/component

HiDeoo:
#### Description

This PR removes a duplicated plugin/component entry in the documentation introduced in #3546.


delucis (approved):
Thanks for fixing that @HiDeoo! I totally missed it when reviewing #3546.

PR #3602, opened 2025-12-02
Add `starlight-links` to recommended extensions

HiDeoo:
#### Description

As previously discussed, this PR adds the Visual Studio Code extension [`starlight-links`](https://github.com/HiDeoo/starlight-links) to the list of recommended extensions in the `.vscode/extensions.json` file.

This only adds it to the Starlight monorepo recommendations for now with the goal to increasingly get more feedback/users before potentially adding it to the examples at some point.


delucis (approved):
Definitely approve! I’d also be happy to add it for our starter templates, but if you’d prefer to wait, then that’s OK too.

PR #3610, opened 2025-12-03
Add preview releases

HiDeoo:
#### Description

As discussed, this PR adds support for preview releases in the monorepo using [pkg.pr.new](https://pkg.pr.new/).

The worflow is heavily inspired by the [one used in the Astro repo](https://github.com/withastro/astro/blob/main/.github/workflows/preview-release.yml) with the following changes:

- Maintainers can trigger a preview release by adding the `pr-preview` label (it's [not possible](https://github.com/stackblitz-labs/pkg.pr.new/issues/187#issuecomment-2322825988) to trigger it directly using a comment).
- A little bit stricter permissions (I don't think we need write access to the `issues` and `id-token` scope).
- I added some comments to the step computing the packages to be included as it was not really obvious to me at first glance to understand exactly what was going on and I needed to run it locally to verify the output.
- I also added a bit more logic to this step to compute the list of package paths as needed by the `pkg-pr-new` command rather than doing it later so all logic is in one place.
- If no packages are published, the workflow errors out with a message indicating to make sure there are changesets included in the PR. I think I would prefer to see an error than waiting for the comment to show up before realizing the action succeeded but nothing was published.

Also checked the workflow using [`zizmor`](https://github.com/zizmorcore/zizmor), and only an info for `prefer trusted publishing for authentication` is reported for `pnpm dlx pkg-pr-new publish` but this sounds like a false positive for `pnpm publish` as there is no trusted publishing possible here afaik.

If I'm not mistaken, I think we need this PR to be merged first before being able to test the entire process though so probably need a thorough review before merging in case I made any mistake.

Note that when we would want to use preview releases with #3572, we would also need to edit this workflow to include a build step.

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -8,7 +8,7 @@ on:
 
 # Automatically cancel in-progress actions on the same branch
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:


.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,90 @@
...
+  cancel-in-progress: true
+
+jobs:
+  preview:
+    name: Preview Release
+    if: ${{ github.repository_owner == 'withastro' && github.event.label.name == 'pr-preview' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
delucis, line 21: Is there a risk here with granting write access to PRs when using the `pull_request` trigger? I guess the fact the PR needs to be labelled means it can’t be _triggered_ by a third party.

But maybe the commenting will fail for PRs from forks? IIRC forks aren’t able to get write permission for PRs?

Just took a look and all our other actions with this permission are either run using the `pull_request_target` trigger, or in limited circumstances like on push to `main` or a cron schedule.

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          # This makes Actions fetch all Git history so that Changesets can find pending changesets
+          fetch-depth: 0
```
delucis: I know we still have this in our usual release workflow but it’s [no longer needed](https://github.com/changesets/action/pull/177):

```suggestion
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Publish preview packages
+        env:
+          PACKAGES: ${{ steps.preview.outputs.packages }}
+        run: |
+          if [ -z "$PACKAGES" ]; then
+            echo "::error::No packages to publish in this preview release."
+            echo "::error::Make sure the pull request includes a changeset."
+            exit 1
+          fi
```
delucis, line 83: One question on errorring here: will we actually see that in the PR? I don’t know whether or not this workflow would show up as one of the PRs “checks” given it runs on the `labeled` event.

delucis (commented):
Thanks @HiDeoo — this will be fun to have and super helpful for sure.

Left a few questions just to make sure I understand everything!

.github/workflows/preview-release.yml
HiDeoo: Hmm, good question.

From my understanding, `pkg.pr.new` does not require any permissions in the workflow as they're all [granted by the GitHub application](https://github.com/stackblitz-labs/pkg.pr.new/issues/305) itself.

The only step in my head that would potentially fail is the label removal for forks (I thought I added a comment about that like in the Astro repo but looks like I instead replaced it with another comment — I will need to fix that). When this happens, this should not be an issue as `actions-ecosystem/action-remove-labels` default behavior is controlled by the `fail_on_error` input which defaults to `false`. So if the label removal fails, the action should just continue. The only downside is that we would have to remove/re-add the label manually if we want to re-trigger a preview on a PR from a fork.

> Just took a look and all our other actions with this permission are either run using the pull_request_target trigger, or in limited circumstances like on push to main or a cron schedule.

Interesting, I picked it for 2 reasons:

- That's the one [we use](https://github.com/withastro/astro/blob/main/.github/workflows/preview-release.yml#L4) in the Astro repo
- I [saw](https://github.com/stackblitz-labs/pkg.pr.new/issues/235#issuecomment-2408401911) that using `pull_request_target` would lead to preview having code from `main` instead of the PR

.github/workflows/preview-release.yml
HiDeoo: And you even told me but I forgot, sorry about that 🙈

.github/workflows/preview-release.yml
HiDeoo: Good question again.

I based my initial research on [this PR](https://github.com/withastro/astro/pull/14889) which triggered some preview releases but as you can see, it doesn't appear in [the checks](https://github.com/withastro/astro/pull/14889/checks). Maybe I did not give enough details, but my idea is more about being able to open the Actions page in a new tab after applying the label, and just wait to see either green (published) or red (failed) rather than always green, closing that tab, going back to the PR, refreshing it, and noticing that no comment was added.

.github/workflows/preview-release.yml
delucis: > I [saw](https://github.com/stackblitz-labs/pkg.pr.new/issues/235#issuecomment-2408401911) that using `pull_request_target` would lead to preview having code from `main` instead of the PR

Oh, well that’s a very good reason not to use it 😅 

OK, so sounds like there might be some minor annoyances, but on balance things will work pretty well 👍

.github/workflows/preview-release.yml
delucis: Thanks for clarifying. Sounds OK. I guess if it becomes a real issue we could maybe use a step that manually posts a comment to the PR if this fails (at least for non-forks).

HiDeoo: Definitely, I would assume some minor tweaks once we start using it over time.

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Get changeset status
+        id: changeset
+        run: |
+          pnpm changeset status --output changeset-status.json
+          echo "status=$(cat changeset-status.json | jq -c .)" >> $GITHUB_OUTPUT
+
+      - name: Get pnpm packages
+        id: pnpm
+        run: echo "packages=$(pnpm list --recursive --depth -1 --json | jq -c .)" >> $GITHUB_OUTPUT
```
trueberryless, line 48: We could filter for public packages only, but I guess since changesets for private packages are not possible, this just makes jq slower:

```suggestion
        run: echo "packages=$(pnpm list --recursive --depth -1 --json | jq -c '[.[] | select(.private == false)]')" >> $GITHUB_OUTPUT
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,88 @@
+name: Preview Release
+permissions: {}
+
+on:
+  pull_request:
+    branches: [main]
+    types: [labeled]
```
trueberryless, line 7: This is just an idea: We could also allow commit triggers to automatically update the preview package on new changes (which would also require checking if the label is currently active on the PR - more difficult) so you don't need to re-add the label each time. At the same time, this would then also not be what you want/expect as packages could update when you just want to try them out or maybe there is also some malicious action potentially involved? Therefore, just an idea... 🤔

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,88 @@
...
+
+# Automatically cancel in-progress actions on the same branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
```
trueberryless: Do we need to update the trigger to `pull_request` here too or not?


```suggestion
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
```

trueberryless (commented):
Very interested to see this in _action_, I left some of my thoughts. Wonderful implementation so far, couldn't spot a single mistake 🙌

.github/workflows/preview-release.yml
HiDeoo: Good catch, definitely a copy/paste mistake 🙌

.github/workflows/preview-release.yml
HiDeoo: Yeah, there is no need to do so in this case.

.github/workflows/preview-release.yml
HiDeoo: We definitely don't want new releases on every commit. Most of the time it'll probably just throwing resources away, specially near the end of PRs when working on final touches and documentation, and also some of the reasons you mentioned.

Manually triggering releases when a maintainer judges it's appropriate seems like the best approach in our case.

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,88 @@
...
+              const packagePathsMap = Object.fromEntries(packages.map(pkg => [pkg.name, relative(process.cwd(), pkg.path)]));
+              # A list of all relative package paths to publish in this preview release.
+              const previewPackagePaths = [...new Set(changes.changesets.map(change => change.releases.map(release => packagePathsMap[release.name])).flat())]
+              # A space-separated string of all relative package paths to publish in this preview release (or an empty string if none).
+              core.setOutput('packages', previewPackagePaths.join(' '));
+
+      # We remove the preview label before publishing so that the label is always removed even if
+      # publishing fails, so we can retry by only adding the label again.
+      - name: Remove Preview Label
+        uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0 # v1.3.0
```
trueberryless, line 71: Does this action work when the checkout does not persist the credentials? Does this action use `git` commands directly? I assume it uses the GitHub API as labels of PRs are not a git thing...

I personally have never used this action, maybe you have more experience.

delucis: Yeah git credentials are unrelated here it, uses the GitHub token instead.

delucis (approved):
Let’s see how this breaks!

Just kidding — I believe in you (but not in GitHub Actions…)

PR #3611, opened 2025-12-05
i18n(de): update guides/authoring-content.mdx

trueberryless:
#### Description

Update German translation of `guides/authoring-content.mdx` with changes from #3575

I have chosen little bit different translations regarding describing _where_ the footnote will be added, as the direct translation was not clearly understandable IMO and felt insufficient. Therefore, the translation includes the additional clarification for what a caret is, as I was also unaware that the German term is the same and wanted to clarify that... Okay, enough talk, you can see the changes yourself 😅

### DIFF
--- a/.github/workflows/file-icons.yml
+++ b/.github/workflows/file-icons.yml
@@ -35,7 +35,7 @@ jobs:
         working-directory: packages/file-icons-generator
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@84ae59a2cdc2258d6fa0732dd66352dddae2a412 # v7.0.9
+        uses: peter-evans/create-pull-request@d4f3be6ce6f4083b7ac7490ab98b48a62db1ee41 # v7.0.10
         with:
           branch: ci/file-icons
           token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
--- a/docs/src/content/docs/de/resources/plugins.mdx
+++ b/docs/src/content/docs/de/resources/plugins.mdx
@@ -99,11 +99,6 @@ Erweitere deine Website mit offiziellen Plugins, die vom Starlight-Team unterst�
 		description="Füge einen netten Verweis auf Starlight oder Astro am Ende des Inhalts&shy;verzeichnisses ein."
 	/>
 	<LinkCard
-		href="https://github.com/trueberryless-org/starlight-contributor-list"
-		title="starlight-contributor-list"
-		description="Zeigt eine Liste aller Mitwirkenden an deinem Projekt an."
-	/>
-	<LinkCard
 		href="https://github.com/dragomano/starlight-giscus"
 		title="starlight-giscus"
 		description="Füge Giscus-Kommentare zu deiner Dokumentations-Website hinzu."


HiDeoo:
> I have chosen little bit different translations regarding describing _where_ the footnote will be added

As a non-native speaker, I don't feel like I have enough tools to properly see what exactly changed compared to the original text regarding the footnote placement so I'll let a native speaker review this one.

trueberryless:
> As a non-native speaker, I don't feel like I have enough tools to properly see what exactly changed compared to the original text regarding the footnote placement so I'll let a native speaker review this one.

I think that's a good decision for this PR. Thank you very much for looking into the PR nonetheless 👍

I'm gonna ping @randomguy-2650 here, maybe they want to take a look (but no hurry or obligation) 🙌

docs/src/content/docs/de/guides/authoring-content.mdx
```diff
@@ -543,6 +543,18 @@ Das Sternbild [Andromeda](<https://de.wikipedia.org/wiki/Andromeda_(Sternbild)>)
 </details>
 ```
 
+## Fußnoten
+
+[Fußnoten](https://www.markdownguide.org/extended-syntax/#footnotes) sind nützlich, um kurze Hinweise oder Erklärungen hinzuzufügen, ohne den Textfluss zu stören. Sie werden hochgestellt als nummerierte Links im Text angezeigt, die zu den entsprechenden Erläuterungen am Ende der Seite führen.
+
+Erstelle eine Fußnote mit eckigen Klammern, einem Caret (`^`) und einer Nummer:
```
randomguy-2650:
```diff
-Erstelle eine Fußnote mit eckigen Klammern, einem Caret (`^`) und einer Nummer:
+Definiere eine Fußnote mit eckigen Klammern, einem Caret (`^`) und einer Nummer:
```

The original used “define” (imperative), so it could be used here, but I think “erstellen” works better here.

docs/src/content/docs/de/guides/authoring-content.mdx
```diff
@@ -543,6 +543,18 @@ Das Sternbild [Andromeda](<https://de.wikipedia.org/wiki/Andromeda_(Sternbild)>)
...
+
+```md
+Hier ist eine Fußnote[^1] mit nachfolgendem Referenztext.
+
+[^1]: Meine ausführliche Referenz.
```
randomguy-2650:
```suggestion
[^1]: Meine Referenz.
```

It seems that the original English file used `[^1]: My reference.` instead of something like `[^1]: My detailed reference.`, so I would change it here.

But it might seem better with the word “ausführlich” (English: “detailed”), so if it makes sense, you could make a PR. 🤔

docs/src/content/docs/de/guides/authoring-content.mdx
randomguy-2650: Caret is correct[^1], but I don’t think ``(`^`)`` is necessary because “Caret” seems like it’ll already self‐explain, if you know what I mean.

[^1]: Here’s a footnote. Some might say circumflex (Zirkumflex) or literally “Dach” or “Hut” (both are informal), but I think caret is good.

docs/src/content/docs/de/guides/authoring-content.mdx
trueberryless: Actually, I translated this as an ironical joke as the reference is not detailed at all. But rereading it now, yeah, I think I'll drop the word, but I would be curious what @delucis thinks about adding it to the English version as well.

docs/src/content/docs/de/guides/authoring-content.mdx
trueberryless: How about that, I think that explains it pretty well without needing the caret symbol:

```suggestion
Erstelle eine Fußnote mit eckigen Klammern, einem Caret (auch bekannt als Zirkumflex oder Dach) und einer Nummer:
```

I'm open for dropping Dach as well, if you think it's too informal.

docs/src/content/docs/de/guides/authoring-content.mdx
HiDeoo: I think we're pretty happy with the English version, and such reference does not have to be detailed at all. In fact, we even say "adding _brief_ references or clarifications" so let's keep it that way.

docs/src/content/docs/de/guides/authoring-content.mdx
randomguy-2650: I think we’ll keep Dach for the random weird people that only know the informal variant for idk what reason.

HiDeoo (approved):
Amazing :tada:

Thanks for the contribution and thanks @randomguy-2650 for jumping in to help review so quickly :raised_hands: Great work all around!

trueberryless:
Thanks HiDeoo for the help 🙌 and huge thanks to randomguy for looking into it. I know I can count on you 💪🚀

PR #3618, opened 2025-12-09
Refactor autogenerated sidebar groups

HiDeoo:
#### Description

- [x] This PR adds a new feature which has been discussed and approved [here](https://github.com/withastro/starlight/discussions/3336).

- Closes #3336

This PR removes the notion of "autogenerated sidebar groups" in favor of simply using "autogenerated sidebar links" which can be used in regular sidebar groups. As discussed in #3336, this simplifies the mental model around groups and also allows for use cases that were not possible before, such as having a sidebar entirely made of autogenerated links for a specific directory without having it nested inside a group.

For users, the gist of the change is:

```diff
{
    label: 'My group',
-   autogenerate: { directory: 'some-dir' },
+   items: [{ autogenerate: { directory: 'some-dir' } }],
}
```

Such refactor removes the 1:1 mapping that was previously available between an autogenerated sidebar group in the Starlight configuration and such autogenerated group in the Starlight route data. To address such change, the Starlight route data has been updated and many solutions are possible to address this, e.g. having multiple types (`link`, `autolink`, `group`, and `autogroup`). This PR has gone through multiple variations of such approach, and the current implementation does not introduce new types but rather adds an `autogenerated` property set to to `true` for links and subgroups that were autogenerated. Of course, other approaches are possible and feedback is welcome.

To make sure such change does not impact plugins with previous features that would no longer be possible due to the changes, I rewrote locally the [`starlight-auto-sidebar`](https://github.com/HiDeoo/starlight-auto-sidebar/) plugin, which was heavily relying on that 1:1 mapping and to my knowledge, may be one of the "largest" plugins tweaking sidebar route data for autogenerated groups with a large enough set of tests. I managed to rewrite it successfully while using the Starlight version from this PR and achieved the same results as before.

Note: I also addded a few tests for some behaviors that were not covered before and were impacted by this change, e.g. the `collapsed` state of autogenerated subgroups.

#### Remaining tasks

- [x] Add changeset


packages/starlight/__tests__/basics/config-errors.test.ts
```diff
@@ -197,11 +197,11 @@ test('errors with bad nested sidebar config', () => {
 		parseStarlightConfigWithFriendlyErrors({
 			title: 'Test',
 			sidebar: [
+				// @ts-expect-error - Testing invalid config
```
HiDeoo, line 199: I need to investigate why this error was hoisted (I can guess why, but need to know the specific reason) and see if we can move it back down at the group item level without losing some other type behaviors we expect.

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -186,22 +186,25 @@ The configuration above generates the following sidebar:
...
+### Autogenerated links
 
-Starlight can automatically generate a group in your sidebar based on a directory of your docs.
+Starlight can automatically generate links in your sidebar based on a directory of your docs.
 This is helpful when you do not want to manually enter each sidebar item in a group.
 
 By default, pages are sorted in alphabetical order according to the file [`id`](/reference/route-data/#id).
 
-Add an autogenerated group using an object with `label` and `autogenerate` properties. Your `autogenerate` configuration must specify the `directory` to use for sidebar entries. For example, with the following configuration:
+Add autogenerated links using an object with an `autogenerate` properties.
```
delucis: Small fix:
```suggestion
Add autogenerated links using an object with the `autogenerate` property.
```

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -186,22 +186,25 @@ The configuration above generates the following sidebar:
...
+Your `autogenerate` configuration must specify the `directory` to use for sidebar entries.
+Autogenerated links can also be nested within groups.
```
delucis: Note for myself that if this is an “also”, we should maybe split up the docs for it. (I’ll have a think how best to phrase it/show the functionality.)

docs/src/content/docs/guides/sidebar.mdx
```diff
@@ -438,22 +441,26 @@ The configuration above generates the following sidebar:
...
+				{
+					autogenerate: {
+						// Autogenerate links for the 'constellations' directory.
+						directory: 'constellations',
+						// Italicize all autogenerated link labels.
+						attrs: { style: 'font-style: italic' },
+					},
+				},
+			],
 		},
```
delucis, line 458: Could be worth simplifying these examples?
```suggestion
		{
			autogenerate: {
				// Autogenerate links for the 'constellations' directory.
				directory: 'constellations',
				// Italicize all autogenerated link labels.
				attrs: { style: 'font-style: italic' },
			},
		},
```
(Preview would need updating to match if we do)

packages/starlight/__tests__/basics/navigation.test.ts
```diff
@@ -85,7 +92,8 @@ describe('getSidebar', () => {
 			const currentItems = items.filter((item) => item.type === 'link' && item.isCurrent);
 			expect(currentItems).toHaveLength(1);
 			const currentItem = currentItems[0];
-			if (currentItem?.type !== 'link') throw new Error('Expected current item to be link');
+			if (currentItem?.type !== 'link' && currentItem?.type !== 'link')
+				throw new Error('Expected current item to be link');
```
delucis: I think this change is maybe left over from some experimentation?

```suggestion
			if (currentItem?.type !== 'link') throw new Error('Expected current item to be link');
```

packages/starlight/__tests__/basics/navigation.test.ts
```diff
@@ -149,12 +158,13 @@ describe('flattenSidebar', () => {
 		// Sidebar should include some nested group items.
 		expect(sidebar.some((item) => item.type === 'group')).toBe(true);
 		// Flattened sidebar should only include link items.
-		expect(flattened.every((item) => item.type === 'link')).toBe(true);
+		expect(flattened.every((item) => item.type === 'link' || item.type === 'link')).toBe(true);
```
delucis: Same here:
```suggestion
		expect(flattened.every((item) => item.type === 'link')).toBe(true);
```

packages/starlight/utils/routing/types.ts
```diff
@@ -23,14 +23,24 @@ export interface SidebarLink {
 	attrs: LinkHTMLAttributes;
 }
 
-export interface SidebarGroup {
+export interface SidebarAutoLink extends SidebarManualLink {
+	autogenerated: true;
+}
+
```
delucis, line 33: @HiDeoo Did you have an intended use case for this new property? We didn’t distinguish autogenerated groups from other groups previously, so maybe we can just keep treating all links the same once they get generated and omit this property?

I see this mentioned in the PR comment but I don’t understand _why_ plugins would need access to this data.

packages/starlight/utils/error-map.ts
```diff
@@ -107,6 +107,12 @@ const errorMap: z.core.$ZodErrorMap = (issue) => {
 			for (const unionError of issue.errors) {
 				const expectedShape: string[] = [];
 				for (const issue of unionError) {
+					// We sometimes use `z.never()` to explicitly error on certain property combinations in
```
delucis: Nit: isn’t it `z.NEVER`?
```suggestion
					// We sometimes use `z.NEVER` to explicitly error on certain property combinations in
```

delucis (commented):
I updated the branch to be aligned with `main` and noted a few things. Most were notes I can use myself, but one question would be good to get feedback on from you @HiDeoo!

docs/src/content/docs/guides/sidebar.mdx
delucis: I actually think we can omit it:

```suggestion
```

docs/src/content/docs/guides/sidebar.mdx
delucis: Done in 009b6ea00d47a7a7f78fa12a888a142653c723aa

packages/starlight/utils/routing/types.ts
HiDeoo: > We didn’t distinguish autogenerated groups from other groups previously

We did not explicitly, but route data middlewares could still infer which entries were autogenerated because the sidebar configuration still mapped predictably to the route data structure.

Before this refactor, it was fairly easy to infer which route data entries were autogenerated or not:

```
sidebar config                  route data
---------------------------     --------------------------------------------
manual link                     link
manual group                    group
autogenerated group   ------>   group  (everything inside is autogenerated)
manual link                     link
```

Now, we no longer have that implicit mapping. With this refactor, autogenerated entries can expand into multiple links and subgroups and can be interleaved with manually authored entries, so a similar approach no longer works.

```
sidebar config                  route data
---------------------------     --------------------------------------------
manual link                     link
manual group                    group
autogenerated item   ------->   group
                                link
                                link
                                link
manual link                     link
autogenerated item   ------->   link
                                group
                                link
```

This would still technically be possible but would require partially re-implementing a lot of Starlight's logic just to identify which groups/links are autogenerated and which are not, which feels fragile, error-prone, and not a pleasant experience imo.

> I see this mentioned in the PR comment but I don’t understand why plugins would need access to this data.

Considering the above, if a plugin wanted to apply some transformation only to autogenerated entries, this becomes a lot more difficult without some indication in the route data of which groups/links were autogenerated and which were manually authored. It could be a very simple plugin changing the label of autogenerated groups for example, or a more advanced plugin like the [`starlight-auto-sidebar`](https://github.com/HiDeoo/starlight-auto-sidebar/) plugin where provided features are only applicable to autogenerated links/groups.

Hope that clarifies a bit the motivation behind this change, but let me know if you have any other questions, suggestions, or want to discuss further.

delucis: Thanks for the explanation! I don’t love adding this extra metadata not directly related to the rendering view — or maybe feel it should be _more_ to be useful? e.g. if you have two groups next to each other how do you distinguish them? In which case should they include e.g. the autogeneration directory as metadata or something?

But I’m also fine with rolling with this for now if you’re confident it covers the known use cases.

delucis (commented):
I think this is in pretty good shape!

There’s this note-to-self https://github.com/withastro/starlight/pull/3618#discussion_r2602177057 that you might want to revisit @HiDeoo?

Otherwise, unless you think we should add any additional `autogenerate` metadata to the sidebar links structure, I’d love to get this out in the next minor release 🚀

packages/starlight/utils/routing/types.ts
HiDeoo: My first draft was to initially make it work without breaking anything, and the smallest amount of change possible.

I'll test later with a bit more metadata for the cases I have and see if it actually helps or not, and report back :thumbsup:

HiDeoo: As the `starlight-auto-sidebar` plugin has evolved quite a lot since my last local attempt to make it compatible with this branch, I ended up starting from scratch and re-updating again the plugin entirely with the new changes from this PR and also test your suggestion at the same time to attach a bit more than just a boolean.

The difference between a simple boolean and the autogenerated directory is a huge improvement, allowing for a lot of simplification, fixing a few issues, e.g. entirely supporting custom sidebar containing autogenerated entries in custom pages (which is supported in Starlight but was not in the plugin), some edge cases and making the plugin more robust in general.

I don't think a lot of plugins would need a lot more than just knowing if something in the sidebar is autogenerated or not, but if ever needed or for a plugin as involved in sidebar tweaking as the `starlight-auto-sidebar` one, this is definitely a huge improvement.

> I don’t love adding this extra metadata not directly related to the rendering view — or maybe feel it should be more to be useful?

On the first part, while I understand the concern and is something we've never done before in Starlight, I don't really think it's an issue if the new metadata provides real value, allow to support a use-case that is otherwise not really possible or would be really hard to support, even if Starlight itself default rendering-path don't use it.
On the second part, this is definitely the case now with your suggestion.

As it's getting late here and have worked on this the entire afternoon, I'll sleep on it and will share the code tomorrow, also add changeset/documentation for this (as this was not the case before as I was still exploring the idea), and we can discuss more about it if needed after that. (I'll also take a look at the remaining note)

delucis: OK, sounds good! Thanks for digging into it. And yeah, happy for us to go with something providing a more functional API if possible rather than adding just a flag and leaving people still having to hack around.

HiDeoo:
Pushed a new update adressing [this comment](https://github.com/withastro/starlight/pull/3618#discussion_r3181432058) and refactoring the previously exposed `autogenerated: boolean` property on route data to an `autogenerate` object with a `directory` property matching the configured autogenerated directory so we can discuss it with a concrete example.

I also included in a dedicated commit some changes adressing this [previous note](https://github.com/withastro/starlight/pull/3618#discussion_r2602177057) regarding type error location. I did that in a dedicated commit so it can be easily reverted as I think it may be getting out of hand and trying to fight TypeScript too much to get best possible error and locations. Not sure if the extra complexity is worth for a workaround that took me a few hours to even figure out and implement. We already have what I would qualify as quite good runtime validation with detailed error messages, and even migration steps. It also feels a bit fragile as it's relying on some specific TypeScript behavior that is not guaranteed to be stable.

packages/starlight/__tests__/basics/navigation-labels.test.ts
```diff
@@ -24,6 +24,9 @@ describe('getSidebar', () => {
 			[
 			  {
 			    "attrs": {},
+			    "autogenerate": {
+			      "directory": "",
```
delucis, line 28: Am I right to understand that the empty string value is for the special case of an undefined `sidebar` config which is implicitly autogenerated? I wonder if it should have some special value like `"."`? Or maybe not if it’s more helpful for people to do be able to do `if (!autogenerate.directory)` instead of `if (autogenerate.directory === '.')`.

Either way, we need to document this in the route data docs a bit better. I was very lazy adding [those docs](https://starlight.astro.build/reference/route-data/#sidebar) so all it says is “Type: `SidebarEntry[]`” which… is not very helpful 😅

(Probably out of scope in this PR, but I guess supporting a special `"."` value also in _user_ config would allow users to mix full autogeneration with other links more easily? e.g. `sidebar: [{ link: '/', label: 'Home' }, { autogenerate: { directory: '.' } }]`)

packages/starlight/utils/navigation.ts
```diff
@@ -348,10 +387,20 @@ function sidebarFromDir(
...
+		dirToItem(
+			dirOrRoute,
+			key,
+			key,
+			currentPathname,
+			locale,
+			collapsed,
+			attrs,
+			autogenerate ?? { directory: '' }
+		)
```
delucis: I wonder if it might be time to convert some of these functions to an options object instead of positional arguments 😅

Very opaque to think oh yeah, `key, key,` of course we pass that those as 2nd and 3rd arguments. (Not the fault of this PR — just an accumulation of us adding one more arg each time.)

delucis (commented):
Thanks for the updates @HiDeoo!

Re: the Typescript stuff — I don’t have too strong an opinion. It’s definitely getting very _arcane_ and hard to understand with these subtle things like `void` vs `never`, but I’ll let you decide if you want to keep it or not.

packages/starlight/__tests__/basics/navigation-labels.test.ts
HiDeoo: > Am I right to understand that the empty string value is for the special case of an undefined `sidebar` config which is implicitly autogenerated? I wonder if it should have some special value like `"."`? Or maybe not if it’s more helpful for people to do be able to do `if (!autogenerate.directory)` instead of `if (autogenerate.directory === '.')`.

Yeah, this is correct and one case where I was not entirely sure if it was the correct approach so used something matching slug for now so I can push the PR and we can discuss it there. I think all the approaches possible are:

1. `autogenerate.directory = ''` (current approach matching slug)
2. `autogenerate.directory = '.'`
3. `autogenerate = {}` (your idea if I understand correctly)

Intuitively, I would have preferred 1 or 3 but considering:

> (Probably out of scope in this PR, but I guess supporting a special `"."` value also in _user_ config would allow users to mix full autogeneration with other links more easily? e.g. `sidebar: [{ link: '/', label: 'Home' }, { autogenerate: { directory: '.' } }]`)

Then I think 2 would be the best approach long term?

I guess indeed a bit out of scope based on the initial goal of the PR but considering there are long term implications and some relations between the two and that it should not be too much work, maybe there is a world where we should do it at the same time? Not sure, at least maybe worth trying it out in a PoC in this PR to experiment and see if one approach ends up being better than the other?

> Either way, we need to document this in the route data docs a bit better. I was very lazy adding [those docs](https://starlight.astro.build/reference/route-data/#sidebar) so all it says is “Type: `SidebarEntry[]`” which… is not very helpful 😅

Already started but I wanted to push the actual implementation first so we can discuss it asap.

packages/starlight/utils/navigation.ts
HiDeoo: Noticed the same thing, specially when the formatter switched it to multiple lines instead of one.

Will be addressed in a future commit.

packages/starlight/__tests__/basics/navigation-labels.test.ts
delucis: Ah, I actually hadn’t considered `autogenerate = {}` but I think `directory` always being defined in _some_ way makes sense because there’s always a directory even if it’s the implicit root one. So the question is just, is the root `''` or `'.'` (or some `Symbol` constant we export I guess would be a third option).

(In terms of the usage I was just pointing out the difference between the falsy `''` and truthy `'.'` values.)

The most common user interaction would be with the Starlight config version of this so the key thing to decide is probably whether it feels more natural to set `directory: '.'` or `directory: ''` in config in the cases where anyone needs this.

On the one hand `'.'` is quite explicit and common for file paths, but on the other hand the directories for this option generally don’t start with `./` like relative file paths do.

> maybe worth trying it out in a PoC in this PR to experiment

We _could_ but I also don’t think the PoC is required to decide which config feels best necessarily. So here we’d just decide what future config feels best and align the route data shape with it.

One nice thing that might be possible with this future is we can probably remove special case code for the empty sidebar case and just have the Zod schema do `.default([{ autogenerate: { directory '.' } }])`.

HiDeoo:
Pushed a few updates:

- Removed the TypeScript changes. While it refined the type error locations, I think it made the error even more obscure imo. The `never` type was already probably confusing but more common than `void` and could still guessed by just reading the error message. With the refactor, reading that something is "not assignable to type 'void'" felt less intuitive. In the end, I think our runtime error message will provide the best guidance with exact location.
  
  SCR-20260505-lhnp

- Updated the configuration reference documentation which was still not updated
- Documented `Astro.locals.starlightRoute.sidebar` in the route data reference page. Right now, I just started by inlining everything in details as a base and we can later restructure/shorten/move it if needed.
- Did some refactor in `utils/navigation.ts` as pointed out in [this comment](https://github.com/withastro/starlight/pull/3618#discussion_r3187597960). I also found some old data that we were still passing around from before sidebar intermediate representations, so I cleaned those up as well.

I also played a bit with [the idea](https://github.com/withastro/starlight/pull/3618#discussion_r3187766864) of allowing either `directory: '.'` or `directory: ''` in the config (not included). After implementing a basic version and actually trying out the `directory: '.'` option by typing it in my Astro config, I found it more confusing than just using `directory: ''`: 

- `.` usually refers to the current directory, so while editing the config, my current directory was the project root but the directory we would be autogenerating from is the `src/conten/docs/` directory.
- If `.` is supported, can I use `..`?

Only my opinion of course, and happy to discuss further, but I kept `directory: ''` for now since it felt more intuitive to me.

.changeset/00-slow-dragons-notice.md
delucis, line 1: Thought: maybe we should rename this changeset to prefix it with `00-` so that this change appears first in the release notes. It’ll be by far the biggest impact change in the release.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -149,7 +149,7 @@ Groups of links are expanded by default. You can change this behavior by setting
 
 Autogenerated subgroups respect the `collapsed` property of their parent group by default. Set the `autogenerate.collapsed` property to override this.
```
delucis: Wondering if this behaviour should change to be more like regular groups? I don’t think a regular group auto-collapses if the parent group is collapsed? So if autogenerated links are more like adding a group, should they behave the same, so not collapsed unless `collapsed: true` is set?

docs/src/content/docs/reference/route-data.mdx
```diff
@@ -100,9 +100,102 @@ Learn more about the shape of this object in [Astro’s Collection Entry Type](h
 
 ### `sidebar`
 
-**Type:** `SidebarEntry[]`
+**Type:** <code>(<a href="#sidebarlink-type">SidebarLink</a> | <a href="#sidebargroup-type">SidebarGroup</a>)[]</code>
```
delucis: Definitely a nit, but for types like these, the `Array<...>` style might be a bit easier to read?

```suggestion
**Type:** <code>Array&lt;<a href="#sidebarlink-type">SidebarLink</a> | <a href="#sidebargroup-type">SidebarGroup</a>&gt;</code>
```

docs/src/content/docs/reference/route-data.mdx
```diff
@@ -100,9 +100,102 @@ Learn more about the shape of this object in [Astro’s Collection Entry Type](h
...
+Badge configuration to render next to the group label, if any.
+See the [`<Badge>` component props reference](/components/badges/#badge-props) for details.
+
+##### `autogenerate`
+
+**Type:** `{ directory: string } | undefined`
+
+For groups created from an [autogenerated sidebar entry](/guides/sidebar/#autogenerated-links), `directory` matches the configured [`autogenerate.directory`](/reference/configuration/#sidebar) value.
+
+When the [`sidebar`](/reference/configuration/#sidebar) is not configured, Starlight generates entries for all documentation pages and uses an empty string (`''`) as the `directory` value.
```
delucis, line 198: I wonder if there’s a good way to avoid the duplication of these sections? 🤔 

I considered a “common properties” section containing `type`, `badge`, and `autogenerate`, but I guess it might just be more confusing? And would take away the ease of linking to `SidebarLink`/`SidebarGroup` as a comprehensive type.

So maybe not and we just have to tolerate the duplication, remembering to keep them in sync.

packages/starlight/schemas/sidebar.ts
```diff
@@ -50,31 +50,65 @@ const SidebarLinkItemSchema = z.strictObject({
...
+			// TODO: not supported by Docusaurus but would be good to have
+			/** How many directories deep to include from this directory in the sidebar. Default: `Infinity`. */
+			// depth: z.number().optional(),
+		}),
+	})
+	.strict()
+	.superRefine((config, ctx) => {
+		if (!('label' in config)) return;
+
+		ctx.addIssue({
```
delucis, line 81: Might be worth adding a TODO here so it’s easy to find come v1 or whenever.

```suggestion
		// TODO: Remove this error message in a future release once most users have migrated
		ctx.addIssue({
```

delucis (commented):
Beautiful work as always @HiDeoo!

I agree that `''` for the root directory option works. It feels natural to go from `'dir/sub-dir'` => `'dir'` => `''` and, like you say, this avoids the hint of the relative path syntax meaning current directory and suggesting `'..'` or `'./dir'` might work.

Left a few comments — mostly pretty minor, although the `collapsed` question might warrant some discussion?

.changeset/00-slow-dragons-notice.md
HiDeoo: Done in [`11802db` (this PR)](https://github.com/withastro/starlight/pull/3618/commits/11802db856de298a35b58399181f1124f4876a7e)

docs/src/content/docs/reference/configuration.mdx
HiDeoo: We definitely could, and personally, most of the time, no matter what, I avoid the "cascade" behavior and manually set `autogenerate.collapsed` when I deal with collapsed groups.

> I don’t think a regular group auto-collapses if the parent group is collapsed?

Correct.

I guess the question may be more about the timing:

- Now, as part of this PR, so all autogenerated changes happen at the same time for users.
- Later, e.g. a v1 behavior change, so users only have to deal with the current changes right now, and some other changes later to avoid overwhelming them with too many changes at once.

docs/src/content/docs/reference/route-data.mdx
HiDeoo: Good call, I also updated the same case below in the page.

docs/src/content/docs/reference/route-data.mdx
HiDeoo: Yeah, when writing it I tried a few variations but ended up reverting to the duplicated version as I didn't really like alternatives I tried. I'll try to see if I can come up with something :thinking:

packages/starlight/schemas/sidebar.ts
HiDeoo: Done in https://github.com/withastro/starlight/pull/3618/changes/33fe43945b45b4fd7d420c0e14f0809a17fb6214

docs/src/content/docs/reference/configuration.mdx
delucis: Is it annoying to support currently passing down the parent state? That would be one reason to pull the plug now and say “because of this change, it’s no longer inherited”.

It also might be better to do sooner rather than later so we’re not in an in between place where someone could do:

```js
sidebar: [
  {
    label: 'Collapsed group',
    collapsed: true,
    items: [
      { autogenerate: { directory: 'examples' } }, // collapsed by default
      { label: 'Nested group', items: [...] }, // not collapsed by default
    ],
  },
],
```

Won’t be super common and obviously no-one has this _currently_, but would definitely be a case where our defaults are contradictory.

HiDeoo: > Is it annoying to support currently passing down the parent state?

Just pushed [`ad5286c` (this PR)](https://github.com/withastro/starlight/pull/3618/commits/ad5286c99a4d30a2f8a992e45eaac217089c09b9) which should be a better answer.

After writing the changeset for this and emphasizing that it's only 1 use-case that needs updating and how close the extra small change is to the other changes required for this PR, I think we should just do it with this PR too.

docs/src/content/docs/reference/route-data.mdx
delucis: I’m happy to stay with this duplicative approach for now. It has a clear structure 👍

delucis (approved):
Perfect! Seems like a good cleanup indeed.

I think this is looking ready for release! Awesome work figuring out this new approach to make things more flexible for users @HiDeoo 👏

HiDeoo:
Thanks for the thorough review and super helpful suggestions as always :raised_hands:

PR #3620, opened 2025-12-09
Fix preview releases

HiDeoo:
#### Description

This PR fixes our [currently broken](https://github.com/withastro/starlight/actions/runs/20076697024/job/57592861853?pr=3534) preview release workflow by ensuring a full Git history is fetched during the checkout step.

While shallow clones are [extended during the release process](https://github.com/changesets/changesets/blob/3cb9982b96aef8e2b46860258d56dbd7a2a02953/packages/git/src/index.ts#L58-L64), this code path seems to not be hit when running `changeset status`. Here is a side-by-side comparison of running that command in a full clone vs a shallow clone:

image

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
+          # While releasing with changesets works with a shallow clone, this is not the case of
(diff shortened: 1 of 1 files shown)


delucis (approved):
Continuing the tradition of a GitHub Actions PR not working the first time 🫡 

Thanks for testing and fixing @HiDeoo!

trueberryless:
> Continuing the tradition of a GitHub Actions PR not working the first time 🫡

To be fair, this is a really tricky / nasty one as it seems like this should be handled by changeset. But TIL! Thanks HiDeoo for the quick investigation 🚀

PR #3623, opened 2025-12-10
Fix preview releases

HiDeoo:
#### Description

Feels like "déjà vu", but this PR is a new attempt to get preview releases working. I noticed 2 issues:

- In the `Get preview packages` step (which we've never hit before), I noticed while reading the code that I improperly named a variable.
- While [reading the changesets CLI code](https://github.com/changesets/changesets/blob/main/packages/cli/src/utils/versionablePackages.ts#L16), I noticed that the comparison is is done using a `ref` that is [defined](https://github.com/changesets/changesets/blob/main/packages/cli/src/commands/status/index.ts#L35-L36) if the [`--since` flag](https://github.com/changesets/changesets/tree/main/packages/cli#status) is used (or if the deprecated `--sinceMaster` is used). This is not our case and when it's not defined, it fallbacks to [`config.baseBranch`](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#basebranch-git-branch-name). After checking out, the Astro repo [sets it to `origin/main`](https://github.com/withastro/astro/blob/main/.changeset/config.json#L7), while our repo uses `main`.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,7 +7,7 @@
   "commit": false,
   "linked": [],
   "access": "public",
-  "baseBranch": "main",
+  "baseBranch": "origin/main",
   "updateInternalDependencies": "patch",
   "ignore": ["starlight-docs", "@example/*", "starlight-file-icons-generator", "@e2e/*"],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {


trueberryless:
Just for traceability purposes, I am gonna quote HiDeoo here regarding the reason for `origin/main` instead of `main`:

> For the `baseBranch`, yeah, we need it for preview releases which [run `pnpm changeset status`](<https://github.com/withastro/starlight/blob/main/.github/workflows/preview-release.yml#L46>) which basically ends up running `git merge-base <baseBranch> HEAD`. In pull requests,` actions/checkout` checks out a detached PR commit instead of creating a local `main` branch, but in the pre-release workflow, we use `fetch-depth: 0` so remote tracking branches are available so `origin/main` works. 

He kindly explained it on [Discord](https://discord.com/channels/830184174198718474/1540371437473234964/1540413008302448730).

PR #3624, opened 2025-12-10
Fix preview releases (again)

HiDeoo:
#### Description

This one is pretty self-explanatory and makes sense in this context, my brain used yaml-style comments while this script is actually run in a JavaScript context 🙈


delucis (approved):
I’m sure this will work `#blessed`

PR #3625, opened 2025-12-10
Fix preview release (maybe)

HiDeoo:
#### Description

I should be forbidden of creating workflows…


delucis (approved):
IF WE UPPERCASE EVERYTHING IT WILL DEFINITELY WORK THIS TIME

PR #3649, opened 2026-01-06
docs: add community resource - gated content blog from Liran Tal

trueberryless:
#### Description

Add link card entry to community resources (recipes and guides) for [this blog from Liran Tal](https://lirantal.com/blog/customizing-astro-starlight-sidebar-gated-content-auth).


HiDeoo:
Thanks for the submission.

It looks like the article is a bit outdated (noticing the old `social` config and Starlight overrides relying on `Astro.props` instead of route data). While it's still a great resource to link to with other links in a conversation or support thread, I think resources added to the docs should be a bit more up-to-date, or at least updated to the current version at the time of the addition.

As this is a great resource, I think in this case, a better approach would be to reach out to the author and see if they would be open to updating the article to work with more recent versions of Starlight, and let them resubmit the article once it's updated if they want to.

trueberryless:
That's a fantastic idea, I'll add it to my to-do list to contact them 👍

trueberryless:
@HiDeoo Fantastic news: Liran updated his blog post 🙌 

The title and link are the same, and IIRC the description has not matched the post's description before and is custom to match the style of other resources.
I think we can reopen this existing PR 👍

PR #3652, opened 2026-01-11
feat: :sparkles: add catchintent starlight docs for showcase

AkashRajpurohit:
#### Description

- Added CatchIntent.com/docs on showcase sites for starlight (ref: https://x.com/akashwhocodes/status/2009929185863823711)


AkashRajpurohit:
Hi @HiDeoo, updated the image and the pipeline seems to have passed now.

PR #3656, opened 2026-01-15
Fix table of contents intersection observer

delucis:
#### Description

This PR fixes an issue with Starlight’s intersection observer used for updating table of contents state, which we discovered while reviewing #3382 during Talking & Doc’ing today.

We found multiple issues:
- After we added anchor link support, headings in sites with this enabled are wrapped in a `<div>` breaking an assumption in the table of contents logic.
- After moving Markdown styles to an external CSS file, we renamed `.content` to `.sl-markdown-content`, but forgot to rename the same in a query selector here.

This caused some subtle issues that were not super obvious, but we noticed them while digging deeper.

This PR also adds some Playwright tests to make sure highlighting behaves as expected in the browser.


trueberryless:
Awesome stuff here 🙌 

Regarding the already amazing new test cases, I wonder if we also should add tests for changing the default heading levels, or is this already implicitly covered with test cases like [this](https://github.com/withastro/starlight/blob/51862b71f93bf51c29f9d3dcea94d74100ee98a5/packages/starlight/__tests__/basics/toc.test.ts#L62)?

I will try to take a closer look and approve the changes in the next few days, but you don't need to wait as Florian just approved as well 😅

ematipico (approved):
Unfortunately I don't much context, but if the tests pass, i'm good

delucis:
> Regarding the already amazing new test cases, I wonder if we also should add tests for changing the default heading levels, or is this already implicitly covered with test cases like [this](https://github.com/withastro/starlight/blob/51862b71f93bf51c29f9d3dcea94d74100ee98a5/packages/starlight/__tests__/basics/toc.test.ts#L62)?

Yeah, I think we can be confident that the unit test there is making sure we use the right min/max levels when generating the ToC. And we just pass those down to the custom element. **If** we had more logic involved on the client to extract that data, then maybe it would be good to test, but the existing tests should cover any case where we break the `data-min-h`/`data-max-h` attributes.

PR #3664, opened 2026-01-18
i18n(de): Fix orthographic errors in `getting-started.mdx` and `index.mdx`

randomguy-2650:
#### Description

This PR updates some “quotes” to be consistent with the rest of the document. Also removes a redundant (optional) comma and fixes a Deppenleerzeichen.

Also a note on some A+ thingy that I am unsure of.

### DIFF
--- a/docs/src/content/docs/de/index.mdx
+++ b/docs/src/content/docs/de/index.mdx
@@ -138,7 +138,7 @@ import Testimonial from '~/components/testimonial.astro';
     handle="J_Everhart383"
     cite="https://twitter.com/J_Everhart383/status/1691900590048292908"
   >
-    Ich habe die WPEngine Atlas Platform Dokumentation neu aufgebaut. Vertrau mir, wenn ich sage, dass Starlight alles hat, was du brauchst, um eine A+ Doku-Plattform zu machen&nbsp;🙌
+    Ich habe die WPEngine Atlas Platform Dokumentation neu aufgebaut. Vertrau mir, wenn ich sage, dass Starlight alles hat, was du brauchst, um eine hervorragende Doku-Plattform zu machen&nbsp;🙌
   </Testimonial>
   <Testimonial
     name="Chloe"


docs/src/content/docs/de/index.mdx
randomguy-2650:
```suggestion
    Ich habe die WPEngine Atlas Platform Dokumentation neu aufgebaut. Vertrau mir, wenn ich sage, dass Starlight alles hat, was du brauchst, um eine hervorragende Doku-Plattform zu machen&nbsp;🙌
```

A+ is just what’s used in the U.S., so I’m not sure if it makes sense. Maybe it makes sense to use 1+.

In _Oberstufe_, you’d (at least in my experience) use the 15–0 system, so maybe 15?

But if you would prefer to avoid this numbering system in the translation altogether, then maybe “erstklassig” or “hervorragend” could work.

**Side note:** It seems like this stupid suggestion thing isn’t fixed yet.

---

<div lang="de">
<p>A+ wird hauptsächlich nur in den Vereinigten Staaten benutzt, also bin ich mir nicht sicher, ob das Sinn macht. Vielleicht passt 1+ besser hin.</p>

<p>In der Oberstufe benutzt man (zumindest nach meiner Erfahrung) das 15 bis 0‐System, also vielleicht 15?</p>

<p>Aber wenn du dieses Nummerierungsystem in der Übersetzung eher vermeiden würdest, könnten wir möglicherweise „erstklassig“ oder „hervorragend“ nutzen.</p>

<p><strong>P.S.:</strong> Irgendwie ist dieses nervige Suggestions‐Bug immer noch nicht gefixt worden …</p>
</div>

docs/src/content/docs/de/index.mdx
```diff
@@ -41,7 +41,7 @@ import Testimonial from '~/components/testimonial.astro';
 		mit integrierte Frontmatter-Validierung.
 	</Card>
 	<Card title="Bringe deine eigenen Komponenten mit" icon="puzzle">
-		Starlight wird als Framework-unabhängige, vollständige
+		Starlight wird als frameworkunabhängige, vollständige
```
trueberryless, line 44: Not sure on this one, I found [many references](https://context.reverso.net/translation/german-english/Framework-unabh%C3%A4ngig) regarding the old translation, but your suggestion seems right [as well](https://context.reverso.net/translation/german-english/frameworkunabh%C3%A4ngig). I guess we can change it because it's an adjective 👍

docs/src/content/docs/de/index.mdx
trueberryless: I didn't know about the "15 numbering system" until now, so I'm opposed to it as I guess many other also wouldn't know how to interpret a "more or less random" 15 in the middle of this comment.

I really like your suggestion of rewriting it to something like "hervorragende" if this is not used in the surrounding comments...

What do you think about this: "…was du brauchst, um eine **1A** Doku-Plattform zu machen…" Do you know this phrase or is this maybe just something that is known in my area?

randomguy-2650: Would you mind clearing up what 1A means? I personally have never seen it used before.

Edit: Missed your ending part. It’s likely something in Austria as I’ve never heard of it.

Edit 2: I tried asking an AI about it because Google didn’t seem to find anything about it. It supposedly comes from Gütekategorien where 1A has no flaws (maybe not true, everything has at least some kind of flaw!), but take this with a grain of salt!

trueberryless: Okay, this might be something Austrian then 😅

Let's go with "hervorragende" then 👍

randomguy-2650:
Should be ready to go by now.

randomguy-2650 (commented):
I can’t review properly for some reason but should we keep “Monorepo” on line 92?

trueberryless:
> I can’t review properly for some reason but should we keep “Monorepo” on line 92?

I also wouldn't translate this, as "repository" is also not translated, and the "mono" definitely doesn't make it easier to find a nice German word. Closest we could do is something like: `Monorepo (ein zentrales Repository für mehrere Projekte)`, but that's a little bit much, isn't it?

But feel free to make a PR adding "monorepo" to our [i18n-guide glossary](https://github.com/withastro/docs/blob/main/i18n-guides%2Fdeutsch.md#%C3%BCbersetzungs-glossar) so we can globally define the (not)-translation 👍

PR #3666, opened 2026-01-19
update hero `actions` to handle `base` prefix

mayank99:
#### Description

This is a one line fix for the issue with `actions` described in https://github.com/withastro/starlight/discussions/3660.

I tested it in my project by patching `@astrojs/starlight` and confirming that BASE_URL is included in the home page CTA.

### DIFF
--- a/.changeset/nervous-fireants-smash.md
+++ b/.changeset/nervous-fireants-smash.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fixes `hero.actions` links to respect the `base` path (if configured).


delucis:
Oh, sorry, I think I maybe wasn’t clear in my reply in #3660 — we generally consider the URLs passed to `actions` to be “user-authored links” (equivalent to e.g. a Markdown link) and so don’t process them for the same reason we don’t for Markdown.

I don’t think we could make this change as a simple patch fix because anyone who is prefixing their `actions` URLs (like me in my many docs sites deployed to GitHub Pages 😅) would end up with links that are _double_ prefixed with the `base` path, breaking them.

Let’s keep discussing in #3660. Maybe there are use cases I didn’t grasp yet which make a manual prefix here impractical.

PR #3667, opened 2026-01-20
fix: Adds `select` element's identifier

oluwatobiss:
This PR adds a `name` attribute on the `select` element in `starlight/components/Select.astro` to fix the browser’s `form field element should have an id or name attribute` warning.

form field browser warning issue

### DIFF
--- a/.changeset/fresh-pigs-pump.md
+++ b/.changeset/fresh-pigs-pump.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Adds `name` attribute on `select` element to fix the browser's `form field element should have an id or name attribute` warning


delucis:
Thanks for the PR @oluwatobiss! However, this change is not actually needed.

The informational tip shown by Chrome there is saying that a `name` or `id` could be important if you want to autocomplete forms correctly, but even says “that this is not strictly needed”.

Because these `<select>` elements are not used in the context of a genuine `<form>` flow where autocomplete might make sense (e.g. picking a country from a dropdown when filling out an address), the advisory is not relevant to Starlight. In fact, to improve cross-browser behaviour we even had to add an explicit `autocomplete="off"` on these menus to avoid any autocomplete behaviour interfering.

So long story short, the browser dev tools are assuming this might be a form and you might _want_ autocomplete, but for these menus this is not the case and no changes are required.

Thanks nevertheless for letting us know you hit this. We’ll keep it in mind if anyone else mentions it.

oluwatobiss:
Thank you for checking the PR, @delucis. Although the `name` attribute is not strictly required, its inclusion enhances user experience by reducing unnecessary DevTools warnings, especially since `autocomplete="off"` already disables autocomplete.

Users should not encounter avoidable or non-essential warnings. Therefore, this PR aims to eliminate unnecessary messages from the browser console, so users don’t have to worry about irrelevant warnings.

delucis:
Hi @oluwatobiss, I do understand that concern, but there are a lot of different tools like this and we can't accommodate every audit's inaccurate analysis. 

The screenshot you shared suggests it's not even a warning level message for Chrome, but just an informational note.

PR #3672, opened 2026-01-26
 improves Starlight performance by implementing lazy loading 

Gorniv:
This PR significantly improves Starlight performance by implementing lazy loading for two major bundles that previously impacted initial load times and rendering metrics:

1. Lazy Load Starlight Built-in Translations: (Commit 89a0c52) Moved from eager to lazy loading for built-in translation JSON files. Only English is loaded by default for type inference and as a fallback, while other languages are dynamically imported only when required by the project locale. This reduces the server-side memory footprint and the size of the initial localization bundle.

2. Lazy Load Pagefind UI: Optimized the Search.astro component to defer the loading of the @pagefind/default-ui bundle (~73kB). Previously loaded on DOMContentLoaded, it now loads only when a user first interacts with the search button or shortcut (Ctrl+K). A prefetch hint has been added to ensure the search UI remains responsive when needed.

**What does this PR change**? Implements on-demand loading for Starlight translations and the Pagefind search interface to optimize Core Web Vitals.

**Did you change something visual**? No. The UI remains identical, but the background loading logic is now significantly more efficient.

### DIFF
--- a/packages/starlight/__tests__/i18n-non-root-single-locale/translations-with-user-config.test.ts
+++ b/packages/starlight/__tests__/i18n-non-root-single-locale/translations-with-user-config.test.ts
@@ -9,9 +9,9 @@ vi.mock('astro:content', async () =>
 );
 
 describe('useTranslations()', () => {
-	test('uses user-defined translations', () => {
+	test('uses user-defined translations', async () => {
 		const t = useTranslations('fr');
 		expect(t('page.editLink')).toBe('Modifier cette doc!');
-		expect(t('page.editLink')).not.toBe(translations.fr?.['page.editLink']);
+		expect(t('page.editLink')).not.toBe((await translations.fr?.())?.['page.editLink']);
 	});
 });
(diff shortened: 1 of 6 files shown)


delucis:
Hi @Gorniv — as described in our PR template, we ask contributors to first [open a discussion](https://github.com/withastro/starlight/discussions/categories/feature-requests) (or an issue for bugs) before moving to the PR stage:

https://github.com/withastro/starlight/blob/879d6b347ffca43e98afae30417582c3921a0c8f/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L5-L7

This avoids losing time in the implementation phase, especially if there is something that may never be accepted in the first place. Given this, I’ll close this PR.

If you have issues with server-side memory caused by the translation files please open an issue to describe this. In practice, I would be surprised if this is an important issue — the entire translations directory is only 55 KB currently.

The search bundle is intentionally loaded on `DOMContentLoaded` so it is available quickly for users opening the search modal. We chose this to delay the load enough that it doesn’t block higher priority requests, but still maintain responsiveness. I don’t think we’re interested in changing this.

PR #3674, opened 2026-01-28
Allow passing user options to `pagefind.createIndex`

controversial:
#### Description

This PR allows the user to specify options in `pagefind.index` that will be passed through to `pagefind.createIndex`, notably:
- [`excludeSelectors`](https://pagefind.app/docs/config-options/#exclude-selectors), which is useful for excluding content from pagefind where the user does not control the markup to add `data-pagefind-ignore`. Configuring this option is the main motivation for this PR.
- [`rootSelector`](https://pagefind.app/docs/config-options/#root-selector), which could be useful if anyone wants to change the default placement of `data-pagefind-body` from its default placement on `<main>`.
- [`forceLanguage`](https://pagefind.app/docs/config-options/#force-language) which seemed like it wouldn’t hurt to include but I don’t have a specific motivation for it, and can remove it if it seems it would work against Starlight’s locale handling.

I excluded the ability to configure logging options here, since they don’t affect the output index, and I excluded the config option for including `/index.html` paths since it doesn’t seem useful within Starlight.

These options weren’t practical to expose in Starlight before switching to the JS API in #3534 but now it seems reasonable to make them configurable.

This is a more generic solution to allow users to address problems like the one described in #3676

---

I’m eager for suggestions on how to include unit tests for this functionality, at a glance it seems like the existing unit tests mock most of the relevant functionality for pagefind, so I haven’t added tests here yet as it didn’t seem straightforward

### DIFF
--- a/packages/starlight/__tests__/basics/pagefind.test.ts
+++ b/packages/starlight/__tests__/basics/pagefind.test.ts
@@ -15,7 +15,7 @@ async function runStarlightPagefind(outputDir: URL) {
 
 	vi.mocked(pagefind.createIndex).mockResolvedValue({ index, errors: [] });
 
-	await starlightPagefind({ dir: outputDir, logger: new TestAstroIntegrationLogger() });
+	await starlightPagefind({ dir: outputDir, logger: new TestAstroIntegrationLogger() }, {});
 
 	return index;
 }
@@ -61,7 +61,7 @@ test('logs Pagefind errors and closes Pagefind', async () => {
 		errors: [errorMessage],
 	});
 
-	await expect(starlightPagefind({ dir: outputDir, logger })).rejects.toThrow();
+	await expect(starlightPagefind({ dir: outputDir, logger }, {})).rejects.toThrow();
 
 	expect(logger.error).toHaveBeenCalledWith(`Pagefind error: ${errorMessage}`);
 
--- a/packages/starlight/integrations/pagefind.ts
+++ b/packages/starlight/integrations/pagefind.ts
@@ -4,10 +4,10 @@ import * as pagefind from 'pagefind';
 import type { StarlightConfig } from '../types';
 
 /** Run Pagefind to generate search index files based on the build output directory. */
-export async function starlightPagefind({
-	dir,
-	logger: starlightLogger,
-}: PagefindIntegrationOptions, userConfig: Extract<StarlightConfig['pagefind'], object>) {
+export async function starlightPagefind(
+	{ dir, logger: starlightLogger }: PagefindIntegrationOptions,
+	userConfig: Pick<Extract<StarlightConfig['pagefind'], object>, 'index'>
+) {
 	const logger = starlightLogger.fork('starlight:pagefind');
 	const options = { dir, logger };
 


packages/starlight/components/Search.astro
```diff
@@ -75,6 +75,7 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 
 <script>
 	import { pagefindUserConfig } from 'virtual:starlight/pagefind-config';
+	const { index: _, ...pagefindClientConfig } = pagefindUserConfig;
```
delucis, line 78: I think this would still end up leaking the index configuration into the client bundle. It might be better to do this here when creating the virtual module so it isn’t included at all:

https://github.com/withastro/starlight/blob/49f153ddd085743ab402fbaeebe350f4b45af292/packages/starlight/integrations/virtual-user-config.ts#L139

And updating the virtual module type needs to be done manually too here:

https://github.com/withastro/starlight/blob/49f153ddd085743ab402fbaeebe350f4b45af292/packages/starlight/virtual-internal.d.ts#L25-L29

delucis:
Thanks for the PR @controversial! We normally ask people first open a [discussion](https://github.com/withastro/starlight/discussions) for features like this to give us a bit of time to chat through the expectations, motivations, use cases etc. and solicit user feedback, which helps us prioritise and understand the value of things.

In this case, I do have some concerns around the usefulness of everything included here. I can see an argument for `excludeSelectors` (although we do already document how to [“Exclude part of a page”](https://starlight.astro.build/guides/site-search/#exclude-part-of-a-page), so I’d like to understand more about the scenarios where that approach is not sufficient). The `rootSelector` feels risky: I’m not quite clear what this would help with. And as you mentioned, the `forceLanguage` setting doesn’t have an obvious usage (I imagine in a basic Pagefind scenario you might have content that doesn’t include `lang` attributes for whatever reason and need to force it, but in Starlight’s case, we already make sure all content is tagged correctly).

We can chat about these here if you like, but given these doubts, it might make more sense to move to a discussion to outline all the use cases etc.

controversial:
I’d be happy to open a discussion!

The primary motivation here was that I was trying to accomplish #3675 in user-land, and was surprised to find that i didn’t have the ability to configure `excludeSelectors`. Having `excludeSelectors` would allow users to fix this type of pagefind indexing issue without waiting for upstream changes, in cases where the user doesn’t control the markup that they’re trying to ignore.

I think that’s useful functionality even when paired with `data-pagefind-ignore`, because you can’t always set `data-pagefind-ignore` where you need it.

I’d be happy to drop the other two parts of this PR (`rootSelector`, `forceLanguage`) as I think the argument to be made for those is less clear

delucis:
> The primary motivation here was that I was trying to accomplish #3675 in user-land, and was surprised to find that i didn’t have the ability to configure `excludeSelectors`.

I guess to me the question is whether there are use cases that wouldn’t be “Fix a Starlight bug”, because if the API mainly allows user-land fixes of bugs on our end, it would be better to fix those bugs upstream rather than requiring individual users to patch things up if they notice. Have you hit other cases where control over `excludeSelectors` would have helped you?

controversial:
I think the other cases would be anywhere that useres are rendering markup that they don’t directly control—this could be markup that Starlight generates, or it could be markup from 3rd-party remark plugins, or from 3rd-party components that are imported and rendered within mdx content. IMO it’s not uncommon to have markup in a page to which it would be difficult to directly attach `data-pagefind-ignore`

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -503,6 +503,7 @@ Set `pagefind` to an object to configure the Pagefind search client:
 
 - See [“Customize Pagefind's result ranking”](https://pagefind.app/docs/ranking/) in the Pagefind documentation for more details about using the `pagefind.ranking` option to control how search result ranking is calculated
 - See [“Searching multiple sites”](https://pagefind.app/docs/multisite/) in the Pagefind documentation for more details about using the `pagefind.mergeIndex` option to control how to search across multiple sites
+- See [“pagefind.createIndex”](https://pagefind.app/docs/node-api/#pagefindcreateindex) in the Pagefind documentation for more details about using the `pagefind.index` option to customize what content is indexed.
```
delucis, line 506: Given we’ll only support `excludeSelectors`, we can probably describe that here in a tiny bit of detail.

docs/src/content/docs/reference/configuration.mdx
```diff
@@ -529,6 +530,10 @@ interface PagefindOptions {
 			termSimilarity?: number;
 		};
 	}>;
+	index?: Pick<
+		pagefind.PagefindServiceConfig,
+		'rootSelector' | 'excludeSelectors' | 'forceLanguage'
+	>;
```
delucis, line 536: I’m not sure it’s useful to document this using a type from Pagefind which people can’t look up. But with the simplification, it’s probably not too bad to show the actual type? (I’m assuming it’s something like this.)
```suggestion
	index?: {
		excludeSelectors?: string[];
	};
```

delucis:
Hi @controversial are you still interested in landing these changes?

controversial:
Yes, I am, sorry that I haven’t been able to work on this for a few weeks! Will try to make these changes today or tomorrow.

delucis:
No worries! Just wanted to make sure if we should close or keep the PR. Take your time 🙌

delucis (approved):
Hey @controversial! I took the time to update the PR with the changes we discussed and also expose the `includeCharacters` option, which was added in the time since we were working on this.

I think it should be ready to go out in our next minor!

PR #3678, opened 2026-01-29
a11y(toc): Change `aria-current` value from `true` to `location`

randomguy-2650:
#### Description

This is more precise in ARIA. While `true` works, `location` is more precise and semantic, just like using `<nav>` instead of a generic `<div>` where appropriate.

#### To‐do

- [ ] Add changeset


randomguy-2650:
Closing pull request because I’ll start a new branch. This is too outdated.

PR #3679, opened 2026-01-29
a11y(toc): Change `aria-current` from `true` to `location`

randomguy-2650:
#### Description

This is more precise in ARIA. While true works, location is more precise and semantic, just like using `<nav>` instead of a generic `<div>` where appropriate.

Although many sites use `aria-current="true"`, WCAG recommends using a more precise option, if one exists, in this case `"location"`.

See [WAI-ARIA 1.2 guidance in section 6.7](https://www.w3.org/TR/wai-aria-1.2/#aria-current).

#### To‐do

- [x] Add changeset

### DIFF
--- a/packages/starlight/components/TableOfContents/TableOfContentsList.astro
+++ b/packages/starlight/components/TableOfContents/TableOfContentsList.astro
@@ -39,7 +39,7 @@ const { toc, isMobile = false, depth = 0 } = Astro.props;
 			padding-inline: calc(1rem * var(--depth) + var(--pad-inline)) var(--pad-inline);
 			line-height: 1.25;
 		}
-		a[aria-current='true'] {
+		a[aria-current='location'] {
 			color: var(--sl-color-text-accent);
 		}
 		.isMobile a {
@@ -58,13 +58,13 @@ const { toc, isMobile = false, depth = 0 } = Astro.props;
 		.isMobile:first-child > li:first-child > a {
 			border-top: 0;
 		}
-		.isMobile a[aria-current='true'],
-		.isMobile a[aria-current='true']:hover,
-		.isMobile a[aria-current='true']:focus {
+		.isMobile a[aria-current='location'],
+		.isMobile a[aria-current='location']:hover,
+		.isMobile a[aria-current='location']:focus {
 			color: var(--sl-color-white);
 			background-color: unset;
 		}
-		.isMobile a[aria-current='true']::after {
+		.isMobile a[aria-current='location']::after {
 			content: '';
 			width: 1rem;
 			background-color: var(--sl-color-text-accent);


randomguy-2650:
The description here by WAI-ARIA 1.2:

> Represents the current location within an environment or context.

In this case, it means where (location) on the same page, unlike the value `page`, which is where on the site.

delucis:
Thanks. I’d understood that was your interpretation, but was curious if you were informed by additional context such as real-world usage patterns, screen reader testing, conversations with impacted users, etc.

ArmandPhilippot:
I'm not an accessibility expert and I rely more on what I read than on tests made on various screen readers (mostly because I'm a Linux user and Orca isn't great and I never spent time to improve my setup to test that...). So, take this with a pinch of salt.

I understand `location` the same way as Chris, and I can't find any trusty recommendations on that subject, except [Sara Soueidan](https://www.sarasoueidan.com/blog/css-scrollspy/#the-semantic-accessibility-of-html-scroll-markers) who also suggests using `aria-current="true"` for a table of content.

What I find interesting from my researches is that some people (including [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent#location)) recommend using `location` instead of `page` for a breadcrumb... I always thought `page` was more appropriate (and MDN contradicts itself by using `page` [in their examples](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent#examples)).

There is also this [opened issue on `w3c/aria-practices`](https://github.com/w3c/aria-practices/issues/2583) which kinda suggests `aria-current="true"` is probably enough (in "The full IRC log of that discussion" tab) even if the context is a bit different?

[Patternfly](https://www.patternfly.org/components/jump-links/accessibility/) suggests that `aria-current="location"` should be used on the destination once a jump link is clicked (like a table of contents I guess), not on the initial link.

So thanks @randomguy-2650 for this suggestion! But do you have other links suggesting `location` is better in this case?

Currently, my opinion would be to wait before applying those changes. I think the specs are not really clear on that subject, and tests with screen readers (and users of screen readers preferably) are necessary...

HiDeoo:
I had similar discussions in the past on the same topic and even questioned myself about it a few times. It's definitely a tricky one as the spec itself may not be specific enough to properly define when to use one or the other.

What made me finally always choose `aria-current="true"` over `location` for a table of contents is listening to what most screen readers announce in French, which is "emplacement actuel". While this is the proper translation of "current location", it always felt a bit off to me in the context of a table of contents because it implies more of a spatial position rather than a position within a document structure. Personally, I only got that feeling in French, probably because English is not my first language, and by questioning a few French-speaking screen reader users, I confirmed that they also found it a bit odd (and very uncommon) so I ended up always avoiding it except in specific contexts like maps, diagrams, etc. where at least 2 axes are involved.

I would be personally in favor of keeping the current implementation as is, and revisiting it if we ever get feedback from screen reader users that would express confusion about it or a real benefit of using `location` instead of `true` in this specific case.

ArmandPhilippot:
> it always felt a bit off to me in the context of a table of contents because it implies more of a spatial position rather than a position within a document structure

This is also where this is announced. The currently highlighted item in a table of contents is not really the location in the page. Its target is the current location. And so, "current location" (whatever the language) in the table of contents also sounds a bit strange to me.

I don't know accessibility-wise if Patternfly suggestion makes sense (this is probably superfluous if the heading is announced?). But, at least, semantically, it does make more sense to me to announce the current location when the heading is reached than having this announcement in the table of contents.

So, yeah, I'd say `aria-current="true"` is probably the best indication in a table of contents.

randomguy-2650:
This topic is just way too confusing to pick anything.

WAI-ARIA APG just does nothing on it. They have literal comboboxes and complicated things but no TOC, which arguably is way simpler.

delucis:
Thanks for all the extra input everyone! Given this, it sounds like it indeed may be best to keep the current implementation. Definitely a tricky one where the WAI ARIA spec is vague and real-world usage isn’t 100% clear. I will say that seeing someone like Sara Soueidan using `"true"` as well as the other feedback is reassuring that that may be the best.

> This topic is just way too confusing to pick anything.

Welcome to accessibility engineering! 😄

Definitely appreciate you taking the time to try to improve things. Unfortunately, it’s definitely an area where things are not entirely clear. The ARIA spec can be vague and we’re left to figuring out best practice based on user testing, real-world behaviour of screenreaders etc. Even the Authoring Practices Guide (if it had an appropriate example) cannot always be trusted as [they themselves say](https://www.w3.org/WAI/ARIA/apg/practices/read-me-first/#browserandassistivetechnologysupport), because it is intended to show “correct” ARIA usage according to the spec and does not guarantee correct real-world behaviour in browsers and assistive technologies.

PR #3701, opened 2026-02-07
Allow toggling mdx optimize mode on & off

zastrowm:
#### Description

Astro has a bug where if mdx optimize=true then it sometimes ignores the customizing/overriding of mdx components via `export const components = `.  To unblock starlight users and allow overriding of mdx components, add an explicit option on the starlight.markdown config that can turn the mdx optimize off.

Astro bug: https://github.com/withastro/astro/issues/14611

The workaround when directly working with Astro is:

```
mdx({
	optimize: true,
}),
```

However, doing this in Starlight requires additional workaround as then you also need to replicate the `starlightExpressiveCode` behavior which must come *before* mdx.  The most straightforward solution is to expose an option so that users can explicitly control `optimize` being true or false, which is what I do in here.

**Note**: I didn't start a discussion first, as this seemed to fall under the ["Make a pull request (PR) to address an open issue or to fix obvious problems" part of the contribution guidelines](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md) but can start a discussion instead.

### DIFF
--- a/packages/starlight/__tests__/basics/user-config.test.ts
+++ b/packages/starlight/__tests__/basics/user-config.test.ts
@@ -12,3 +12,20 @@ test('preserve social config order', () => {
 	});
 	expect((config.social || []).map(({ icon }) => icon)).toEqual(['twitch', 'github', 'discord']);
 });
+
+test('markdown.optimize defaults to true', () => {
+	const config = StarlightConfigSchema.parse({
+		title: 'Test',
+	});
+	expect(config.markdown.optimize).toBe(true);
+});
+
+test('mdxStrictMode can be set to false', () => {
+	const config = StarlightConfigSchema.parse({
+		title: 'Test',
+		markdown: {
+			optimize: false,
+		},
+	});
+	expect(config.markdown.optimize).toBe(false);
+});
--- a/packages/starlight/index.ts
+++ b/packages/starlight/index.ts
@@ -102,7 +102,7 @@ export default function StarlightIntegration(
 					integrations.push(starlightSitemap(starlightConfig));
 				}
 				if (!allIntegrations.find(({ name }) => name === '@astrojs/mdx')) {
-					integrations.push(mdx({ optimize: true }));
+					integrations.push(mdx({ optimize: starlightConfig.markdown.optimize }));
 				}
 
 				// Add Starlight directives restoration integration at the end of the list so that remark
--- a/packages/starlight/utils/user-config.ts
+++ b/packages/starlight/utils/user-config.ts
@@ -285,6 +285,15 @@ const UserConfigSchema = z.object({
 				.describe(
 					'Define additional directories where files should be processed by Starlight’s Markdown pipeline. Default: `[]`.'
 				),
+			/**
+			 * Define whether to optimize the MDX output for faster builds and rendering. Default: `true`.
+			 *
+			 * See https://docs.astro.build/en/guides/integrations-guide/mdx/#optimize
+			 */
+			optimize: z
+				.boolean()
+				.default(true)
+				.describe('Define whether to optimize the MDX output for faster builds and rendering. Default: `true`.'),
 		})
 		.default({})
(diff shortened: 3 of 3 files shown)


delucis:
Hi @zastrowm! Firstly, thanks for reading the contribution guidelines!

Starlight doesn’t support passing components in that way, so I’m not sure it makes sense to make this configurable. Are you using an override to be able to pass components? If so, I think you should be able to add the MDX integration yourself to work around this (we skip adding it with our `optimize: true` config if the user already added it).

zastrowm:
> Starlight doesn’t support passing components in that way, so I’m not sure it makes sense to make this configurable.

[Overriding mdx components](https://docs.astro.build/en/guides/integrations-guide/mdx/#assigning-custom-components-to-html-elements) can be done by using the following in an mdx file:

```ts
export const components = { a: AOverride };
```

and this *should* work in Astro & Starlight both.  However, due to [this astro bug](withastro/astro/issues/14611), it doesn't work consistently when MDX(optimize=true). 

An example of the failure can be seen in: https://stackblitz.com/edit/github-fcapnzw7?file=src%2Fcontent%2Fdocs%2Findex.mdx

> If so, I think you should be able to add the MDX integration yourself to work around this 

You *can* work around this by manually configuring MDX - and by extension `astroExpressiveCode` since it has to come before `mdx` - but it because [starlight is configuring expressiveCode with additional options](https://github.com/withastro/starlight/blob/0426f292f2621c138f00f029385afb71c2cbd14e/packages/starlight/index.ts#L98), you end up needing to replicate starlights default configuration for `astroExpressiveCode`.  If you do not, you get different behavior in code blocks:

**Before manually configuring mdx**

image

**After manually configuring mdx** (notice how the syntax highlighting color changed0)

image

(can be toggled by uncommented the lines in https://stackblitz.com/edit/github-fcapnzw7?file=astro.config.mjs).

This is a small visual change, but it speaks to the added complexity that is being added when the desire is to turn off mdx optimizations - I now have to figure out how to re-configure syntax highlighting to get it back to the starlight default.

It seemed like a more maintainable/robust method was allowing configuring the starlight mdx optimize on/off instead of forcing it to be `true`.  

----

LMK if you feel that this should be moved to a discussion.

delucis:
Hey @zastrowm, am I missing something? If I try the `export const components` approach, it works fine: https://stackblitz.com/edit/github-1p1jfzoj?file=src%2Fcontent%2Fdocs%2Findex.mdx

zastrowm:
@delucis it works in that case, but it's inconsistent - if you remove the newline before the export then it doesn't work - but it does work in that case with optimizations turned off.  Locally I've seen cases where if the export was at the end of the doc it worked, but not if it was at the beginning.  But if I moved it down 3 lines, then all the following lines were transformed but but not the ones above it 🫤.

I presume that https://github.com/withastro/astro/issues/14611 might be related to whitespace or newlines or something, but ultimately I haven't root caused it yet, thus the proposed workaround here.

delucis:
Got it — thanks for the extra info. Sounds like any issues there are might be a bug? In that case I’d lean towards not changing anything in Starlight and seeing if the actual bug can be fixed rather than working around it. What do you reckon?

It would also be good to document the exact cases you’re hitting in that issue — the actual issue and reproduction are about the `<Content components={{...}} />` approach, which is separate from the `export const components = ...` one. (`optimize` is supposed to work with `export const components` but does not support passing the `components` prop to `<Content>`.)

zastrowm:
> It would also be good to document the exact cases you’re hitting in that issue — the actual issue and reproduction are about the <Content components={{...}} /> approach, which is separate from the export const components = ...

Oh that's a good callout; I had my blinders on matching my problem with an existing issue. I should definitely add my use case and/or file a new issue.

> seeing if the actual bug can be fixed rather than working around it. What do you reckon?

In the ideal case I'd agree; practically I like being able to workaround it thus the PR, but I understand the reluctance.

PR #3704, opened 2026-02-10
feat: improve components exports with wildcard

florian-lefebvre:
#### Description

- Uses exports wildcards in package.json to improve maintenance
- Required moving non exported things to another non nested directory. I chose `components-internals` but it could be anywhere else really


.changeset/beige-ties-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Improves how components are exported from `@astrojs/starlight/components/*`
```
delucis: I wonder if we even need a changeset: there should be no user-facing impact in theory, right?

Or maybe it should be really clear about that?

```suggestion
Refactors how components are exported from `@astrojs/starlight/components/*` internally
```

packages/starlight/components-internals/TableOfContents/starlight-toc.ts
delucis, line 1: Would like to hear from @HiDeoo too on this, but thinking about it I wonder if this should be a `minor`: while not officially supported, I guess _some_ people might be importing files we considered internal and this would break for them?

Not possible via an `@astrojs/starlight/components/*` path, but I have seen people sometimes use a relative path import like `../node_modules/@astrojs/starlight/components/*` to get around the package manager Node resolution limitation.

Obviously that is technically unsupported behaviour so we could decide we don’t care if we want.

delucis: One more thought on this front: could we avoid the move with a different pattern in the `package.json`?

```json
"./components/*": "./components/*.astro",
```

This would not match the `*.ts` files nor the `*.astro` in a subdirectory, I think?

It’s kind of nice to have the related code together (e.g. `SidebarPersistState.ts` and `SidebarPersister.astro` are currently next to each other).

packages/starlight/package.json
```diff
@@ -31,142 +31,7 @@
...
-    },
-    "./components/TwoColumnContent.astro": {
-      "types": "./components/TwoColumnContent.astro.tsx",
-      "import": "./components/TwoColumnContent.astro"
-    },
-    "./components/Search.astro": {
-      "types": "./components/Search.astro.tsx",
-      "import": "./components/Search.astro"
-    },
+    "./components/*": "./components/*",
```
delucis, line 33: As per my last comment, this would then become:

```suggestion
    "./components/*": "./components/*.astro",
```

if it works.

.changeset/beige-ties-scream.md
florian-lefebvre: Yeah it's not changing anything user-facing, it depends if we want to play it safe. Your call!

packages/starlight/components-internals/TableOfContents/starlight-toc.ts
florian-lefebvre: Last time I checked (was a while ago), the support for wildcards is pretty limited. And even if we said `*.astro` as value, AFAIK it would take all component no matter the depth, exposing `packages/starlight/components/TableOfContents/TableOfContentsList.astro`

florian-lefebvre: Actually this is interesting, at <https://nodejs.org/api/packages.html#package-entry-points> search for the "Subpath patterns" heading. Looks like there's a way to exclude stuff. So yeah an option then would be:
- Only re-export .astro files
- Move internal stuff to an `internal` folder inside components
- Exclude this internal folder

HiDeoo: Indeed, as [documented](https://nodejs.org/api/packages.html#subpath-patterns), `*` maps expose nested subpaths as it is a string replacement syntax only.

I also did some experimentations and this also currently brings behavior difference. For example, here is the current behavior that I was personally not even aware of when triggering auto-completion on such paths (I wonder if it was always like this or we regressed at some point):


https://github.com/user-attachments/assets/0fe76a7e-b945-47a1-9edf-ce7c81cd4eba

With the PR current implementation, here is the behavior:


https://github.com/user-attachments/assets/3c419f95-8752-4076-a87c-62f6385ee6f7

With Chris's suggestion, we get the following without any extension:


https://github.com/user-attachments/assets/4b015562-34f7-4a61-ae50-8f0e0807bcaf

Another approach would be to use the current one in this PR without moving any files, as it's possible to exclude private subfolders from patterns using `null`:

```json
"./components/*": "./components/*",
"./components/*.ts": null,
"./components/TableOfContents/*": null,
```

Unfortunately, such exclusions do trigger an error when importing one of the excluded path, altho, they're not excluded from auto-completion:


https://github.com/user-attachments/assets/4ce55875-a63a-4ad4-bd9a-ca28c30ffb3a

In the end, I think the best DX ends up being the current PR implementation, at least for now.

Release-wise, I think I would personally feel more confident with a minor. There is no rush into such refactor and just our own example (in `docs/src/components/icons-list.astro`) shows that some people may be doing some weird shenanigans like you suggested, or even in a Vite plugin.

Note that considering the improved DX, I would not make the changeset "internal" only but focus on the improved DX.

Any thoughts following up all this?

florian-lefebvre: I agree with you! Interesting that exclusions still show in completions, so weird

delucis: Great investigation. Yes, agreed, worth it for the autocomplete 100%. I also feel like this used to work probably? But maybe not — I interact less frequently with the overrides, often copy-pasting the import statement, so could be that I just never noticed. Thanks for sharing that!

.changeset/beige-ties-scream.md
delucis: OK, based on @HiDeoo’s feedback I think we only need to update the changeset.

```suggestion
---
'@astrojs/starlight': minor
---

Fixes autocomplete for components exported from `@astrojs/starlight/components/*`

**⚠️ Potentially breaking change:** This change moves some files used in Starlight’s component internals out of the `components/` directory. Direct use of these files was not and is not officially supported. If you previously imported `TableOfContents/starlight-toc.ts`, `TableOfContents/TableOfContentsList.astro`, `Icons.ts`, or `SidebarPersistState.ts`, please review your code when updating.
```

Not sure what level of detail to give here. We usually provide migration guidance, but it feels weird to do so for something we don’t officially support. I think this should be enough for someone to figure it out hopefully if they need?

florian-lefebvre: LGTM but I think at this stage I'm not very useful anymore

PR #3711, opened 2026-02-15
docs: add rehype-gfm-components plugin

claylo:
Write GFM, view it on GitHub, deploy it with Starlight. The same Markdown file for both -- all content easily viewable on GitHub, still transforms to rich and interactive components on your Starlight site. No MDX, no imports.

#### Description

Just a plugin for those who want to use Starlight ... and also want their docs to be fully readable on GitHub. (It can't be _just me_, right?)

Thank you for all your efforts on Starlight! I'm loving it.

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -287,4 +287,9 @@ These community tools and integrations can be used to add features to your Starl
 		title="starlight-contributor-list"
 		description="Display a list of all contributors to your project."
 	/>
+	<LinkCard
+	  href="https://claylo.github.io/rehype-gfm-components"
+	  title="rehype-gfm-components"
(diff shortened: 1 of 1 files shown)


claylo:
@delucis Thank you for your feedback. On investigation into the rendering issue with the file tree, I realized I have a design problem -- I went out of my way to avoid processing the Markdown files twice. In doing so, I introduced a number of rendering bugs. I'll open a new pull request when I have this sorted out. I'm taking a completely different approach.

Let this be a cautionary tale about opening pull requests on code developed while sick with the flu! :grin:

Thanks again.

delucis:
Haha, oh no! Didn't mean to send you down another rabbit hole, it mostly looked like it was working well! I could still imagine people finding it handy even if they didn't use the file tree.

But wishing you good luck on the rewrite!

Issue #3713, opened 2026-02-18
FileTree python files like `__init__.py` keep getting highlighted


joshistoast:
### What version of `starlight` are you using?

0.37.6

### What version of `astro` are you using?

5.6.1

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chromium/Brave

### Describe the Bug

Using the `FileTree` component to describe a python environment is made more difficult by the fact that python files often like to have `__` on either sides of the name, and the component then interprets this as a highlight.
So instead of rendering it as `__init__.py`, it's rendered as `init` with a highlight.

<img width="545" height="377" alt="Image" src="https://github.com/user-attachments/assets/181018dd-5528-442f-84c2-9ee847a34f30" />

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hyzxqoph?file=src%2Fcontent%2Fdocs%2Findex.mdx

### Participation

- [ ] I am willing to submit a pull request for this issue.

joshistoast:
I can work around this by escaping each individual underscore, but I'd rather not need to do this for such a commonly used python naming format:

`__init__.py` -> `\_\_init\_\_.py`

delucis:
Aha, interesting edge case, thanks for sharing!

It’s not strictly speaking just a file-tree issue — it’s a Markdown issue because file trees are just wrapped around regular Markdown lists, e.g. we can see it on GitHub too:

```md
- __init__.py
- \_\_init\_\_.py
- `__init__.py`
```
The above renders as:

> - __init__.py
> - \_\_init\_\_.py
> - `__init__.py`

To avoid that for the `<FileTree>` component, we’d probably have to use a quite different syntax approach, like using a code block where the `__` wouldn’t be interpreted that way. But that would mean we don’t get the benefits of using the list style (renders fine in plain Markdown rendering, compatible with editors, familiar, etc.)

I think maybe wrapping the filename in backticks could be a good enough workaround for your case (instead of using `\` to escape each underscore)? That would also then render correctly in a regular Markdown renderer. Here’s a small example:

<img width="648" height="178" alt="source code of a file tree showing an __init__.py file rendered in backticks, and a rendered version of that tree side by side" src="https://github.com/user-attachments/assets/8483ce2a-cdb3-4aa9-b8f2-2b5a4133c31e" />

delucis:
[...]

PR #3714, opened 2026-02-19
FileTree: document escaping special characters in filenames

delucis:
#### Description

- Follow up to #3713 
- Documents the use of code backticks to escape filenames like `__init__.py` that otherwise break because the `_` is interpreted as Markdown syntax

Docs preview: https://deploy-preview-3714--astro-starlight.netlify.app/components/file-tree/#escape-special-characters

### DIFF
--- a/docs/src/content/docs/components/file-tree.mdx
+++ b/docs/src/content/docs/components/file-tree.mdx
@@ -224,7 +224,7 @@ Escape special characters such as underscores or spaces in filenames by wrapping
 
 <Preview>
 
-```mdx {8}
+```mdx
 import { FileTree } from '@astrojs/starlight/components';
 
 <FileTree>
@@ -238,7 +238,7 @@ import { FileTree } from '@astrojs/starlight/components';
 
 <Fragment slot="markdoc">
 
-```markdoc {5}
+```markdoc
 {% filetree %}
 - `__init__.py`
 - `Hello world.txt`


docs/src/content/docs/components/file-tree.mdx
```diff
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
 
 </Preview>
 
+### Escape special characters
+
+Escape special characters such as underscores or spaces in filenames by wrapping the filename in code backticks.
+
+<Preview>
+
+```mdx {8}
```
HiDeoo:
```suggestion
```mdx
```

I guess it's a leftover of the previous example and the goal is not to highlight anything considering all lines are an example?

Or maybe highlight the actual backticks?


```suggestion
```mdx /\x60/
```

docs/src/content/docs/components/file-tree.mdx
```diff
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
...
+- `__init__.py`
+- `Hello world.txt`
+- **`READ ME FIRST.md`**
+
+</FileTree>
+```
+
+<Fragment slot="markdoc">
+
+```markdoc {5}
```
HiDeoo:
```suggestion
```markdoc
```

```suggestion
```markdoc /\x60/
```

docs/src/content/docs/components/file-tree.mdx
delucis: Ah yeah, I missed this after copy/pasting. Thanks for catching it!

Tried the backticks locally but I think it’s maybe more confusing/hard to read than helpful?

image

So I think I’ll go with just removing the pointless highlight meta.

PR #3722, opened 2026-02-23
i18n(fr): update `resources/community-content.mdx`

ArmandPhilippot:
#### Description

Adds changes from #3649 to the French translation of `resources/community-content.mdx`.

### DIFF
--- a/docs/src/content/docs/de/components/file-tree.mdx
+++ b/docs/src/content/docs/de/components/file-tree.mdx
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
 
 </Preview>
 
+### Sonderzeichen umgehen
+
+Sichere Sonderzeichen wie Unterstriche oder Leerzeichen in Dateinamen, indem du den Dateinamen in Code-Backticks einschließt.
+
+<Preview>
+
+```mdx
+import { FileTree } from '@astrojs/starlight/components';
+
+<FileTree>
+
+- `__init__.py`
+- `Hallo Welt.txt`
+- **`LIES MICH ZUERST.md`**
+
+</FileTree>
+```
+
+<Fragment slot="markdoc">
+
+```markdoc
+{% filetree %}
+- `__init__.py`
+- `Hallo Welt.txt`
+- **`LIES MICH ZUERST.md`**
+{% /filetree %}
+```
+
+</Fragment>
+
+<FileTree slot="preview">
+
+- `__init__.py`
+- `Hallo Welt.txt`
+- **`LIES MICH ZUERST.md`**
+
+</FileTree>
+
+</Preview>
+
 ## `<FileTree>`-Eigenschaften
 
 **Implementation:** [`FileTree.astro`](https://github.com/withastro/starlight/blob/main/packages/starlight/user-components/FileTree.astro)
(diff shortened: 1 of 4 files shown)


docs/src/content/docs/fr/resources/community-content.mdx
```diff
@@ -169,6 +169,11 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 		title="Ajouter une bannière pour l'ensemble du site avec Starlight"
 		description="Découvrez comment créer une bannière pour l'ensemble du site grâce à la fonctionnalité intégrée de bannière par page de Starlight."
 	/>
+	<LinkCard
+		href="https://lirantal.com/blog/customizing-astro-starlight-sidebar-gated-content-auth"
+		title="Personnalisation de la barre latérale de Starlight pour le contenu protégé par authentification"
+		description="Sécurisez votre matériel de cours en utilisant Clerk et SSR pour masquer ou afficher dynamiquement les chapitres dans la barre latérale de Starlight en fonction de l'état de connexion de l'utilisateur."
```
ArmandPhilippot: I'm not necessarily fan of what I came up with a literal translation... If we're okay to take some freedom, maybe the following is enough (shorter and clearer)?

```suggestion
		description="Sécurisez votre matériel de cours en utilisant Clerk et SSR pour afficher dynamiquement des chapitres dans la barre latérale de Starlight quand l'utilisateur est connecté."
```

docs/src/content/docs/fr/resources/community-content.mdx
```diff
@@ -169,6 +169,11 @@ Explorez le contenu communautaire produit par les utilisateurs de Starlight :
 		title="Ajouter une bannière pour l'ensemble du site avec Starlight"
 		description="Découvrez comment créer une bannière pour l'ensemble du site grâce à la fonctionnalité intégrée de bannière par page de Starlight."
 	/>
+	<LinkCard
+		href="https://lirantal.com/blog/customizing-astro-starlight-sidebar-gated-content-auth"
+		title="Personnalisation de la barre latérale de Starlight pour le contenu protégé par authentification"
```
HiDeoo:
```suggestion
		title="Personnalisation de la barre latérale de Starlight pour du contenu protégé par authentification"
```

Not sure if me or not, "le" sounds kinda weird in this title as-is 🤔

docs/src/content/docs/fr/resources/community-content.mdx
HiDeoo: I think your suggestion is indeed way better. I'm also not a huge fan of the "chapters" in the English version, but let's keep it, small nit tho: I think "des chapitres" rather than "les chapitres" (kinda like my previous comment) basically would sound better? Any thought?

ArmandPhilippot: Ah, I guess it depends on how we read it!
I was going to say, for me "les chapitres" refers to "cours" (*"les chapitres (du cours)"*), so "les" sounds natural to me.
But, yes, if we only hide some chapters, then "des" makes sense I think.

So, yeah, no issue to update my suggestion with that!

docs/src/content/docs/fr/resources/community-content.mdx
ArmandPhilippot: However, here, "pour du contenu" sounds a bit weird to me (but maybe it's only me 😅 ).
I'd say here "contenu" is defined by "protégé par authentification" so using a definite article sounds more natural to me (even if this is whatever content... protected).

I don't have a strong opinion on this... Maybe that's fine! So, if you think this is better, I'm okay with your suggestion!

HiDeoo (approved):
Well, not sure if my suggested changes are better or not, but let's roll with them 😅  (and we can always revert if needed)

Thanks for the update

PR #3723, opened 2026-02-23
i18n(fr): update `components/file-tree.mdx`

ArmandPhilippot:
#### Description

Adds changes from #3714 to the French translation of `components/file-tree.mdx`.

### DIFF
--- a/docs/src/content/docs/de/components/file-tree.mdx
+++ b/docs/src/content/docs/de/components/file-tree.mdx
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
 
 </Preview>
 
+### Sonderzeichen umgehen
+
+Sichere Sonderzeichen wie Unterstriche oder Leerzeichen in Dateinamen, indem du den Dateinamen in Code-Backticks einschließt.
+
+<Preview>
+
+```mdx
+import { FileTree } from '@astrojs/starlight/components';
+
+<FileTree>
+
+- `__init__.py`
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/fr/components/file-tree.mdx
```diff
@@ -218,6 +218,46 @@ import { FileTree } from '@astrojs/starlight/components';
 
 </Preview>
 
+### Échapper les caractères spéciaux
+
+Échappez les caractères spéciaux tels que les traits de soulignement ou les espaces dans les noms de fichiers en enveloppant le nom de fichier avec des accents graves.
```
HiDeoo:
```suggestion
Échappez les caractères spéciaux tels que les tirets bas ou les espaces dans les noms de fichiers en entourant le nom de fichier avec des accents graves.
```

- For consistency, we use ["tiret bas"](https://fr.wikipedia.org/wiki/Tiret_bas) in other pages, e.g. the config reference: https://starlight.astro.build/fr/reference/configuration/#docsloader
- Isn't it a bit more easy to understand by using "entourer" here?

ArmandPhilippot: Ah, I didn't knew there was a precedent! Yep, makes sense then!
And I don't have a preference between "entourant" ou "enveloppant"...  they sound like synonyms to me, so let's go with your suggestion!

HiDeoo (approved):
Thanks for the quick update 🙌

Issue #3712, opened 2026-02-17
Using {% steps %} in Mark Doc with {% tabs %}


3goats:
### What version of `starlight` are you using?

0.37.6 

### What version of `astro` are you using?

5.17.2

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

This is working as expected" 

``````
---
title: Eval Guide
description: todo
---

{% steps %}

1. Import the component into your MDX file:

   ```js
   import { Steps } from '@astrojs/starlight/components';
   ```

2. Wrap `<Steps>` around your ordered list items.

{% /steps %}
`````` 

However when I add tabs to the step items like this: 

``````
---
title: Eval Guide
description: todo
---

{% steps %}

1. Import the component into your MDX file:

   ```js
   import { Steps } from '@astrojs/starlight/components';
   ```

    {% tabs syncKey="pkg" %}
    {% tabitem label="npm" %}
      ```bash
      npm create astro@latest
      ```
    {% /tabitem %}
    {% tabitem label="pnpm" %}
      ```bash
      pnpm create astro
      ```
    {% /tabitem %}
    {% /tabs %}

2. Wrap `<Steps>` around your ordered list items.

{% /steps %}
``````
Starlight throws this error: 

```
11:17:36 [ERROR] [AstroUserError] The `<Steps>` component expects its content to be a single ordered list (`<ol>`) but found multiple child elements: `<script>`, `<ol>`.
  Hint:
    To learn more about the `<Steps>` component, see https://starlight.astro.build/components/steps/
    
    Full HTML passed to `<Steps>`:
    
    <script type="module" src="/node_modules/.pnpm/@astrojs+starlight@0.37.3_astro@5.17.2_jiti@2.6.1_lightningcss@1.30.2_rollup@4.57.1_typescript@5.9.3_/node_modules/@astrojs/starlight/user-components/Tabs.astro?v=998bbfe6&#x26;astro&#x26;type=script&#x26;index=0&#x26;lang.ts"></script>
    <ol>
      <li>
        <p>Import the component into your MDX file:</p>
        <div class="expressive-code"><link rel="stylesheet" href="/_astro/ec.v4551.css">
          <script type="module" src="/_astro/ec.0vx5m.js"></script>
          <figure class="frame not-content">
            <figcaption class="header"></figcaption>
[...]

PR #3644, opened 2026-01-02
Astro 6 support

HiDeoo:
#### Description

This PR adds support for Astro 6 (currently requires at least version `6.0.0`)

See https://github.com/withastro/starlight/pull/3644#issuecomment-3855392365 for preview releases.

#### Known issues

| Symbol | Status                          |
| ------ | ------------------------------- |
| 🟢     | Fixed                           |
| 🟡     | Fix pending review |
| 🔴     | No fix yet, under investigation |

- 🟢 ~~Build issue on Windows when importing an Astro component with a `<script>` in MDX using an import alias~~
  - ~~[Repro](https://github.com/HiDeoo/astro-6-mdx-import-alias-windows-repro)~~
  - _Fixed by https://github.com/withastro/astro/pull/15112_
- 🟢 ~~The Markdoc example is failing to build as an updated version of `@astrojs/markdoc` was not published as part of the alpha.~~
- 🟢 ~~Incompatible with Vitest 4~~
  - ~~[Issue](https://redirect.github.com/vitest-dev/vitest/issues/9324)~~
  - _Fixed by https://github.com/vitest-dev/vitest/pull/9584_
  - _Fix available in version [`4.1.0-beta.4`](https://github.com/vitest-dev/vitest/releases/tag/v4.1.0-beta.4) at the moment_
  - ~~_Not yet used in this PR due to unrelated mocking issues: https://redirect.github.com/vitest-dev/vitest/issues/9742 - https://redirect.github.com/vitest-dev/vitest/issues/9771_~~
- 🟢 ~~SSR build issue with explicit URL imports, e.g. `import styles from '../styles/custom.css?url&no-inline'`~~
  - ~~[Repro](https://github.com/HiDeoo/astro-6-url-import-ssr-repro)~~
  - _Fixed by https://github.com/withastro/astro/pull/15093_
- 🟢 ~~Extra CSS emitted in dev depending on the order of visited pages~~
  - ~~[Repro](https://github.com/HiDeoo/astro-6-dev-server-css-repro)~~
  - _Fixed by https://github.com/withastro/astro/pull/15167_
- 🟢 ~~Adding CSS to a component that did not previously had CSS requires a restart of the dev server to take effect~~
  - ~~[Repro](https://github.com/HiDeoo/astro-6-add-css-dev-repro)~~
  - _Fixed by https://github.com/withastro/astro/pull/15133_
- 🟢 ~~SSR issue where visiting some pages matching pre-rendered route patterns can crash the server~~
[...]

PR #3724, opened 2026-02-24
Update monorepo version of Astro to latest

delucis:
#### Description

This PR updates the versions of Astro and other `@astrojs/*` dependencies in the monorepo to latest.

Will cause conflicts with #3644 but probably not too bad.

Draft for now. I mainly did this to have a new version of Astro to work with while making a hot fix for #3712 but thought I might as well open a PR while I was at it.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,7 +10,7 @@ importers:
     devDependencies:
       '@astrojs/check':
         specifier: ^0.9.6
-        version: 0.9.6(prettier-plugin-astro@0.14.1)(prettier@3.3.3)(typescript@5.6.3)
+        version: 0.9.6(prettier-plugin-astro@0.14.1)(prettier@3.8.1)(typescript@5.6.3)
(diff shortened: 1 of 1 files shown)


delucis:
> /index.html	7.31 KB (+0.29% 🔺)

Looks like this is due to `<Image>` now including `fetchpriority="auto"` everywhere (since [5.10.0](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#5100) I think)? (At least it’s the only thing I could spot while diffing the output between `main` and this branch.)

Doesn’t matter too much, but it is a bit silly for Astro to output that given it’s the default value.

(**Edit:** I opened https://github.com/withastro/astro/pull/15646 to remove this redundant attribute)

delucis:
We'll be updating to v6 in #3644 very soon so going to close this one. Main purpose for me was testing #3726 which it did well (plus I found that image issue which is now fixed in Astro V6), so no need to push it forwards to be merged.

PR #3730, opened 2026-02-28
docs: add starlight-telescope and starlight-announcement to plugins s…

frostybee:
#### Description

This PR adds starlight-telescope and starlight-announcement to the community plugins list in the Starlight docs.


frostybee:
Thanks, HiDeoo, for the feedback. I now have a clear understanding of where the content flashing is happening, and I'll look into it.

Issue #3748, opened 2026-03-11
Global `tableOfContents` config ignored due to zod v4 `.default().optional()` behavior change


pyxelr:
### What version of `starlight` are you using?

0.38.0

### What version of `astro` are you using?

6.0.2

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Description

In Starlight 0.38.0 (with Astro 6), the global `tableOfContents` configuration in `astro.config.mjs` is silently ignored. All pages use the schema defaults (`minHeadingLevel: 2`, `maxHeadingLevel: 3`) regardless of the global config.

### Reproduction

```js
// astro.config.mjs
starlight({
  tableOfContents: { maxHeadingLevel: 6 },
  // ...
})
```

**Expected:** ToC shows headings up to h6.
**Actual:** ToC only shows h2–h3. The built HTML always has `data-max-h="3"`.

### Root cause

In [`schema.ts#L36`](https://github.com/withastro/starlight/blame/c8ef30f3b599dcabe714a3e513c9670c57bff366/packages/starlight/schema.ts#L36), the frontmatter field is defined as:

```ts
tableOfContents: TableOfContentsSchema().optional(),
```

`TableOfContentsSchema()` includes `.default({ minHeadingLevel: 2, maxHeadingLevel: 3 })`. In zod v3, chaining `.default(x).optional()` returned `undefined` for `undefined` input. In zod v4 (shipped with Astro 6), it returns the default value instead.

This means `entry.data.tableOfContents` is **always** `{ minHeadingLevel: 2, maxHeadingLevel: 3 }`, even when frontmatter doesn't set it. The check in `getToC()`:

```ts
entry.data.tableOfContents !== undefined
    ? entry.data.tableOfContents   // ← always taken
    : config.tableOfContents;      // ← never reached
```

always takes the frontmatter branch, so the global config is never used.

### Suggested fix

1. Use an inline schema **without** `.default()` for the frontmatter field, so it returns `undefined` when not set in frontmatter:

```ts
tableOfContents: z
    .union([
        z.object({
            minHeadingLevel: z.int().min(1).max(6).optional(),
            maxHeadingLevel: z.int().min(1).max(6).optional(),
        }),
        z.boolean(),
    ])
    .optional(),
```

2. Update `getToC()` to merge partial frontmatter values with the global config:

```ts
const frontmatterToC = entry.data.tableOfContents;
let tocConfig;
[...]

PR #3751, opened 2026-03-11
fix: respect global `tableOfContents` config with zod v4

pyxelr:
#### Description

- Closes https://github.com/withastro/starlight/issues/3748
- With zod v4, `.default(value).optional()` returns the default value for `undefined` input instead of `undefined` itself. This caused the frontmatter `tableOfContents` field to always resolve to the schema default (`{ minHeadingLevel: 2, maxHeadingLevel: 3 }`), preventing the global Starlight config from being used as a fallback.

#### Changes

- **`schemas/tableOfContents.ts`**: Extracted a shared `TableOfContentsBaseSchema` and created two separate exports:
  - `UserConfigTableOfContentsSchema()` — wraps the base with `.default(defaults)` for the user config.
  - `FrontmatterTableOfContentsSchema()` — wraps the base with `.optional()` so `undefined` is preserved when frontmatter does not set the field.
- **`schema.ts`**: Updated the frontmatter schema to use `FrontmatterTableOfContentsSchema()`.
- **`utils/user-config.ts`**: Updated to use `UserConfigTableOfContentsSchema()`.

#### Testing

All 511 existing unit tests pass. The fix can be verified by setting a global `tableOfContents` config (e.g., `maxHeadingLevel: 4`) and confirming it applies to pages that do not override it in frontmatter.

A [StackBlitz reproduction](https://stackblitz.com/edit/github-jkv5cdpe?file=astro.config.mjs) is available in the linked issue.

### DIFF
--- a/.changeset/fix-toc-global-config-zod-v4.md
+++ b/.changeset/fix-toc-global-config-zod-v4.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Fixes global `tableOfContents` config being ignored due to a zod v4 behavior change where `.default().optional()` returns the default value for `undefined` input instead of `undefined` itself.
+Fixes a regression causing global `tableOfContents` config to be ignored


packages/starlight/schemas/tableOfContents.ts
```diff
@@ -17,3 +17,22 @@ export const TableOfContentsSchema = () =>
...
+		.union([
+			z.object({
+				/** The level to start including headings at in the table of contents. Default: 2. */
+				minHeadingLevel: z.int().min(1).max(6).optional(),
+				/** The level to stop including headings at in the table of contents. Default: 3. */
+				maxHeadingLevel: z.int().min(1).max(6).optional(),
+			}),
+			z.boolean(),
+		])
+		.optional();
```
delucis: I think we can fix this entirely at the schema level and also avoid needing to duplicate things:


```suggestion
const TableOfContentsBaseSchema = z
	.union([
		z.object({
			/** The level to start including headings at in the table of contents. Default: 2. */
			minHeadingLevel: z.int().min(1).max(6).optional().default(2),
			/** The level to stop including headings at in the table of contents. Default: 3. */
			maxHeadingLevel: z.int().min(1).max(6).optional().default(3),
		}),
		z.boolean().transform((enabled) => (enabled ? defaults : false)),
	])
	.refine((toc) => (toc ? toc.minHeadingLevel <= toc.maxHeadingLevel : true), {
		error: 'minHeadingLevel must be less than or equal to maxHeadingLevel',
	});

export const UserConfigTableOfContentsSchema = () => TableOfContentsBaseSchema.default(defaults);

export const FrontmatterTableOfContentsSchema = () => TableOfContentsBaseSchema.optional();
```

(This suggestion also renames `TableOfContentsSchema` to `UserConfigTableOfContentsSchema` for consistency, so references to it will also need to be updated.)

packages/starlight/utils/routing/data.ts
```diff
@@ -61,12 +61,26 @@ export function generateRouteData({
...
+	} else if (typeof frontmatterToC === 'object' && globalToC !== false) {
+		// Partial override — merge with global config for missing values.
+		tocConfig = {
+			minHeadingLevel: frontmatterToC.minHeadingLevel ?? globalToC.minHeadingLevel,
+			maxHeadingLevel: frontmatterToC.maxHeadingLevel ?? globalToC.maxHeadingLevel,
+		};
+	} else {
+		tocConfig = false;
+	}
+
```
delucis: That allows us to avoid any need for a change in logic here:
```suggestion
	const tocConfig =
		entry.data.template === 'splash'
			? false
			: entry.data.tableOfContents !== undefined
				? entry.data.tableOfContents
				: config.tableOfContents;
```

(It is true that we don’t currently support the “partial override” scenario. Maybe we should, but that would be new behaviour, so we should not change it in this PR which should only patch the regression.)

delucis (changes requested):
Thanks for the PR @pyxelr! I played around a bit and I think we can actually scope the changes entirely to the schema (which is the origin of the problem) and avoid logic changes for now.

delucis (approved):
Perfect, thank you! I just pushed up some tests I was using to diagnose the issue locally, but this should be good to go!

Thanks again for contributing 🙌

voytxt:
this seems to fix the issue in dev (`astro dev`), but not in prod (`astro build && astro preview`)

delucis:
Do you have an example you could point to @voytxt for us to debug? In theory at least there should be no dev/build difference to this API 🤔

voytxt:
I would presume this is some obscure caching bug? Could very easily be something completely different I have no idea... **I managed to solve it however, both local prod and Vercel prod work properly now!** I'm not exactly sure what solved it, but I can reitarate what steps I did.

For local prod, I first tried deleting all gitignored directories and also building with `astro build --force`, that didn't seem to solve it, but when I commented and later uncommented most options in the astro config file, that somehow fixed it, and I could no longer reproduce the bug on machine whatsoever. Note that there was no git-tracked files changed in the end, nothing.

For Vercel prod, nothing seemed to work, until I bumped Astro from 6.0.3 to 6.0.4. That seemed to magically fix it and I now also can't reproduce it on Vercel, even if I deploy the version with Astro 6.0.3.

If it's not vite/astro directly, my hunch would be on pnpm (I use pnpm 10.32.1) doing some weird shenanigans, or possibly some dependency (tho I'm not sure if that's possible). For the sake of completeness, here they are: (latest versions of everything)

```json
  "devDependencies": {
    "@astrojs/starlight": "^0.38.1",
    "@astrojs/vercel": "^10.0.0",
    "astro": "^6.0.4",
    "astro-vtbot": "^2.1.11",
    "prettier": "^3.8.1",
    "prettier-plugin-astro": "^0.14.1",
    "sharp": "^0.34.5",
    "starlight-sidebar-topics": "^0.7.1"
  }
  ```

delucis:
Ohh, interesting. It is theoretically possible that the content collection cache could have caused an issue here! I would have expected that updating Starlight to 0.38.1 would have been enough to invalidate that but maybe not for some reason. Glad it’s working now anyway and thanks for letting us know — I’ll keep an eye out for anyone else hitting it.

PR #3759, opened 2026-03-16
fix: use region-specific translations for default locale

MilesChou:
## Summary

- Fix default locale `buildResources` using `stripLangRegion` result as a separate layer that overwrites region-specific translations
- For example, a monolingual site with `lang: 'zh-TW'` gets Simplified Chinese (`zh`) UI strings instead of Traditional Chinese (`zh-TW`)

## Root Cause

In `createTranslationSystem.ts`, the default locale translation loading passes both `builtinTranslations[defaultLocale]` and `builtinTranslations[stripLangRegion(defaultLocale)]` as separate arguments to `buildResources`. Since `buildResources` applies later layers over earlier ones, `zh` (Simplified Chinese) overwrites `zh-TW` (Traditional Chinese).

The locales loop (line 37) correctly uses `||` as a fallback, but the default locale path (line 25-26) does not.

## Fix

Align the default locale behavior with the locales loop by using `||` so the `stripLangRegion` result is only used as a fallback when the specific locale has no built-in translations:

```diff
-  builtinTranslations[defaultLocale],
-  builtinTranslations[stripLangRegion(defaultLocale)],
+  builtinTranslations[defaultLocale] || builtinTranslations[stripLangRegion(defaultLocale)],
```

## Test plan

- [x] Added test case for `zh-TW` monolingual site verifying `search.label` returns `搜尋` (Traditional Chinese) instead of `搜索` (Simplified Chinese)
- [x] All existing i18n tests pass

## Reproduction

```js
// astro.config.mjs
starlight({
  locales: {
    root: {
      label: '繁體中文',
      lang: 'zh-TW',
    },
  },
})
```

The UI strings (search, theme selector, etc.) incorrectly show Simplified Chinese instead of Traditional Chinese.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4.4.0
+      - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version: ${{ env.NODE_VERSION }}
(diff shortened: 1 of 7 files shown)


delucis (approved):
Thank you @MilesChou! This change makes sense to me.

I initially wondered if it should only flip the order of the two resources (so you’d e.g. apply `zh-TW` on top of `zh` instead of the reverse like it is currently). That way e.g. a partial `en-US` dictionary could be combined with a complete `en` dictionary.

But I think your change is the right call:

- These are our built-in translations and we require complete dictionaries for these, so there is no possibility of a “partial” dictionary.
- While the `en` + `en-US` example might make sense, there are cases like `zh` + `zh-TW` where combining the dictionaries would not make sense.

Issue #3760, opened 2026-03-16
perf: sidebar trees are deep cloned on every page causes build overhead


MohamedH1998:
### What version of `starlight` are you using?

0.37.6

### What version of `astro` are you using?

5.13.7

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

In our [developer docs](https://developers.cloudflare.com/), we have 7,303 pages and currently and our local builds are currently taking 350-400 seconds (locally) - after profiling, we found that for every page rendered, Starlight deep-clones the entire sidebar trees - with large 7,303 pages and a large sidebar this adds up.

| Metric | Value |
|--------|-------|
| `structuredClone` self-time | 41.8s |
| Measured build time saving after fix | **~49–56s (~14%)** |

### What's happening
Every page triggers a full `structuredClone` of the entire intermediate sidebar tree inside `getSidebarFromIntermediateSidebar` (`utils/navigation.ts:409`). The only thing done with that clone is flipping `isCurrent: true` on a single link entry to mark the current page.
Immediately after, a second full clone of the entire route data (including the sidebar) is made via `klona(routeData)` in the middleware pipeline - and this is the copy that components actually read from.
**Clone 1 is redundant in static builds.** The `klona` in middleware already provides full isolation for all downstream consumers. The `structuredClone` exists solely to prevent mutating the cached sidebar — a concern that can be solved with a simple in-place set/reset of one boolean.

### Proposed fix
Add an internal non-exported fast path (`getSidebarForRender`) gated on `config.prerender === true` that replaces `structuredClone` with an in-place set/reset of `isCurrent`. Static builds are sequential — there is no concurrent access risk. The exported `getSidebar()` and `getSidebarFromConfig()` APIs are unchanged in all modes. Only the internal render pipeline takes the fast path.

### Link to Minimal Reproducible Example

N/A

### Participation

- [x] I am willing to submit a pull request for this issue.

delucis:
Hi @MohamedH1998 thanks for the issue! Would be great to optimise things here.

I’m a bit wary of forking the logic for maintainability purposes but happy to take a look at a PR and see what it looks like!

MohamedH1998:
I appreciate you looking into this so quickly, I've raised the PR @delucis

PR #3761, opened 2026-03-16
perf: avoid structuredClone in sidebar generation for prerendered builds

MohamedH1998:
#### Description

During prerendered builds, `generateRouteData` was calling `getSidebar` on every page, which does a full `structuredClone` of the intermediate sidebar tree on each call. For large sites this is O(n) allocations × number of pages (in our case 7303 page).
This PR introduces a fast path  `getSidebarForRender` — that mutates the shared intermediate sidebar in place by flipping a single `isCurrent` pointer per call, reducing sidebar work to O(1) per page during static builds.

As a POC, I've implemented a patch to the package in our repo which saw a significant decrease in build times, see [here](https://github.com/cloudflare/cloudflare-docs/pull/28970)

- Closes issue #3760 

### Changes

utils/navigation.ts
- Add `getSidebarForRender`: mutates the cached intermediate sidebar in place, tracking the previously-current link per locale so it can be reset on the next call without a full tree scan.
- Add `resetCurrentEntries`: ensures the clone path (`getSidebar` / `getSidebarFromConfig`) always starts from a clean state, guarding against stale `isCurrent` flags if the intermediate cache was previously mutated by the fast path.
- Add `findAndMarkCurrentEntry`: shared helper for the in-place mutation path.
utils/routing/data.ts
- Branch on `config.prerender` in `generateRouteData`: use `getSidebarForRender` for static builds, `getSidebar` for SSR.
- `structuredClone` the result of `getSidebarForRender` before handing it off to the route data object, so each page's sidebar snapshot is isolated from subsequent mutations.
relevant tests


delucis:
@MohamedH1998 I opened an alternative approach to removing `structuredClone` here which avoids some of the things I was concerned about in this PR.

Would you be able to test it on your Cloudflare Docs branch to see how it performs?

You can install a preview release with:
```sh
pnpm add https://pkg.pr.new/@astrojs/starlight@3768
```

MohamedH1998:
@delucis FYI, we've bumped our starlight version and saw the same improvements! Appreciate you getting it sorted so quickly

delucis:
Awesome, thanks for letting us know!

PR #3768, opened 2026-03-19
Improve performance of sidebar generation logic

### LINKED ISSUE #3760: perf: sidebar trees are deep cloned on every page causes build overhead

delucis:
#### Description

- Closes #3760 
- Closes #3761 which it supersedes

This PR refactors our sidebar generation logic to avoid using `structuredClone()` each time a sidebar is created. Especially on large sidebars, the clone can be expensive.

#2252 added the current sidebar approach and introduced the structured clone partially to defend against user mutations in component overrides. Since then, we refactored our route data system in #2390, which introduced `klona` to create a clone of the entire route data object. This means it is now safe to remove the extra `structuredClone()` in the sidebar logic.

The new strategy tracks the sidebar entries with `isCurrent` manually so we can patch those in a performant way on a single persistent object without relying on the structured clone to avoid mutations.

I haven’t tested on Cloudflare’s docs (mentioned in #3760) but I tested in Starlight’s docs:

- Starlight docs: 1% performance increase (~178ms)
- Starlight docs modified to add several thousand extra sidebar entries: 6% performance increase (~2.8s)

I would assume the increase should be even more significant in the CF docs given their high page count and sidebar size. And the fix here is essentially the same as the one tested in https://github.com/cloudflare/cloudflare-docs/pull/28970 — `structuredClone()` was removed in both — this PR just avoids introducing some unnecessary code duplication and forking of code paths.

### DIFF
--- a/packages/starlight/utils/navigation.ts
+++ b/packages/starlight/utils/navigation.ts
@@ -361,7 +361,7 @@ function sidebarFromDir(
  * When generating the final sidebar for a page, the current page entry in the sidebar is marked
  * with `isCurrent` and cached. Subsequent runs then reset the previous current entry before marking
  * the new current page.
- * 
+ *
  * Sidebars, like all route data, are deep cloned before the data is passed to users for mutation,
  * so optimising with a single mutable object per locale is safe.
  *
@@ -377,7 +377,8 @@ export function getSidebar(pathname: string, locale: string | undefined): Sideba
 		intermediateSidebar = getIntermediateSidebarFromConfig(config.sidebar, pathname, locale);
 		intermediateSidebars.set(locale, intermediateSidebar);
 	}
-	return getSidebarFromIntermediateSidebar(intermediateSidebar, pathname, locale);
+	setIntermediateSidebarCurrentEntry(intermediateSidebar, pathname, locale);
+	return intermediateSidebar;
 }
 
 /** Get the sidebar for the current page using the specified sidebar config. */
@@ -386,8 +387,10 @@ export function getSidebarFromConfig(
 	pathname: string,
 	locale: string | undefined
 ): SidebarEntry[] {
-	const intermediateSidebar = getIntermediateSidebarFromConfig(sidebarConfig, pathname, locale);
-	return getSidebarFromIntermediateSidebar(intermediateSidebar, pathname, locale);
+	const sidebar = getIntermediateSidebarFromConfig(sidebarConfig, pathname, locale);
+	const currentEntry = getSidebarCurrentEntry(sidebar, pathname);
+	if (currentEntry) currentEntry.isCurrent = true;
+	return sidebar;
 }
 
 /** Get the intermediate sidebar for the current page using the specified sidebar config. */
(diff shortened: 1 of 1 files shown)


ematipico (approved):

packages/starlight/utils/navigation.ts
HiDeoo: We should probably update this comment as the part about cloning the intermediate sidebar is no longer true, at least at this level.

packages/starlight/utils/navigation.ts
```diff
@@ -404,29 +405,50 @@ function getIntermediateSidebarFromConfig(
...
+		entry.isCurrent = true;
+		lastCurrentEntryByLocale.set(locale, entry);
+	}
+}
+
+/** Finds the current page in an intermediate sidebar. */
+function getIntermediateSidebarCurrentEntry(
+	intermediateSidebar: SidebarEntry[],
+	pathname: string,
+	locale: string | undefined
```
HiDeoo: I don't think the `locale` is ever used in this function (except when calling this function recursively)?

packages/starlight/utils/navigation.ts
HiDeoo, line 385: This path is hit when rendering a Starlight custom page with a custom sidebar. For such cases, we recompute an intermediate sidebar but never cache hit.

I wonder if it could be confusing that this path does not cache this intermediate sidebar but still cache the `lastCurrentEntryByLocale` 🤔 (I tested by writing a few regression tests for sanity interleaving `getSidebar` and `getSidebarFromConfig` calls, but I don't think it's possible that this would cause an issue).

packages/starlight/utils/navigation.ts
delucis: Good catch — I previously was updating the cache inside here, but didn’t like that it was separate from the reset logic and pulled it back out. Will fix.

packages/starlight/utils/navigation.ts
delucis: Good call out! Had a go at this in 3767bb7f359af8b38b5d6d93390f31588fe50a16.

delucis: I’ll take a look. I thought I’d accounted for all usages but I missed that one. It does feel weird to cache anything for those, so maybe we can skip it or have a separate function for that usage.

PR #3769, opened 2026-03-21
refactor(ci): use original first interaction action again

trueberryless:
#### Description

In https://github.com/withastro/starlight/pull/1139 we decided to use a fork of [`actions/first-interaction`](https://github.com/actions/first-interaction) because the Node version was outdated and the fork has more robust features in general (read the detailed benefits in HiDeoo's list in the PR).

Since then, the original action has been updated quite well, and seems to be maintained actively again, whereas now the fork has not been updated for about a year. I have not done a comparison performance- or feature-wise in depth. But I recommend using the original and actively maintained action.

Therefore, I reverted the commit and also removed the checkout step, as it is actually not necessary.

### DIFF
--- a/.github/workflows/welcome-bot.yml
+++ b/.github/workflows/welcome-bot.yml
@@ -13,11 +13,10 @@ jobs:
     name: Welcome First-Time Contributors
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - uses: zephyrproject-rtos/action-first-interaction@58853996b1ac504b8e0f6964301f369d2bb22e5c
+      - uses: actions/first-interaction@1c4688942c71f71d4f5502a26ea67c331730fa4d # v3.1.0
         with:
           repo-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
-          pr-opened-message: |
+          pr-message: |
             Hello! Thank you for opening your **first PR** to Starlight! ✨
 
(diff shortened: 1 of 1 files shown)


delucis:
@trueberryless looks like runs are failing since we merged this PR, see https://github.com/withastro/starlight/actions/workflows/welcome-bot.yml

Could you take a look?

trueberryless:
Not at PC right now, but seems like we must replace the parameter hyphens with underscores. Can you do this @delucis ? Otherwise , I can do it later...

delucis:
OK, I can do it!

BTW this error is quite clear in VS Code with the official GitHub Actions extension:
editor with welcome-bot.yml open showing inline errors for misconfigured keys

If you don’t have that installed yet, I’d highly recommend it: https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions

trueberryless:
> If you don’t have that installed yet, I’d highly recommend it: [marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions)

I'm using Zed, but I'll install [it's plugin](https://zed.dev/extensions/github-actions), thx 👍 

Why is `issue_comment` [mandatory](https://github.com/withastro/starlight/actions/runs/23445507499/job/68207982024?pr=3776) 😭 ?
We might wanna revert this, maybe the fork is still the better option, alto outdated 😢

delucis:
Yeah, let’s revert 😄 Can you make a PR?

PR #3772, opened 2026-03-23
i18n(fr): update `manual-setup.mdx`

ArmandPhilippot:
#### Description

Adds changes from #3644 to the French translation of `manual-setup.mdx`.

### DIFF
--- a/.changeset/big-kangaroos-brush.md
+++ b/.changeset/big-kangaroos-brush.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Adds examples to the inline documentation for `title` in the Starlight configuration object
--- a/.github/workflows/welcome-bot.yml
+++ b/.github/workflows/welcome-bot.yml
@@ -13,10 +13,10 @@ jobs:
     name: Welcome First-Time Contributors
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/first-interaction@1c4688942c71f71d4f5502a26ea67c331730fa4d # v3.1.0
+      - uses: zephyrproject-rtos/action-first-interaction@58853996b1ac504b8e0f6964301f369d2bb22e5c
         with:
           repo-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
-          pr-message: |
+          pr-opened-message: |
             Hello! Thank you for opening your **first PR** to Starlight! ✨
 
             Here’s what will happen next:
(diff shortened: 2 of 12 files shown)


docs/src/content/docs/fr/manual-setup.mdx
```diff
@@ -77,8 +77,8 @@ export const collections = {
 };
 ```
 
-Starlight est également compatible avec [l'option `legacy.collections`](https://docs.astro.build/fr/reference/legacy-flags/) où les collections sont gérées en utilisant l'implémentation de collections de contenu héritée.
-Celà est utile si vous avez un projet Astro existant et que vous ne pouvez pas apporter de modifications aux collections pour utiliser un chargeur.
+Starlight est également compatible avec [l'option `legacy.collectionsBackwardsCompat`](https://docs.astro.build/fr/reference/legacy-flags/#collectionsbackwardscompat) qui préserve certaines fonctionnalités héritées des collections de contenu de la v4.
+Celà est utile si vous avez un projet Astro existant et que vous ne pouvez pas apporter de modifications aux collections pour le moment afin de migrer vers l'API de couche de contenu introduite dans la v5.0.
```
HiDeoo, line 81: Q: should we use the [same as in Astro Docs](https://docs.astro.build/fr/reference/legacy-flags/#collectionsbackwardscompat), e.g. `l’API Content Layer`?

ArmandPhilippot: Ah, I thought this was translated everywhere... I probably rushed this translation. Too tiny. 😅 

I checked the Astro Docs repo locally, and it seems this is the only page where we use "Content Layer". The other pages (e.g. [Content Loader Reference](https://docs.astro.build/fr/reference/content-loader-reference/) or [Content Collections](https://docs.astro.build/fr/guides/content-collections/) use "couche de contenu"). This is also what the [i18n guide](https://github.com/withastro/docs/blob/main/i18n-guides/fran%C3%A7ais.md#concepts-et-api-dastro) suggests.

So, I'd say we keep the translation and I'll update the Astro Docs file to also use the translated version?

HiDeoo: Oh, perfect, ignore me then. I just remembered seeing this translation here and assumed it was the one used everywhere 🙈

ArmandPhilippot: Astro Docs PR ready https://github.com/withastro/docs/pull/13582

I hadn't checked the Starlight repository! So, it seems we have 0 "couche de contenu" and two "Content Layer" in a card on [Community content](https://starlight.astro.build/fr/resources/community-content/#recettes-et-guides). Well, the card is pretty long already so I don't mind the inconsistency here... But, we could update that if we want. This doesn't seem to make the card bigger on desktop:

Screenshot of the updated card

PR #3775, opened 2026-03-23
test welcome bot (temp)

trueberryless:
now I am quite interested, if pull_request_target will take the current main branch or the old commit I target with this...


trueberryless:
Okay, it takes the target, as I would have expected from the name 😅

trueberryless:
And now the PR open event doesnt trigger anymore 😭

PR #3812, opened 2026-04-08
Update Website Carbon stats

delucis:
#### Description

- Updates the comparisons table on our environmental impact page
- Uses latest data from websitecarbon.com
- Direct link: https://deploy-preview-3812--astro-starlight.netlify.app/environmental-impact/#comparisons


trueberryless:
The changes look good to me. ✅ 

I am just not approving yet, as while we are at it, we could maybe also add the stats for [Vocs](https://vocs.dev/), which gets an [`A` with `0.05g`](https://www.websitecarbon.com/website/vocs-dev-docs/), so I think it deserves to be in this list. It is also decently popular with `1.5k` stars on GitHub and 3 years old, but seems to still be maintained based on the commit history - if not as good as other competitors.

What do you think?

delucis:
Thanks for the review @trueberryless! I think we can add any additional tools in a follow-up PR if we want to.

PR #3824, opened 2026-04-15
fix(i18n): allow getLocaleDir fallback to compile under TypeScript 6

### LINKED ISSUE #3822: TypeScript v6: Property 'language' does not exist on type 'never'

0.38.3

6.1.6

pnpm

mac

N/A

### Describe the Bug

After updating TypeScript to v6, I got the following error when running `tsc --noEmit`:

```
> tsc --noEmit

node_modules/.pnpm/@astrojs+starlight@0.38.3_astro@6.1.6_@types+node@24.12.2_rollup@4.60.1_typescript@6.0.2_/node_modules/@astrojs/starlight/utils/i18n.ts:196:38 - error TS2339: Property 'language' does not exist on type 'never'.

196  return wellKnownRTL.includes(locale.language) ? 'rtl' : 'ltr';
                                         ~~~~~~~~


Found 1 error in node_modules/.pnpm/@astrojs+starlight@0.38.3_astro@6.1.6_@types+node@24.12.2_rollup@4.60.1_typescript@6.0.2_/node_modules/@astrojs/starlight/utils/i18n.ts:196
```

It seems that this is caused by the fact that TypeScript v6 has a newer DOM library; thus, `locale` becomes `never` in [this line](https://github.com/withastro/starlight/blob/db10448ef5faa8c00a5474d5d572c9e00f667708/packages/starlight/utils/i18n.ts#L196) after the two `if` statements above.

For a short-term workaround, we can just add a `@ts-ignore` comment for this line; In the long term, https://github.com/withastro/starlight/pull/3572 should be the correct path to avoid such issues.

https://github.com/issueset/starlight-ts-6-issue.git

### Participation

SAY-5:
Per #3822, after upgrading to TypeScript v6 (which ships a newer DOM lib that knows about both `Intl.Locale.textInfo` and `Intl.Locale.getTextInfo`), `tsc --noEmit` fails with:

```
Property 'language' does not exist on type 'never'.
```

The two `if` branches in `getLocaleDir` narrow `locale` to `never` because TypeScript now sees both branches as exhaustive, which leaves the fallback unreachable from a type-narrowing standpoint.

Cast `locale` back to `Intl.Locale` in the fallback so the runtime path (Firefox before either property is supported) still compiles. Runtime behaviour is unchanged. The longer-term cleanup tracked in #3572 is still the right path; this is the small forward fix users need today.

Closes #3822


delucis (changes requested):
Thanks for the PR @SAY-5! Looks like there’s a lint error due to the “unnecessary” `as` (from TypeScript v5’s perspective). Could you take a look?

To ensure compatibility across users with both TS 5 and TS 6, it might be better to use a `// @ts-ignore` comment instead of type casting in this case. We do that in a few other places in the code base where there’s TS code that can error is some projects but not in others (so we can’t use `// @ts-expect-error`).

SAY-5:
Thanks @delucis! Switched to `@ts-ignore` on the property access and dropped the cast. That keeps TS 5 happy (no unnecessary-cast lint) and TS 6 happy (narrowed type no longer carries `language`, directive silences it). The runtime property is always present because the `in` guards above only narrow statically.

delucis:
@SAY-5 Looks like the lint job is still failing. You can see how we ignore this lint rule for `@ts-ignore` elsewhere in the codebase — I’d suggest doing the same here.

You can run `pnpm lint` locally to double check these kinds of things.

PR #3828, opened 2026-04-19
fix: empty aside alignment

MangelMaxime:
#### Description

Hello,
this PR make it so Asides with empty content have the right alignment.

**Before**

image

**After**

image

### DIFF
--- a/.changeset/cold-maps-behave.md
+++ b/.changeset/cold-maps-behave.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thanks @MangelMaxime! Not sure what the use case is for empty asides but I will assume you hit one 😀 

I’ve added a changeset and will get this published in our next patch release.

MangelMaxime:
@delucis Indeed 😅

I am working on a plugin for generating F# API documentation in Starlight. And it can happens that an API is deprecated without a message.

image

PR #3836, opened 2026-04-21
chore(docs): add Vocs to env comparison stats

trueberryless:
#### Description

This PR is a follow-up of #3812 where I [suggested the addition of Vocs](https://github.com/withastro/starlight/pull/3812#issuecomment-4225948301). Since the date below the table is now newer for Vocs, I updated all the stats for the other websites as well (sorry for the translators 😅).

> [!NOTE]
> GitBook is jumping up and down 😆 

I also thought if it would make sense to automate the updates every half a year or yearly with a GitHub action. It "shouldn't" be that hard as the table is the same in all languages and the date can be localized 🤔 What do you think?

### DIFF
--- a/docs/src/content/docs/environmental-impact.md
+++ b/docs/src/content/docs/environmental-impact.md
@@ -97,15 +97,16 @@ These tests with the [Website Carbon Calculator][wcc] compare similar pages buil
 | [docsify][dy-carbon]        | 0.04g              |   A    |
 | [Sphinx][sx-carbon]         | 0.04g              |   A    |
 | [mdBook][md-carbon]         | 0.05g              |   A    |
+| [Vocs][vs-carbon]           | 0.05g              |   A    |
 | [MkDocs][mk-carbon]         | 0.08g              |   A    |
-| [Fumadocs][fs-carbon]       | 0.08g              |   B    |
+| [Fumadocs][fs-carbon]       | 0.09g              |   B    |
 | [Docusaurus][ds-carbon]     | 0.09g              |   B    |
 | [Nextra][nx-carbon]         | 0.10g              |   B    |
 | [Docus][dc-carbon]          | 0.11g              |   B    |
-| [GitBook][gb-carbon]        | 0.28g              |   E    |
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks for the addition and re-checking all the values. Bit annoying for translators who were quickly onto the last update, but I guess we’ll survive 😁 

Re: avoiding some of that — yes, I think we could probably design a component that uses a custom translation string for the “Data collected on…” line and `Date#toLocaleDateString()` for the date formatting. It could then be dropped into the page as something like 

```mdx
<ComparisonsTable
  labels={{
    framework: 'Framework',
    co2: 'CO₂ per page visit',
    rating: 'Rating'
  }}
/>
```

(similar to how we use the color theme editor on the CSS page)

I’m less sure about automating the updating of the data though. In my experience, automation tasks need to run fairly frequently, otherwise they tend to break and all you end up doing is fixing them each time they run, which is just as much effort as doing this task manually. And that’s assuming it’s reliable to begin with: they retired the API for the carbon calculator, so it would likely require trying to scrape the value somehow, which is probably not worth the fragility. In this case, updating e.g. every 1–2 weeks would be excessive, so I think a very occasional manual collection is fine. And for readers who do want to get more up-to-date data, they can use the links we provide.

delucis:
Oh, and another localization thing to consider would be the values in grams. In fact, I notice now I’m looking that for example the German table is showing values with the style `0.01g` although it should probably be `0,01 g`. That could be handled in the component with stuff like
```js
Intl.NumberFormat(currentLocale, { style: 'unit', unit: 'gram', unitDisplay: 'narrow' }).format(value)
```

trueberryless:
> Oh, and another localization thing to consider would be the values in grams. In fact, I notice now I’m looking that for example the German table is showing values with the style `0.01g` although it should probably be `0,01 g`. That could be handled in the component with stuff like
> ```js
> Intl.NumberFormat(currentLocale, { style: 'unit', unit: 'gram', unitDisplay: 'narrow' }).format(value)
> ```

Great catch, I'll fix this in the upcoming translation 🙌

PR #3838, opened 2026-04-22
i18n(ko-KR): update `environmental-impact.md`

jsparkdev:
#### Description

- #3836


trueberryless (approved):
Thanks for the second update - this week I think 😅

LGTM 🎉

delucis (approved):
Thanks!

And indeed, a lot of changes to this table, and I have one more coming in #3842 😅 But at least this one should allow us to avoid needing translation updates any time this data changes.

PR #3841, opened 2026-04-22
i18n(fr): update `environmental-impact.md`

ArmandPhilippot:
#### Description

Adds changes from #3836 to the French translation of `environmental-impact.mdx`.

I also updated the format used in the CO₂ column. In French, this should be a comma and there should be a non breaking space before the unit. Thanks Chris for noticing that for the German version, this totally slipped my mind 😄


delucis (approved):
LGTM! 🚀 

There’ll be one more (and **final** 🤞) update for this table in #3842 but hopefully then this page will calm down.

PR #3842, opened 2026-04-22
Use component to display carbon comparisons in environmental impact page

delucis:
#### Description

- Follow-up to @trueberryless’s suggestion in #3836 
- Replaces the current Markdown table with a component that displays the same data.
- Translators only need to provide labels for the table headings and last-updated note. The data itself and the last-updated date get localized using JS APIs.
- The output is identical to what we had before but will reduce translation overhead from updating the data. This can now be done once in the component and automatically reflected across languages.

> [!TIP]
>  Note for translators, the `note` label uses i18next-style interpolation syntax, so the `{{date}}` placeholder should not be translated. For example, the German localization might look like:
> ```js
> note: 'Daten erhoben am {{date}}. Klicke auf einen Link, um aktuelle Zahlen zu sehen.',
> ```


ematipico (approved):

trueberryless (approved):
Nice one, love it ❤️ 

Thanks for taking the suggestion immediately to implementation 🚀

PR #3858, opened 2026-04-24
Update i18next

delucis:
#### Description

- This PR updates i18next from v23 to v26
- I reviewed their changelog and [migration guide](https://www.i18next.com/misc/migration-guide) and I don’t think any of the changes should impact users as most of the changes were either internal API things on the side Starlight itself sets up or removing deprecated formats and environments which we already didn’t support
- Nevertheless, I’ve included a minor changeset just to be safe and flag it in case someone is doing something elaborate with i18next.

### DIFF
--- a/.changeset/proud-schools-peel.md
+++ b/.changeset/proud-schools-peel.md
@@ -4,4 +4,4 @@
 
 Updates `i18next`, used for Starlight’s localization APIs, from v23 to v26
 
-There should be no user-facing changes from this update
+There should not be any user-facing changes from this update
--- a/.changeset/proud-schools-peel.md
+++ b/.changeset/proud-schools-peel.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/starlight': minor
+---
+
+Updates `i18next`, used for Starlight’s localization APIs, from v23 to v26
+
+There should be no user-facing changes from this update
--- a/packages/starlight/utils/createTranslationSystem.ts
+++ b/packages/starlight/utils/createTranslationSystem.ts
@@ -72,11 +72,11 @@ export async function createTranslationSystem<T extends i18nSchemaOutput>(
 
 		const t = i18n.getFixedT(lang, I18nextNamespace) as I18nT;
 		t.all = () => i18n.getResourceBundle(lang, I18nextNamespace) as ReturnType<I18nT['all']>;
-		// Since i18next 25.10.4 the `ExistsFunction` type has a signature including a predicate. We
-		// need the explicit return type to help TS narrow and infer the type correctly.
+		// Since i18next 25.10.4 the `ExistsFunction` type has a signature including a predicate that
+		// TS cannot preserve when composing a new function to add default options.
 		// See: https://github.com/i18next/i18next/issues/2425
-		t.exists = (key, options): key is keyof object =>
-			i18n.exists(key, { lng: lang, ns: I18nextNamespace, ...options });
+		t.exists = ((key, options) =>
+			i18n.exists(key, { lng: lang, ns: I18nextNamespace, ...options })) as ExistsFunction;
 		t.dir = (dirLang = lang) => i18n.dir(dirLang);
 
 		return t;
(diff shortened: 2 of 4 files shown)


.changeset/proud-schools-peel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/starlight': minor
+---
+
+Updates `i18next`, used for Starlight’s localization APIs, from v23 to v26
+
+There should be no user-facing changes from this update
```
trueberryless: I would prefer a slightly different wording, what do you think?

```suggestion
There should not be any user-facing changes from this update
```

ArmandPhilippot (dismissed):
Well, if Typescript is happy with the workaround, I'm happy too. 😄

packages/starlight/utils/createTranslationSystem.ts
```diff
@@ -72,7 +72,11 @@ export async function createTranslationSystem<T extends i18nSchemaOutput>(
 
 		const t = i18n.getFixedT(lang, I18nextNamespace) as I18nT;
 		t.all = () => i18n.getResourceBundle(lang, I18nextNamespace) as ReturnType<I18nT['all']>;
-		t.exists = (key, options) => i18n.exists(key, { lng: lang, ns: I18nextNamespace, ...options });
+		// Since i18next 25.10.4 the `ExistsFunction` type has a signature including a predicate. We
+		// need the explicit return type to help TS narrow and infer the type correctly.
+		// See: https://github.com/i18next/i18next/issues/2425
+		t.exists = (key, options): key is keyof object =>
+			i18n.exists(key, { lng: lang, ns: I18nextNamespace, ...options });
```
HiDeoo: I'm not entirely sure about the approach and comment here :thinking:

Adding `key is keyof object` does 2 things:

- `key is keyof object` transforms the regular function in a type-predicate function making it compatible with new predicate overload in `ExistsFunction` added in the latest version.
- `key is keyof object`, because `keyof object` is `never` becomes `key is never` and `never` is a subtype of any narrowed type, including `TKeys & SelectorKey` in `ExistsFunction`.

The runtime behavior is not what we just described at all, so basically, the new annotation ends up only silencing TypeScript but I think it requires a lot of internal knowledge of how TypeScript works to understand what it does exactly and why.

I think it would be just more straightforward and easier to understand if we just casted the function to `I18nT['exists']`, similarly to what we do in the previous line.


```suggestion
		// Since i18next 25.10.4 the `ExistsFunction` type has a signature including a predicate that
		// TS cannot preserve when composing a new function to add default options.
		// See: https://github.com/i18next/i18next/issues/2425
		t.exists = ((key, options) =>
			i18n.exists(key, { lng: lang, ns: I18nextNamespace, ...options })) as I18nT['exists'];
```

delucis: Yeah, we could do that too — I think I tried a variant on that while experimenting but ESLint complained (because it said we were pointlessly casting something that was already that type). But I think I just forgot the `()` wrapping the function 😅

It doesn’t matter too much what we choose given this is all the same result for keeping TypeScript happy internally in a way users never see. Happy to go with a cast if its preferable. (I think `as ExistsFunction` would be clearer in this case?)

delucis: Done in c4c2786eeb159c7dc68e00f77922faeace66c7c1

HiDeoo (dismissed):

packages/starlight/utils/createTranslationSystem.ts
HiDeoo: > I think `as ExistsFunction` would be clearer in this case?

Not sure one is actually clearer 😅 My train of thoughts was "let's cast it as we want it to be, which is `I18nT['exists']` in the end so I don't have to know what it is actually` is" but I can see why the other way around can also makes sense.

Anw, thanks for considering, I think it's a bit more obvious (at least to me) what is happening.

matthewp (approved):

PR #3862, opened 2026-04-28
Add bottom margin to nested lists

### LINKED ISSUE #3860: Compact list inside of an expanded list should have bottom margin

0.38.4

6.1.9

npm

Windows

Edge

### Describe the Bug

Found this [discussion](https://discord.com/channels/830184174198718474/1113047413779869728/1113053005558849617) on Discord.

When a compact list is the last item within an expanded list (e.g. `<Steps>`), the list element does not have a bottom margin set.

```mdx
<Steps>

1. This is the first step.

2. This is the second step.
   - This is sub-step **a**.
   - This is sub-step **b**.
   - This is sub-step **c**.

3. This is the third step.

</Steps>
```

### Actual Output

Image

### Expected Output

Image

https://stackblitz.com/edit/github-oapspcza?file=src%2Fcontent%2Fdocs%2Fguides%2Fexample.md

### Solution

~~Would recommend removing the `ol` and `ul` exclusion [here](https://github.com/withastro/starlight/blob/ccf6000549c580796a8a8730637074c25e3ccc44/packages/starlight/style/markdown.css#L39-L40).~~

Addressed in #3862 with custom CSS.

### Participation

itrew:
Closes #3860 

#### Description

When a compact list is nested in an expanded list it doesn't have a bottom margin applied, making the list look uneven. Often times I like to utilize an expanded ordered list for `<Steps>` components, but might want to include a bulleted `ul` in there. ~~This change just removes the `ul` and `ol` exclusions for applying a bottom margin that exist in list items.~~

After looking into it more, there were times when the original solution was adding margins when they shouldn't be added (e.g. a compact list containing another compact list). New CSS now applies the margin to only the list elements that are the last child of an `li` and are preceded by some other non-inline element.

### DIFF
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -36,8 +36,6 @@
 	> :is(
 		:last-child:not(
 				li,
-				ul,
-				ol,
 				a,
 				strong,
 				em,
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -54,7 +54,7 @@
 			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
 			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
 			 */
-		:not(script):has(~ script:last-child):not(:has(~ :not(script))),
+		:not(script):has(~ script:last-child):not(:has(~ :not(script)))
 	) {
 		margin-bottom: 1.25rem;
 	}
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -36,6 +36,8 @@
 	> :is(
 		:last-child:not(
 				li,
+				ul,
+				ol,
 				a,
 				strong,
 				em,
@@ -52,8 +54,18 @@
 			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
 			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
 			 */
-		:not(script):has(~ script:last-child):not(:has(~ :not(script)))
+		:not(script):has(~ script:last-child):not(:has(~ :not(script))),
 	) {
+		margin-bottom: 1.25rem;
+	}
+
+	/* Apply a margin to a list element if it's in an li with other non-inline elements. */
+	.sl-markdown-content
+	li:has(
+		> :not(a, strong, em, del, span, input, code, br) + :not(li, a, strong, em, del, span, input, code, br)
+	)
+	:is(ol,ul):last-child
+	{
 		margin-bottom: 1.25rem;
 	}
 
--- a/packages/starlight/style/markdown.css
+++ b/packages/starlight/style/markdown.css
@@ -31,41 +31,31 @@
 		overflow-wrap: anywhere;
 	}
 
+	/*
+	 * This complex selector manages spacing inside lists.
+	 *
+	 * - Finds all lists containing items with non-inline direct children (e.g.
+	 *   ul > li > p)
+	 * - Applies a bottom margin to the last non-inline child of each list item
+	 *   in this list
+	 *
+	 * This ensures that even if only one list item contains a block element
+	 * (most commonly a paragraph), siblings of that list item are spaced
+	 * consistently.
+	 */
 	.sl-markdown-content
-	li
-	> :is(
-		:last-child:not(
-				li,
-				ul,
-				ol,
-				a,
-				strong,
-				em,
-				del,
-				span,
-				input,
-				code,
-				br,
-				script,
-				:where(.not-content *)
-			),
-			/**
-			 * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
-			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
-			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
+	 :is(ol, ul):has(> li > :not(a, strong, em, del, span, input, code, br, script, ol, ul))
+	 > li
+	 > :is(
+			:last-child:not(a, strong, em, del, span, input, code, br, script, :where(.not-content *)),
+			/*
+			 * For list items ending with 1 or multiple script elements (`:has(~
+			 * script:last-child)`), we need to style the last non-script
+			 * element (`:not(script)`) that doesn't have a subsequent sibling
+			 * that is not a script (`:not(:has(~ :not(script)))`).
 			 */
-		:not(script):has(~ script:last-child):not(:has(~ :not(script)))
-	) {
-		margin-bottom: 1.25rem;
-	}
-
-	/* Apply a margin to a list element if it's in an li with other non-inline elements. */
-	.sl-markdown-content
-	li:has(
-		> :not(a, strong, em, del, span, input, code, br) + :not(li, a, strong, em, del, span, input, code, br)
-	)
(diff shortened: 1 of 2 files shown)


packages/starlight/style/markdown.css
```diff
@@ -59,6 +59,16 @@
 		margin-bottom: 1.25rem;
 	}
 
+	/* Apply a margin to a list element if it's in an li with other non-inline elements. */
+	.sl-markdown-content
+	li:has(
+		> :not(a, strong, em, del, span, input, code, br) + :not(li, a, strong, em, del, span, input, code, br)
```
delucis: Couple of notes/queries here:

1. We should probably include `script` as the rule above does so that a script doesn’t trigger a change.

```suggestion
		> :not(a, strong, em, del, span, input, code, br, script) + :not(li, a, strong, em, del, span, input, code, br, script)
```

2. What’s the purpose of the `+` sibling clause? Would this have the same impact by ignoring lists so you’re effectively checking for a list with a non-list sibling preceding it?

```suggestion
	li:has(
		> :not(a, strong, em, del, span, input, code, br, script, ol:last-child, ul:last-child)
	)
```

packages/starlight/style/markdown.css
```diff
@@ -59,6 +59,16 @@
 		margin-bottom: 1.25rem;
 	}
 
+	/* Apply a margin to a list element if it's in an li with other non-inline elements. */
```
delucis: Might be worth merging these two declarations to keep the margin value in sync/shared between the two?

```suggestion
),
	/* Apply a margin to a list element if it's in an li with other non-inline elements. */
```

delucis (commented):
Thanks @itrew! Left a couple of small notes — would love to hear what you think.

delucis:
OK, so I was curious and spent some time playing with these styles/selectors. I think I may have come up with an alternative approach to replace the existing selector and your addition here. Here’s a StackBlitz demo showing the changes: https://stackblitz.com/edit/github-oapspcza-9dwcdwcs?file=src%2Fcontent%2Fdocs%2Findex.md

The proposed change would be to use this:

```css
.sl-markdown-content
   :is(ol, ul):has(> li > :not(a, strong, em, del, span, input, code, br, script, ol, ul))
   > li
   > :is(
      :last-child:not(:where(.not-content *)),
      /**
       * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
       * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
       * sibling that is not a script (`:not(:has(~ :not(script)))`).
       */
      :not(script):has(~ script:last-child):not(:has(~ :not(script)))
   ) {
      margin-bottom: 1.25rem;
}
```

Logic:
- Find all lists that contain items that contain non-inline direct children (with exception for nested lists which will handle their own children)
- For all items in these lists, apply the bottom margin to the last child (with caveat for scripts as currently)

IIRC we didn’t consider this kind of logic previously because `:has()` was not widely supported enough and we wanted an approach that worked for our full browser support matrix. We later introduced the script bug fix in #3026 and updated our support matrix to drop official support for browsers that don’t support `:has()` (see https://github.com/withastro/starlight/pull/3026#discussion_r2016292813), released in v0.33 a year ago.

Would be good to do more testing with my proposed change, but at least in the Starlight docs this updated rule causes no visible impact. Will try on a few more docs sites.

delucis:
Running a test with my suggestion against Astro’s docs I spotted [a couple](https://docs.astro.build/en/guides/view-transitions/#client-side-navigation-process) [of places](https://docs.astro.build/en/guides/deploy/aws-via-flightcontrol/#how-to-deploy) this change would improve that contains nested lists similar to your issue:

| | |
|---|---|
| image | image |


But also spotted one issue with the new styles:

image

Can be fixed by tweaking to exclude inline elements:

```css
.sl-markdown-content
   :is(ol, ul):has(> li > :not(a, strong, em, del, span, input, code, br, script, ol, ul))
   > li
   > :is(
      :last-child:not(a, strong, em, del, span, input, code, br, script, :where(.not-content *)),
      /**
       * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
       * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
       * sibling that is not a script (`:not(:has(~ :not(script)))`).
       */
      :not(script):has(~ script:last-child):not(:has(~ :not(script)))
   ) {
      margin-bottom: 1.25rem;
}
```

So then the logic is:

- Find all lists that contain items that contain non-inline direct children (with exception for nested lists which will handle their own children)
- For all items in these lists, apply the bottom margin to the last **non-inline** child (with caveat for scripts as currently)

Given the complexity and size of the Astro docs, they give me a fairly high confidence to suggest we go with this. I’ll add a proper review suggestion.

packages/starlight/style/markdown.css
```diff
@@ -59,6 +59,16 @@
...
+	.sl-markdown-content
+	li:has(
+		> :not(a, strong, em, del, span, input, code, br) + :not(li, a, strong, em, del, span, input, code, br)
+	)
+	:is(ol,ul):last-child
+	{
+		margin-bottom: 1.25rem;
+	}
```
delucis, line 60:
```suggestion
	/*
	 * This complex selector manages spacing inside lists.
	 * - Finds all lists containing items with non-inline direct children (e.g. ul > li > p)
	 * - Applies a bottom margin to the last non-inline child of each list item in this list
	 * This ensures that even if only one list item contains a block element (most commonly a
	 * paragraph), siblings of that list item are spaced consistently.
	 */
	.sl-markdown-content
	 :is(ol, ul):has(> li > :not(a, strong, em, del, span, input, code, br, script, ol, ul))
	 > li
	 > :is(
			:last-child:not(a, strong, em, del, span, input, code, br, script, :where(.not-content *)),
			/**
			 * For list items ending with 1 or multiple script elements (`:has(~ script:last-child)`), we
			 * need to style the last non-script element (`:not(script)`) that doesn't have a subsequent
			 * sibling that is not a script (`:not(:has(~ :not(script)))`).
			 */
			:not(script):has(~ script:last-child):not(:has(~ :not(script)))
	 ) {
		margin-bottom: 1.25rem;
	}
```

delucis (changes requested):
OK, proposing this change to handle all list-item spacing consistently:

packages/starlight/style/markdown.css
itrew: This looks fantastic, appreciate the suggestion! CSS isn't my strong suit; your solution is much more elegant.

delucis (approved):
Thanks for nerdsniping me into thinking through this! 😁 

Given the change has a small risk of being disruptive if someone has custom styles, I’ve made this a minor and we’ll get it published in v0.39 soon.

HiDeoo (approved):
Tested a few websites using the visual regression toolkit we built a while ago for cascade layers. The only differences spotted were actually expected improvements.

For example, one of the website I tested was the [Biome website](https://biomejs.dev/) and mostly 2 pages had some differences and they were also expected:

- https://biomejs.dev/linter/rules/use-naming-convention/
- https://biomejs.dev/internals/changelog_v1/

Sharing below the resulting visual diff for the `/linter/rules/use-naming-convention/` pages only (not sharing the one for the changelog page as it's a 18.4MB image 🥵). Differences start around the middle of the page, after the nested list before the "const, let, using, and var declarations" text but it's all ends up being expected improvements piling up until the end of the page.

Image

PR #3872, opened 2026-05-04
Enable `text-autospace` in Chinese and Japanese documents

tats-u:
#### Description

- Closes #3871 

Especially in Chinese and Japanese, the CSS property `text-autospace: normal` is preferred for better looking like books by automatically inserting spaces with the moderate width.

- https://developer.chrome.com/blog/css-i18n-features#inter-script_spacing_text-autospace
- https://developer.mozilla.org/docs/Web/CSS/text-autospace
- https://webkit.org/blog/16574/webkit-features-in-safari-18-4/#:~:text=Text%20auto%20space
- https://github.com/vuejs/vitepress/issues/4996
- https://github.com/web-infra-dev/rspress/pull/2920
- https://github.com/web-infra-dev/rspress/releases/tag/v2.0.0-rc.3

Without this, some users and AIs try to insert such spaces manually as workarounds but they are too wide and prevent keywords from searched.

Japanese:

Before:

image

After:

image

Note: the current Chinese content of Starlight translation is not affected because it took the workarounds like manually inserting spaces.

The other languages are not affected.

### DIFF
--- a/.changeset/gentle-hats-help.md
+++ b/.changeset/gentle-hats-help.md
@@ -4,9 +4,7 @@
 
 Enables [the CSS property `text-autospace`](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-autospace) in Chinese and Japanese documents.
 
-The specification is mature enough to meet most requirements, though there could be edge cases where it causes issues. You can disable this until the specification and browser implementations are more stable.
-
-If you would prefer to revert this change, you can add the following custom CSS to your site:
+The specification is mature enough to meet most requirements, though there could be edge cases where it causes issues. Most users should be satisfied or unaffected by this change and no additional measures should be necessary, but if you would prefer to revert this change, you can add the following custom CSS to your site:
 
 ```css
 [lang] {
--- a/.changeset/gentle-hats-help.md
+++ b/.changeset/gentle-hats-help.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/gentle-hats-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Enables [the CSS property `text-autospace`](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-autospace) in Chinese and Japanese documents.
```
delucis, line 5: We often try to provide guidance in case there’s anyone trying to preserve backwards compatibility. Maybe something like this?

````suggestion
Enables [the CSS property `text-autospace`](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-autospace) in Chinese and Japanese documents.

If you would prefer to disable autospacing in Chinese and Japanese pages, you can add the following custom CSS to your site:

```css
[lang]:where(:lang(zh, ja)) {
	text-autospace: initial;
}
```
````

I don’t imagine it’s super likely but helpful in case anyone does hit it.

delucis (commented):
Thanks @tats-u — this looks good to me. Just one small changelog suggestion.

tats-u:
Based on your feedback, I have updated the changeset description.

delucis (dismissed):
Thanks again @tats-u! We’ll get this released in our next minor.

delucis (approved):

tats-u:
Oh, my fuChrome, the notation `:lang(zh, ja)` is only supported by Firefox. Shame on them.

image

I didn't notice it because I'm a Firefox fanboy. I'll send you a fix.

PR #3887, opened 2026-05-07
Add new icons to Starlight’s built-in set

delucis:
#### Description

Adds generic clock, desktop, mobile phone, app window, database, server, git branch, notes, question mark, analytics, and padlock icons. Also adds a new logo icon for SolidJS. The mobile icon has a bit of a funny name — `mobile-android` — but I kept it to match the source library’s name.

Closes #3881
Closes #3863

Which icons to include built-in is always a bit subjective, but I went through our icon library and tried to add a handful that might meet some more common needs on docs sites given I was already adding a couple based on user requests.

For the SolidJS logo, I [asked on their Discord server](https://discord.com/channels/722131463138705510/1501984081360195745/1501987176580382912) to confirm which style they preferred for a monochrome icon (as it deviates from their more common blue gradient style).


matthewp (approved):
I didn't pull this down and look at the SVGs so trusting your judgment here!

delucis:
Ah yeah, I could have attached some screenshots to simplify reviewing. The new icons are available in the deploy preview icon reference: https://deploy-preview-3887--astro-starlight.netlify.app/reference/icons/

They’re mixed in with other icons, but here:

image

image

HiDeoo (commented):
Neat, they all looks good to me. Just one tiny question regarding `mobile-android`. Does this name match the original one? Does other `mobile-*` exist? I don't personally see what makes this icon Android-specific 🤔

delucis:
> Neat, they all looks good to me. Just one tiny question regarding `mobile-android`. Does this name match the original one? Does other `mobile-*` exist? I don't personally see what makes this icon Android-specific 🤔

Heh, just recently updated the PR description about this.

Yes, this is the source library’s name, which I wanted to keep for consistency. And no, they do not have a `mobile-ios` or whatever 😅 They have `mobile-android-alt` and `mobile-vibrate`: https://icones.js.org/collection/uil?s=mobile

(I guess the original naming is to help search in case someone searches “android”? But still, definitely an odd choice.)

We could rename it though if we think it’s an issue? We already don’t 100% match the source naming IIRC, but it’s an easy default to try to keep for consistency.

HiDeoo (approved):
Ah, sorry, I didn't refresh the PR page after checking out the PR locally.

Well, definitely odd indeed, specially if they don't have other platforms, but let's keep it for consistency 🤷 

Thanks 🙌

PR #3890, opened 2026-05-08
Make sure to enable `text-autospace` even in Chromium-based browsers

tats-u:
#### Description

- Fixes https://github.com/withastro/starlight/pull/3872#issuecomment-4405904913  

`:lang(zh, ja)` used in #3872 shamefully has not been supported Chromium ~~and Safari~~. As a result, `text-autospace` has been enabled only in Firefox. Replaced it with the combination of `:lang(zh)` and `:lang(ja)`.

Screenshot in Chrome (Pay attention to the slight presence of whitespace around "Astro" and "Starlight"):

Before (https://starlight.astro.build/ja/getting-started/):

image

After:

image

### DIFF
--- a/.changeset/empty-cases-pick.md
+++ b/.changeset/empty-cases-pick.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Make sure to enable `text-autospace` even in non-Firefox browsers
+Fixes CSS selector for `text-autospace` styles in Chromium browsers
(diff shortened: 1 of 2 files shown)


tats-u:
https://deploy-preview-3890--astro-starlight.netlify.app/ja/getting-started/

delucis (approved):
Thanks for catching that! Based on the table on MDN it looks like Safari also supports the comma-separated list, but Chromium browsers do not. Easy fix though in this case luckily.

tats-u:
Thank you but we forgot to fix the title of this issue and the comment.

PR #3893, opened 2026-05-09
Fix inaccurate comment on `:lang` selectors

tats-u:
#### Description

- Fixes https://github.com/withastro/starlight/pull/3890#pullrequestreview-4253703763 & https://github.com/withastro/starlight/pull/3890#issuecomment-4411190048

Tweak a comment: "non-Firefox" → "Chromium-based"

Safari has already supported the syntax `:lang(zh, ja)`.

image

The position of Opera is the culprit.

This change is so trivial that we should ship it with other changes.

### DIFF
--- a/.changeset/happy-kings-know.md
+++ b/.changeset/happy-kings-know.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fix inaccurate comment on `:lang` selectors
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks for circling back on that one! I removed the changeset as this is not a user-facing change — it’s mainly internal so we can keep track of our decision making.

tats-u:
I see. Thank you for the merge.

Issue #3875, opened 2026-05-05
Approve postinstall scripts of esbuild & sharp


tats-u:
### What version of `starlight` are you using?

0.38.4 (main)

### What version of `astro` are you using?

N/A

### What package manager are you using?

pnpm

### What operating system are you using?

N/A (Windows)

### What browser are you using?

N/A

### Describe the Bug

https://github.com/pnpm/pnpm/releases/tag/v10.0.0
https://github.com/pnpm/pnpm/releases/tag/v11.0.0
https://pnpm.io/cli/approve-builds

pnpm has disabled postinstall scripts since v10 to prevent supply chain attacks (e.g. axios, chalk, debug). However, `pnpm-workspace.yaml` in this repository has not explicitly approved or rejected any build scripts in `esbuild` or `sharp` yet. They are famous packages that utilize postinstall scripts and [listed in Bun's allowlist](https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt).

You should approve their build scripts by adding their names to `allowBuild` in `pnpm-workspace.yaml`.

1. Clone this repo
2. `pnpm i`
3. Expected: no messages / Actual: pnpm emits a warning message that it declined their build scripts and you should approve (or decline) them in `pnpm-workspace.yaml`

### Link to Minimal Reproducible Example

https://github.com/withastro/starlight

### Participation

- [x] I am willing to submit a pull request for this issue.

delucis:
I don't think we want to allow this — these packages use postinstall scripts to build their native code on platforms that don't have a prebuilt binary. However, binaries are available for all common platforms so it's very rare a contributor would need the scripts to run. 

Although we can *probably* trust sharp and esbuild as projects, running postinstall scripts is still an attack surface I'd prefer to avoid in the case of a supply chain compromise, which can impact even a trusted, well-managed package in theory.

Given the risks and given this is not necessarily fixing anything for anyone, I'm not sure I'm convinced we should do this.

tats-u:
This repository has not activated some features to prevent supply chain attacks yet.

https://pnpm.io/supply-chain-security

- Block unmatured versions that have just been released (`minimumReleaseAge`)
- Block versions that have not been published via GitHub Actions (`trustPolicy`) Note: recent versions of both packages have certificates that have been released via GitHub Actions. Versions compromised in supply chain attacks are published without going through GitHub Actions, essentially using the credentials of stolen tokens under the guise of compromised maintainers.

[...]

PR #3894, opened 2026-05-09
Explicitly disallow postinstall scripts of esbuild & sharp

tats-u:
#### Description

Opposite of:

- https://github.com/withastro/starlight/issues/3875
- https://github.com/withastro/starlight/pull/3876

but the goal is the same: to silence the Postinstall Scripts Unapproved notification when `pnpm install`.

References:

| Package | URL | Quote |
| -- | -- | -- |
| esbuild | https://github.com/evanw/esbuild/issues/4085#issuecomment-2696009329 | <q>The install script still optimizes the command by replacing the JavaScript shim by the actual binary executable, so that running the esbuild command directly doesn't have to run node first just to shell out to the executable. So it's not necessary for correctness (and you're welcome to disable the install script if you'd like to and if it still works for you) but it might result in esbuild being slower than it would have been otherwise.</q> |
| sharp | https://sharp.pixelplumbing.com/install/ | <q>When using `pnpm`, add `sharp` to [ignoredBuiltDependencies](https://pnpm.io/settings#ignoredbuiltdependencies)[^pnpm-11] to silence warnings.</q> |

[^pnpm-11]: equivalent to `allowBuilds` = `false` in pnpm v10.26+.

Changesets are **NOT** required.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -3,6 +3,11 @@ packages:
   - 'examples/**/*'
   - 'docs'
 
+# https://github.com/withastro/starlight/issues/3875
+allowBuilds:
+  esbuild: false # https://github.com/evanw/esbuild/issues/4085#issuecomment-2696009329
+  sharp: false # https://sharp.pixelplumbing.com/install/: "When using pnpm, add sharp to ignoredBuiltDependencies to silence warnings."
+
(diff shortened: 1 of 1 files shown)


delucis (approved):
I think with pnpm 11 this will become required in any case, so happy to add it for now!

PR #3900, opened 2026-05-12
fix: improve theme select legibility in dark mode on Windows Firefox

### LINKED ISSUE #3426: Legibility problem with Theme select in Windows Firefox

0.35.3

5.13.7

pnpm

Windows 11 Pro

Firefox 142.0.1 (64-bit)

### Describe the Bug

In firefox, the theme select menu while in dark mode has numerous problems. It is fine in light mode. The background around the options is white when it should be dark. The text on the current hover / checked option is white, instead of inverted.

Image

See the following screenshot. This is true of all new starlight installs and has for at least since starlight 0.35.2, but likely much older. You can see it on the official https://docs.astro.build astro docs website as well.

My workaround was to make a custom css file in my install to fix it with the following:
```css
	[data-theme='dark'] starlight-theme-select {
		& select,
		& select option,
		& select option:checked {
			background-color: var(--sl-color-gray-6);
			color: var(--sl-color-text);
		}
		
		& select option:hover {
			background-color: var(--sl-color-accent-high);
			color: var(--sl-color-text-invert);
		}
	}
```

My workaround produces the following in firefox dark mode. I did not extensively test this workaround in other browsers:

Image

While I believe it maintains functionality in light mode.

https://astro.new/latest/preview/starlight-basics/

### Participation

\- 🦥

dashitongzhi:
## Description

Fixes the theme select dropdown having illegible colors in dark mode on Windows Firefox.

On Windows, Firefox renders native `<select>` dropdown options with white backgrounds in dark mode, making the text nearly invisible. The hovered/checked option text color is also white-on-white.

### Before
The theme select in dark mode on Windows Firefox shows white option backgrounds with poor contrast, making it hard to read available options.

### After
Explicit dark-theme color rules are applied to the select and option elements, using the existing Starlight design tokens (`--sl-color-gray-6`, `--sl-color-text`, `--sl-color-accent-high`, `--sl-color-text-invert`).

## Changes

- Added dark-mode-specific styles for `select`, `option`, and `option:checked` in `Select.astro`
- Added hover state styles for options in dark mode
- Uses existing CSS custom properties for consistency with the rest of the theme

Fixes #3426

### DIFF
--- a/packages/starlight/components/Select.astro
+++ b/packages/starlight/components/Select.astro
@@ -78,6 +78,25 @@ interface Props {
 			color: var(--sl-color-gray-1);
 		}
 
+		/*
+		 * Firefox on Windows renders native select dropdowns with incorrect
+		 * colors in dark mode — option backgrounds appear white and
+		 * checked/hovered text is invisible. Explicitly set colors for the
+		 * dark theme to ensure legibility.
+		 * See: https://github.com/withastro/starlight/issues/3426
+		 */
+		:global([data-theme='dark']) select,
+		:global([data-theme='dark']) select option,
+		:global([data-theme='dark']) select option:checked {
+			background-color: var(--sl-color-gray-6);
+			color: var(--sl-color-text);
+		}
+
+		:global([data-theme='dark']) select option:hover {
+			background-color: var(--sl-color-accent-high);
+			color: var(--sl-color-text-invert);
+		}
+
 		@media (min-width: 50rem) {
 			select {
 				font-size: var(--sl-text-sm);
--- a/packages/starlight/components/Select.astro
+++ b/packages/starlight/components/Select.astro
@@ -79,22 +79,25 @@ interface Props {
 		}
 
 		/*
-		 * Firefox on Windows renders native select dropdowns with incorrect
-		 * colors in dark mode — option backgrounds appear white and
-		 * checked/hovered text is invisible. Explicitly set colors for the
-		 * dark theme to ensure legibility.
+		 * Firefox on Windows renders native select option dropdowns with
+		 * incorrect colors in dark mode — option backgrounds appear white
+		 * and checked/hovered text is invisible. Use @supports to target
+		 * Firefox (-moz-appearance) so other browsers are unaffected.
+		 * Only fix option backgrounds, not the select itself, to preserve
+		 * transparent backgrounds on pages like the homepage.
 		 * See: https://github.com/withastro/starlight/issues/3426
 		 */
-		:global([data-theme='dark']) select,
-		:global([data-theme='dark']) select option,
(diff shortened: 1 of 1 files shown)


HiDeoo (changes requested):

If I'm not missing anything, the suggested solution hit the limitation described in [this comment](https://github.com/withastro/starlight/issues/3426#issuecomment-3291344829), or [this one](https://github.com/withastro/starlight/discussions/3180#discussioncomment-13098623) where I link to the exact Firefox bug, which prevent Starlight itself from setting a solid background colour, which may or may not match the nav bar.

For example, the [homepage](https://deploy-preview-3900--astro-starlight.netlify.app/) of the deploy preview of the PR now renders as such on Chrome:

Image

Let me know if I'm missing something.

dashitongzhi:
## Fix: Scope dark mode select fix to Firefox only

HiDeoo, thanks for the review! You're right — the fix was too broad.

### Changes made:

1. **Firefox-only targeting**: Wrapped in `@supports (-moz-appearance: none)` so only Firefox is affected. Chrome/Safari/Edge are completely untouched.

2. **Removed `select` from the selector**: Only targeting `option` and `option:checked` now, NOT the `<select>` element itself. This preserves the transparent background on the homepage and other pages.

3. **Use semantic token**: Changed `--sl-color-gray-6` → `--sl-color-bg-nav` for better theme consistency (Copilot's suggestion).

The dropdown options will still have proper dark backgrounds in Firefox, but the select element itself stays transparent — so the homepage hero nav bar looks correct in all browsers.

HiDeoo:
Thanks for the update.

As I mentioned in [this comment](https://github.com/withastro/starlight/discussions/3180#discussioncomment-13093821), I was personally never able to reproduce the issue on Windows at the time. Before potentially discussing this new approach further and its potential drawbacks, I would like to make sure it's still an issue or the issue didn't change.

I'm mentioning this because I just tested to reproduce again now on a Windows machine, and while I still can't reproduce the issue, I did notice on the latest Firefox version, the dropdown menu now renders differently from the previous screenshots I shared [here](https://github.com/withastro/starlight/discussions/3180#discussioncomment-13093821). Now, in dark mode, the dropdown background is the expected color and no longer white which was not the case before.

Would you be able to share the Firefox version you're currently using for testing, and if not the latest, could you try updating and testing the default version (e.g. on https://starlight.astro.build/ and not this PR) and share a new screenshot to see if it matches what I'm seeing?

I know it's a bit annoying but issues that we can't easily reproduce or don't know all the requirements to reproduce can be very difficult to fix, and need extra care.

delucis:
Closing due to lack of response. Happy to reopen in the future if there are answers to HiDeoo’s questions.

yeshuabradford80-afk (approved):

PR #3901, opened 2026-05-12
CI/E2E tests improvements

HiDeoo:
#### Description

This PR contains various improvements to the E2E tests mostly focusing on CI performance.

Here is a table summarizing some of the most impactful changes comparing the before, before average, and after times on my personal machine and CI:

| Task       | Host         | Before[^1]                                                                                | [Before average](https://github.com/withastro/starlight/actions/metrics/performance?tab=jobs&filters=workflow_file_name%3Aci.yml) | After                                                                                     |
| ---------- | ------------ | ----------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
| E2E tests  | Personal Mac | 43.6s                                                                                     | N/A                                                                                                                               | 28.9s                                                                                     |
| E2E tests  | CI - Linux   | [2m 27s](https://github.com/withastro/starlight/actions/runs/25486117899/job/74782187861) | 2m 34s                                                                                                                            | [1m 29s](https://github.com/withastro/starlight/actions/runs/25737206045/job/75577362094) |
| E2E tests  | CI - Windows | [5m 9s](https://github.com/withastro/starlight/actions/runs/25486117899/job/74782187895)  | 5m 11s                                                                                                                            | [3m 11s](https://github.com/withastro/starlight/actions/runs/25737206045/job/75577362071) |
| A11y tests | Personal Mac | 1m 1s                                                                                     | N/A                                                                                                                               | 41s                                                                                       |
| A11y tests | CI - Linux   | [2m 39s](https://github.com/withastro/starlight/actions/runs/25486117899/job/74782187917) | 2m 43s                                                                                                                            | [1m 47s](https://github.com/withastro/starlight/actions/runs/25737206045/job/75577362076) |

You can find below [a review](https://github.com/withastro/starlight/pull/3901#pullrequestreview-4273143722) were I try to comment on the reasoning behind most of the changes.

[^1]: For CI times, the "before" times are from [merging to `main` the recent autogenerated sidebar groups refactor](https://github.com/withastro/starlight/actions/runs/25486117899).


.github/workflows/ci.yml
```diff
@@ -51,6 +51,9 @@ jobs:
               - '!docs/src/content/docs/**'
             packages:
               - 'packages/**'
+              - 'package.json'
+              - 'pnpm-lock.yaml'
+              - 'pnpm-workspace.yaml'
```
HiDeoo, line 56: Realized when tweaking filters that we were not tracking these files but an operation could technically change only one or more of these files in a way that could have an impact on Starlight but we would not re-run some jobs.

.github/workflows/ci.yml
```diff
@@ -81,30 +84,14 @@ jobs:
...
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      # Uninstall man-db to prevent man page updates taking a long time after package installations
-      # on Ubuntu 24.04.
-      # https://github.com/actions/runner/issues/4030
-      # https://github.com/actions/runner-images/issues/10977
-      - name: Uninstall man-db
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          sudo apt-get update
-          sudo apt-get remove man-db
```
HiDeoo, line 92: We are now using browsers already pre-installed on the images ([Linux](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#browsers-and-drivers) / [Windows](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md#browsers-and-drivers) reference) which are regularly updated so as we are no longer installing anything, we can remove this workaround.

.github/workflows/ci.yml
```diff
@@ -81,30 +84,14 @@ jobs:
...
-      - name: Test packages (Linux)
-        if: matrix.os == 'ubuntu-latest'
-        run: pnpm -r test:e2e:chrome
-      # Firefox is used for Windows E2E tests as Chrome on Windows requires the installation of the
-      # Server Media Foundation windows feature, which can take up to 3 minutes to install on CI.
-      # https://github.com/microsoft/playwright/blob/e7bff526433b6dcb02801763ab5b1c6407902d47/packages/playwright-core/src/server/registry/dependencies.ts#L79-L89
-      - name: Test packages (Windows)
-        if: matrix.os == 'windows-latest'
-        run: pnpm -r test:e2e:firefox
+      - run: pnpm i --filter @astrojs/starlight --filter "@e2e/*"
```
HiDeoo, line 92: Probably quite small impact, but I tested on my Windows machine, and without a cache hit, this was a ~4s improvement when running `pnpm i`, so probably a bit more on CI but happy to remove it if we think it's not worth it.

.github/workflows/ci.yml
```diff
@@ -81,30 +84,14 @@ jobs:
...
-      - run: pnpm i
-      - name: Test packages (Linux)
-        if: matrix.os == 'ubuntu-latest'
-        run: pnpm -r test:e2e:chrome
-      # Firefox is used for Windows E2E tests as Chrome on Windows requires the installation of the
-      # Server Media Foundation windows feature, which can take up to 3 minutes to install on CI.
-      # https://github.com/microsoft/playwright/blob/e7bff526433b6dcb02801763ab5b1c6407902d47/packages/playwright-core/src/server/registry/dependencies.ts#L79-L89
-      - name: Test packages (Windows)
-        if: matrix.os == 'windows-latest'
-        run: pnpm -r test:e2e:firefox
```
HiDeoo, line 107: We are now using browsers already pre-installed on the images, so we can remove the Firefox usage which was only added to workaround a Chrome dependency installation on Windows that was taking a long time (note that Playwright can use pre-installed Chrome but it's not the same for Firefox that would still require an extra installation).

.github/workflows/ci.yml
```diff
@@ -172,10 +150,12 @@ jobs:
 
       - name: Run accessibility audit
         working-directory: ./docs
-        run: pnpm t
+        run: pnpm test:ci
 
   windows-smoke:
     name: Docs site builds on Windows
+    needs: changes
+    if: ${{ needs.changes.outputs.packages == 'true' || needs.changes.outputs.ci == 'true' || needs.changes.outputs.docs_not_content == 'true' }}
```
HiDeoo, line 158: Noticed we didn't have a filter on the Windows smoke build, and a scenario where only modifying the documentation content in a way that would only break on Windows and not on Linux during deployment seemed unlikely, so I added a filter. Happy to revert if I didn't think of some obvious cases.

docs/__a11y__/docs.test.ts
```diff
@@ -1,22 +1,18 @@
...
-test('does not report accessibility violations on the docs site', async ({ docsSite }) => {
-	let violationsCount = 0;
+const urls = await getDocsSiteUrls();
 
-	const urls = await docsSite.getAllUrls();
+for (const url of urls) {
+	const { pathname } = new URL(url);
 
-	for (const url of urls) {
+	test(`does not report accessibility violations: ${pathname}`, async ({ docsSite }, testInfo) => {
```
HiDeoo, line 8: Instead of looping overs URLs in a test, we're looping over URLs to create multiple tests so they can be parallelized.

docs/__a11y__/reporter.ts
HiDeoo, line 1: This entire new file exists because we are now parallelizing a11y tests and our previous approach, in case of errors, to just log the violations as soon as they are found, can now lead to interleaved logs from multiple tests, which is hard to read. This custom reporter now uses the Playwright API to save the violations during tests, and only emit logs at the end of the tests.

docs/__a11y__/test-utils.ts
```diff
@@ -64,6 +60,40 @@ process.env.ASTRO_DISABLE_UPDATE_CHECK = 'true';
 
 const locales = lunariaConfig.locales.map((locale) => locale.lang);
 
+export async function getDocsSiteUrls() {
```
HiDeoo, line 63: This is the same code as before, just hoisted out of the fixture so we can access it outside of a test and loop over URLs.

docs/__a11y__/test-utils.ts
```diff
@@ -121,16 +116,16 @@ class DocsSite {
...
+		if (violations.length === 0) return;
 
-		if (violations.length > 0) {
-			console.error(`> Found ${violations.length} violations on ${url}\n`);
-			await reportViolations(violations, new DefaultTerminalReporter(true, true, false));
-			console.error('\n');
-		} else {
-			console.log(`> Found no violations on ${url}`);
-		}
+		await testInfo.attach(A11yReportAttachmentName, {
```
HiDeoo, line 125: We now save violations instead of immediately reporting them and they will be consumed later by our custom reporter.

docs/playwright.config.ts
```diff
@@ -1,28 +1,33 @@
...
-	forbidOnly: !!process.env['CI'],
+	forbidOnly: isCI,
+	fullyParallel: true,
 	projects: [
 		{
 			name: 'Chrome Stable',
 			use: {
 				...devices['Desktop Chrome'],
+				// Re-use system Chrome on CI to avoid re-installing it on every run.
+				channel: isCI ? 'chrome' : undefined,
```
HiDeoo, line 14: On CI, we now specify a channel which means we can use pre-installed browsers on the image. Locally, it's the same as before and Playwright will download a browser if needed.

packages/starlight/__e2e__/fixtures/basics/src/noNodeModule.ts
HiDeoo, line 1: This plugin has been moved to the `basics` fixture instead of having a dedicated one just for this test. I measured locally and running it in a dedicated fixture which requires to run `astro build` (which is the most expensive part of the test) is not worth and it's faster to just add it to another fixture.

packages/starlight/__e2e__/fixtures/basics/package.json
```diff
@@ -4,6 +4,6 @@
   "private": true,
   "dependencies": {
     "@astrojs/starlight": "workspace:*",
-    "astro": "^6.2.2"
+    "astro": "^6.3.1"
```
HiDeoo, line 7: E2E test files are now run in parallel. This Astro version bump across the monorepo is required due to a bug being fixed in this specific new version.

Our E2E tests using the production build rely on the preview server which can change ports when the default port is in use. When this happens, a bug caused the new port to not be properly surfaced. This was not an issue for us before because we were running the E2E test files sequentially, but with the new parallelization, we started to hit this issue.

Note that the [release notes](https://github.com/withastro/astro/releases/tag/astro%406.3.1) don't mention this bug fix as it was [released without a changeset](https://github.com/withastro/astro/pull/16382/changes#diff-ed03375837ca0bf96c6865baa67c6f7109c041aa3533696ae585ddeb6fc748dbR111-R119).

packages/starlight/__e2e__/fixtures/ssr/astro.config.mjs
```diff
@@ -9,13 +9,8 @@ export default defineConfig({
...
 	adapter: node({ mode: 'standalone' }),
 	compressHTML: false, // for easier debugging
-	// Output to different folders and expose on different ports
-	// on each case so the servers don't conflict with
-	// each other during tests.
+	// Output to different folders so the servers don't conflict with each other during tests.
 	outDir: prerendering ? 'dist' : 'build',
-	server: {
-		port: prerendering ? 4322 : 4321,
-	},
```
HiDeoo, line 18: We can no longer use this approach with parallel tests, as some other test files may already have used these ports so we rely on automatic port assignment instead. More details on this in a comment in `packages/starlight/__e2e__/test-utils.ts`.

HiDeoo (commented):
Tried to explain the reasoning behind most changes, but let me know if you have any questions or want me to clarify anything.

packages/starlight/playwright.config.ts
```diff
@@ -1,27 +1,20 @@
...
-				headless,
+				// Re-use system Chrome on CI to avoid re-installing it on every run.
+				channel: isCI ? 'chrome' : undefined,
+				headless: true,
 			},
 		},
 	],
 	testMatch: '__e2e__/*.test.ts',
 	timeout: 40 * 1_000,
-	workers: 1,
```
HiDeoo, line 26: We now parallelize file tests, and the amount of workers now [defaults](https://playwright.dev/docs/api/class-testconfig#test-config-workers) to half of the number of logical CPU cores (and standard GitHub-hosted runners for public repositories are [not single-core machines](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)).

.github/workflows/ci.yml
delucis: I think this is the only one that gave me pause, but on balance I think you’re right. Windows issues are usually things like path resolution etc. which are unlikely to come up in our docs code.

delucis (approved):
Wow — big PR! The comments walking through the changes are super helpful. I think it all looks good!

PR #3902, opened 2026-05-12
Update package manager to pnpm 11

delucis:
#### Description

This PR updates to pnpm 11 and configures a 3-day minimum release age for dependencies to mirror the main `astro` monorepo: https://github.com/withastro/astro/blob/3abcc86c092ca2a3455fc0ffc72b6b7ca300f1a2/pnpm-workspace.yaml#L32

This helps defend a little against supply chain attacks, but also avoids us installing deps that then cause issues in the `astro` monorepo CI because they’re newer than their config allows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "typescript": "^6.0.3",
     "typescript-eslint": "^8.59.2"
   },
-  "packageManager": "pnpm@10.33.3",
+  "packageManager": "pnpm@11.1.1",
   "size-limit": [
     {
       "name": "/index.html",
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -22,3 +22,6 @@ peerDependencyRules:
     - '@algolia/client-search'
     - 'playwright'
     - 'search-insights'
+
+# Prevent installing packages published in the last 3 days
+minimumReleaseAge: 4320


package.json
```diff
@@ -28,7 +28,7 @@
     "typescript": "^6.0.3",
     "typescript-eslint": "^8.59.2"
   },
-  "packageManager": "pnpm@10.33.3",
+  "packageManager": "pnpm@11.1.1",
```
ematipico, line 31: You should add `engines` and declare `node` >= 22

ematipico (approved):

package.json
delucis: Oh, why? We use `.nvmrc` for Node version, which is set to 24.14.1.

pnpm-workspace.yaml
```diff
@@ -22,3 +22,6 @@ peerDependencyRules:
     - '@algolia/client-search'
     - 'playwright'
     - 'search-insights'
+
+# Prevent installing packages published in the last 3 days
+minimumReleaseAge: 4320
```
HiDeoo, line 27: 1. Should we update the Renovate config too?

According to https://docs.renovatebot.com/key-concepts/minimum-release-age/#my-package-manager-has-support-for-minimum-release-age-how-does-renovate-work-with-that-do-i-need-to-set-it-in-both

> We recommend specifying minimum release age in both your Renovate and package manager configuration.

2. Should we have some `minimumReleaseAgeExclude`, e.g. `@astrojs/*`, etc.?

delucis: Re: 1, yes! #3904 will do that.

Re: 2, hmm, good question. I lean towards only adding exemptions if we find they’re needed? Easy enough to add later if we find we want to update an Astro dep before the threshold is passed (I can imagine this happening e.g. at a major Astro release).

The `astro` monorepo notably has [numerous exemptions](https://github.com/withastro/astro/blob/0f59ba44e7b9a81f515bed1b861fe55969ef6f3e/pnpm-workspace.yaml#L33-L62) which get quite out of date, so I’d like to avoid that if we can.

HiDeoo: Ah sorry, missed the other PR. And for point 2, makes sense, fine with that.

HiDeoo (approved):

package.json
ematipico: It's a native check done by the package manager, so you don't need nvmrc

PR #3904, opened 2026-05-12
Wait 3 days for Renovate updates

delucis:
#### Description

- Companion to #3902 
- Tells Renovate to wait for 3 days before applying updates to match our pnpm policy
- In practice not super important as we mainly use Renovate for GitHub Actions, which don’t involve pnpm, but still good to keep these in sync

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -10,6 +10,7 @@
 	],
 	rangeStrategy: 'bump',
 	ignorePaths: ['**/node_modules/**'],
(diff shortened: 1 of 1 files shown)


HiDeoo (approved):
Looks good to me, thanks for the follow-up.

I guess there may be a day where we may want an exception for `withastro/automation`, but like you said, we can add it later if we ever need it, and this won't prevent us from updating it manually if ever needed anw.

PR #3907, opened 2026-05-17
feat: add blank template

steveiliop56:
#### Description

This pull request, adds a simple new page template called `blank` that allows the end-user to fully customize what's rendered in the `ContentPanel` without having to worry about CSS hacks to hide things like the `PageTitle` which are present in the `splash` template. For more information see #3906

Additionally, it refactors a small section in the `Page.astro` that handles conditional rendering for pages with a hero image. Hopefully it improves visibility and maintainability in comparison to the previous version.

This pull request does not impact the default Starlight look.

### DIFF
--- a/.changeset/tricky-wings-sit.md
+++ b/.changeset/tricky-wings-sit.md
@@ -2,16 +2,6 @@
 '@astrojs/starlight': minor
 ---
 
-Adds support for `blank` page template allowing for full control over the content panel of the page.
+Adds support for an empty page template to allow full control over the content panel of the page.
 
-By default, the most basic page template is `splash` which is intended to be used for landing pages and other pages that don't require a sidebar. However it still renders things like the page title which can be annoying to hide in case you want to customize the entire page. The new `blank` template allows you to do just that, giving you a completely blank canvas to work with. You can use this template for any page where you want to have full control over the content panel.
-
-When you use the `blank` template, you will be assigned a simple slot under a content panel:
-
-```astro
-<ContentPanel>
-  <slot />
-</ContentPanel>
-```
-
-You will then be responsible for rendering the page title and any other content you want to include in the page. This allows you to create custom layouts and designs for your pages without having to worry about the default page structure getting in the way.
+Read more about the new `'blank'` template in the ["Page layout" guide](https://starlight.astro.build/guides/customization/#page-layout).
--- a/.changeset/tricky-wings-sit.md
+++ b/.changeset/tricky-wings-sit.md
@@ -1,0 +1,17 @@
+---
+'@astrojs/starlight': minor
+---
+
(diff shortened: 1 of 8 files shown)


.changeset/tricky-wings-sit.md
```diff
@@ -0,0 +1,17 @@
...
+
+When you use the `blank` template, you will be assigned a simple slot under a content panel:
+
+```astro
+<ContentPanel>
+  <slot />
+</ContentPanel>
+```
+
+You will then be responsible for rendering the page title and any other content you want to include in the page. This allows you to create custom layouts and designs for your pages without having to worry about the default page structure getting in the way.
```
trueberryless:
```suggestion
Adds support for an empty page template to allow full control over the content panel of the page.

Read more about the new `'blank'` template in the ["Page layout" guide](https://starlight.astro.build/guides/customization/#page-layout).
```

trueberryless (commented):
Thanks for this amazing first time contribution 🥳 

Overall the changes look good, I would just like to point out some docs changes:
1. Could you add a new section under https://starlight.astro.build/guides/customization/#page-layout which explains the new `'blank'` template similar to how you currently explain it in the changeset. One difference to the current changeset explanation tho is that the docs should be more "stateless" and unemotional, meaning that they reflect the current state of how things work, so no need for phrases like "**new** `'blank'` template" or any introduction to reason about the third template.

    I think since this is a new feature, we can add an entire new heading, e.g. something like that (I'm sorry that I can't suggest it in the code directly, since it is outside the diff range):
  
    ````md
	## Page layout
	
	By default, Starlight pages use a layout with a global navigation sidebar and a table of contents that shows the current page headings.

	Starlight offers two additional layouts, which can be configured by setting the [`template` field](/reference/frontmatter/#template) in a page’s frontmatter.
	
	The `'splash'` template applies a wider page layout without sidebars.
	This works particularly well for landing pages and you can see it in action on the [homepage of this site](/).
	
	```md {5}
	---
	# src/content/docs/index.md
	
	title: My Landing Page
	template: splash
	---
	```
	
	The `'blank'` template leaves only the navigation header and creates a blank canvas, which can be useful if you want to fully customize your landing page.
	
	```md {5}
	---
	# src/content/docs/index.md
	
	title: Empty Page
	template: blank
	---
	```
	````
	
2. The advantage of documenting the new template there is that we can then just link to it from the changeset and we don't need to document all those details in the changeset, like my suggestion.

With these changes, I think the PR looks pretty solid, and we can wait for maintainers to look at your proposal 👍

.changeset/tricky-wings-sit.md
steveiliop56: Good idea, fixed!

PR #3911, opened 2026-05-18
fix(MobileMenuToggle): trap keyboard focus inside the mobile menu

### LINKED ISSUE #2697: mobile menu does not contain focus to the menu on tab navigation

0.31.1

5.0.5

npm

linux

chrome, firefox

### Describe the Bug

When navigating the mobile menu with the `tab` key, focus is not contained to the menu resulting in tab navigation continuing to the current page after the last item in the menu is reached.

**Steps to reproduce**
1. Open [Starlight Docs](https://starlight.astro.build/getting-started/)
2. Reduce viewport to < 800px
3. Open mobile menu
4. Tab through all elements until the last menu element is reached (theme switcher)
5. Tab again

**Expected Result**
The logo should be focused

**Actual Result**
Focus continues to the underlying page, first with the `summary` tag, then the `anchor` tag for "Astro framework", etc.

https://stackblitz.com/edit/withastro-starlight-a96pv7eh

### Participation

timothyjordan:
#### Description

Fixes #2697.

When the mobile menu is open on narrow viewports, Tab navigation currently escapes the menu after the last focusable element (the theme switcher) and lands on the underlying page content.

This PR sets the native `inert` attribute on `.main-frame` and the skip-to-content link while the mobile menu is expanded, so the browser keeps keyboard focus inside the header and sidebar without a hand-maintained list of focusable selectors. A `matchMedia('(min-width: 50em)')` listener collapses the menu and clears `inert` automatically if the viewport grows past the mobile breakpoint — so rotating or resizing a device cannot leave the page locked behind a stale trap.

#### Test plan

Verified manually against `pnpm --filter starlight-docs dev` at mobile (600×900) and desktop (1024×900) widths:

- Open the menu: `aria-expanded="true"`, `inert` set on both `.main-frame` and `.sl-skip-link`.
- Tab through the menu: focus visits sidebar `<summary>` group toggles (e.g. "Start Here", "Guides") and never lands inside `.main-frame`. Shift+Tab from the first sidebar item returns to the toggle/close button.
- Programmatic `.focus()` on a link inside `.main-frame` is refused by the browser while inert.
- Resize to 1024px wide while open: `aria-expanded` drops to `"false"`, `inert` is cleared, article links are interactive again.
- Press Escape: menu closes and focus returns to the toggle (regression check).

Per maintainer request, no dedicated e2e tests are added in this PR — if we add coverage later, it should live in `__e2e__/basics.test.ts` so it shares the basics fixture build.

`pnpm lint`, `pnpm typecheck`, and `pnpm --filter @astrojs/starlight test` (519/519) all pass.

#### Changeset

Patch bump for `@astrojs/starlight` included.


packages/starlight/__e2e__/mobile-menu.test.ts
HiDeoo, line 1: We don't really have E2E tests right now for this part of the code or similar interactions. I'm not sure yet if we want to add some or not, but if we do, we would probably add them to the `packages/starlight/__e2e__/basics.test.ts` file to avoid rebuilding the `basics` fixture twice.

Maybe until we agree on the approach, we can omit them and add them later when we have a better idea of the final implementation and know for sure that we want to add E2E tests for this part of the code?

HiDeoo (commented):

At first glance, I notice a few potential issues with the current approach:

- The current code would exclude the close button from the focus loop, which feels a bit odd to me when quickly testing it out.
- The manually maintained list of interactive elements could be difficult to maintain. For example, right now it misses `<summary>` that we are using for sidebar groups. I think it would also need to be extended to include all possible interactive elements as users may add new ones in their own project.
- If we change the viewport width to a point where the mobile menu becomes hidden, e.g. changing the orientation on a tablet or resizing the window, the focus trap would still be active and prevent users from accessing the rest of the page.

Thinking outside of the box, I wonder if we could use the `inert` attribute to achieve a similar effect in this case. I _think_ we would only need to set `inert` on the `.main-frame` and potentially the "Skip to content" link to achieve a similar effect without needing to maintain a list of interactive elements. This could end up being a tiny method called from `setExpanded()` and a media query listener to handle the viewport width changes.

What do you think? Would you be open to exploring this approach so we can discuss the pros and cons of each approach in more detail?

packages/starlight/__e2e__/mobile-menu.test.ts
timothyjordan: Great feedback! Updated PR incoming.

timothyjordan:
Rewrote the focus containment around the `inert` attribute as you suggested.

`MobileMenuToggle` now marks both `.main-frame` and the skip-to-content link `inert` from `setExpanded()`, and a `matchMedia('(min-width: 50em)')` listener collapses the menu (clearing `inert`) when the viewport grows past the mobile breakpoint. That addresses all three concerns from your initial review:

1. The close/toggle button is no longer special-cased — the browser handles it as part of the surrounding `<nav>`.
2. `<summary>` toggles (and any user-added interactive elements inside the sidebar) are now covered automatically since the browser is doing the focusable enumeration.
3. Rotating/resizing past the mobile breakpoint cleanly resets the state instead of leaving the page locked.

`SkipLink` picked up a `.sl-skip-link` class so it can be targeted for `inert` alongside `.main-frame`. Also dropped `__e2e__/mobile-menu.test.ts` per your other comment — happy to add tests under `__e2e__/basics.test.ts` instead if/when you'd like coverage after this lands.

farrosfr:
Hi @timothyjordan and @HiDeoo 👋 — your `fix(upgrade): avoid downgrading older dist-tags`-style approach with the `inert` attribute is great work on #2697.

I noticed your description explicitly leaves the door open for follow-up test coverage ("Per maintainer request, no dedicated e2e tests are added in this PR — if we add coverage later, it should live in `__e2e__/basics.test.ts`"). I'd like to write that coverage as a complementary PR.

Before I do, quick coordination question:

1. **Where should the tests live?** Your description says `__e2e__/basics.test.ts` — confirm that's still preferred?
2. **Scope:** I'm planning four tests covering:
   - `.main-frame` gets `inert` when the menu opens
   - `.sl-skip-link` gets `inert` when the menu opens
   - Both `inert` attributes are cleared when the menu closes (Escape / button toggle)
   - `matchMedia` listener collapses the menu and clears `inert` when viewport grows past 50em
3. **Where should the PR target?** Options: (a) a separate PR against `withastro/main` that I link in a comment on #3911 (you can then merge them together or treat as follow-up), or (b) push as commits directly to this PR's branch (timothyjordan:fix/mobile-menu-focus-trap-2697) so it ships as one atomic PR. (b) is cleaner but requires your permission since it's your branch.

Also a related gap I noticed while reading the code: the `'.sl-skip-link'` selector in `syncInert()` is silently a no-op when users override the `SkipLink` component via `components.SkipLink` schema (their custom component won't carry the class). Happy to address that in the same PR if you'd like — or leave it for a separate one.

Let me know how you'd like to proceed 🙏

HiDeoo:
Thanks for your interest @farrosfr 

With Astro v7 around the corner, I did not get the time yet to prioritize this PR since it was last updated, my apologies for that.

As commented [here](https://github.com/withastro/starlight/pull/3911#discussion_r3287341738), I think we will first want to agree on the approach, make sure it properly fixes the issue, does not have any regressions, etc. before moving forward with any tests.

Once we have a clear direction, we can then evaluate the best way to implement tests for this change, if any. Considering the cost of E2E tests, we will definitely want to focus tests that really provide value to us and our users, and we would probably want to avoid adding too many tests and rather focus on a few high-value tests instead if needed.

In the meantime, the most helpful feedback could be to actually manually test the current implementation on many devices and browsers, making sure it works as expected, and report such tests, e.g. which device, which browser, which browser version, etc. have been tested, and what the results were.

Thanks everyone for your patience and help in making sure we get this right.

farrosfr:
## Manual test report - `inert` rewrite of mobile menu focus trap

Ran the current `inert`-based implementation on the deploy preview against the 3 concerns from @HiDeoo's 2026-05-22 review. All 3 are resolved.

### Environment

| Device (emulated) | Viewport   | Browser                          |
|-------------------|------------|----------------------------------|
| iPhone X          | 375 x 812  | Chromium 144 (Playwright, headed) |
| iPad portrait     | 768 x 1024 | Chromium 144 (Playwright, headed) |
| Android (mid)     | 412 x 915  | Chromium 144 (Playwright, headed) |

Caveat: this is **Playwright emulation**, not real-device testing - I don't have a Safari/iOS device at hand. Happy to re-run if anyone has a real-device matrix to add.

### Results

| # | Test                                                | iPhone X | iPad portrait | Android mid |
|---|-----------------------------------------------------|:--------:|:-------------:|:-----------:|
| 1 | Hamburger opens menu                                |    ✓     |      ✓       |     ✓      |
| 2 | `aria-expanded` flips true                          |    ✓     |      ✓       |     ✓      |
| 3 | `.sl-skip-link` + `.main-frame` get `inert`         |    ✓     |      ✓       |     ✓      |
| 4 | 20x Tab stays inside sidebar                        |    ✓     |      ✓       |     ✓      |
| 5 | Shift+Tab wraps inside                              |    ✓     |      ✓       |     ✓      |
| 6 | Resize while open releases `inert`                  |    ✓     |      ✓       |     ✓      |

18/24 pass. The 6 fails are on 2 edge-case tests (Escape, link click) and are test-script ordering issues in my Playwright harness, not code bugs:
- Escape test fired on a menu already auto-closed by the prior resize test
- Link-click test asserted state on the page after navigation

### HiDeoo's 2026-05-22 concerns

1. → "Manually maintained list of interactive elements could be difficult to maintain" - **fixed**: `inert` on `.main-frame` covers any future element automatically.
2. → "Current code would exclude the close button from the focus loop" - **fixed**: close button lives inside `.sl-sidebar`, not `.main-frame`, so it stays tabbable.
3. → "If we change the viewport width... the focus trap would still be active" - **fixed**: `matchMedia` listener fires on the resize test above; both `inert` flags go false and `aria-expanded` flips to `false`.

### Real-device tests still needed (I can't do these from a PC)

- iPhone Safari / iOS WebKit (historical inert support gaps)
- Tablet orientation flip (portrait ↔ landscape) while menu open
- NVDA / VoiceOver announcement of sidebar role + active element
- 200% zoom
- Samsung Internet

Screenshots and re-runnable harness are at the PR's deploy-preview mirror if useful during review; happy to push them anywhere maintainers prefer (PR comment repo, gist, branch, etc.).

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -45,6 +56,17 @@ import Icon from '../user-components/Icon.astro';
 				this.btn.focus();
 			}
 		}
+
+		// Mark the main content and skip link `inert` while the mobile menu is
+		// expanded so the browser keeps keyboard focus inside the header and
+		// sidebar without relying on a hand-maintained list of focusable selectors.
+		syncInert() {
+			const active =
+				!this.mobileMql.matches && document.body.hasAttribute('data-mobile-menu-expanded');
```
delucis: I think we can probably simplify this: we already know the expanded state when calling `syncInert()`, so can just pass it in instead of re-computing it:
```suggestion
		syncInert(active: boolean) {
```

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -28,11 +29,21 @@ import Icon from '../user-components/Icon.astro';
...
+				if (e.matches && this.getAttribute('aria-expanded') === 'true') {
+					this.setExpanded(false);
+				}
+			});
 		}
 
 		setExpanded(expanded: boolean) {
 			this.setAttribute('aria-expanded', String(expanded));
 			document.body.toggleAttribute('data-mobile-menu-expanded', expanded);
+			this.syncInert();
```
delucis: Passing it in here would then toggle the `inert` attribute whenever the menu is expanded:

```suggestion
			this.syncInert(expanded);
```

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -28,11 +29,21 @@ import Icon from '../user-components/Icon.astro';
 			if (parentNav) {
 				parentNav.addEventListener('keyup', (e) => this.closeOnEscape(e));
 			}
+
+			// Collapse the menu (which also clears `inert`) when the viewport
+			// grows past the mobile breakpoint so users on rotating or resizing
+			// devices are not locked out of the page content.
+			this.mobileMql.addEventListener('change', (e) => {
```
delucis: That also means we can inline the media watcher and avoid the indirection of the top-level class property:
```suggestion
			matchMedia('(min-width: 50em)').addEventListener('change', (e) => {
```

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -17,6 +17,7 @@ import Icon from '../user-components/Icon.astro';
 <script>
 	class StarlightMenuButton extends HTMLElement {
 		btn = this.querySelector('button')!;
+		mobileMql = matchMedia('(min-width: 50em)');
```
delucis:
```suggestion
```

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -28,11 +29,21 @@ import Icon from '../user-components/Icon.astro';
...
 				parentNav.addEventListener('keyup', (e) => this.closeOnEscape(e));
 			}
+
+			// Collapse the menu (which also clears `inert`) when the viewport
+			// grows past the mobile breakpoint so users on rotating or resizing
+			// devices are not locked out of the page content.
+			this.mobileMql.addEventListener('change', (e) => {
+				if (e.matches && this.getAttribute('aria-expanded') === 'true') {
+					this.setExpanded(false);
+				}
```
delucis: This can be simplified down to this:

```suggestion
				this.setExpanded(e.matches);
```

It means `setExpanded()` always runs including in “redundant” cases, but that’s harmless and keeps the code simple.

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -45,6 +56,17 @@ import Icon from '../user-components/Icon.astro';
...
+
+		// Mark the main content and skip link `inert` while the mobile menu is
+		// expanded so the browser keeps keyboard focus inside the header and
+		// sidebar without relying on a hand-maintained list of focusable selectors.
+		syncInert() {
+			const active =
+				!this.mobileMql.matches && document.body.hasAttribute('data-mobile-menu-expanded');
+			for (const selector of ['.main-frame', '.sl-skip-link']) {
+				document.querySelector<HTMLElement>(selector)?.toggleAttribute('inert', active);
+			}
```
delucis: Bit of a nit, but this can be done with a single selector:

```suggestion
			document.querySelectorAll<HTMLElement>('.main-frame, .sl-skip-link').forEach((el) => {
				el.toggleAttribute('inert', active);
			});
```

.changeset/mobile-menu-focus-trap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Keeps keyboard focus inside the mobile menu while it is open by marking the rest of the page `inert`. Focus containment now covers all interactive elements (including sidebar group `<summary>` toggles and any custom components) and is released automatically if the viewport grows past the mobile breakpoint.
```
delucis: Small suggestion here to keep the focus on the user-facing change (users don’t need to know the how/why of implementation details, but the what of the change that impacts them):

```suggestion
Keeps keyboard focus inside the mobile menu while it is open, preventing focus moving to hidden interactive elements in page content.
```

packages/starlight/components/SkipLink.astro
```diff
@@ -2,17 +2,17 @@
...
 	@layer starlight.core {
-		a {
+		.sl-skip-link {
 			clip: rect(0, 0, 0, 0);
 			position: fixed;
 			top: 0.75rem;
 			inset-inline-start: 0.75rem;
 		}
-		a:focus {
+		.sl-skip-link:focus {
```
delucis: Thanks to Astro’s scoping, there’s no need to change these even if we do introduce a class for other purposes:
```suggestion
		a {
clip: rect(0, 0, 0, 0);
position: fixed;
top: 0.75rem;
inset-inline-start: 0.75rem;
}
		a:focus {
```

delucis (changes requested):
Thanks very much for the PR @timothyjordan! I think the implementation is basically sound 🎉 

I did a quick test in the following browsers on macOS:

- Firefox 152
- Safari 26.2
- Chrome 149
- Edge 148

In each, it seems to work as expected: in the current production deployment of the Starlight docs, tabbing would continue into page content; in the preview deployment for this PR, tabbing would end at the end of the menu either moving to browser chrome or back to the top of the page (behaviour varies slightly here depending on the browser, but in each case I believe its the standard behaviour and working correctly).

I would obviously love more tests on different browsers and OSes, but as a baseline it gives me some confidence that this approach works.

I left a few suggestions to clean up the code — I’d also be happy to apply them myself directly if that would be helpful.

FrancoKaddour:
Came across this while working on a similar fix (#4029). The inert approach is definitely cleaner — tested the deploy preview on mobile and focus containment works correctly. Happy to help with any additional testing if needed.

FrancoKaddour:
Thanks for the refactor! One thing that looks off to me in the latest commit: the media query handler was changed from setExpanded(false) to setExpanded(e.matches). For (min-width: 50em), e.matches is true when the viewport crosses into desktop territory — so rotating from mobile to desktop now calls setExpanded(true), which sets .main-frame and .sl-skip-link to inert on desktop and cuts off keyboard access to the main content.

I think the original alwaysfalse was correct: regardless of which direction the viewport changes, we want to reset to a closed state. Should this be reverted to setExpanded(false)?

delucis (dismissed):
Thanks again for the PR @timothyjordan! I took the liberty of applying my review suggestions (and refactoring a little more) so we can get this released. I also added some end-to-end tests that check the expected behaviour using our basic fixture to help avoid regressions in the future.

@FrancoKaddour I think you may have been looking at an isolated intermediary commit? The code is still using `setExpanded(false)` when the viewport changes.

delucis (approved):
Made a last couple of updates to testing based on some feedback from @HiDeoo.

Should be good to go — let’s trap that focus 🚀

PR #3914, opened 2026-05-22
Disable minimum release age for `withastro/automation` updates

HiDeoo:
#### Description

This PR disables the minimum release age check of 3 days [recently added](https://github.com/withastro/starlight/pull/3904)  for `withastro/automation` updates.

While reviewing #3903, and even after retrying the PR, I was confused why the `renovate/stability-days` check was still pending, even though all the updates were released around 10 days ago ([last change for `withastro/automation`](https://github.com/withastro/automation/commit/53c85fabe6ac4ed7a96e49a005f03fa6cc63d2ac) - [last release for `pnpm/action-setup`](https://github.com/pnpm/action-setup/releases/tag/v6.0.8)).

I'm not a Renovate expert, but I _think_ I finally understand the reason:

- The `withastro/automation` is a digest update as we're using the latest commit SHA on the `main` branch in workflow files without a version tag (`@<sha> # main`).
- As documented [here](https://docs.renovatebot.com/key-concepts/minimum-release-age/#which-update-types-take-minimumreleaseage-into-account), for digest updates, having a release timestamp that can have `minimumReleaseAge` enforced is "Generally not supported".
- In Renovate 42, the absence of a release timestamp will be treated as if the release is not yet past the timestamp, which provides a safer default.

I'm not 100% sure, but I think this is the reason why the `renovate/stability-days` check is still pending, and will probably be always pending, preventing us from merging other updates.

A few extra points:

- If merging this PR and retrying #3903 does not resolve the issue, we can revert this PR and investigate further.
- If this works, it could still be a short-term solution and maybe a more long term solution would be to add proper releases with tags to `withastro/automation`.
- No matter what we end up doing, we will probably need to do the same to the Docs repo as I think https://github.com/withastro/docs/pull/13873 may have the same issue.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -34,5 +34,11 @@
 			matchDepTypes: ['uses-with'],
 			enabled: false,
 		},
+		{
+			description: 'Disable minimum release age for withastro/automation digest updates',
+			matchManagers: ['github-actions'],
+			matchPackageNames: ['withastro/automation'],
+			minimumReleaseAge: null,
+		},
 	],
 }


ArmandPhilippot (approved):
Not a Renovate expert, but this makes sense to me! I had a similar issue when I enabled by mistake `lockFileMaintenance` in a repo.

And thanks for digging into this, I noticed the docs PR still waiting but I hadn't yet taken the time to look into what was wrong.

HiDeoo:
Thanks for the review 🙌 Perfect, let's try it in this repo, and if it doesn't work, we can revert it, and if it works, we can discuss if we should change how we release `withastro/automation`.

PR #3919, opened 2026-05-25
Update Lunaria workflow Node.js version

HiDeoo:
#### Description

This PR bumps the Node.js version used in the Lunaria workflow to the latest Node.js 22 version [like in our other workflows](https://github.com/withastro/starlight/blob/main/.github/workflows/ci.yml#L16) instead of the specific `22.12.0` version.

We recently updated to pnpm 11 which requires Node.js 22.13 or higher and would [error out](https://github.com/withastro/starlight/actions/runs/26351811215/job/77571272653?pr=3918) with the following message when running the workflow:

```
ERROR: This version of pnpm requires at least Node.js v22.13
The current version of Node.js is v22.12.0
```

### DIFF
--- a/.github/workflows/lunaria.yml
+++ b/.github/workflows/lunaria.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22.12.0
+          node-version: 22
           cache: pnpm
 
       - name: Install dependencies
--- a/.github/workflows/lunaria.yml
+++ b/.github/workflows/lunaria.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22
+          node-version: 22.22.3
           cache: pnpm
 
       - name: Install dependencies


.github/workflows/lunaria.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22.12.0
+          node-version: 22
```
delucis: Would it be better to specify the latest v22 patch to pin to for consistency across runs?

```suggestion
          node-version: 22.22.3
```

([Version 22.22.3](https://nodejs.org/en/download/archive/v22.22.3) was released on May 13, 2026, so is old enough for us to use following our general >3 days old logic.)

HiDeoo: We definitely can as I only picked `22` to match our `ci.yml` (we could also revisit updating it).

trueberryless (approved):
LGTM 🎉

.github/workflows/lunaria.yml
delucis: Ah, that’s true. I think we opted for the unpinned versions because we’re not using auto-updates from Renovate, so it allowed us to use latest “for free”. Maybe something to reconsider there too to see if we’d like to let Renovate manage more stuff giving us the guarantee of stable versions until we explicitly update.

delucis (approved):

PR #3920, opened 2026-05-25
Use relative hero link in starter templates

sebdanielsson:
Fixes broken links when deployed on a subpath.


delucis:
Thanks for the PR @sebdanielsson! In general we don’t recommend relative links like this as their behaviour can be confusing and unpredictable (even changing once deployed depending on whether a page is loaded with or without a trailing `/`).

Obviously the current starters do break when deployed to a subpath, but I think it’s still better to keep the current absolute paths even if it means the links will need manually updating for subpath deployments.

PR #3923, opened 2026-05-27
feat: add support for Astro 6.4 and Sätteri

### LINKED ISSUE #3934: Tables not being rendered anymore

0.39.2

6.4.2

pnpm

Linux

Firefox

### Describe the Bug

Tables are not being renderede anymore.

Throw a table like this anywhere: https://pastes.dev/X4MDET7mtJ

### Participation

Princesseuh:
#### Description

This PR adds support for Astro 6.4 and the Sätteri processor introduced in that version. It does so by checking what  (if any) `markdown.processor` is set, and adding plugins accordingly to that.

Versions earlier than 6.4 are also still supported, when `markdown.processor` isn't present, Starlight pushes plugins to the now deprecated remarkPlugins / rehypePlugins options instead.

This PR **does not** remove the hard dependency on `@astrojs/markdown-remark`, it would be possible to with similar patterns as the ones done for Sätteri, but I figured that it wasn't up to me to decide that.

- Closes https://github.com/withastro/starlight/issues/3934


packages/starlight/package.json
ocavue: Should we also update `@astrojs/mdx` from v5 to v6?

packages/starlight/index.ts
```diff
@@ -105,9 +106,41 @@ export default function StarlightIntegration(
...
+				// the (now-deprecated) `markdown.{remark,rehype}Plugins` config below.
+				const processor = config.markdown?.processor;
+				if (processor?.name === 'satteri') {
+					const { isSatteriProcessor, starlightSatteriPlugins } = await import(
+						'./integrations/satteri'
+					);
+					if (isSatteriProcessor(processor)) {
+						const { mdastPlugins, hastPlugins } = starlightSatteriPlugins(remarkRehypeOptions);
+						processor.options.mdastPlugins.push(...mdastPlugins);
+						processor.options.hastPlugins.push(...hastPlugins);
```
HiDeoo: Don't we need to handle [features](https://satteri.bruits.org/docs/features/) too?

For example, if a user configure its processor using:

```ts
markdown: { processor: satteri() },
```

By default, the `directive` feature is disabled but Starlight relies on it so I guess we would need to enable it?

Extra question: unrelated to this PR, but it seems defaults are different between Unified and Sätteri, e.g. `markdown.smartypants` and `unified({ smartypants })` defaulting to `true` while Sätteri `smartPunctuation` [defaults to `false`](https://satteri.bruits.org/docs/features/). I guess this would need to be fixed in Sätteri/Astro/Starlight to not be breaking?

.changeset/satteri-support.md
```diff
@@ -0,0 +1,19 @@
...
+// astro.config.mjs
+
+import { defineConfig } from 'astro/config';
+import { satteri } from '@astrojs/markdown-satteri';
+
+export default defineConfig({
+  markdown: {
+    processor: satteri(),
+  },
+});
```
HiDeoo, line 19:
````suggestion
});
```
````

I didn't review the changeset or docs yet, but noticed this code block was not closed.

File: packages/starlight/integrations/satteri.ts [hunk 3]
```diff
@@ -0,0 +1,241 @@
...
+import { Icons } from '../components-internals/Icons';
+import { throwInvalidAsideIconError } from './asides-error';
+import type { StarlightIcon } from '../types';
+
+// Re-exported so callers can narrow `markdown.processor` to a Sätteri processor through the same
+// lazy import that loads the optional `@astrojs/markdown-satteri` peer dependency.
+export { isSatteriProcessor };
+
+/** Sätteri mdast/hast plugins applied to Starlight content. */
+export function starlightSatteriPlugins(options: RemarkRehypePluginOptions): {
```
HiDeoo: Maybe we should move/rename some type/helpers to a `markdown-process.ts` file to easily identify what's common to both processors?

- `RemarkRehypePluginOptions` → `MarkdownProcessorPluginOptions`
- `getRemarkRehypePaths()` → `getMarkdownProcessorPaths()`
- etc.

packages/starlight/__tests__/remark-rehype/anchor-links.test.ts
HiDeoo, line 1: Not related to this specific test file but mostly all tests and the testing strategy.

I wonder if instead of duplicating tests, we shouldn't rename the `remark-rehype` test suite to `markdown-processor` and have some kind of utility wrapping our tests, creating 2 processors and running the exact same tests, e.g. using `describe.each` for the 2 processors. And if needed, a specific processor could have a `satteri.test.ts` file fo non-common tests.

The few reasons coming to mind:

- Not all tests are ported, only some of them and it's not really obvious why. I'm also a bit afraid, if in the future we end up going all in on Sätteri and removing Unified, we end up losing some tests and a lot of them are actually regression tests that I don't think we want to lose.
- Some tests that are not ported seems to produce errors at render time, e.g. nested asides:
  ```
  patch targets node 4 inside a removed subtree
  ```
- Some tests uses slightly different/simplified assertions making the test pass but it seems the output is wrong.
- It would also simplify maintenance for us if we don't have to duplicate all tests.

packages/starlight/__tests__/satteri/asides.test.ts
```diff
@@ -0,0 +1,102 @@
...
+test('localizes the default label for the file’s language', async () => {
+	const res = await renderMarkdown(`:::note\nbody\n:::`, {
+		fileURL: docFileURL('fr/index.md'),
+	});
+	expect(res.code).includes('aria-label="Note"');
+});
+
+test('unknown variants pass through unchanged', async () => {
+	const res = await renderMarkdown(`:::question\nbody\n:::`);
+	expect(res.code).not.includes('starlight-aside');
```
HiDeoo: Seems that `res.code` is empty so the assertion pass but by not testing the actual content it's missing the fact we don't get back the unknown variant as-is.

packages/starlight/__tests__/satteri/asides.test.ts
```diff
@@ -0,0 +1,102 @@
...
+		expect(res.code).includes('starlight-aside--note');
+		expect(res.code).not.includes(':::note');
+	});
+});
+
+test('skips files outside the docs collection', async () => {
+	const res = await renderMarkdown(`:::note\nbody\n:::`, {
+		fileURL: nonDocFileURL(),
+	});
+	expect(res.code).not.includes('starlight-aside');
```
HiDeoo: Same here, `res.code` seems to be empty.

packages/starlight/integrations/satteri.ts
```diff
@@ -0,0 +1,241 @@
...
+			// wrapper gives each document a fresh slugger.
+			...(options.starlightConfig.markdown.headingLinks
+				? [() => satteriHeadingIdsPlugin(), satteriAutolinkHeadingsPlugin(options, allowedPaths)]
+				: []),
+		],
+	};
+}
+
+/** Restore unhandled text/leaf directives to their source Markdown form. */
+export function satteriDirectivesRestoration(): MdastPluginDefinition {
```
HiDeoo, line 47: It seems that this always happen for unhandled directive whereas before, we would only restore [if `node.data` is `undefined`](https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/asides.ts#L251)?

packages/starlight/integrations/remark-rehype.ts
```diff
@@ -90,10 +90,14 @@ function shouldTransformFile(file: VFile, remarkRehypePaths: string[]) {
...
 	if (!file?.path) return false;
 
-	const normalizedPath = normalizePath(file.path);
+	return shouldTransformPath(file.path, remarkRehypePaths);
+}
 
-	// If the document is not part of the allowed remark/rehype paths, skip it.
-	return remarkRehypePaths.some((path) => normalizedPath.startsWith(path));
+/** Path-string variant of {@link shouldTransformFile} for callers without a VFile. */
+export function shouldTransformPath(path: string | undefined, remarkRehypePaths: string[]) {
```
HiDeoo: For Sätteri, wouldn't we need a different helper, as `ctx.filename` [seems to be `fileURL.pathname`](https://github.com/withastro/astro/blob/main/packages/markdown/satteri/src/satteri-processor.ts#L296), so this could be an issue with some characters, e.g. a space being encoded as `%20`?

I think we may also have the same issue with the path we pass down to `absolutePathToLang()`.

Altho, the Sätteri docs does not really say anything about the format of `filename` so I guess this could also be fixed in Astro or documented?

HiDeoo (dismissed):
Thanks a lot :raised_hands: This is definitely one impressive first PR :sweat_smile:

I didn't get the time to go over all the changes yet or play enough with the code, but I'm sharing some initial thoughts and questions.

On a more global level, I'm a bit confused on how to use the new processor with our docs, e.g. if I install `@astrojs/markdown-satteri` in our `docs/` directory, and set `markdown: { processor: satteri() },` I get the following error:

```
❯ nr dev
$ astro dev
17:33:37 [ERROR] [@astrojs/starlight] An unhandled error occurred while running the "astro:config:setup" hook
Vite module runner has been closed.
  Location:
    /Users/hideo/Temp/starlight/node_modules/.pnpm/vite@7.3.2_@types+node@24.12.4_jiti@2.6.1_lightningcss@1.32.0_yaml@2.8.3/node_modules/vite/dist/node/module-runner.js:1081:26
  Stack trace:
    at ModuleRunner.getModuleInformation (file:///Users/hideo/Temp/starlight/node_modules/.pnpm/vite@7.3.2_@types+node@24.12.4_jiti@2.6.1_lightningcss@1.32.0_yaml@2.8.3/node_modules/vite/dist/node/module-runner.js:1081:26)
    at request (file:///Users/hideo/Temp/starlight/node_modules/.pnpm/vite@7.3.2_@types+node@24.12.4_jiti@2.6.1_lightningcss@1.32.0_yaml@2.8.3/node_modules/vite/dist/node/module-runner.js:1103:94)
    at astro:config:setup (/Users/hideo/Temp/starlight/packages/starlight/index.ts:98:73)
    at async runHookInternal (file:///Users/hideo/Temp/starlight/node_modules/.pnpm/astro@6.4.2_@types+node@24.12.4_jiti@2.6.1_lightningcss@1.32.0_rollup@4.53.3_yaml@2.8.3/node_modules/astro/dist/integrations/hooks.js:52:5)
    at async createContainer (file:///Users/hideo/Temp/starlight/node_modules/.pnpm/astro@6.4.2_@types+node@24.12.4_jiti@2.6.1_lightningcss@1.32.0_rollup@4.53.3_yaml@2.8.3/node_modules/astro/dist/core/dev/container.js:21:14)
[ELIFECYCLE] Command failed with exit code 1.
```

One last point, if someone configures `markdown: { processor: unified() },` or not configure any processor at all, I wonder if the deprecation log could be a bit annoying:

```
[astro] `markdown.remarkPlugins`, `markdown.rehypePlugins`, and `markdown.remarkRehype` are deprecated. Pass them to `unified({...})` from `@astrojs/markdown-remark` directly instead.
```

packages/starlight/index.ts
Princesseuh: `smartPunctuation` defaults to `false` in Sätteri, but the Astro processor enables it by default

packages/starlight/integrations/remark-rehype.ts
Princesseuh: I don't think I really like that `shouldTransformPath` thing in general, I think perhaps a better option would be to give access to the filename inside the factory of a plugin? Because right now it causes unnecessary nodes to be materialized anyway

packages/starlight/index.ts
HiDeoo: Ah, I see, thanks. I guess we will probably want to tweak this [section](https://docs.astro.build/en/guides/markdown-content/#switching-to-the-s%C3%A4tteri-processor) to not redirect to the Sätteri docs for everything, e.g. default values:

> See the Sätteri documentation for the available plugins and features.

packages/starlight/integrations/remark-rehype.ts
HiDeoo: Yeah, I noticed that in the Sätteri implementation, really obvious in `satteriRtlCodeSupportPlugin()`, would be nice if it could be avoided but I don't know enough to know if there is some kind of limitation to that (altho I guess the issue I pointed out would still be the same).

Princesseuh:
Quick status update here, Starlight's tests found an actual issue in Sätteri that I'm fixing here: https://github.com/bruits/satteri/pull/68

packages/starlight/integrations/remark-rehype.ts
Princesseuh: Leaving it as-is for now because the solution I wanted to do doesn't actually work right now, will need to think further.

packages/starlight/integrations/satteri.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * Add `dir="ltr"` to the opening tag of a raw `<pre>` HTML string, unless it already declares a
+ * `dir`. Returns `null` when the value isn’t a `<pre>`, leaving unrelated raw HTML untouched.
+ */
+function ltrRawPre(value: string): string | null {
+	const openTag = value.match(rawPreOpenTag)?.[0];
+	if (!openTag || /\sdir\s*=/.test(openTag)) return null;
+	return value.replace(openTag, () => `<pre dir="ltr"${openTag.slice(4)}`);
+}
```
Princesseuh, line 220: This is something unfortunate I didn't realize, when I did shiki highlighting in Astro for Sätteri, I figured that using `toHTML` (or whatever it's called) is fine (and much faster), but I didn't realize it meant plugins can't interact with the shiki HAST in other plugins. Something to fix in Astro, or tolerate for now, I guess...

packages/starlight/package.json
```diff
@@ -48,19 +48,27 @@
...
+    "@astrojs/markdown-satteri": "^0.2.0",
     "astro": "^6.0.0"
   },
+  "peerDependenciesMeta": {
+    "@astrojs/markdown-satteri": {
+      "optional": true
+    }
+  },
   "devDependencies": {
+    "@astrojs/markdown-satteri": "https://pkg.pr.new/@astrojs/markdown-satteri@16955?t=1780458151",
```
Princesseuh: temporary until Astro's next patch is merged https://github.com/withastro/astro/pull/16950 will of course update before merging.

packages/starlight/integrations/markdown-plugins.ts
```diff
@@ -0,0 +1,88 @@
+import { isUnifiedProcessor } from '@astrojs/markdown-remark';
+import type { AstroConfig, AstroIntegration, AstroIntegrationLogger } from 'astro';
+import { remarkDirectivesRestoration } from './asides';
+import type { MarkdownProcessorPluginOptions } from './markdown-process';
+import { starlightRehypePlugins, starlightRemarkPlugins } from './remark-rehype';
+
+// `@astrojs/markdown-satteri` is an optional peer dependency, so its integration is loaded lazily.
+// The import is started here, at module evaluation, while Astro's config module runner is still
+// alive — a dynamic `import()` of this relative module from inside `astro:config:setup` fails with
+// "Vite module runner has been closed". When the optional dependency is absent, the import rejects
```
HiDeoo: > a dynamic `import()` of this relative module from inside `astro:config:setup` fails with "Vite module runner has been closed"

Maybe it's me, but I'm not sure to understand the comment. We say a dynamic `import()` will fail but we use a dynamic `import()`? Or maybe the comment explains the `.catch(() => null)`?

packages/starlight/integrations/markdown-plugins.ts
```diff
@@ -0,0 +1,88 @@
...
+ * 6.0–6.3) so the caller falls back to the legacy `markdown.{remark,rehype}Plugins` keys.
+ */
+export function applyStarlightMarkdownPlugins(
+	processor: MarkdownProcessor | undefined,
+	options: MarkdownProcessorPluginOptions,
+	satteri: SatteriIntegration,
+	logger: Pick<AstroIntegrationLogger, 'warn'>
+): boolean {
+	if (!processor) return false;
+	if (processor.name === 'satteri' && satteri?.isSatteriProcessor(processor)) {
```
HiDeoo: Q: Any reason why we check the processor name here but not for `unified()`?

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	test('applies `dir="ltr"` to Shiki-highlighted fenced code blocks', async () => {
+		const res = await ctx().render('```js\nconsole.log("test")\n```');
+		expect(res.code).includes('dir="ltr"');
+		expect(res.code).includes('astro-code');
+	});
+
+	test('applies `dir="ltr"` to fenced code blocks when Shiki is off', async () => {
+		const res = await ctx().render('```\nconsole.log("test")\n```', { processor: plain });
+		expect(res.code).includes('<pre dir="ltr">');
```
HiDeoo, line 31:
```suggestion
		expect(res.code).includes('<pre dir="ltr">');
		expect(res.code).includes('<code>');
```

We should probably assert that `<code>` inside `<pre>` are skipped like we did before by asserting the entire content.

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+	});
+
+	test('does not override an existing `dir` on inline code', async () => {
+		const res = await ctx().render(`<code dir="rtl">manual</code>`);
+		expect(res.code).includes('dir="rtl"');
+		expect(res.code).not.includes('dir="auto"');
```
HiDeoo, line 52:
```suggestion
		expect(res.code).not.includes('dir="auto"');
		expect(res.code).includes('<code>');
```

Same here, but this will fail with Sätteri.

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+	});
+
+	test('does not override an existing `dir` on fenced code blocks', async () => {
+		const res = await ctx().render('<pre dir="rtl"><code>manual</code></pre>', {
+			processor: plain,
+		});
+		expect(res.code).includes('dir="rtl"');
+		expect(res.code).not.includes('dir="ltr"');
```
HiDeoo, line 61:
```suggestion
		expect(res.code).not.includes('dir="ltr"');
		expect(res.code).includes('<code>');
```

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+		const res = await ctx().render('```\nconsole.log("test")\n```', { processor: plain });
+		expect(res.code).includes('<pre dir="ltr">');
+	});
+
+	test('does not override an existing `dir` on inline code', async () => {
+		const res = await ctx().render(`<code dir="rtl">manual</code>`);
+		expect(res.code).includes('dir="rtl"');
+		expect(res.code).not.includes('dir="auto"');
+	});
+
```
HiDeoo, line 54: I think we may be missing a test for the default case with Shiki-highlighted fenced code blocks:

```ts
	test('applies `dir="ltr"` to Shiki-highlighted fenced code blocks', async () => {
		const res = await ctx().render('<pre><code>manual</code></pre>', {
			processor: plain,
		});
		expect(res.code).includes('<pre dir="ltr">');
		expect(res.code).includes('<code>');
	});
```

Altho, this one seems to fail with Unified altho the code did not change. Maybe something different in the way we setup the test 🤔

packages/starlight/integrations/satteri.ts
```diff
@@ -0,0 +1,268 @@
...
+			{
+				filter: ['pre'],
+				visit(node, ctx) {
+					if (!shouldTransformPath(ctx.fileURL, allowedPaths)) return;
+					if (node.properties && 'dir' in node.properties) return;
+					ctx.setProperty(node, 'dir', 'ltr');
+				},
+			},
+			{
+				filter: ['code'],
```
HiDeoo: I think this would now match `<code>` inside `<pre>` (which may or may not be a direct child). I suggested a test now failing for this as we had one before.

This is not an issue with Remark as we [stop traversing](https://github.com/withastro/starlight/blob/main/packages/starlight/integrations/code-rtl-support.ts#L28) when encountering a `<pre>` or `<code>`.

packages/starlight/package.json
```diff
@@ -48,19 +48,27 @@
...
+    "astro": "^6.4.0",
     "linkedom": "^0.18.12",
+    "satteri": "^0.8.0",
     "vitest": "^4.1.5"
   },
   "dependencies": {
-    "@astrojs/markdown-remark": "^7.1.1",
-    "@astrojs/mdx": "^5.0.4",
+    "@astrojs/markdown-remark": "^7.2.0",
+    "@astrojs/mdx": "https://pkg.pr.new/@astrojs/mdx@16955?t=1780458151",
```
HiDeoo: The new version will have a peer dependency of [`^6.4.0` for Astro](https://github.com/withastro/astro/blob/main/packages/integrations/mdx/package.json#L54), so I guess we kinda need to bump our own peer dep for Astro to this version and advertise this in the changeset as a breaking change requiring an Astro update?

For the changeset part, we usually do that by adding this at the end of the changeset:

````md
⚠️ **BREAKING CHANGE:** The minimum supported version of Astro is now v6.4.0.

Please update Starlight and Astro together:

```sh
npx @astrojs/upgrade
```
````

File: pnpm-workspace.yaml [hunk 18]
```diff
@@ -24,4 +24,13 @@ peerDependencyRules:
     - 'search-insights'
 
 # Prevent installing packages published in the last 3 days
-minimumReleaseAge: 4320
+# TEMP: disabled to install the Sätteri 0.7.0 preview packages (astro#16955); restore before commit.
+# minimumReleaseAge: 4320
```
HiDeoo: Just flagging to not forget about this.

packages/starlight/integrations/markdown-plugins.ts
```diff
@@ -0,0 +1,88 @@
...
+export function registerDirectivesRestoration(
+	processor: MarkdownProcessor | undefined,
+	satteri: SatteriIntegration
+): boolean {
+	if (processor?.name === 'satteri' && satteri?.isSatteriProcessor(processor)) {
+		processor.options.mdastPlugins.push(satteri.satteriDirectivesRestoration());
+		return true;
+	}
+	if (processor && isUnifiedProcessor(processor)) {
+		processor.options.remarkPlugins.push(remarkDirectivesRestoration);
```
HiDeoo, line 50: I guess this wouldn't work and break in the case of a user using the `unified()` processor, and having a Starlight plugin adding an Astro integration using `addIntegration()` and this plugin using `updateConfig({ markdown: { remarkPlugins: [customRemarkPluginDealingWithDirective })`?

The restoration plugin would be pushed and the `customRemarkPluginDealingWithDirective` would be pushed after that, meaning it would never see directives?

packages/starlight/integrations/satteri.ts
```diff
@@ -0,0 +1,268 @@
...
+											type: 'element',
+											tagName: 'svg',
+											properties: { width: '16', height: '16', viewBox: '0 0 24 24' },
+											children: [
+												{
+													type: 'element',
+													tagName: 'path',
+													properties: {
+														fill: 'currentcolor',
+														d: 'm12.11 15.39-3.88 3.88a2.52 2.52 0 0 1-3.5 0 2.47 2.47 0 0 1 0-3.5l3.88-3.88a1 1 0 0 0-1.42-1.42l-3.88 3.89a4.48 4.48 0 0 0 6.33 6.33l3.89-3.88a1 1 0 1 0-1.42-1.42Zm8.58-12.08a4.49 4.49 0 0 0-6.33 0l-3.89 3.88a1 1 0 0 0 1.42 1.42l3.88-3.88a2.52 2.52 0 0 1 3.5 0 2.47 2.47 0 0 1 0 3.5l-3.88 3.88a1 1 0 1 0 1.42 1.42l3.88-3.89a4.49 4.49 0 0 0 0-6.33ZM8.83 15.17a1 1 0 0 0 1.1.22 1 1 0 0 0 .32-.22l4.92-4.92a1 1 0 0 0-1.42-1.42l-4.92 4.92a1 1 0 0 0 0 1.42Z',
```
HiDeoo: Maybe it could be worth extracting this path too like aside icons? We now have 3 places where we rely on it:

- `packages/starlight/integrations/heading-links.ts`
- `packages/starlight/integrations/satteri.ts`
- `packages/starlight/components/AnchorHeading.astro`

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -0,0 +1,60 @@
...
+	test('applies `dir="auto"` to inline code', async () => {
+		const res = await ctx().render('Some text with `inline code`.');
+		expect(res.code).includes('<code dir="auto">inline code</code>');
+	});
+
+	test('applies `dir="ltr"` to Shiki-highlighted fenced code blocks', async () => {
+		const res = await ctx().render('```js\nconsole.log("test")\n```');
+		expect(res.code).includes('dir="ltr"');
+		expect(res.code).includes('astro-code');
```
HiDeoo, line 24: Q: Does "to Shiki-highlighted" mean with Expressive Code disabled and relying on Astro's Shiki highligher? Or Expressive Code (that also uses Shiki)?

I'm asking because the `astro-code` expected is added by Astro's Shiki highlighter which is not Starlight default setup. If yes, maybe we should make that more obvious in the test names?

HiDeoo (dismissed):
Thanks for the update :raised_hands: The new changes looks really great and I really like the non-duplicated tests :clap: Adding some new comments and suggestions for now and I'll continue to play around with the PR tomorrow.

Also, I can now set `markdown: { processor: satteri() },` in the `docs/` and I no longer get the error I mentioned in my previous review (I installed `"@astrojs/markdown-satteri": "https://pkg.pr.new/@astrojs/markdown-satteri@16955?t=1780458151"`). Altho, it looks code blocks are not processed by Expressive Code, or using Astro's Shiki highlighter?

Once this is fixed, I'll run our visual diff checker on all our docs pages (and same on Astro Docs) for sanity (doing it now would highlight all code blocks on every pages).

packages/starlight/integrations/markdown-plugins.ts
Princesseuh: I honestly have no idea why it works this way. This seems related to a regression in vite that started happening a long time ago, like in Astro 5. The comment is not amazing though, I'll rephrase

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
Princesseuh: Yeah, we need https://github.com/expressive-code/expressive-code/pull/445 for expressive-code really. The Sätteri plugin I made works, but there's a bunch of Astro specific things in `astro-expressive-code` that are needed.

Until then I thought it to be useful to test nonetheless the behavior with expressive code disabled, since that's an option possible

HiDeoo:
Just pushed 2 new changes to the PR:

**1. An error is now thrown when Sätteri is used with Starlight saying that the Sätteri Markdown processor is not yet compatible with Starlight.**

For the reasoning behind this, we discussed with @delucis the current state and came up with the following conclusions:

- Expressive Code is a core feature of Starlight, it brings some features that users expect (e.g. copy button) and some that are key Starlight features (e.g. ensuring proper color contrast of code blocks). It feels a bit weird to say "We support Sätteri but to get the full Starlight experience, you cannot have code blocks".
- Right now, if we were to support Sätteri, we would have a very strong signal (an error being thrown) to signal to users that they must disable Expressive Code to use Sätteri. Once Expressive Code is updated and compatible with Sätteri, we wouldn't have an equally strong signal to say "You can now use Expressive Code again". While it can obviously be communicated in the changeset, it would be missed by many users (and we can't obviously just check if `expressiveCode` is disabled and then log a warning as some users might have it disabled for other reasons).
- All the work in this PR and reviews is not wasted, and as soon as Expressive Code is compatible with Sätteri, we can just update it, remove the error, and have proper Sätteri support in Starlight released very quickly.
- It would probably not give a good first impression to users to have Sätteri "kinda work" with Starlight but then have key features not working. Having a future release where we can just say "Sätteri is now supported" and have it work well with all features seems like a better experience and would probably lead to a better reception imo.

**2. The second change I did is remove all the logic related to Astro 6.0–6.3 support and the nwo legacy `markdown.{remark,rehype}Plugins` keys.**

- If the PR initially aimed to support such versions, we discovered this PR requires a peer dependency on Astro 6.4, which means we should be able to drop that logic and code paths now.
- May need to confirm this one with @Princesseuh to make sure I didn't forget about any edge cases.

Let me know what you all think about these changes.

Princesseuh:
Sure, no worries. Will try to work on finishing the expressive code PR as soon as possible. Maybe by then I'll also have fixed a few of the rough edges this PR works around right now.

As I'm currently on sick leave (one must wonder why I'm on GitHub dot com) it might take time, but hopefully there's not much work left on the currently open PR there.

HiDeoo:
Why are you on GitHub dot com? 😠 (well, I tagged you, but still…)

> Will try to work on finishing the expressive code PR as soon as possible.

I was thinking about taking a look at it Monday. Not sure if I can do much, but at least on paper, it looks like it's in my competence. At the very least, I can get familiar with the existing code to later be able to review it if I fail to do more.

Princesseuh:
The code is very similar to this PR because the Sätteri part is done already, so it's mostly Astro stuff.

hippotastic:
Sätteri is now supported by `astro-expressive-code` as of [version `0.43.0`](https://expressive-code.com/releases/#0430).

HiDeoo:
Thanks to the bug fixes in the latest patch version of Sätteri, we can now properly build the Starlight Docs and also [the Astro Docs](https://github.com/withastro/docs/pull/14022) using Sätteri.

Considering this, I re-enabled support for Sätteri, removed the previous Expressive Code disclaimer, and a preview release is now available [here](https://github.com/withastro/starlight/pull/3923#issuecomment-4658230501).

I still want to take a final look at all the recent changes, ~~make a tiny patch to Expressive Code~~ (done in https://github.com/expressive-code/expressive-code/pull/449), but I think we are very close to have a PR ready to be merged. Feel free to test the preview release and/or review the PR in the meantime.

Edit: forgot to mention, I ran our visual diff checker on all pages and no visual differences were found.

.changeset/satteri-support.md
```diff
@@ -0,0 +1,30 @@
+---
+'@astrojs/starlight': minor
+---
+
+Adds support for Astro 6.4 and the new Sätteri Markdown processor.
+
+It is now possible to opt-into using [Astro's 6.4 Sätteri Markdown processor](https://astro.build/blog/astro-640/#faster-markdown-builds-with-s%C3%A4tteri) by installing the `@astrojs/markdown-satteri` package and configuring it in your `astro.config.mjs` file:
```
delucis: Small typo:
```suggestion
It is now possible to opt into using [Astro's 6.4 Sätteri Markdown processor](https://astro.build/blog/astro-640/#faster-markdown-builds-with-s%C3%A4tteri) by installing the `@astrojs/markdown-satteri` package and configuring it in your `astro.config.mjs` file:
```

docs/src/content/docs/guides/authoring-content.mdx
```diff
@@ -561,7 +561,7 @@ Starlight supports all other Markdown authoring syntax, such as lists and tables
 
 ## Advanced Markdown and MDX configuration
 
-Starlight uses Astro’s Markdown and MDX renderer built on remark and rehype. You can add support for custom syntax and behavior by adding `remarkPlugins` or `rehypePlugins` in your Astro config file. See [“Markdown Plugins”](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) in the Astro docs to learn more.
+Starlight renders Markdown and MDX using Astro’s configurable [`markdown.processor`](https://docs.astro.build/en/reference/configuration-reference/#markdownprocessor), which defaults to the remark and rehype-based `unified()` processor. You can add support for custom syntax and behavior by passing plugins and options to your configured processor. See [“Markdown Plugins”](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) in the Astro docs to learn more.
```
delucis: Maybe we don’t need to mention the Astro default here and leave that to Astro’s docs? (Especially as the default will soon change, so this avoids us updating anything then.)
```suggestion
Starlight renders Markdown and MDX using Astro’s configurable [`markdown.processor`](https://docs.astro.build/en/reference/configuration-reference/#markdownprocessor). You can add support for custom syntax and behavior by passing plugins and options to your configured processor. See [“Markdown Plugins”](https://docs.astro.build/en/guides/markdown-content/#markdown-plugins) in the Astro docs to learn more.
```
One thing that does come to mind: we support the `unified` and `satteri` processors, but I guess we may not support a custom third-party processor? Maybe not an issue today, but if third-party processors become a thing, we may need to specify our support one day.

docs/package.json
```diff
@@ -21,7 +21,7 @@
     "@astrojs/starlight": "workspace:*",
     "@lunariajs/core": "^0.1.1",
     "@types/culori": "^4.0.1",
-    "astro": "^6.3.1",
+    "astro": "^6.4.0",
```
delucis: The latest release is 6.4.4
Should we update these all (and the peer dep?) to latest before releasing?

packages/starlight/integrations/anchor-icon.ts
```diff
@@ -0,0 +1,5 @@
+// SVG `d` for the anchor-link icon rendered next to headings. Shared by the unified
+// (`heading-links.ts`) and Sätteri (`satteri.ts`) autolink plugins and the `<AnchorHeading>`
+// component so the three stay in sync.
+export const anchorLinkIconPath =
+	'm12.11 15.39-3.88 3.88a2.52 2.52 0 0 1-3.5 0 2.47 2.47 0 0 1 0-3.5l3.88-3.88a1 1 0 0 0-1.42-1.42l-3.88 3.89a4.48 4.48 0 0 0 6.33 6.33l3.89-3.88a1 1 0 1 0-1.42-1.42Zm8.58-12.08a4.49 4.49 0 0 0-6.33 0l-3.89 3.88a1 1 0 0 0 1.42 1.42l3.88-3.88a2.52 2.52 0 0 1 3.5 0 2.47 2.47 0 0 1 0 3.5l-3.88 3.88a1 1 0 1 0 1.42 1.42l3.88-3.89a4.49 4.49 0 0 0 0-6.33ZM8.83 15.17a1 1 0 0 0 1.1.22 1 1 0 0 0 .32-.22l4.92-4.92a1 1 0 0 0-1.42-1.42l-4.92 4.92a1 1 0 0 0 0 1.42Z';
```
delucis, line 5: **[Can be tackled in a follow-up.]**
Should we move this to our main Icons.ts exposing it publicly and share it from there?

Those use `<path>` instead of the raw `d` attribute, but maybe we can make that work?

(Not new in this PR to be clear — we already had a similar thing for the asides.)

packages/starlight/integrations/aside-icons.ts
delucis, line 1: **[Can be tackled in a follow-up.]**
Similar thought here: the `<path>`s for these are already in `Icons.ts`

packages/starlight/integrations/satteri.ts
delucis: Just to double check, @Princesseuh marked this as resolved, but I don’t see any comments/code changes, so wanted to confirm what happened here?

Also, I vaguely recall some discussion about Sätteri handling this itself? Or am I misremembering?

pnpm-workspace.yaml
```diff
@@ -25,3 +25,15 @@ peerDependencyRules:
...
+  - '@astrojs/internal-helpers'
+  - '@astrojs/markdown-satteri'
+  - '@astrojs/mdx'
+  - '@astrojs/markdown-remark'
+  - satteri
+  - '@bruits/*'
+  - 'astro-expressive-code'
+  - 'rehype-expressive-code'
+  - 'expressive-code'
+  - '@expressive-code/*'
```
delucis, line 35: May be worth double checking if we can remove any of these before merging.

packages/starlight/integrations/markdown-plugins.ts
```diff
@@ -0,0 +1,68 @@
+import { isUnifiedProcessor } from '@astrojs/markdown-remark';
+import type { AstroConfig, AstroIntegration, AstroIntegrationLogger } from 'astro';
+import { remarkDirectivesRestoration } from './asides';
+import type { MarkdownProcessorPluginOptions } from './markdown-process';
+import { starlightRehypePlugins, starlightRemarkPlugins } from './remark-rehype';
+
+// For some reason, trying to `await import("./satteri")` in the integration module causes a Vite
+// error about the module runner being closed. This shape works, not sure why!
+export const satteriIntegration = import('./satteri').catch(() => null);
```
delucis, line 9: Any more information about why this error was happening we could add here?

Also, do we really need to export this, import it in `index.ts`, only to then pass it into the `applyStarlightMarkdownPlugins` export from this file?

If we loaded it in the apply function, we could gate loading of the module on if it’s actually needed? i.e.
```js
const satteriIntegration = processor.name === 'satteri' ? await import('./satteri') : null;
```
That might also give us a chance to avoid catching the error which could be silencing something meaningful?

delucis (commented):
Thanks for the awesome work @Princesseuh and @HiDeoo! Left some thoughts, although I will admit I did not read through all the tests 😅

packages/starlight/integrations/anchor-icon.ts
HiDeoo: Good call, added myself a todo to investigate and tackle this in a follow-up PR 👍

packages/starlight/integrations/satteri.ts
HiDeoo: Checks were added to visitors, you can see `if (node.data !== undefined) return;` below.

packages/starlight/integrations/markdown-plugins.ts
HiDeoo: Done in [`c9ff3e6` (this PR)](https://github.com/withastro/starlight/pull/3923/commits/c9ff3e689a8a881609696e0880359645d2e0e80b), seems to work fine even in a project not having `@astrojs/markdown-satteri`, but I'll let @Princesseuh  have the final say as this was part of the initial implementation and not sure if I may be missing any edge cases.

packages/starlight/integrations/markdown-plugins.ts
```diff
@@ -0,0 +1,70 @@
...
+ * plugin set for the configured engine.
+ */
+export async function applyStarlightMarkdownPlugins(
+	processor: MarkdownProcessor,
+	options: MarkdownProcessorPluginOptions,
+	logger: Pick<AstroIntegrationLogger, 'warn'>
+) {
+	const satteri = await getSatteriIntegration(processor);
+
+	if (satteri?.isSatteriProcessor(processor)) {
```
delucis: We already check we have a Sätteri processor in the `getSatteriIntegration()` function, so the repeated check here is superfluous (`isSatteriProcessor()` just checks the `name` field like we do):

```suggestion
	const satteri = await getSatteriIntegration(processor);

	if (satteri) {
```

delucis: Alternatively — and maybe preferably? — we could do this? And refactor the get function to not bother with the check?
```suggestion
	if (processor.name === 'satteri') {
		const satteri = await getSatteriIntegration(processor);
```
(Also not sure if it _has_ to be a top-level variable? Normally this method should only be called once, right? So there’s not really an advantage to caching the imported module like that? It just adds a bit of complexity.)

HiDeoo: `isSatteriProcessor()` (and also `isUnifiedProcessor()`) are type guards, e.g. without them, TypeScript would complain, for example when trying to assign `processor.options.features.directive` that is only possible with Sätteri.

No matter what, I ended reverting my commit changing the logic as while testing a bit more, I discovered that this was also crashing with the same error this PR initially hit ("Vite module runner has been closed" and nothing else) but surprisingly when using Sätteri with `@astrojs/markdown-satteri` installed…

So far, only Erika's syntax seems to work without any crash in all cases. I think a Vite bug was mentioned but I haven't had the time to investigate it further yet.

delucis: I’ve seen the “Vite module runner has been closed” error around a few places now. I think from the initial Vite instance that is spun up for config parsing closing and some things were relying on it? Would be good to chase down as the current version feels fragile. But don’t want to block this PR on that obviously.

This Astro issue with the error also had to do with dynamic imports: https://github.com/withastro/astro/issues/12689

I suspect other user code with dynamic imports could hit the same issue?

delucis (approved):
Thanks again everyone for the great work here! 👏 

There’s definitely some details it would be nice to clean up eventually, but I think we should get this published to fix the tables bug and let people start playing with Sätteri.

PR #3926, opened 2026-05-28
Manage Node.js versions with Renovate

HiDeoo:
#### Description

Follow-up to [this comment](https://github.com/withastro/starlight/pull/3919#discussion_r3298847606), this PR configures Renovate to manage Node.js versions in our CI workflows so we can pin them, have some consistency across all Node.js versions used in our workflows, and respect our `minimumReleaseAge` configuration.

- All workflows use the latest Node.js LTS version (respect the `minimumReleaseAge` configuration)
- Except `.github/workflows/ci.yml` which uses the latest patch of the minimum Node.js major supported by Astro (matches the current behavior of the CI workflow except versions are now pinned and respect the `minimumReleaseAge` configuration)

I also configure Node.js Renovate PRs to be roughly every two weeks only, so we don't get too many PRs but open to feedback on this.

I validated the config using [`npx --yes --package renovate -- renovate-config-validator`](https://docs.renovatebot.com/config-validation/#default-behavior).


CONTRIBUTING.md
```diff
@@ -42,7 +42,7 @@ You can [develop locally](#developing-locally) or use an online coding developme
 
 #### Developing locally
 
-**Prerequisites:** Developing Starlight requires [Node.js](https://nodejs.org/en) (v16 or higher) and [pnpm](https://pnpm.io/) (v8.2 or higher). Make sure you have these installed before following these steps.
+**Prerequisites:** Developing Starlight requires [Node.js](https://nodejs.org/en) (v22.13.0 or higher) and [pnpm](https://pnpm.io/) (using the version specified by the `packageManager` field in `package.json`). Make sure you have these installed before following these steps.
```
HiDeoo, line 45: `v22.13.0` is different from the minimum Node.js version supported by Astro to match the requirements of the `pnpm` version we use which is required in this case of "Developing locally".

delucis (approved):
Looks great! And +1 to limiting the updates to not **too** frequently.

PR #3932, opened 2026-05-30
fix!: use logical class names to improve RTL support

babakfp:
#### Description

When using this project to build a right-to-left site, seeing classes like `right-sidebar` is a bit confusing because the sidebar in a right-to-left site is placed on the left side.

.right-* -> .end-*

Example classes that are renamed:

.right-group -> .end-group
.right-sidebar -> .end-sidebar
.right-sidebar-panel -> .end-sidebar-panel
.right-sidebar-container -> .end-sidebar-container

This is a breaking change because people use these classes to customize starlight and releasing under a minor/patch could break the UI of their website.


delucis:
Thanks for the PR @babakfp. This PR changes slot names (not class names) and would be a breaking and disruptive change to users. While breaking changes can obviously be justified and we make them in minor releases, we normally only make them to enable some new behaviour. This PR only renames the API but does not change RTL support — the styling of Starlight is already fully RTL compatible and the slot names do not impact that.

I agree that perhaps start/end could have been better than left/right for the API naming in this case, but I don’t think renaming at this point is helpful enough to justify the disruption given it doesn’t introduce new functionality.

(Please also note for the future that we require posts to our discussions forum for changes like this before making a PR, so there’s a chance to gather feedback before moving to the implementation stage.)

PR #3936, opened 2026-05-31
fix: invalid option error when modifying enums in docsSchema's extend

### LINKED ISSUE #3935: **hero.actions.1.variant**: Invalid option: expected one of "primary"|"secondary"|"minimal"

0.39.2

6.3.1

npm

Windows

Firefox

### Describe the Bug

In the starlight docs, it mentions `docsSchema()` can modify fields, however it is unable to modify `z.enum` fields The likely cause is from how `ZodIntersection` are treated.

### Why this happens

The Zod documentation explicitly notes that intersections (a logical "AND") produce a `ZodIntersection` instance rather than a `ZodObject`.

From the [Zod v4 docs](https://zod.dev/api#intersections):
> When merging object schemas, prefer `A.extend(B)` over intersections. Using `.extend()` will give you a new object schema, whereas `z.intersection(A, B)` returns a `ZodIntersection` instance which lacks common object methods like pick and omit.

This limitation also affects any tooling that expects to operate on a `ZodObject`, including schema‑amending utilities like `docsSchema()`.

### Current workarounds
1. Introduce an additional field to represent the `enum` option outside the intersection, or
2. To Rebuild the full schema.

https://stackblitz.com/edit/github-kzc2ffxo?file=src%2Fcontent.config.ts

### Participation

miichom:
#### Description
In the starlight docs, it mentions `docsSchema()` can modify fields, however it is unable to modify `z.enum` fields The likely cause is from how `ZodIntersection` are treated.

### Why this happens

The Zod documentation explicitly notes that intersections (a logical "AND") produce a `ZodIntersection` instance rather than a `ZodObject`.

From the [Zod v4 docs](https://zod.dev/api#intersections):
> When merging object schemas, prefer `A.extend(B)` over intersections. Using `.extend()` will give you a new object schema, whereas `z.intersection(A, B)` returns a `ZodIntersection` instance which lacks common object methods like pick and omit.

This limitation also affects any tooling that expects to operate on a `ZodObject`, including schema‑amending utilities like `docsSchema()`.

- Closes #3935 
This PR replace the `and` intersection in `docsSchema` with `extend` to support the modifying of `ZodEnum`.

Nothing was changed.


delucis (changes requested):
Thanks for the PR @miichom! And apologies for the slow response — I knew this would be a tricky one and needed time to play around with it to be sure the fix would work.

I’ve pushed up some tests for schema extension — previously we only touched that logic very minimally in an unrelated test. I included one for your case of extending a `z.enum()` property, which now passes on this branch (running that test against `main` fails, as you reported in #3935). However, the second part of this is **type** testing: this helps us ensure that the types inferred from an extended docs schema are correct.

Your changes seem to introduce an issue with the types when extending an object, e.g.
```js
docsSchema({
	extend: z.object({ sidebar: z.object({ custom: z.number() }) }),
})
```
The default properties of the `sidebar` field are missing, with only the custom extension present. AFAICT this issue only impacts the types — you can see that in the failing test in CI. At runtime, all properties are parsed by Zod as expected.

Would you be able to take a look?

HiDeoo:
> Your changes seem to introduce an issue with the types when extending an object, e.g.

Just to make sure, is only the type check an issue? Is this snapshot correct [here](https://github.com/withastro/starlight/pull/3936/changes#diff-a9f29476ff42ae9cd69bc75e908f85c410b52561f9e4d8e90cb24beefda30f6bR295) and the result with `sidebar` default properties being dropped is expected?

delucis:
> Is this snapshot correct [here](https://github.com/withastro/starlight/pull/3936/changes#diff-a9f29476ff42ae9cd69bc75e908f85c410b52561f9e4d8e90cb24beefda30f6bR295) and the result with `sidebar` default properties being dropped is expected?

Oh damn, you’re 100% correct. I thought I remembered checking this and seeing that everything expected was there, but nope, they’re dropped. So, yes, the issue is present for both runtime values AND types.

I updated that snapshot to reflect the runtime breakage too.

packages/starlight/__tests__/basics/schema.test.ts
```diff
@@ -171,3 +173,132 @@ describe('HeadConfigSchema', () => {
...
+			{
+			  "draft": false,
+			  "editUrl": true,
+			  "head": [],
+			  "hero": {
+			    "actions": [
+			      {
+			        "variant": "custom",
+			      },
+			    ],
```
delucis, line 250: Oh and the same issue is present here which is the case reported in #3935:
```suggestion
			    "actions": [
			      {
			        "link": "",
			        "text": "",
			        "variant": "custom",
			      },
			    ],
```

packages/starlight/__tests__/basics/schema.test-d.ts
```diff
@@ -66,7 +66,6 @@ describe('docs schema', () => {
 			image: () => ({}) as any,
 		});
 		const parsed = schema.parse({});
-		expectTypeOf(parsed).toExtend<PartialDocsSchemaOutput>();
 		expectTypeOf(parsed.sidebar.custom).toBeNumber();
 	});
 });
```
delucis, line 74: @miichom This test was correct btw — we shouldn’t remove it. This test should pass as it was. The fact it was failing was a sign of an issue with the `docsSchema()` types.

HiDeoo:
Played around with a potential fix in https://github.com/withastro/starlight/pull/4040.

I did that in a different PR to not mess around with the current PR where I cherry-picked all the commits from this PR in a single commit and added what I worked on top of it (also added 1 test and tweaked an existing one).

This is a draft as I'm not yet sure if this is the right approach or there are better ones, and I did not yet review it properly as I was so head down in the code that I did not yet take a step back. It's also overly commented as I was sometime getting lost when implementing it and I did not clean anything up yet.

If we ever decide to move forward with this, we should cherry pick only the relevant commits from https://github.com/withastro/starlight/pull/4040 in this PR so author credit is preserved and the history is cleaner.

miichom:
> It's also overly commented as I was sometime getting lost when implementing it and I did not clean anything up yet.

I personally think for now keeping it commented is a good idea, I will have a proper look when I get home.

miichom:
Okay, it took sometime but I preferred @HiDeoo schema merging over my one - I did remove the comments, but I am happy to re-add them if it helps others in the future. I added another test as the `icon` must start with `<svg` in the schema and would error without these PR changes.

delucis (commented):
Thanks for your patience on this one @miichom!

I think we’re happy to accept this change with HiDeoo’s implementation. I wonder if it would be good to move most of this new code into a file under `packages/starlight/utils/` to help keep the main `schema.ts` file focused on its primary purpose of declaring the schema. What do you think?

miichom:
I think that is a great idea!

.changeset/all-badgers-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Replace "and" in "docsSchema" with "extend" to support the modifying of ZodEnum.
```
delucis: We aim for our changelogs to focus on the user issue rather than the internal implementation, so I think something like this may be more useful for users:

```suggestion
Fixes support for modifying Zod enums when passing an [`extend` option](https://starlight.astro.build/reference/frontmatter/#extend) to Starlight’s `docsSchema()`
```

More on how we document changes in case you’re curious: https://contribute.docs.astro.build/docs-for-code-changes/changesets/

packages/starlight/utils/zodDeepMerge.ts
delucis, line 1: Maybe a file name like `zodDeepMerge.ts` might be more explicit?

delucis (commented):
Thanks @miichom — looking good! Just left a couple of last small notes.

delucis (dismissed):
Thanks @miichom! I think this looks good to me. Will give @HiDeoo a chance to leave a final review in case he has time today, but otherwise we’ll get this released soon 🙌

HiDeoo (commented):
Thanks for pushing this over the finish line and cleaning up the code. This looks perfect to me and ready to be shipped :rocket:

One tiny nit: now that this new utility and associated type live in their dedicated file, maybe we should rename the exported function and type, e.g. `deepMergeSchemas()` and `DeepMergedSchema` so the intent when reading other files is clearer?

miichom:
> One tiny nit: now that this new utility and associated type live in their dedicated file, maybe we should rename the exported function and type, e.g. `deepMergeSchemas()` and `DeepMergedSchema` so the intent when reading other files is clearer?

I am for this! It makes a little more sense than `mergeWithDefaultSchema`.

delucis (approved):
Thanks again @miichom and @HiDeoo for the collaboration on this one 🙌

PR #3948, opened 2026-06-10
Improve Sätteri import comment

HiDeoo:
#### Description

Follow-up to https://github.com/withastro/starlight/pull/3923#discussion_r3380597541, this PR improves the comment describing how we import the Sätteri integration to workaround an Astro limitation.

The comment should be clear enough to explain the issue and the workaround. Regarding a potential fix in Astro for the underlying issue, I would definitelly need to think more about it. So far, the only idea I had is to have the Vite module runner used for loading the config to stay open instead of closing immediately, and return some kind of disposable. The disposable would be used to properly close it only when needed, e.g. after a build or when the dev server is closed, when the dev server is restarted, after a sync, etc. The `getViteConfig()` case would also need to be handled. On top of the extra cost of keeping the module runner open, it's also a lot of orchestration to pass around such disposable with the current architecture.

A workaround that I only briefly tested for fun is to by-pass Vite and rely on a native dynamic import to load `@astrojs/markdown-satteri`, e.g. using `() => new Function('return import("@astrojs/markdown-satteri")')` but that's definitely not something we want to do (even tho it seems to work) :sweat_smile:

### DIFF
--- a/packages/starlight/integrations/markdown-plugins.ts
+++ b/packages/starlight/integrations/markdown-plugins.ts
@@ -5,9 +5,10 @@ import type { MarkdownProcessorPluginOptions } from './markdown-process';
 import { starlightRehypePlugins, starlightRemarkPlugins } from './remark-rehype';
 
 // This file is imported by the Starlight integration that is used in Astro configuration files.
-// Astro loads its configuration using a temporary Vite module runner and closes it before running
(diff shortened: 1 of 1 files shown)


delucis (approved):
Appreciate the investigation — would be nice to fix properly in Astro as I guess this could bite users and the error message is fairly cryptic. But more than happy that we understand this and document it properly — thank you!

PR #3951, opened 2026-06-11
Astro 7 support

HiDeoo:
#### Description

This PR adds support for Astro 7.

#### Known issues

| Symbol | Status                          |
| ------ | ------------------------------- |
| 🟢     | Fixed                           |
| 🟡     | Fix pending review or release              |
| 🔴     | No fix yet, under investigation |

- 🟢 Many warnings (1 in dev, 6 in build) due to the use of `resolve.alias[].customResolver` [now deprecated](https://vite.dev/guide/migration#other-related-deprecations) and probably related to the Astro `astro:tsconfig-alias` vite plugin [still using it](https://github.com/withastro/astro/blob/next/packages/astro/src/vite-plugin-config-alias/index.ts#L103-L111).

  ```sh
  [WARN] [vite] `resolve.alias` contains an alias with `customResolver` option. This is deprecated and will be removed in Vite 9. Please use a custom plugin with a resolveId hook and `enforce: 'pre'` instead.
  ```
  - _Fixed by https://github.com/withastro/astro/pull/17090_
- 🟢 Warning at build time due to some plugins taking significant time during the build process which is due to the [`pluginTimings` built-in Rolldown check](https://rolldown.rs/reference/InputOptions.checks#plugintimings) enabled by default. Feels weird in a release about speed to get warnings about some things being slow. Maybe we should disable this check by default?

  ```sh
  [WARN] [vite] [PLUGIN_TIMINGS] Your build spent significant time in plugins. Here is a breakdown:
  - @mdx-js/rolldown (41%)
  - astro:head-metadata-build (36%)
  - vite-plugin-astro-expressive-code (4%)
  - astro:server-islands (3%)
  - vite:css-post (2%)
  ```
  - _Fixed by https://github.com/withastro/astro/pull/17090_
- 🟢 Type-checking issue in `packages/starlight/integrations/remark-asides.ts`
  - _Fixed by https://github.com/bruits/satteri/pull/97_
- 🟢 Some duplicated heading IDs in the Sätteri databag
  - _Fixed by https://github.com/withastro/astro/pull/17165_

#### Remaining tasks

- [x] Investigate `size-limit` CSS size change (probably Vite 8/Rolldown related but need to see what and confirm the hypothesis)
  - Experimenting with [lazy barrel pptimization](https://rolldown.rs/in-depth/lazy-barrel-optimization#configuration)
- [x] Bump Astro dependencies to latest stable versions once released.
- [x] Bump some peer dependencies versions once ready to release.
- [x] Search and address any `TODO(HiDeoo)` comments.
- [x] Add changesets.
  - [x] Make sure changesets use the correct new Starlight version number (e.g. `0.41.0`).
  - [x] Make sure to have updated links to v7 migration guide.
- [x] Update the banner link in `docs/src/content/docs/index.mdx` to point to the correct version tag.
- [ ] Re-enable link validation 
  - Will probably happen post-release once the plugin is updated
- [x] Merge with the following Lunaria directive:
  ```
  @lunaria-ignore:docs/src/content/docs/pt-pt/index.mdx;docs/src/content/docs/uk/index.mdx;docs/src/content/docs/zh-cn/index.mdx
  ```


.changeset/fancy-swans-tickle.md
```diff
@@ -0,0 +1,17 @@
...
+
+⚠️ **BREAKING CHANGE:** Astro v6 is no longer supported. Make sure you [update Astro](https://docs.astro.build/en/guides/upgrade-to/v7/) and any other official integrations at the same time as updating Starlight:
+
+```sh
+npx @astrojs/upgrade
+```
+
+_Community Starlight plugins and Astro integrations may also need to be manually updated to work with Astro v7. If you encounter any issues, please reach out to the plugin or integration author to see if it is a known issue or if an updated version is being worked on._
+
+⚠️ **BREAKING CHANGE:** This release drops official support for Chromium-based browsers prior to version 111 (released 07 March 2023) and Safari-based browsers prior to version 16.4 (released 27 March 2023). You can find a list of currently supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+111%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+16.4%2C+iOS+%3E%3D+16.4%2C+not+op_mini+all).
```
HiDeoo, line 17: Aligns with Vite 8 [default browser target changes](https://vite.dev/guide/migration#default-browser-target-change) and uses the [`baseline-widely-available` target](https://web-platform-dx.github.io/baseline/).

In our case, the most impactful change seems to be the new [range media query syntax](https://web.dev/articles/media-query-range-syntax) (e.g. `@media (width >= 400px)`) being used.

If we ever want to avoid this, we would need to update the [`build.target` Vite configuration option](https://vite.dev/config/build-options#build-target).

HiDeoo:
Updated the branch to the latest Astro beta version (`7.0.0-beta.4`) which includes the change of the default Markdown processor from Unified to Sätteri. This will probably be our biggest change in this PR.

I also took the liberty of reorganizing some files in the `integrations/` directory, as I was getting very confused over time, e.g. is `asides.ts` related to Unified or Sätteri? Now, files are prefixed with `remark-` or `rehype-` if they are Unified plugins.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
HiDeoo, line 1: Initially added because with Rolldown, what is bundled with our components barrel file would slightly bring us over our CSS `size-limit` threshold. After testing it in our docs, Astro Docs, and a few other projects, I couldn't find any downsides and brings us way below the threshold so it ends up being a net positive.

packages/starlight/integrations/rehype-code-rtl-support.ts
HiDeoo, line 1: I took the liberty of reorganizing some files in the `integrations/` directory, as I was getting very confused over time, e.g. is `asides.ts` related to Unified or Sätteri? Now, files are prefixed with `remark-` or `rehype-` if they are Unified plugins (and `vite-` for Vite plugins).

.changeset/fancy-swans-tickle.md
delucis: Interesting 👀 

I’m generally a little sceptical of Baseline (there can sometimes be important cases where older browsers are still in use, so the two-year cut-off is sometimes a bit arbitrary). But it doesn’t really make sense for Starlight to override the Vite default would be my guess. Happy to go with this and if we get feedback that people are hitting compatibility issues we can reconsider.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
```diff
@@ -0,0 +1,40 @@
+import { fileURLToPath } from 'node:url';
+import type { ViteUserConfig } from 'astro';
+
+// https://vite.dev/guide/api-plugin#hook-filters
+const componentsBarrelIdFilter = /components\.ts(?:\?.*)?$/;
```
delucis: Could this be narrowed any further? For example at least with a leading slash to avoid matching files like `my-components.ts`?

Or could it even be the `@astrojs/starlight/components` path?

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const backSlashRegex = /\\/g;
+const queryStringRegex = /\?.*$/;
+
+const starlightComponentsBarrelId = normalizeId(
+	fileURLToPath(new URL('../components.ts', import.meta.url))
+);
```
delucis, line 12: Just to double check: is the `id` in the handler always the same resolved file path? I don’t want to block this PR on this detail, but it could be good to add a test to ensure this is working as expected on Windows/Unix given all the fun around paths.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
delucis: Very cool! Is there a reason not to use the simpler approach of adding `"sideEffects": false` to our `package.json`?

delucis (dismissed):
This looks fantastic! I left some queries around the lazy barrel optimization, but I don’t think they need to block releasing this.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
HiDeoo: Good call, just did that in https://github.com/withastro/starlight/pull/3951/changes/73626e16f6a1d454a40b128d9c4cc40f8c9a4a4b

We cannot do that with `@astrojs/starlight/components` as in the `transform` hook, `id` is already the resolved module id, not necessarily the original `@astrojs/starlight/components` specifier. Starlight code also forces this resolution internally via relative imports, so the stable approach is to match the resolved `components.ts` file path.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
HiDeoo: `id` in the handler will be resolved file paths matching our `componentsBarrelIdFilter` but we still need to normalize it and compare it against the resolved `starlightComponentsBarrelId` so we only apply the transform to our barrel file and nothing else.

delucis: Ah yeah, I was not concerned about it applying to other stuff. I was more wondering whether there could be a scenario where the normalized file path here _doesn’t_ match the ID Vite resolves to (so our transform wouldn’t apply and the optimization wouldn’t work).

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
HiDeoo: Basically fear 😅 

Using a plugin approach, we can say "mark this module and only this one as side-effect free so the lazy barrel optimization feature can be used on it". This makes me more confortable than doing that at the package level for all files without knowing all kind of other optimizations could be done based on that.

packages/starlight/integrations/vite-lazy-barrel-optimization.ts
HiDeoo: I see, I added a few tests that should cover path shenanigans (and I also tested on a Windows machine before and after the recent change).

delucis (approved):
Approving again — thanks for adding those changes @HiDeoo!

PR #3953, opened 2026-06-12
Fix Markdown processing path matching

HiDeoo:
#### Description

While implementing support for [`markdown.processedDirs`](https://starlight.astro.build/reference/configuration/#processeddirs) in a Starlight plugin, I realized on our implementation was a bit fragile. I would guess nobody ever ran into this issue in practice as this behavior existed since we initially introduced heading links.

For example, a configured path like `./src/content/custom/` could process `./src/content/custom-test/index.md`.

### DIFF
--- a/.changeset/true-buses-juggle.md
+++ b/.changeset/true-buses-juggle.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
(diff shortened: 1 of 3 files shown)


delucis (approved):
Ah yeah, easy one to miss — I’ve hit it before in similar scenarios. Fix looks good to me!

PR #3954, opened 2026-06-14
i18n(fa): Improve and update Persian translations

Goudarz:
#### Description

Improve and update Persian translations.

PR #3955, opened 2026-06-15
Add Persian language support and initial translations

Goudarz:
### Description
This PR adds Persian as a new language for the Starlight documentation and includes the initial translated pages.

### Changes
* Added `fa` to the locales configuration in `docs/astro.config.mjs`, `docs/lunaria.config.json`
* Added the Persian language subtag to `.github/labeler.yml`
* Added the Persian translation for:

  * `docs/src/content/docs/fa/index.mdx`
  * `docs/src/content/docs/fa/404.mdx`

### Notes
This is the initial step for Persian localization. Additional documentation pages will be translated and submitted in future pull requests.


docs/src/content/docs/fa/404.md
```diff
@@ -0,0 +1,14 @@
...
+title: صفحه پیدا نشد
+template: splash
+editUrl: false
+lastUpdated: false
+hero:
+  title: '۴۰۴'
+  tagline: <strong>آخ، به مشکلی برخوردیم.</strong> صفحه مورد نظر پیدا نشد.<br>نشانی وب را بررسی کنید یا از نوار جست‌و‌جو کمک بگیرید.
+  actions:
+    - text: بازگشت به صفحه اصلی
+      icon: right-arrow
```
delucis: I would guess we probably want to flip the arrow here:

```suggestion
      icon: left-arrow
```

Currently it looks like this:

button in farsi with a button on the left pointing right against the flow of text

docs/astro.config.mjs
```diff
@@ -21,6 +21,7 @@ export const locales = {
 	hi: { label: 'हिंदी', lang: 'hi' },
 	da: { label: 'Dansk', lang: 'da' },
 	uk: { label: 'Українська', lang: 'uk' },
+	fa: { label: 'فارسی', lang: 'fa-IR', dir: /** @type {'rtl'} */ ('rtl') },
```
delucis: AFAIK we don’t need the `IR` region subtag here — Persian Wikipedia for example just uses `fa`:

```suggestion
	fa: { label: 'فارسی', lang: 'fa', dir: /** @type {'rtl'} */ ('rtl') },
```

docs/src/content/docs/fa/index.mdx
```diff
@@ -0,0 +1,157 @@
...
+editUrl: false
+lastUpdated: false
+hero:
+  title: مستندات خود را با Starlight درخشان کنید
+  tagline: هر آنچه برای ساخت یک وبگاه مستندات حرفه‌ای نیاز دارید؛ سریع، در دسترس و آسان برای استفاده.
+  image:
+    file: ~/assets/hero-star.webp
+  actions:
+    - text: آغاز کنید
+      icon: right-arrow
```
delucis: Oh and same with the arrow here:
```suggestion
      icon: left-arrow
```

(Unfortunately we don’t have a reversed external link icon, which I guess may also be expected to have the arrow pointing up and left instead of up and right)

docs/__a11y__/test-utils.ts
```diff
@@ -41,6 +41,7 @@ const config: Config = {
 		locales: {
 			// N.B. If adding more locales here, also update the changed files filters in
 			// `.github/workflows/ci.yml` to ensure tests run when files for those locales change.
+			fa: ['guides/project-structure', 'reference/overrides'],
```
delucis, line 44: I chose two pages excluded from the English test suite and different from the Japanese ones for variety, but we could also align these if we prefer. (Neither of these are actually translated currently, so we could also choose to switch one to the homepage if that was important.)

docs/astro.config.mjs
```diff
@@ -21,6 +21,7 @@ export const locales = {
 	hi: { label: 'हिंदी', lang: 'hi' },
 	da: { label: 'Dansk', lang: 'da' },
 	uk: { label: 'Українська', lang: 'uk' },
+	fa: { label: 'فارسی', lang: 'fa', dir: /** @type {'rtl'} */ ('rtl') },
```
HiDeoo: Nit, but can't suggest as it's outside of review range, maybe we could use `/** @satisfies {import('@astrojs/starlight/types').StarlightUserConfig['locales']} */` before `export const locales` instead of sprinkling JSDoc inside the object.

HiDeoo (dismissed):
Looks really great to me, only left a tiny nit, but other than that, I think we're ready to have Persian language in the documentation 🚀 

Thanks for kicking this off @Goudarz and @delucis for bringing it to the finish line.

docs/astro.config.mjs
delucis: Good thinking! Done in 717889e78f0caee622091195a407a3ba88f9a06c

delucis (approved):
Thank you for the contribution @Goudarz! I removed the changes to the `fa.json` file that are already in progress in #3954 and made a few small tweaks, so I think this is ready to merge. We decided to include a couple of Persian pages in our accessibility test suite as we don’t currently test any RTL languages that way, so it’s a nice extra test to have, very helpful.

One small thing you could fix in a follow-up PR if you have time: the Starlight landing page (`index.mdx`) now includes a banner announcing the release of Astro v7. It would be good to add this to the Persian translation too.

Issue #3957, opened 2026-06-15
Manual setup instructions have a broken link for the integrations page - redirect not working


willklein:
### What page of the docs did you find an issue on?

Broken link for Integrations on the Manual Setup page

### Describe the issue

Under [Add the Starlight integration](https://starlight.astro.build/manual-setup/#add-the-starlight-integration), this link goes to a 404:

> Starlight is an [Astro integration](https://docs.astro.build/en/guides/integrations-guide/).

It should instead go to: https://docs.astro.build/en/guides/integrations/

~~I'll be posting a PR momentarily to fix this.~~
EDIT: the team mentioned there's a redirect not working that they want to address

### What operating system are you using?

Mac, but any OS will have the same behavior

### What browser are you using?

Edge, but any browser will have the same behavior

### Participation

- [x] I am willing to submit a pull request for this issue.

willklein:
This appears to be an issue with a redirect not working: https://github.com/withastro/starlight/pull/3958#pullrequestreview-4500648751

> Regarding the broken link, we actually have [a redirect in Astro Docs from /:lang/guides/integrations-guide to /:lang/guides/integrations/](https://github.com/withastro/docs/blob/main/public/_redirects#L19) which seems to be broken right now. We are currently investigating the issue to fix it in Astro Docs and then we can remove the changes to this link in this PR (and it even help us catch issues like this in the future).

HiDeoo:
Ah, I totally forgot to close this issue. This should have been fixed by https://github.com/withastro/docs/pull/14069

Would you be able to check now if https://docs.astro.build/en/guides/integrations-guide/ work for you? It seems to be fixed on my end.

HiDeoo:
Asked a few extra people and all confirmed that the URL is now working correctly.

Considering this, I'll go ahead and close the issue for now.

PR #3958, opened 2026-06-15
Add link to Astro install and setup guide on manual setup page

willklein:
#### Description

EDIT: I removed the "fixes (issue number)", as that is a redirect issue and will be addressed separately (#3957).

I was doing a manual setup with an existing project to add Starlight. I first noticed the broken link for integrations, which this PR addresses. I also found I needed to setup Astro first, but there wasn't a link to it in the document.

This PR:
- ~~fixes the broken link for the integrations page on the manual setup page~~
- adds a link to the manual setup for Astro in the neighboring content
- ~~makes these changes across all translations with a manual-setup.mdx page~~

I made these changes by hand, but relied on AI to identify the correct text to update for adding the manual setup for Astro link. I checked with translation tools and they look correct, but I cannot say I have 100% confidence the added link encapsulates the correct text perfectly across all languages. I have tested this locally.

This is my first contribution to Astro - let me know if you have any suggestions. I can always remove the added link for Manual Setup of Astro if desired. I realize a discussion may have been warranted for that type of improvement.

### DIFF
--- a/docs/src/content/docs/manual-setup.mdx
+++ b/docs/src/content/docs/manual-setup.mdx
@@ -10,7 +10,7 @@ If you want to add Starlight to an existing Astro project, this guide will expla
 
 ## Set up Starlight
 
-To follow this guide, you’ll need an existing Astro project.
+To follow this guide, you’ll need an [existing Astro project](https://docs.astro.build/en/install-and-setup/).
 
 ### Add the Starlight integration
 


HiDeoo (changes requested):
Thanks for the report and the PR :raised_hands:

Regarding the broken link, we actually have [a redirect in Astro Docs from `/:lang/guides/integrations-guide` to `/:lang/guides/integrations/`](https://github.com/withastro/docs/blob/main/public/_redirects#L19) which seems to be broken right now. We are currently investigating the issue to fix it in Astro Docs and then we can remove the changes to this link in this PR (and it even help us catch issues like this in the future).

> I also found I needed to setup Astro first, but there wasn't a link to it in the document.

I guess in the case you don't have Astro set up, the easier approach would be to follow our ["Create a new project" guide](https://starlight.astro.build/getting-started/#create-a-new-project) but linking to the Astro Docs installation guide wouldn't hurt either, altho I guess linking to https://docs.astro.build/en/install-and-setup/ without the `#manual-setup` anchor would be better since users may be interested in the prerequisites or previous sections of the guide.

Note that this change should only be done in the English version of the guide, and our translators will take care of the other languages independently.

willklein:
Thanks, @HiDeoo! I removed all the other changes, and changed the link to the manual setup for Astro to just be https://docs.astro.build/en/install-and-setup/. I also updated the PR description and title to reflect your input.

I think that added link would remove a small bit of friction when someone wants to get started with Starlight but needs to do the Astro setup first, as was my case. If you still think that's a good idea, that should do it for this PR.

HiDeoo (approved):
> I think that added link would remove a small bit of friction when someone wants to get started with Starlight but needs to do the Astro setup first, as was my case.

Definitely, thanks for updating the PR, looks perfect to me :star2:

PR #3959, opened 2026-06-16
[i18nIgnore] i18n(zh-cn): Update `css-and-tailwind.mdx` translation

Ying-Luan:
#### Description

- Update outdated zh-cn translation in `css-and-tailwind.mdx`.
- Other parts of the zh-cn translation may also be outdated, but I'm not familiar enough with the project to translate them accurately, so they are left untouched for now.

PR #3961, opened 2026-06-17
docs: remove some outdated Tailwind related translations

HiDeoo:
#### Description

Supersedes #3959

Should be merged after #3959 (so the author get credit) with the following Lunaria directive:

```
@lunaria-ignore:docs/src/content/docs/es/guides/customization.mdx;docs/src/content/docs/es/reference/configuration.mdx;docs/src/content/docs/zh-cn/guides/customization.mdx;docs/src/content/docs/zh-cn/reference/configuration.mdx
```

This PR removes some outdated Tailwind related translations for the following files:

- `docs/src/content/docs/es/guides/css-and-tailwind.mdx` (last updated Sep 11, 2024)
- `docs/src/content/docs/zh-cn/guides/css-and-tailwind.mdx` (last updated Mar 5, 2025)

The specific reason to focus these 2 files is that they are still documenting Tailwind v3 (v4 was released January 22, 2025) and link to [this page](https://docs.astro.build/en/guides/integrations-guide/tailwind/) which can be understood as Starlight not supporting Tailwind v4 (this is the [exact reason](https://github.com/withastro/starlight/pull/3959#issuecomment-4718412522) why #3959 was opened). The installation instructions are also very different, e.g. using a Vite plugin, using CSS configuration files instead of `tailwind.config.mjs` files, etc.

#3959 tried to fix the `zh-cn` page by only updating the link, but I think in this case and considering the huge difference between Tailwind v3 and v4, and that these pages have not been updated for more than 1 year, it is better to remove them entirely. If someone wants to add them back and update them, they can still rely on the git history to find the old content.

This may also be the time to do another round of outdated translation removal, but with Astro 7 around the corner, just targetting these 2 files for now seems like a good idea.


ArmandPhilippot (dismissed):
Right, I remember the Tailwind headache. 😄 This sounds like a good idea to me! Outdated translations are often more harmful than helpful.

delucis (dismissed):
Thanks for thinking this through @HiDeoo! The plan makes sense to me 👍

louisescher (approved):

ArmandPhilippot (approved):

PR #3962, opened 2026-06-17
docs: fix stale Astro integration links in manual setup

Kyzenkms:
## Summary
- fix the broken Astro integrations docs link on the manual setup page
- fix the stale Cloudflare adapter docs link in the same guide

## Testing
- verified both links were updated in the docs source


HiDeoo:
Thanks for the contribution.

Are both links not working for you? They're actually redirects that had an issue a few days ago, but should be [fixed now](https://github.com/withastro/docs/pull/14069) and they both appears to be working for me.

HiDeoo:
Asked a few extra people and all confirmed that the URLs are working correctly.

Considering this, I'll go ahead and close the issue for now. If you encounter any further problems with the URLs, feel free to open an issue first, and we can investigate further.

Thanks for your contribution nonetheless.

PR #3964, opened 2026-06-17
docs: add starlight-md-txt plugin

max-ostapenko:
Adds the starlight-md-txt plugin to the community plugins list. starlight-md-txt allows exposing documentation pages as raw, agent-friendly Markdown at .md.txt URLs.


HiDeoo (changes requested):
Thanks for the new addition :star2:

- It looks like the URL is either invalid or maybe the repo is not public? I'm personally getting a 404 error when trying to access it.
- Could you append the new entry at the end of the existing list as it's the approach we use for such pages?

max-ostapenko:
@HiDeoo thanks for feedback ✅

HiDeoo (approved):
Amazing, thanks for the quick follow-up and new plugin!

Everything looks perfect to me :rocket:

PR #3965, opened 2026-06-18
Use canonical /integrations/ URL on manual setup page

### LINKED ISSUE #3957: Manual setup instructions have a broken link for the integrations page - redirect not working

farrosfr:
Fixes #3957.

The link to the Astro "integrations" reference on the manual setup page was pointing at the legacy `/guides/integrations-guide/` path. That path was returning 404 due to a broken redirect in the Astro docs site.

The redirect itself has now been fixed upstream in [withastro/docs#14069](https://github.com/withastro/docs/pull/14069), so the existing link works again. This PR goes one step further and switches the link to the canonical `/guides/integrations/` URL so Starlight no longer depends on that redirect staying healthy.

## Changes

- `docs/src/content/docs/manual-setup.mdx` and the `de`, `es`, `fr`, `ja`, `ko`, `ru`, and `zh-cn` translations: the "Starlight is an Astro integration" link now uses `/<lang>/guides/integrations/` directly.

## Notes

- The Cloudflare adapter link on the same page (`/guides/integrations-guide/cloudflare/`) was checked and is the actual canonical URL for that page, not a redirect — left unchanged.
- The `uk` translation was intentionally skipped. `docs.astro.build/uk/guides/integrations/` does not exist yet (404), and the current file already links Astro docs at the `/en/` prefix as a fallback. Fixing that is a separate translation gap.
- `pnpm linkcheck` passes locally (592 pages built, all internal links valid). No changeset needed per `CONTRIBUTING.md` (docs-only change).

### DIFF
--- a/docs/src/content/docs/de/manual-setup.mdx
+++ b/docs/src/content/docs/de/manual-setup.mdx
@@ -14,7 +14,7 @@ Um dieser Anleitung folgen zu können, benötigst du ein bestehendes Astro-Proje
 
 ### Hinzufügen der Starlight-Integration
 
-Starlight ist eine [Astro-Integration](https://docs.astro.build/de/guides/integrations-guide/). Füge sie zu deiner Website hinzu, indem du den Befehl `astro add` im Projektstamm&shy;verzeichnis deines Projekts ausführst:
+Starlight ist eine [Astro-Integration](https://docs.astro.build/de/guides/integrations/). Füge sie zu deiner Website hinzu, indem du den Befehl `astro add` im Projektstamm&shy;verzeichnis deines Projekts ausführst:
 
 <Tabs syncKey="pkg">
     <TabItem label="npm">
--- a/docs/src/content/docs/es/manual-setup.mdx
+++ b/docs/src/content/docs/es/manual-setup.mdx
@@ -14,7 +14,7 @@ Para seguir esta guía, necesitarás un proyecto de Astro existente.
 
 ### Agregando la integración de Starlight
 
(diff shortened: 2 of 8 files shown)


HiDeoo:
As [mentioned in the issue](https://github.com/withastro/starlight/issues/3957#issuecomment-4733621649), this was a temporary issue with redirects that has since been [resolved](https://github.com/withastro/docs/pull/14069).

I just tested the URLs and they are working correctly now and also asked a few extra people and all confirmed that the URLs are working correctly. We are also perfectly fine relying on redirects as it avoids updating many pages (we also cannot update translations like in this PR as this would mark them as up-to-date while they may not be), and it also gives us a chance to catch such issues in the future if redirects are broken.

Thanks for your contribution nonetheless.

farrosfr:
Thanks for the detailed feedback, that makes sense. I did not realize the Lunaria side-effect of touching translations; marking them as "complete" purely because a URL happens to match in English was a bad signal to broadcast. The redirect-as-tripwire argument for leaving things alone is also a good one I had not considered.

Appreciate you taking the time to explain it. Lesson learned.

PR #3967, opened 2026-06-18
Add link icons and dedupe Markdown/component icons

HiDeoo:
Follow-up to #3923, this PR cleans up various parts regarding aside and heading link icons:

- The Unified/Sätteri aside plugins now uses the same Starlight icons as the existing ones already rendered by the `<Aside>` component
- The Unified/Sätteri heading link plugins and `<AnchorHeading>` component now uses a new common Starlight icon (`link-alt` to match the Unicons naming)
- As it felt weird to only add new `link-alt` icon, I also added the matching `link` icon

The diff also shows how much more minified the aside component icons are compared to our previous Markdown plugin ones. I wonder if we forgot to minify them or just used a different preset 🤷 

~~Not quite sure if a changeset for the deduping part of the PR should be added. Visually, nothing changed, and it's mostly SVG path changes. And if yes, not quite sure how to phrase it as my attempts so far kinda hinted wrong icons were rendered which is not the case 😅~~ Finally managed to write something I'm happy with.


packages/starlight/__tests__/markdown-processor/snapshots/satteri/generates-anchor-link-markup.html
HiDeoo, line 1: The diff here is mostly from using SVGOMG with the Unicons [`link-alt` icon](https://iconscout.com/unicons/free-line-icon/link-alt) which produces a slightly different result compared to the one we had previously.

One thing to note is that `fill="currentColor"` is now applied to the SVG rather the `<path>` as we are re-using the path from a Starlight icon now and we re-use the same approach used for all Starlight icons.

packages/starlight/__tests__/markdown-processor/snapshots/satteri/generates-aside.html
HiDeoo, line 1: For this diff and the following ones, the changes are mostly us now re-using the `<Aside>` component icons for the Markdown asides, and it looks like our Markdown aside icons never got optimized by SVGOMG.

packages/starlight/components/AnchorHeading.astro
HiDeoo, line 1: We are now using the new and shared `link-alt` Starlight icon for this component.

packages/starlight/integrations/anchor-icon.ts
HiDeoo, line 1: Removed in favor of a shared `link-alt` Starlight icon.

packages/starlight/integrations/aside-icons.ts
HiDeoo, line 1: Icon paths removed in favor of using the same Starlight icons used by the `<Aside>` component we already had.

packages/starlight/integrations/aside-utils.ts
HiDeoo, line 1: A set of utility functions for asides now regrouped in this single file, that are used either by the Markdown pipeline (both Unified and Sätteri) and also the `<Aside>` component.

packages/starlight/integrations/asides-error.ts
HiDeoo, line 1: Moved and internalized in `packages/starlight/integrations/aside-utils.ts`.

packages/starlight/integrations/markdown-icon.ts
HiDeoo, line 1: Mardown pipeline (Unified and Sätteri) specific icons and helpers to consume Starlight icons (extracted from `packages/starlight/integrations/remark-asides.ts`).

packages/starlight/integrations/remark-asides.ts
HiDeoo, line 1: Starting from this file, we are now mostly re-using the new (Unified and Sätteri) helpers we have, and removing dead-code.

HiDeoo:
Updated the PR and added some comments to hopefully make the review easier to follow 🤞

packages/starlight/integrations/remark-asides.ts
```diff
@@ -125,13 +122,6 @@ function makeSvgChildNodes(children: Result['children']): P[] {
  * ```
  */
 export function remarkAsides(options: MarkdownProcessorPluginOptions): Plugin<[], Root> {
-	const iconPaths: Record<AsideVariant, ReturnType<typeof s>[]> = {
-		note: asideIconPathAttrs.note.map((attrs) => s('path', attrs)),
-		tip: asideIconPathAttrs.tip.map((attrs) => s('path', attrs)),
-		caution: asideIconPathAttrs.caution.map((attrs) => s('path', attrs)),
-		danger: asideIconPathAttrs.danger.map((attrs) => s('path', attrs)),
-	};
```
delucis, line 133: Should we maybe keep this map and call `parseIconChildren` for each icon early here? That way we have to parse each icon only once instead of re-parsing for every aside instance. (I’d guess it’s a very tiny overhead, given the SVGs are pretty small, but still could add up across all instances in big sites?)

packages/starlight/style/anchor-links.css
```diff
@@ -110,6 +110,7 @@ how we do it.
 	.sl-markdown-content .sl-anchor-icon > svg {
 		display: inline;
 		width: var(--sl-anchor-icon-size);
+		margin-top: 0;
```
delucis: Is there a reason this was needed? Obviously don’t mind adding it, just couldn’t spot why this would now be required.

.changeset/deep-plants-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Adds 2 new icons: `link`, and `link-alt`.
```
delucis: Grammar nit:
```suggestion
Adds 2 new icons: `link` and `link-alt`.
```
(An Oxford comma before “and” is only used for lists of 3+ items)

.changeset/wicked-spoons-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Reuses the same optimized icons as the `<Aside>` component for Markdown aside icons.
```
delucis: Maybe could even be this thinking of the user-facing change? Maybe the fact they’re the same is not super important?
```suggestion
Optimizes the icons in the `<Aside>` component.
```

delucis (commented):
Looks great — thank you for following up on this idea!

I left a few small queries/nits, but nothing major.

packages/starlight/integrations/remark-asides.ts
HiDeoo: Good call, just added it back in [`5ce68d9` (this PR)](https://github.com/withastro/starlight/pull/3967/commits/5ce68d9f5fbd2b0d1ddd1ff1484a28a9ca302998)

packages/starlight/style/anchor-links.css
HiDeoo: Great catch, this was from an initial attempt to use our `<Icon>` component to render the icon in our `<AnchorHeading>` but I decided against it and this slipped through my review.

.changeset/wicked-spoons-own.md
HiDeoo: Well, it's actually the opposite, the icon used by the component were optimized, it's the Markdown one using directives that were not. I went with "Optimizes the icons of Markdown asides." but happy to update it again if needed.

delucis: Oh, I got it backwards. But I’ll take it as a signal that the simpler changeset is an improvement and not that I can’t read properly 😅

delucis (approved):
Perfect — thanks @HiDeoo 🙌

delucis:
Going to merge this to get it out in the same patch with #3988, hope that’s ok!

moeriki:
Hi! Since the last update (`0.41.1`) I've got Starlight rendering raw SVGs for aside icons.

CleanShot 2026-06-29 at 15 53 16@2x

This happens only in MDX files, not MDs.

Currently I have this patch working for me.

```patch
diff --git a/integrations/satteri.ts b/integrations/satteri.ts
index 32a0907d8928a678a02187019509fd34277aeb41..9bd133278877c7488530e021070f9ad734935628 100644
--- a/integrations/satteri.ts
+++ b/integrations/satteri.ts
@@ -112,7 +112,18 @@ function satteriAsidesPlugin(
 				},
 				[
 					paragraphElement('p', { class: 'starlight-aside__title', 'aria-hidden': 'true' }, [
-						{ type: 'html', value: iconSvg },
+						// PATCH: a bare `{ type: 'html' }` node renders in `.md` but is
+						// escaped to literal source in `.mdx` (where raw HTML must be a
+						// `set:html` Fragment). Emit Sätteri's MDX-safe Fragment form so
+						// the icon renders in MDX. See upstream withastro/starlight #3967.
+						{
+							type: 'mdxJsxTextElement',
+							name: 'Fragment',
+							attributes: [
+								{ type: 'mdxJsxAttribute', name: 'set:html', value: iconSvg },
+							],
+							children: [],
+						},
 						...titleNode,
 					]),
 					paragraphElement('div', { class: 'starlight-aside__content' }, children),
```

It looks like it might be related to this change.

HiDeoo:
👋 @moeriki 

Would you be able to share a minimal reproduction of the issue, e.g. on [StackBlitz](https://astro.new/starlight-basics?on=stackblitz)? So far, I'm not able to reproduce such behavior, for example here is StackBlitz with many asides in MDX: https://stackblitz.com/edit/github-egjmzxro?file=src%2Fcontent%2Fdocs%2Findex.mdx

moeriki:
https://stackblitz.com/edit/github-egjmzxro-hf1kf6yk?file=package.json,astro.config.mjs

I found the issue lies in the fact we integrate our own `@astrojs/mdx`.

```js
{
  integrations: [
    starlight({ /* … */ }),
    mdx({
      optimize: true,
    }),
  ],
}
```

Setting `optimize: true` now also fixed the issue (without the patch).

delucis:
Thanks @moeriki! Looks like the `type: "html"` approach was working under the assumption of `optimize: true` (which Starlight sets for users by default in its own copy of the MDX integration if users themselves don’t provide one).

But we should make it work in your scenario too.

delucis:
Opened a PR to fix this: https://github.com/withastro/starlight/pull/4015

PR #3976, opened 2026-06-20
[search] run onPageLoad also for Astro view transition

### LINKED ISSUE #3975: `kbd` disappears in `<Search>` with astro's ClientRouter

0.40.0

6.4.8

pnpm & npm

Mac & Linux

Chrome but doesn't matter I think?

### Describe the Bug

When going from 1 page to another one with `ClientRouter` enabled, the `kbd` in the Search disappear

https://github.com/user-attachments/assets/90bfdcfc-98a5-4850-b174-0e728ea7eb93

But it reappears on page reload:

https://github.com/user-attachments/assets/d934bee3-8913-4c68-8f45-8b088a30053f

The reason is because we have code running on page load, but the HTML gets reloaded (which has `display: none` and the JS removes it). We need to move this JS to run on both page mount and `document.addEventListener("astro:page-load", () => {`

https://stackblitz.com/edit/github-qarqfjfb?file=src%2Fcomponents%2FHead.astro,astro.config.mjs&on=stackblitz

### Participation

Ayc0:
#### Description

- Closes #3975
- Runs the same fn as before, but instead of just at page load, we run it for both the regular page load, and the virtual page loads related to View Transition

Before | After
-|-
<video src="https://github.com/user-attachments/assets/bc0a5074-1171-48ba-918a-aeab8f3d8e3a"> | <video src="https://github.com/user-attachments/assets/0b1f6250-fbea-4a20-a702-d14b90071b26">

### DIFF
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -60,7 +60,12 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 	 */
 }
 <script is:inline>
-	(() => {
+{
+	const onPageLoad = () => {
+		if (!shortcut.style.display) {
+			// Avoid running twice for the initial load
+			return;
+		}
 		const openBtn = document.querySelector('button[data-open-modal]');
 		const shortcut = openBtn?.querySelector('kbd');
 		if (!openBtn || !(shortcut instanceof HTMLElement)) return;
@@ -70,7 +75,11 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 			openBtn.setAttribute('aria-keyshortcuts', 'Meta+K');
 		}
 		shortcut.style.display = '';
-	})();
+	});
+
+	onPageLoad(); // For regular page loads
+	document.addEventListener("astro:page-load", onPageLoad); // For Astro's View Transition system
+}
 </script>
 
 <script>
--- a/packages/starlight/components/Search.astro
+++ b/packages/starlight/components/Search.astro
@@ -62,20 +62,20 @@ if (project.trailingSlash === 'never') dataAttributes['data-strip-trailing-slash
 <script is:inline>
 {
 	const onPageLoad = () => {
+		const openBtn = document.querySelector('button[data-open-modal]');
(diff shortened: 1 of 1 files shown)


Ayc0:
I noticed that this issue was related to an IIFE, but there are <a href="https://github.com/search?q=repo%3Awithastro%2Fstarlight%20%2F%5C)%5C(%5C)%2F&type=code">3 other files with IIFEs in the code</a>:
image

I don't know if those suffer from the same issue

HiDeoo:
Thanks for the contribution 🙌 

As mentioned in https://github.com/withastro/starlight/issues/3975#issuecomment-4759214591, we rely on a extra package dedicated to client-side routing and view transitions as such requirements are not common in documentation projects.

To get started, this [guide](https://events-3bg.pages.dev/jotter/starlight/) walks through two possible approaches to add proper support in Starlight.

Ayc0:
Hi @HiDeoo 👋 

I also tried out `astro-vtbot` but it seemed that the issue remained even with it :/ Which could make sense as the issue seems to be the JS itself of a starlight file

Also in the guide you mentioned (that I checked before sending this issue. And same for the linked comment that you shared, I previously read it before opening this issue / PR), it is mentioned:

> [Option 1: With Astro’s <ClientRouter />](https://events-3bg.pages.dev/jotter/starlight/#option-1-with-astros-clientrouter-)

So I figured that we could fix it upstream s that both versions of the Guide could work. 

A patch could be applied in `astro-vtbot` where this function would be called twice there instead. But then the code would have to be duplicated across 2 repos

Ayc0:
(Btw, thank you very much for the very very quick answer!!)

HiDeoo:
> I also tried out `astro-vtbot` but it seemed that the issue remained even with it

👋 Oh, interesting, I have not encountered the same behavior so far personally.

In this case, I would definitely recommend opening an issue on the `astro-vtbot` repository with a repro using the package. Starlight code is purposely not handling Astro view transition lifecycle events (so having the fix upstream would break that contract) but `astro-vtbot` is designed to handle those events, so it should work as expected.

Ayc0:
Noted 👍 
I'll transfer it over there, thanks!

PR #3985, opened 2026-06-24
i18n(ko-KR): update ` index.mdx`

jsparkdev:
#### Description

- #3951


delucis (approved):
LGTM! Thanks @jsparkdev 🚀 

(Just realising this means we don’t have localized upgrade guidance because we’re linking to the English changelog, but hopefully that’s OK.)

PR #3990, opened 2026-06-25
docs: bring back links validation

HiDeoo:
#### Description

This PR brings back links validation to the docs. This was temporarily removed in https://github.com/withastro/starlight/pull/3951 as the `starlight-links-validator` plugin was not yet compatible with Sätteri.

Note that I am completely fine waiting a few days so that we don't have to change our `minimumReleaseAgeExclude` configuration but just wanted to open the PR and we can merge it whenever we want, it being now or in 3 days.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -37,3 +37,6 @@ minimumReleaseAgeExclude:
   - 'expressive-code'
   - 'astro-expressive-code'
   - 'rehype-expressive-code'
+  - 'starlight-links-validator'
+  - '@napi-rs/wasm-runtime'
+  - '@tybys/wasm-util'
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -37,3 +37,6 @@ minimumReleaseAgeExclude:
   - 'expressive-code'
   - 'astro-expressive-code'
   - 'rehype-expressive-code'
+  - 'starlight-links-validator'
+  - '@napi-rs/wasm-runtime'
+  - '@tybys/wasm-util'
```
HiDeoo, line 42: [Explanation](https://discord.com/channels/830184174198718474/1484341972885049645/1519749428825821274) for this 2 packages.

delucis (approved):
Thanks for getting the plugin updated so quickly!

PR #3995, opened 2026-06-26
i18n(de): update manual-setup.mdx

trueberryless:
#### Description

This PR updates the German translation of `manual-setup.mdx` with the changes from #3958.
I decided to link to the `/de` route alto it is not translated in Astro docs yet, but if we will ever translate it (😅), we don't need to update it here. And currently, the English version is shown as fallback.


delucis (approved):
Thanks! And I think it makes sense to use `/de/` here 👍

PR #4000, opened 2026-06-27
docs: move FeelBack from `plugins.mdx` to `community-content.mdx`

trueberryless:
#### Description

This PR updates all non-GitHub links in `plugins.mdx` to point to the GitHub repositories instead.
As a user, I expect to be able to learn common practices when clicking on listed plugin cards and since most of them link to GitHub, all non-GitHub links break this promise. Moreover, titles have been made consistent as well to match the repo name.

The most tricky one by far is the "FeelBack" tool. The current links points to a blog post explaining how to integrate FeelBack into Starlight, but there is no dedicated package for Starlight, only for Astro. I decided to move it to the `community-content.mdx` list because of this reason. We could maybe also link to https://github.com/giuseppelt/feelback-integrations/tree/master/samples/astro-starlight, but it contains too little information compared to the blog post. Open for suggestions.


HiDeoo (changes requested):
Thanks for the suggestions.

> all non-GitHub links break this promise

This is not a promise we are currently making or can even make. First, someone could use a different service to host their code.

For themes, we ask for a demo/documentation link as it's very visual, but for a plugin I think it's perfectly reasonable to let authors decide if they prefer to link to either a repository or a documentation website (_as long as such website links to the repository_). It may be a matter of preference, a plugin being very visual and it makes more sense to link straight to a demo embedded in a documentation website, an author having put a lot of effort into a documentation website and wanting to link to that, lack of proper documentation in the repository, monorepo structure where it could be tricky to link to the proper documentation in nested folders, etc.

> titles have been made consistent as well to match the repo name.

I don't think it's a good idea to enforce such a rule. The name of a plugin is not necessarily the same as the repository name, specially for monorepos, and because a repository/package have naming restrictions, a plugin title could be branded differently from them. For example, for themes, lots of them are following the "Starlight Theme Name" pattern, but we already have themes that don't follow this pattern and we don't enforce it. If someone wants to brand their plugin with a different name than the repository or different casing, I think it's perfectly reasonable to let them do so.

Regarding trailing slash changes, I don't think it's something we can enforce either. Some websites may redirect to a URL with a trailing slash, some may not. It's also something we're not validating automatically (and probably don't want to) so this would end up being a change that adds more work to translators to update them, more work to maintainers in the future to check them (and will probably be forgotten with time or just not known by some maintainers).

I think this leaves us with the FeelBack change which I think make sense to move as it's a blog post/guide rather than Starlight specific tooling.

trueberryless:
Thanks for taking the time to write such a detailed explanation!

<details>
<summary>Table of contents</summary>
<br/>

In advance: I am very sorry for how long this response is. It wasn't meant to be that long, it just happened while thinking about ideas that came to me while reading your reply. To make it easier to read, I provided this overview:

1. Reasons for change: My take on why I think we should rethink some parts of our plugin page (and theme + community content).
2. Concrete new suggestions for improvements: 2 ideas based on your provided feedback
3. Closing thoughts

</details>

### 1. Reasons for change

After reading through my own description again, I can see that I provided somewhat misleading reasoning for why I think some of the rejected changes would be beneficial - but not necessarily in the way I currently represented them in this PR. So let me clarify that:

I do not want to enforce titles or links. I do not aim to only _allow_ linking to GitHub and neither do I want that all titles strictly have to follow the pattern. For example, our ["How to add a new theme" contribution guide](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#themes) clearly does not specify any rules for the `title`.

The reason I am suggestion these changes is that with the current catalogs (themes/plugins/community content), which are listed on those pages, there are only very little which stand out of line. And by revisiting existing entries with the knowledge we gained over the years since those pages have been created, we have the potential to present each entry with much more care and love. This doesn't mean that it is against any regulation to not follow any pattern. In fact, I would be very happy to make the titles in general more readable with "Title Case" instead of "kebab-case".

The point I am trying to make is: I believe that since we created those pages and added entries over time, we have learned more about the content that people create in the ecosystem around Starlight. So I think now is the right time to take a look back at all the things we created and crafted together and give each one more meaning and the chance to get discovered more easily through a visual overhaul.

### 2. Concrete new suggestions for improvements

While my initial commit here was more about just consistency, I am happy that I brought this PR up and discuss my ideas for improvements:

Since many of the plugins listed have their dedicated docs sites, I propose (in contrast to my initial commit) to link to those specific documentations in cases where possible. Especially plugins that got crafted with [your generator](https://github.com/HiDeoo/generator-starlight-plugin), often do not have much documentation in their `README.md` files. Instead, the main source to get to know the plugin is their own site. Such a change would also encourage plugin authors to link to the plugin's documentation site instead of following a semi-pattern - linking to GitHub - that grew over the years because the first entries showcased it that way.

One counterargument is that GitHub is sometimes a more reliable hosting source, whereas the guarantee that the plugin author will be able to host the plugin's docs site is not given. Nonetheless, since a large portion of plugins are authored by you, Chris, Frostybee and some by me, I know that the main hosting platforms are either GitHub pages, Netlify or Vercel. At least for those I am confident in suggestion to link to documentation directly.

It may also be a personal taste whether to link to GitHub or the docs site, because in the end, both link to each other in most cases. Considering that you basically have almost a monopoly on Starlight plugins (thanks for all this amazing stuff), if you agree to link to documentation, I suggest that we could only change your and my links for now or ask other authors as well.

It's just a suggestion, I can understand if you want to leave it as is. But I have to say that it often happens that I wait impatiently for GitHub to load, just so I can click the link to the docs right away.

---

Regarding titles of plugins, I also want to suggest the complete opposite now: Why don't we just make all titles "Title Case". The official plugin is already human-readable, so why should we present displayed text in an URL-safe format? I can just speak for myself here: I always wondered why we do it that way, and once again I believe that it just happened to be that way because the first plugins paved this way. And while we are already discussion about reworking and -considering the cards, we could as well think about that decision as well.

### 3. Closing thoughts

I think that my new suggestions definitely need some discussion. I would be interested in how other plugin authors see this and whether they also just "followed the trend" like I did in the past. 

I would be happy to focus on your suggestion to just move the "FeelBack" card in this PR, move this whole discussion over to Discord and ask the top plugin authors for their opinion. And if we can't decide, I think this is also good, because then we can showcase that we do not strictly enforce anything. Once again, I believe that many authors are also just not aware that they can choose the format they want, which led us to have like over 95% having the same "kebab-case". If we split them more like 50/50, no plugin would step out of line because there is no pattern to begin with.

Although you probably already know that I am a big fan of consistency, in this case I would be fine if we showcase multiple variants: title case, kebab case, more creative theme names, etc. That way, we encourage people to also be more creative with their titles and maybe motivate them to create a dedicated documentation for their plugin as well.

I'll update this PR in the meantime to only included the changes you asked for, removing all the others. Nonetheless, I'm looking forward to your response (no matter how long it is 😅).

HiDeoo (approved):
Thanks for updating the PR, looks great to me.

---

> Although you probably already know that I am a big fan of consistency

While I also value consistency, I believe that flexibility can be just as important in certain situations, specially in the case of user-authored content.

Let's take for example the idea, which changed between the initial proposal and now, to change the cases of all titles or update all URLs. This would require to:

- Move the discussion somewhere else
- Ping most plugin authors to discuss about it
- Have maintainers to moderate such discussions, chiming in on the topic and providing feedback
- Have maintainers to make a final decision on the matter
- Either:
  - Author new rules describing the new preferred case style or URL
  - Potentially author and maintain automated tools to enforce this
- Have maintainers be updated on the new recommendations
- Update all existing titles/URLs in the source language
- Update all existing titles/URLs in all translated languages
- Have maintainers remember to review for such criteria when reviewing new content

Now considering the context of an open-source project like Starlight where all contributors are volunteers, this is a lot of work and involvement to ask from many people, for a change that is not critical to the project and wouldn't provide any significant benefit to the end users or maintainers. I would personally prefer to use our limited resources on more impactful changes whenever we have some. Being mindful of the time and effort of our contributors is important.

trueberryless:
Okay. Understood 👍

PR #4004, opened 2026-06-29
i18n(fr): update `index.mdx` (v7 banner)

ArmandPhilippot:
#### Description

Updates the French translation of the homepage to add the v7 banner (see #3951).

The second sentence is a bit verbose, but I thought it might sound more natural than "Découvrez comment mettre à niveau" or "Apprenez à mettre à niveau". And it seems to be fine on small viewports:

But, I don't have a strong opinion if you prefer something else.


HiDeoo (approved):
Personally perfectly happy with your choices, perfect as always 🌟

PR #3997, opened 2026-06-26
Remove 'starlight-utils' LinkCard from plugins documentation

lorenzolewis:
Removed the LinkCard for 'starlight-utils' from the documentation as I no longer have time to maintain this (and it's very out of date). In the documentation site I offer some alternatives for the functionality that this package included.

PR #3993, opened 2026-06-26
Add Axiom Studio for Starlight to community tools

pixelizing:
#### Description

- **What does this PR change? Give us a brief description.**
Hey everyone! 👋 I recently spent a couple of weeks building a local-first browser editor for Starlight called **Axiom Studio**. I originally built it to make my own MDX and frontmatter editing workflow easier and safer without needing any external CMS or database. 

Thought it might be useful for other indie makers and developers in the Astro ecosystem, so I added it to the community tools list in `plugins.mdx`. (I also included a short demo video on the link if you're curious to see how the local editing actually works in action). Let me know if the format looks good!

- **Did you change something visual? A before/after screenshot can be helpful.**
Nope, no visual changes to the docs themselves. Just added the new `<LinkCard>` to the bottom of the grid.

PR #4005, opened 2026-06-29

ArmandPhilippot:
#### Description

Adds changes from #3964, #3993, and #3997 to the French translation of `resources/plugins.mdx`.

I tweaked a bit the translation for "local-first browser editor" (ie. "fonctionnant localement dans le navigateur" instead of something like "éditeur dans le navigateur privilégiant le fonctionnement en local" even if it could be more accurate) to shorten the card:

### DIFF
--- a/docs/src/content/docs/fr/resources/plugins.mdx
+++ b/docs/src/content/docs/fr/resources/plugins.mdx
@@ -59,11 +59,6 @@ Les [modules d'extension](/fr/reference/plugins/) peuvent personnaliser la confi
 		description="Ajouter des fonctionnalités de zoom aux images de votre documentation."
 	/>
 	<LinkCard
-		href="https://github.com/lorenzolewis/starlight-utils"
-		title="starlight-utils"
-		description="Étendre Starlight avec une collection d'utilitaires courants."
-	/>
-	<LinkCard
 		href="https://github.com/trueberryless/starlight-view-modes"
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/fr/resources/plugins.mdx
```diff
@@ -317,4 +312,14 @@ Ces outils et intégrations communautaires peuvent être utilisés pour ajouter
...
 	/>
+	<LinkCard
+		href="https://github.com/max-ostapenko/starlight-md-txt"
+		title="starlight-md-txt"
+		description="Exposer vos pages de documentation Starlight sous forme de Markdown brut adapté aux agents, via des URL se terminant par .md.txt."
+	/>
+	<LinkCard
+		href="https://github.com/pixelizing/axiom-studio-for-starlight"
+		title="Axiom Studio pour Starlight"
+		description="Un éditeur fonctionnant localement dans le navigateur pour modifier les pages Markdown et MDX, gérer le frontmatter et prévisualiser la documentation Starlight sans toucher aux fichiers sources."
```
HiDeoo, line 323: If ever needed, by checking the documentation, local-first in this case seems to mean local and that's it, so we could shorten it to "Un éditeur local…"

HiDeoo (approved):
Thanks for the update, looks perfect to me 🚀

PR #4008, opened 2026-06-29
fix(css): prevent TOC overflow with custom --sl-content-width

### LINKED ISSUE #3513: Table of Contents overflows with custom content width

0.36.2

5.6.1

npm

Mac

Firefox (Zen)

### Describe the Bug

When setting `--sl-content-width` to a custom value greater than the default `45rem`, the Table of Contents (ToC) overflows outside the right side of the screen.

Example: 
- [Sharp Docs](https://sharp.pixelplumbing.com/install/) (resize window width to around `1200px` to see the overflow): [They set the content width to `60rem`](https://github.com/lovell/sharp/blob/040b73ca746f4b8e71950708de4a464c7ba6a188/docs/src/styles/custom.css#L6)
- [Aspire Docs](https://aspire.dev/get-started/what-is-aspire/) (resize window width to around `1400px` to see the overflow): They set the content width to `68rem` (no public repo found)
- [Second Life Creation](https://create.secondlife.com/script/) (resize window width to around `1450px` to see the overflow): [They set the content width to `70rem`](https://github.com/secondlife/create/blob/115aaef84c1bd534fec4ffdf1e640421ea9ece40/src/styles/custom.css#L2)
- Here are other examples I found: https://docs.xtdb.com/index.html

I'm not sure how to fix this cleanly, as I do not fully understand why the width of the ToC is not calculated correctly. I mean, [this calculation](https://github.com/withastro/starlight/blob/dad3709c5220bce66eb84b7005228a9e467f2721/packages/starlight/components/TwoColumnContent.astro#L25) or [this one](https://github.com/withastro/starlight/blob/0a615ec5ed76049ba2f224b2efcf81265a413ddb/packages/starlight/components/PageSidebar.astro#L48-L55) should make sure that the ToC is not overflowing.

One potential solution that came to my mind is making the breakpoint of when the ToC will be in the mobile version dynamic. In other words: If the content width is higher, the ToC has less space and should have a higher `@media (min-width: <higher>rem`, so that it is longer displayed over the main content. However, variables are not supported in media queries as far as I now: [`width` spec](https://drafts.csswg.org/mediaqueries/#width) ↦ [Unit spec](https://drafts.csswg.org/mediaqueries/#units) ↦ [overall Unit definition](https://drafts.csswg.org/css-values-3/).

I'm open for implementing the fix when we have discussed the best way to handle this bug. Maybe we do not even need to handle it as the user themselves are changing the default values, but I think since the calculation is already dynamic, we should not just support the default values from Starlight.

https://stackblitz.com/edit/github-fmckqvbh?file=src%2Fstyles%2Fcustom.css

### Participation

### Suggested labels

- 🐛 Bug

FrancoKaddour:
Fixes #3513

When --sl-content-width is set to a value larger than the default, the
two-column layout at 72rem viewport can overflow. The calc() for
.main-pane resolves to a value exceeding the container width because
.right-sidebar-container's calculated width goes negative (clamped to
0 by CSS), leaving no room for the sidebar.

Fix: use max(var(--sl-sidebar-width), calc(...)) for the sidebar
container to ensure it never drops below its intended width, and
min(calc(100% - var(--sl-sidebar-width)), calc(...)) for the main pane
to ensure it never exceeds available space.

### DIFF
--- a/.changeset/fix-toc-overflow-custom-width.md
+++ b/.changeset/fix-toc-overflow-custom-width.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix TOC overflowing the right edge of the viewport when a custom `--sl-content-width` value exceeds available space
--- a/.changeset/fix-toc-overflow-custom-width.md
+++ b/.changeset/fix-toc-overflow-custom-width.md
@@ -2,4 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
-Fix TOC overflowing the right edge of the viewport when a custom `--sl-content-width` value exceeds available space
+Fixes the table of contents overflowing the right edge of the viewport when a custom `--sl-content-width` value exceeds available space
--- a/packages/starlight/components/TwoColumnContent.astro
+++ b/packages/starlight/components/TwoColumnContent.astro
@@ -21,8 +21,9 @@
 			.right-sidebar-container {
 				order: 2;
 				position: relative;
-				width: calc(
-					var(--sl-sidebar-width) + (100% - var(--sl-content-width) - var(--sl-sidebar-width)) / 2
+				width: max(
+					var(--sl-sidebar-width),
+					calc(var(--sl-sidebar-width) + (100% - var(--sl-content-width) - var(--sl-sidebar-width)) / 2)
 				);
 			}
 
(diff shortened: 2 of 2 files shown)


.changeset/fix-toc-overflow-custom-width.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fix TOC overflowing the right edge of the viewport when a custom `--sl-content-width` value exceeds available space
```
delucis: Small nit on the change description:

```suggestion
Fixes the table of contents overflowing the right edge of the viewport when a custom `--sl-content-width` value exceeds available space
```

delucis (dismissed):
Thanks @FrancoKaddour — this change LGTM!

I tested various `--sl-content-width` values between 70 and 130rem on macOS in Firefox 152, Chrome 149, and Safari 26.2, and the fix worked well for me!

.changeset/fix-toc-overflow-custom-width.md
FrancoKaddour: Done, thanks!

delucis (approved):
Thanks again for the contribution @FrancoKaddour 🙌

PR #4009, opened 2026-06-29
docs: updates links validator and fix link

HiDeoo:
#### Description

This PR updates the links validator which includes a fix (and a regression test now) for validation of frontmatter-only pages with Sätteri. Turns out, in the meantime, we merged https://github.com/withastro/starlight/pull/3955 which introduced an invalid link matching this scenario, so this PR also fixes that invalid link.


delucis (approved):
Oh, good catch! Thanks @HiDeoo 🙌

PR #4011, opened 2026-06-29
feat: integrate zizmor

trueberryless:
#### Description

This PR adds [`zizmor`](https://zizmor.sh/) to make Starlight's GitHub actions more robust and follow best practice security recommendations.

I adapted all actions to comply with `zizmor`'s basic rules and added ignore comments in special cases where the static analyzer can not deduce the actual safety of the workflow correctly (like usages of `pull_request_target` triggers in two workflow files).

I also added a new `.github/workflows/zizmor.yml` action, which validates all actions to continue to follow best practices.

> [!WARNING]
> Additionally, I adapted a small logic in `.github/workflows/preview-release.yml` as well as `.github/workflows/file-icons.yml` by removing an external action that can be replaced with a simple gh CLI command (suggested by `zizmor` as well).


.github/workflows/ci.yml
```diff
@@ -15,11 +15,14 @@ concurrency:
...
   ASTRO_TELEMETRY_DISABLED: true
 
+permissions: {}
+
 jobs:
   changes:
     runs-on: ubuntu-latest
     permissions:
-      pull-requests: read
+      contents: read # checkout repository
```
HiDeoo: Are comments part of a `zizmor` rule?

.github/workflows/preview-release.yml
```diff
@@ -11,14 +10,17 @@ concurrency:
...
   preview:
     name: Preview Release
     if: ${{ github.repository_owner == 'withastro' && github.event.label.name == 'pr-preview' }}
     runs-on: ubuntu-latest
     permissions:
-      contents: read
-      pull-requests: write
+      contents: read # checkout repository
+      pull-requests: write # remove pr-preview label (gh CLI) and comment on PR with preview links (pkg-pr-new)
+      id-token: write # authenticate with npm using OIDC trusted publishing provenance
```
HiDeoo: I'm pretty sure this is wrong (and potentially dangerous) and may be a false positive matching `pnpm dlx pkg-pr-new publish` somehow (either a bug in Zizmor or just not properly handled case that should maybe be investigated and reported). `pkg.pr.new` is a dedicated registry different from the npm registry and does not use its credentials or use trusted publishing.

.github/workflows/preview-release.yml
```diff
@@ -71,9 +73,9 @@ jobs:
       # We remove the preview label before publishing so that the label is always removed even if
       # publishing fails, so we can retry by only adding the label again.
       - name: Remove Preview Label
-        uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0 # v1.3.0
-        with:
-          labels: pr-preview
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh pr edit ${{ github.event.pull_request.number }} --remove-label pr-preview
```
HiDeoo, line 77: I think there may be an undesired change of behavior by applying the `zizmor` suggestion here as-is. Before, when using `actions-ecosystem/action-remove-labels`, when failing to remove the label, the behavior was defined by the action input `fail_on_error` which defaults to `false`. This means that when failing to remove the label, e.g. for a fork, the job would continue and publish a preview release.

I think switching to the GitHub would make it return a non-zero value when failing which would fail the entire job immediately. Maybe we would need a [`continue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error) here.

.github/workflows/size-limit.yml
```diff
@@ -34,8 +40,13 @@ jobs:
     if: ${{ github.event.pull_request.head.repo.full_name == 'withastro/starlight' }}
     name: Check build output is within performance budget
     runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout repository
+      pull-requests: write # post the size comparison report as a PR comment (size-limit-action)
```
HiDeoo: Is this a new permission?

.github/workflows/zizmor.yml
HiDeoo, line 1: I wonder if this shouldn't run as part of the `ci` workflow and re-use our `changes` job output to only run when a workflow file is updated 🤔

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,33 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor 🌈
+    runs-on: ubuntu-latest
```
HiDeoo, line 19: I wonder if this could run on `ubuntu-slim`?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,33 @@
...
+    permissions:
+      contents: read # checkout repository
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor 🌈
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
```
HiDeoo: Looks like pinning the action does not pin the `zizmor` version used and the [`version` input](https://github.com/zizmorcore/zizmor-action#version) defaults to `latest`. Feels like we should pin its version too to avoid to potentially be randomly blocked on some updates and investigate if/how we could use renovate to update it.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          # Use annotations instead of SARIF as this doesn't need special permissions
+          annotations: true
```
HiDeoo: Any reason to use annotations rather than the `security-events: write` permission considering the limitations of annotations (and I guess it's the only reason why we set `advanced-security: false`?)

.github/zizmor.yml
HiDeoo, line 1: I wonder if we should rather disable some rules [inline](https://docs.zizmor.sh/usage/#with-comments) rather than entirely for a file. If we ever add new jobs or steps to the workflow where we entirely disabled some rules, we may miss some important issues?

I guess an alternative could be specifying which specific lines to ignore, but feels this may be a bit annoying to maintain.

.github/zizmor.yml
```diff
@@ -0,0 +1,10 @@
+# Existing privileged PR automation is intentionally isolated to these workflows:
```
HiDeoo: Using an inline approach would also allow us to have targeted comments as I feel like this one only cover the first ignored rule?

.github/zizmor.yml
```diff
@@ -0,0 +1,10 @@
...
+# they do not checkout pull request head code, and they only comment or set status.
+rules:
+  dangerous-triggers:
+    ignore:
+      - lunaria.yml
+      - welcome-bot.yml
+  secrets-inherit:
```
HiDeoo: Why was this ignored? Shouldn't we follow `zizmor` recommended remediation?

HiDeoo (changes requested):
Thanks for the PR, just left a first batch of comments and questions to address.

.github/workflows/ci.yml
trueberryless: Yes, the [audit rule "undocumented-permissions"](https://docs.zizmor.sh/audits/#undocumented-permissions) recommends to document the purpose of additional permissions.

Two things to note regarding this however (I am very glad you brought it up): 
1. This is a pedantic rule. With the current `zizmor.yml` setup in CI -- which runs in [`regular` persona per default](https://github.com/zizmorcore/zizmor-action#persona) -- this would technically not be enforced. Admittedly, I aimed for `pedantic` persona at first, but realized that it would [need concurrency rules in _every_ workflow](https://docs.zizmor.sh/audits/#concurrency-limits), which is `pedantic` persona only -- note that this is not mentioned in docs, only in [code](https://github.com/zizmorcore/zizmor/blob/3a46aaade8a6005c92e8f9dc43c34be560682022/crates/zizmor/src/audit/concurrency_limits.rs#L114) and [GitHub comments](https://github.com/zizmorcore/zizmor/issues/1225#issuecomment-3385868738) --, that's why [I lowered the persona](https://github.com/withastro/starlight/pull/4011/changes/0360032ea6382409b89d6b1d26c152face8fa456). After thinking about it again, I think it makes sense to use pedantic mode, which is why I revert the commit in https://github.com/withastro/starlight/pull/4011/changes/fab5e9e5d4aeef8cd2e31093b1e985cea9b6cbe7 and fixed the remaining warnings and notes in the following four commits (https://github.com/withastro/starlight/pull/4011/changes/31ae5b12385837305a4553072a6dd0116c4a6ac4, https://github.com/withastro/starlight/pull/4011/changes/97f90e0c0bc72ce095e281e52bf88b4cf9a21c47, https://github.com/withastro/starlight/pull/4011/changes/9d04e7f6782eb42357ea02e47e23b3b96712083f, https://github.com/withastro/starlight/pull/4011/changes/208f370faa7887d5820850c6b637257577259f31).
2. [As noted in the docs](https://docs.zizmor.sh/audits/#undocumented-permissions), `contents: read` does not need comments, as it is a "common, self-explanatory permission". I initially added it to be consistent with the other permissions, but since it doesn't add much value, I removed all the `contents: read` comments in https://github.com/withastro/starlight/pull/4011/changes/9dbca18780e0ef5d83c148bd699271a92ea6a220

.github/workflows/preview-release.yml
trueberryless: Great catch!

Alternatively we could also append `|| true` to "swallow" the error. But defining `continue-on-error` is more declarative and clear, so I updated it in https://github.com/withastro/starlight/pull/4011/changes/8c7960a3baad9b478386ffc8f7c3f67a52ecfabb 👍

.github/workflows/size-limit.yml
trueberryless: Ah yes it is. Sorry that was an oversight by me, it is not necessary because we use a custom token. I removed it in https://github.com/withastro/starlight/pull/4011/changes/4ec3b643aa453e56c7e3612d610564b014cd84d7

.github/workflows/zizmor.yml
trueberryless: I personally like to keep things separated as much as possible (separation of concerns). However, given that only 13.86% (70 out of 505 as of 14<sup>th</sup> Aug 2026) of last years merged PRs touch workflow files (analysed with [a little automated script](https://gist.github.com/trueberryless/665917af2aa45dd26ca2d4368b380a35)), I really like the idea of reusing the `changes` job. I didn't merge them yet because it makes the next review easier given a decent amount of existing comments in this new workflow. But I am open to combining them later 👍

.github/zizmor.yml
trueberryless: Thanks! I didn't realize it can be fixed that easily. Updated it in https://github.com/withastro/starlight/pull/4011/changes/0a28d565d5d6493055699a108cd2bef0499c2a49

.github/zizmor.yml
trueberryless: Oh nice! Inline comments are definitely better in our use case. Changed in https://github.com/withastro/starlight/pull/4011/changes/694174ef165109a7331cdd1e47c57fc882178bdb

.github/zizmor.yml
trueberryless: I tried to add suited but concise comments to both inline comments in https://github.com/withastro/starlight/pull/4011/changes/694174ef165109a7331cdd1e47c57fc882178bdb

Let me know if you wanna adjust the phrasing 👍

.github/workflows/pr-labeler.yml
```diff
@@ -1,12 +1,15 @@
 name: 'Pull Request Labeler'
+
 on:
   - pull_request_target
```
trueberryless, line 4: If anyone wonders why no zizmor ignore rule is necessary for this `pull_request_target`, this is because workflows, which **only** contain the `actions/labeler` step, are hard-coded in zizmor to not throw any error.

See https://github.com/zizmorcore/zizmor/pull/1956 for more information 🙌

.github/workflows/preview-release.yml
trueberryless: Oh yeah. I improved it here for now by adding a new inline ignore comment in https://github.com/withastro/starlight/pull/4011/changes/b64f066a6ba36f2b3ac450848abe509c1a5987f6

I'm not sure if Zizmor does any checks other than looking for `"publish"` somewhere in the npm/pnpm command, but I can look into it and report if necessary 👍

.github/workflows/zizmor.yml
trueberryless: That's a lovely optimization we could probably do if we would keep the action in the separated workflow file. I'll leave this idea open for now until we decide whether we want zizmor and `ci.yml` to be combined or separated.

.github/workflows/zizmor.yml
trueberryless: I pinned it to `1.29.0` (currently latest) for now in https://github.com/withastro/starlight/pull/4011/changes/fc0f5b61030f6b009d975a4ee55fc81b138ca303 👍 

Regarding automatic updates via Renovate, I assume that a [`customManager`](https://docs.renovatebot.com/configuration-options/#custommanagers) would be necessary with some tricky RegEx to extract the version string 🤔
Do you want me to do this in this PR or rather a later follow-up?

.github/workflows/zizmor.yml
trueberryless: > I guess it's the only reason why we set `advanced-security: false`?

Yes, that was the reason.

> Any reason to use annotations rather than the `security-events: write` permission considering the limitations of annotations?

Mostly because the action does not _fail_ with `--format=sarif`, even if it finds security leaks.

Additionally, I am also more used to annotations and never worked with security events + SARIF before. Also I couldn't find exactly, _which_ limitations annotations have. I guess ["Advanced Security"](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security) is the whole advantage of the recommended option and since we are OS, we can use it.

Therefore, I dropped the parameters and added the new permission in https://github.com/withastro/starlight/pull/4011/changes/459dbd063231bfeace52fdd4980d7c49ab09eb70, which means the action defaults back to use GitHub's Advanced Security feature. Ideally, we can setup some [Code scanning alerts](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#pull-request-check-failures-for-code-scanning-alerts) if Zizmor reports something. Otherwise, the repots could go unnoticed if I am not mistaken (since the action succeeds nonetheless).

trueberryless (commented):
Thanks for the first few comments. I addressed all ideas and questions -- splitting everything in small commits for easier review. I didn't close any threads yet tho, feel free to do so if you are satisfied.

.github/workflows/zizmor.yml
trueberryless: Actually, after the Zizmor workflow failed with these changes [here](https://github.com/withastro/starlight/actions/runs/31839008994/job/94891610621?pr=4011) and some good night's sleep, I realized that the action ships a [list of versions](https://github.com/zizmorcore/zizmor-action/blob/main/support/versions), which means that pinning the action actually means that `latest` always resolves to the same SHA of the Zizmor version itself, which makes pinning `zizmor` itself unnecessary. Therefore, I removed it again in https://github.com/withastro/starlight/pull/4011/changes/7fc34d78f5292e489bdcc53366bbdfeebe283475 because it is just additional overhead and would require custom RegEx if we want Renovate to update it.

Coincidentally, someone in the Zizmor Discord noted that it would be great to improve the changelog of the action -- ideally including the changelog of Zizmor itself in there whenever it gets updated:

image

This would mean that when Renovate updates the action for us, we also see the changelog for Zizmor 👍

HiDeoo:
Thanks for the update.

> > Are comments part of a `zizmor` rule?
>
> Yes, the [audit rule "undocumented-permissions"](https://docs.zizmor.sh/audits/#undocumented-permissions) recommends to document the purpose of additional permissions.
>
> Two things to note regarding this however (I am very glad you brought it up):
>
> 1. This is a pedantic rule. With the current `zizmor.yml` setup in CI -- which runs in [`regular` persona per default](https://github.com/zizmorcore/zizmor-action?rgh-link-date=2026-08-14T11%3A51%3A46Z#persona) -- this would technically not be enforced. Admittedly, I aimed for `pedantic` persona at first, but realized that it would [need concurrency rules in _every_ workflow](https://docs.zizmor.sh/audits/#concurrency-limits), which is `pedantic` persona only -- note that this is not mentioned in docs, only in [code](https://github.com/zizmorcore/zizmor/blob/3a46aaade8a6005c92e8f9dc43c34be560682022/crates/zizmor/src/audit/concurrency_limits.rs#L114) and [GitHub comments](https://github.com/zizmorcore/zizmor/issues/1225#issuecomment-3385868738) --, that's why [I lowered the persona](https://github.com/withastro/starlight/pull/4011/changes/0360032ea6382409b89d6b1d26c152face8fa456). After thinking about it again, I think it makes sense to use pedantic mode, which is why I revert the commit in [fab5e9e](https://github.com/withastro/starlight/commit/fab5e9e5d4aeef8cd2e31093b1e985cea9b6cbe7) and fixed the remaining warnings and notes in the following four commits ([31ae5b1](https://github.com/withastro/starlight/commit/31ae5b12385837305a4553072a6dd0116c4a6ac4), [97f90e0](https://github.com/withastro/starlight/commit/97f90e0c0bc72ce095e281e52bf88b4cf9a21c47), [9d04e7f](https://github.com/withastro/starlight/commit/9d04e7f6782eb42357ea02e47e23b3b96712083f), [208f370](https://github.com/withastro/starlight/commit/208f370faa7887d5820850c6b637257577259f31)).
> 2. [As noted in the docs](https://docs.zizmor.sh/audits/#undocumented-permissions), `contents: read` does not need comments, as it is a "common, self-explanatory permission". I initially added it to be consistent with the other permissions, but since it doesn't add much value, I removed all the `contents: read` comments in [9dbca18](https://github.com/withastro/starlight/commit/9dbca18780e0ef5d83c148bd699271a92ea6a220)

Well, at this stage, this was mostly only a question, and in the end, I think to get started, we should stick to the regular [persona](https://docs.zizmor.sh/usage/#using-personas):

> the user wants high-signal, low-noise, actionable security findings. This persona is best for ordinary local use as well as use in most CI/CD setups, which is why it's the default.

I think it's a good persona to start with, limit the number of changes in this first iteration, make sure all changes are working properly, and set us up properly if we ever want to switch to a more sensitive persona in the future.

It also looks like since my first review, new unrelated (at least to me) changes have been made to the PR, e.g. [this commit](https://github.com/withastro/starlight/pull/4011/changes/9d04e7f6782eb42357ea02e47e23b3b96712083f) removing the `peter-evans/create-pull-request` action (and I don't know if it's the only one or not), and this has been done without any mention on why (at least I can't find it if it's mentioned somewhere). This is a bit concerning and makes the review a bit more difficult as I don't know why it's been removed, e.g. if that was forgotten in the initial PR, if this is related to the change of persona, is it even reported by Zizmor and if so why wasn't it reported before, etc.

Considering all this, I think we should switch back to the regular persona and version I initially reviewed, address the first wave of feedback, try to minimize the number of changes, see where this gets us, and try to get a first version of the PR merged and if ever needed/decided, iterate on it.

trueberryless:
Good point! I was focused on fixing all additional issues which were flagged by the pedantic persona that I didn't even consider the necessity of it. My bad!

Regarding my commits which were not brought up by you: They were all brought up by Zizmor because I switched to the pedantic persona. In that sense they can seem unrelated, yeah.

What do you propose is the easiest way to review further changes? Would you rather want me to revert all changes regarding the persona change? Or completely rebase and force push, so you there are less commits for you to review (I'm personally opposed to this)? 

The first wave of feedback has already been addressed. I just didn't resolve the threads, as I already wrote [here](https://github.com/withastro/starlight/pull/4011#pullrequestreview-4936857496). The only thing left to decide is whether to merge the workflow into the `ci.yml` or keep it separated (see [this thread](https://github.com/withastro/starlight/pull/4011#discussion_r3784672830)). This is also why I didn't commit any changes regarding the [runner specs](https://github.com/withastro/starlight/pull/4011#discussion_r3710483951) yet. I'm just waiting on your thoughts regarding this.

HiDeoo:
> What do you propose is the easiest way to review further changes? Would you rather want me to revert all changes regarding the persona change?

Yeah, just push w/e needed to bring us back to the previous version as I mentioned and I'll take a look at the changes that address the previous wave of feedback while re-taking a complete look 👍

trueberryless:
Alright, I reverted and removed some changes so we have a smaller diff. Nonetheless, I'd like to point out that I would still be in favor of readding these two commits to the changes:

1. The documentation of permissions except `content: read`, which helps to reason about permissions a lot. This was now removed in https://github.com/withastro/starlight/pull/4011/changes/2cc3c28e82ba54c525133817ee72d43fa2ac9daf
2. Nitpicky, but there are only two jobs in all of our workflows, which do not have names. The fix was [included previously](https://github.com/withastro/starlight/pull/4011/changes/31ae5b12385837305a4553072a6dd0116c4a6ac4), but was also removed now: https://github.com/withastro/starlight/pull/4011/changes/0a170259d3f8579dacd176d788f5382a3cc3ac77

.github/workflows/ci.yml
```diff
@@ -9,16 +9,19 @@ on:
 
 # Automatically cancel in-progress actions on the same branch
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
```
HiDeoo: I think this change may be wrong.

Before, `github.event_name == 'pull_request_target'` was never `true` so `github.ref` was always used. With such change, it can now be `true`, and in such case, `github.head_ref` will now be used and it's just the branch name. This means that 2 forks using the same branch name (e.g. the `patch-something` created by GitHub) would collide. I think in such case, we should use `group: ${{ github.workflow }}-${{ github.ref }}` just like some other workflows.

This led me down a rabbit hole and review all our concurrency group:

- `preview-release.yml`: has the same issue
- `size-limit.yml` and `zizmor.yml`: have the dead condition with `pull_request_target` but never use `pull_request_target`

Given this, I think all our existing concurrency group should just use `group: ${{ github.workflow }}-${{ github.ref }}`.

.github/workflows/preview-release.yml
```diff
@@ -71,14 +72,15 @@ jobs:
...
+        continue-on-error: true
```
HiDeoo: Maybe we should add a small comment why this is needed.

.github/workflows/zizmor.yml
HiDeoo: I think we can do it now.

.github/workflows/zizmor.yml
HiDeoo: > if we would keep the action in the separated workflow file

Why can't we do that even if we don't keep the action in a separate workflow file?

HiDeoo (changes requested):
Thanks for the update, I left a new batch of comments and follow-ups :raised_hands:

.github/workflows/zizmor.yml
HiDeoo: While technically valid, I'm not sure I totally agree.

If I need to find the current Zizmor version being used on the repo, e.g. let's say to run the exact same version locally, do I need to know about this specific behavior, find the correct version branch on the repo and check that version list? If I need to update the action, e.g. a patch fixing an issue, and that same patch update Zizmor too (looks like the version can change between patch of the action) and for some reason I don't want to update Zizmor. Also, like you said, the action changelogs are not the best yet.

I think specifying the version of Zizmor in the workflow itself makes it the version self-documented, easier to track and makes it easy to pin to a specific version without relying on the action.

Any thoughts?

.github/workflows/ci.yml
trueberryless: Great find! For some reason, I assumed that `head_ref` includes the fork owner -- which would make it unique -- but thanks for the correction and detailed summary of the 🐰🕳️

.github/workflows/zizmor.yml
trueberryless: Oh wow. Why did I think that this is only specified once for all steps 👀 

Anyway, [it seems](https://github.com/withastro/starlight/actions/runs/32747899011/job/97497840796?pr=4011) that the docker daemon isn't running on `ubuntu-slim`, so we might need `ubuntu-latest` nonetheless.

trueberryless:
Thanks for the awesome review, HiDeoo!

Unfortunately, I'm not sure if our `ubuntu-slim` optimisation is working out. The docker daemon isn't running and IMHO it would be too much workaround just to make it work with `ubuntu-slim`. Wdyt? Should we switch back to `ubuntu-latest`?

HiDeoo:
> Unfortunately, I'm not sure if our `ubuntu-slim` optimisation is working out. The docker daemon isn't running and IMHO it would be too much workaround just to make it work with `ubuntu-slim`. Wdyt? Should we switch back to `ubuntu-latest`?

Ah, definitely, this was worth trying even if it didn't end up working in the end. Thanks for trying it out.

I guess this leaves us with this, [this comment](https://github.com/withastro/starlight/pull/4011#discussion_r3842184878), and we should be good. 

Note that I didn't review the latest changes, I'll do a final review at the end 👍

.github/workflows/zizmor.yml
trueberryless: Valid arguments to include the version pin. What could be tricky is setting up Renovate to keep the versions updated + making sure that the action version is available in the list -- otherwise the action would fail as we saw [here](https://github.com/withastro/starlight/actions/runs/31839008994/job/94891610621?pr=4011).

I'm gonna include the SHA pin for now and we can figure out how to utilize Renovate here later 👍

trueberryless:
Amazing! Thanks for the guidance and thoughts!

.github/workflows/format.yml
```diff
@@ -5,14 +5,20 @@ on:
...
   format:
     if: github.repository_owner == 'withastro'
+    permissions:
+      contents: write
     uses: withastro/automation/.github/workflows/format.yml@8a5969576beae8ae7bdce64167a95237878f4901 # v1.0.1
     with:
       command: "format"
-    secrets: inherit
+    secrets:
+      FREDKBOT_GITHUB_TOKEN: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
```
HiDeoo, line 24: Re-reading the action documentation, I think this change would need an upstream change too.

If I'm reading correctly, to explicitly pass one or more secrets, they need to be defined just like inputs using [`on.workflow_call.secrets`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callsecrets).

What we could do is declare it as non-required, meaning we could use it like this, and other workflows using the `inherit` approach wouldn't break.

I guess we could even name the secret `token` and use it like `token: ${{ secrets.token || secrets.FREDKBOT_GITHUB_TOKEN }}`.

Something to discuss and make sure I'm 100% correct on this, but would you be up to investigate and if possible, make the change upstream?

HiDeoo (changes requested):
Ok, left what I think would be my last comment.

Regarding Renovate and the Zizmor version, thinking about it, I don't think we would need to do anything. Renovate updates only our GH Actions, so if the defined Zizmor version is dropped during an update, the action will fail in the Renovate PR and we can investigate, fix it, bump if needed, etc. 

We also check changelogs of all GH Actions Renovate PRs so if we see a version is dropped, it's easy to see if it's the same as the version number 2 lines below.

.github/workflows/format.yml
trueberryless: Great find, I think this line in the docs [you linked to](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callsecrets) confirms it:

> If a caller workflow passes a secret that is not specified in the called workflow, this results in an error.

<details>
<summary>
Further evidence can be found in these parts of the GitHub docs
</summary>

- [This section](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow) describes exactly what you suspect.
- `secrets: inherit` is special because it allows references without explicit definitions in the `on` block, as explicitly stated in [the information note under step 2 here](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow):

  > If the secrets are inherited by using `secrets: inherit` in the calling workflow, you can reference them even if they are not explicitly defined in the `on` key. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idsecretsinherit).
  
  This link then also provides [this example](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-of-jobsjob_idsecretsinherit) for the special `inherit` case.
  
- [This reusable workflow example](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callsecrets) together with [this caller workflow example](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#example-caller-workflow) also showcase the necessity of defining the key in the reusable workflow.

</details>

So to sum up: Yes, this requires an upstream change and I can take care of it. I couldn't find an explicit explanation or example in the documentation of what will happen if we define the secret in our reusable workflow and then use it with `secrets: inherit`. However, given the existing docs, they strongly imply that such a scenario will perfectly work as we suspect.

The only thing I would like to discuss before adapting the code is your idea of making the secret input `non-required`. I think that we actually have to make it a required input, as the action would otherwise fail if the `FREDKBOT_GITHUB_TOKEN` is not passed. [Reusable workflows run in the context of the caller workflow](https://docs.github.com/en/actions/concepts/workflows-and-actions/reusing-workflow-configurations#:~:text=If%20you%20reuse%20a%20workflow%20from%20a,not%20the%20called%20workflow.), meaning that even if this token is defined for the whole `withastro` organisation, it will not be available in the reusable format workflow if we do not pass it and the `action/checkout` will fail because the token is an empty string.

I strongly _believe_ that defining this secret input as required would not be a breaking change because the [five actions using it](https://grep.app/search?q=withastro%2Fautomation%2F.github%2Fworkflows%2Fformat.yml) would otherwise never have succeeded. However, I can not confirm if the `FREDKBOT_GITHUB_TOKEN` is really defined for all those repos, which is why I would need your approval.

HiDeoo: Ah, if I understand correctly, you mean declaring a `FREDKBOT_GITHUB_TOKEN` required secret in the shared workflow. Yes, in this case, I think that would work an be non-breaking afaik.

My idea was mostly to avoid having a defined secret named `FREDKBOT_GITHUB_TOKEN` in the shared workflow, and instead have a new more generic (no Fred in it) and non-required secret named e.g. `token` that would be used with `token: ${{ secrets.token || secrets.FREDKBOT_GITHUB_TOKEN }}`.

- If a new caller specifys `token`, it will use that value.
- If an old caller does not specify the `token` secret, `secrets.token` will be empty and it will fall back to `secrets.FREDKBOT_GITHUB_TOKEN`.
  - If they use `secrets: inherit` like before, `secrets.FREDKBOT_GITHUB_TOKEN` will be available as inherited secrets can be referenced without being declared, and basically behave the same as before.
  - If they don't use `secrets: inherit`, then `secrets.FREDKBOT_GITHUB_TOKEN` will also be unavailable and the workflow will fail like it did before.

Altho, considering we should be the only one using these workflows and we advertise them as internal, I guess your approach should be fine :thumbsup:

trueberryless: Interesting. Naming the input something different is also a great call -- I didn't consider that option. However, I would implement my suggestion nonetheless for two reasons:

- It keeps the logic simple. Having two possible token keys with fallbacks in a specific order could be confusing and hard to debug.
- I don't fully see any big benefit because we wouldn't get rid of the `FREDKBOT_GITHUB_TOKEN` anyways. I don't know if we'll ever plan to rename the token to something more generic (especially because Matthew is steward now), but IMHO this is definitely out of scope for now and [would require bigger adaptions across our repos](https://grep.app/search?q=FREDKBOT_GITHUB_TOKEN).

So thanks for the confirmation, I'll open a PR with a required input named `FREDKBOT_GITHUB_TOKEN` in the automation repo 👍

HiDeoo: Perfect, thanks.

And to avoid any confusion, my suggestion was not about removing `FREDKBOT_GITHUB_TOKEN` from anywhere, just to make calling the workflow with an explicit token with a more obvious/generic secret name:

```yml
uses: withastro/automation/.github/workflows/format.yml@8a5969576beae8ae7bdce64167a95237878f4901 # v1.0.1
with:
  command: "format"
secrets:
  token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
```

Rather than:

```yml
uses: withastro/automation/.github/workflows/format.yml@8a5969576beae8ae7bdce64167a95237878f4901 # v1.0.1
with:
  command: "format"
secrets:
  FREDKBOT_GITHUB_TOKEN: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
```

trueberryless: Amazing 🥳 

I created [this PR](https://github.com/withastro/automation/pull/81).

trueberryless:
I just pushed two commits to update the PR with the latest changes from `main` as I noticed the new `benchmark.yml` workflow from last week. I adapted it to fit our changes here. Should I also change the concurrency group to `${{ github.workflow }}-${{ github.ref }}` so it uses the same as every other workflow?

.github/workflows/format.yml
trueberryless: The PR has been merged, so I am going to mark this as resolved 👍

HiDeoo: Thanks, just coming back to this PR, small question: don't we need to switch to the version 1.3.0 of the shared workflow too for this change to work?

HiDeoo:
Update: we just [setup and tested](https://github.com/withastro/docs/pull/14597) a ruleset in the Docs repo to prevent merging on error for an equivalent Zizmor action using SARIF.

Once this PR is merged, we'll import or re-create an equivalent ruleset in this repository (I added a TODO to the PR body too).

.github/workflows/format.yml
trueberryless: Thanks for the reminder, I'd have totally forgotten that  🙄

trueberryless:
> Update: we just [setup and tested](https://github.com/withastro/docs/pull/14597) a ruleset in the Docs repo to prevent merging on error for an equivalent Zizmor action using SARIF.
> 
> Once this PR is merged, we'll import or re-create an equivalent ruleset in this repository (I added a TODO to the PR body too).

Amazing. Looking forward to having more secure workflows and ruleset enabled in Starlight as well 🚀

PR #4013, opened 2026-06-30
Update playwright

HiDeoo:
#### Description

This PR updates the Playwright version used in the monorepo to workaround a browser installation issue on Node.js 24 and Node.js 26 (reported [on Discord](https://discord.com/channels/830184174198718474/1070481941863878697/1521516402396102719))

Basically, [looks like](https://github.com/microsoft/playwright/issues/40724#issuecomment-4514503775) Node.js 24.16.0 bumped libuv which surfaced this [early-exit issue](https://github.com/thejoshwolfe/yauzl/pull/168) which was fixed in https://github.com/microsoft/playwright/pull/40747.

I confirmed the issue by running locally `npx playwright uninstall --all` and then `pnpm test:e2e` on Node.js 24.17.0 which hang indefinitely. After this change and running both commands again, the installation and tests run successfully.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -26,7 +26,7 @@
     "sharp": "^0.34.5"
   },
   "devDependencies": {
-    "@playwright/test": "^1.59.1",
+    "@playwright/test": "^1.61.1",
     "axe-playwright": "^2.2.2",
     "sitemapper": "^4.1.5",
     "starlight-links-validator": "^0.25.2"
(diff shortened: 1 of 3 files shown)


delucis (approved):
Awesome find @HiDeoo! Just tried this branch locally with Node 24 and it successfully installed Playwright and ran tests as expected 🥳

PR #4014, opened 2026-06-30
Allow `Page` component to be overridden

delucis:
#### Description

- This PR is a preview of officially supporting “headless” use of Starlight where users can override the entire HTML rendering step with a custom Astro component that consumes Starlight’s data layer.
- Sharing it to gather early feedback.
- Docs preview: https://deploy-preview-4014--astro-starlight.netlify.app/guides/overriding-components/#override-the-whole-page
- Preview release:
  ```sh
  pnpm add https://pkg.pr.new/withastro/starlight/@astrojs/starlight@4014
  ```
- Really basic demo project: https://stackblitz.com/edit/github-peilyxr6?file=astro.config.mjs,src%2FMyPage.astro&on=stackblitz

#### Context

When we added the overrides feature in https://github.com/withastro/starlight/pull/709 we avoided adding this as it can be a “big hammer” to take over the entirety of page rendering including `<head>` content etc. We reckoned that a big reason for picking Starlight was to get a well tested, accessible, battle hardened and SEO-ready frontend, so completely replacing it might not be something people would need. And due to some of the complexities of a wholesale override we opted not to consider it at that time.

For most users that remains the case, but we have seen adoption in some more complex sites where Starlight’s data and routing layers are attractive foundations but where users want full control over the final rendering step for various reasons. Officially supporting this would give them that.

There may be some details to iron out. For example, user components (like `<Aside>`, `<Tabs>` etc.) should ideally work even when “ejecting” to a custom page component. We’ll likely need to tweak some CSS imports to ensure Starlight’s base layer of CSS variables is available to these (currently these components assume the CSS base layer will be available via the page component).

I’ve added some docs for now, but we can also discuss if the extra section in the “Overriding components” guide is necessary — the override process for the full page is identical to other components, just with a bigger impact around what it does.

### DIFF
--- a/.changeset/loose-pugs-grow.md
+++ b/.changeset/loose-pugs-grow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': minor
+---
+
+Adds support for overriding Starlight’s `Page` component
--- a/docs/src/content/docs/guides/overriding-components.mdx
+++ b/docs/src/content/docs/guides/overriding-components.mdx
@@ -150,3 +150,47 @@ const isHomepage = Astro.locals.starlightRoute.id === '';
 ```
 
 Learn more about conditional rendering in [Astro’s Template Syntax guide](https://docs.astro.build/en/basics/astro-syntax/#dynamic-html).
+
+## Override the whole page
+
+You can completely replace Starlight’s UI and bring your own frontend implementation by overriding the `Page` component:
+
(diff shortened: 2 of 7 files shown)


delucis:
We haven’t really received the feedback on this that I hoped we might, so I’m going to close this for now.

If we hear more requests for this feature we can revisit.

PR #4015, opened 2026-06-30
Fix aside icons when MDX optimization is disabled

delucis:
#### Description

- Fixes https://github.com/withastro/starlight/pull/3967#issuecomment-4833368829
- We were relying on a `type: "html"` AST node for icons in our Markdown aside plugin (which provides support for our `:::note` type syntax). This happens to work when the MDX integration is configured with `optimize: true` (perhaps accidentally based on chatting to @Princesseuh). Starlight enables optimization by default, but if a user manually adds `mdx()`, optimization can be disabled, breaking this behaviour.
- The fix is to use a `mdxJsxTextElement` AST node when processing MDX files that represents `<Fragment set:html={icon}>`.
- Sätteri doesn’t currently provide a way to know if a plugin is running against an MDX file or a regular Markdown file, so for now this PR uses a basic check of the file’s extension. Not sure if we’re happy to publish with this to get the bug fixed or wait on an official flag.

### DIFF
--- a/.changeset/tricky-dolls-cover.md
+++ b/.changeset/tricky-dolls-cover.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes an issue where aside icons were rendered incorrectly in projects where Astro’s MDX integration had optimization disabled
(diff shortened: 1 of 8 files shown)


HiDeoo (approved):
Thanks for following up on this :raised_hands:

The change makes sense to me and thoroughly tested, ready to be shipped imo 🚀

PR #4007, opened 2026-06-29
Add Starlight MD3 theme listing

Axiaobo7788:
#### Description

- Closes # 
- What does this PR change? Give us a brief description.
- Did you change something visual? A before/after screenshot can be helpful.

Adds Starlight MD3 to the community themes list.

This PR adds:

- a new theme entry for Starlight MD3 in `docs/src/content/docs/resources/themes.mdx`
- light and dark preview images for the theme
- a link to the live demo site

Theme demo:
https://axiaobo7788.github.io/starlight-material-design-theme/

Theme source:
https://github.com/aXiaobo7788/starlight-material-design-theme

npm package:
https://www.npmjs.com/package/starlight-theme-md3

This is a docs/resources update only and does not change Starlight runtime behavior.

#### Testing

Not run in this fork. The change only adds a community theme listing and static preview assets.

The linked theme package was separately built, packaged, and published as `starlight-theme-md3@0.1.3`.

PR #4019, opened 2026-07-01
i18n(fr): update `resources/themes.mdx`

ArmandPhilippot:
#### Description

Adds changes from #4007 to the French translation of `resources/themes.mdx`.

Note that "tonal" is more of a term associated with music in French (see [Le Robert](https://dictionnaire.lerobert.com/definition/tonal) for example). So, I think "contrasté" might be more suited here.
Also, I wasn't sure for "state layers"/"couches d'état" but it seems to be a concept from [Material](https://m3.material.io/foundations/interaction/states/state-layers), so I think it might be fine.


HiDeoo (approved):
Thanks for the thorough research on how to properly translate this description (as always) 🙌 

Makes perfect sense to me 🚀

PR #4029, opened 2026-07-02
fix(a11y): trap focus within mobile sidebar when menu is open

### LINKED ISSUE #2697: mobile menu does not contain focus to the menu on tab navigation

FrancoKaddour:
Tab and Shift+Tab currently escape the mobile sidebar and cycle through the rest of the page — this makes the menu unusable for keyboard-only users.

Added a handleKeyDown listener in StarlightMenuButton that intercepts Tab/Shift+Tab when the menu is expanded and loops focus between the toggle button and all focusable elements inside #starlight__sidebar. A getComputedStyle check skips the trap on desktop viewports where the toggle button is hidden and the sidebar is always visible.

Closes #2697

### DIFF
--- a/.changeset/fix-mobile-menu-focus-trap.md
+++ b/.changeset/fix-mobile-menu-focus-trap.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Trap focus within the mobile sidebar when the menu is open. Pressing Tab or Shift+Tab now cycles through the toggle button and all sidebar links without escaping to the rest of the page.
--- a/packages/starlight/components/MobileMenuToggle.astro
+++ b/packages/starlight/components/MobileMenuToggle.astro
@@ -28,6 +28,9 @@ import Icon from '../user-components/Icon.astro';
 			if (parentNav) {
 				parentNav.addEventListener('keyup', (e) => this.closeOnEscape(e));
 			}
+
+			// Trap focus within the sidebar when the mobile menu is open.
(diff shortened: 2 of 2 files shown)


HiDeoo:

Looks like this may be a duplicate of #3911 which uses a different approach with the [`inert` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) to achieve the same goal.

The other PR is waiting final changes, tweaks, and a bit more user testing before it can be merged. It uses an approach that may be a bit more lightweight so I think we should close this one in favor of that one. If you want to take a look at that PR and provide feedback, feel free to do so! Thanks again for your contribution.

FrancoKaddour:
Makes sense, thanks for the context. I'll take a look at #3911 — the inert approach is cleaner.

PR #4035, opened 2026-07-02
docs: format plugin name to match conventions

pixelizing:
#### Description

- Closes # 


delucis:
Thanks for the PR @pixelizing, but I actually think it’s fine as is! We don’t have a standard rule. Most of the other things here use the hyphenated style because that’s the name of a package on npm, but it’s not a requirement or anything. So I think the more readable current version is good given your tool is not an npm package.

PR #4040, opened 2026-07-04
fix: doc schema extend

HiDeoo:
#### Description

Temporary draft PR, **do not merge**.

See https://github.com/withastro/starlight/pull/3936#issuecomment-4882068195 for context and conversations.

### DIFF
--- a/.changeset/all-badgers-grin.md
+++ b/.changeset/all-badgers-grin.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Replace "and" in "docsSchema" with "extend" to support the modifying of ZodEnum.
--- a/packages/starlight/__tests__/basics/schema.test-d.ts
+++ b/packages/starlight/__tests__/basics/schema.test-d.ts
@@ -1,0 +1,72 @@
+import { docsSchema } from '@astrojs/starlight/schema';
+import { z } from 'astro/zod';
+import { describe, expectTypeOf, test } from 'vitest';
+
+interface PartialDocsSchemaOutput {
+	title: string;
+	description?: string | undefined;
+	template: 'doc' | 'splash';
+	sidebar: {
+		hidden: boolean;
+	};
+}
+
+describe('docs schema', () => {
+	test('docs schema should parse to expected shape', () => {
+		const schema = docsSchema()({ image: () => ({}) as any });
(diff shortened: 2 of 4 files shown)


HiDeoo:
I spent many hours this week trying to find a more "built-in" approach to this solution, even dug into the Zod codebase, but in the end, I personally didn't find a more elegant way to handle this.

Considering this, I'm opening up this PR for review to see if anyone has any other ideas, discuss if we want to use this approach, etc. (but we still don't want to merge this PR if we go with this approach).

delucis:
Thanks @HiDeoo! I also opened https://github.com/colinhacks/zod/discussions/6191 in case someone might have ideas/suggestions, but I have the feeling Zod is a) massively used and gets a lot of incoming questions and b) not super actively maintained, so not sure if we’ll ever get a response.

I’m torn — on the one hand this fixes the issue, but on the other it’s a headache to maintain. It might be good to figure out exactly which things it’s impossible to extend currently and decide if that behaviour is acceptable or not? There’s one side of me that thinks “`extend` is intended mainly to add properties or make them required, not to override things like built-in enums” (after all, the intent of the schema is to ensure user frontmatter produces data compatible with Starlight’s code and components, and adding e.g. a new value to the `variant` enum breaks that contract) but on the other hand I can understand the utility to a user of extending that when implementing CTAs that support more variants.

HiDeoo:
I'm definitely also torn and that'a also why I tried to find a better solution. Like you said, it's annoying to maintain, I'm pretty sure it's not 100% perfect and we could probably get issues about edge cases in the future (altho I would expect them to be minimal).

Considering the use case is also understandable (while still niche), I wonder if I'm not tempted to say "let's try it" as a fix for `extend`, and if it ends up being a problem, we can restrict the use of `extend` in a v1 or major?

But if someone has a strong opinion about this, stronger than me, I think I would definitely be to go along with it.

delucis:
Yeah, I’m OK with that. For code sanity it might be good to move the new code into a utility file so that `schema.ts` is still primarily concerned with the schema and not Zod internals but I can comment that on #3936.

HiDeoo:
Perfect. Considering this, I'll now close this PR for now (and we still have it in GH if ever needed for reference or something).

delucis:
Thanks again for figuring this out!

PR #4042, opened 2026-07-04
[i18nIgnore] Removed Broken Link

SavPhill:
Removed broken link to ⁠ecoping.earth⁠. The domain appears to be expired/inactive, which poses a potential security risk for documentation users.


delucis (approved):
Thanks @SavPhill!

(For reference this was their website when the link was added: https://web.archive.org/web/20250303002455/https://ecoping.earth/)

SavPhill:
> Thanks @SavPhill!
> 
> (For reference this was their website when the link was added: https://web.archive.org/web/20250303002455/https://ecoping.earth/)

Thanks @delucis !

I also considered replacing with the web archive link, but chose to remove it. 

If you think it's better to replace the link with an archived copy in future let me know and I'll do so instead.

delucis:
> If you think it's better to replace the link with an archived copy in future let me know and I'll do so instead.

No, given the service itself seems to be defunct, I think removing it was the right call — just wanted to share the archive link here for future reference for anyone curious.

PR #4045, opened 2026-07-05
docs: add starlight-biel to community plugins

dgarcia360:
Adds [starlight-biel](https://github.com/TechDocsStudio/starlight-biel) to the community plugins list: a plugin that adds an Ask AI chatbot answering from your site content, with citations to the source pages. Published on npm as [`starlight-biel`](https://www.npmjs.com/package/starlight-biel) with the `starlight-plugin` keyword. Added at the end of the list, matching the existing card format.

PR #4039, opened 2026-07-04
docs: add Starlight mdBook to community themes

2u841r:
## Summary
- Adds `starlight-theme-mdbook` to the community themes showcase page, following the steps in [CONTRIBUTING.md](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#themes).
- Theme: a Starlight plugin that recreates rust-lang's [mdBook](https://github.com/rust-lang/mdBook) look and feel (layout, chrome, and mdBook's built-in color schemes).
- Repo: https://github.com/2u841r/starlight-theme-mdbook
- Live demo: https://starlight-theme-mdbook.netlify.app/

## Test plan
- [x] Screenshots (`mdbook-light.png`, `mdbook-dark.png`) added to `docs/src/assets/themes/`, PNG, 1280x720
- [x] New entry appended at the end of the `themes` list in `docs/src/content/docs/resources/themes.mdx` with `title`, `description`, `href`, `previews`

***

My discussion / request 9 months ago - #3499

PR #4048, opened 2026-07-07
i18n(ko-KR): update `resources/`

jsparkdev:
#### Description

- #4045 
- #4039

I translated **"chrome"** as **"UI"** because I think it's easier for Korean readers to understand while still conveying the intended meaning in this context.


HiDeoo (approved):
Thanks for the new update :raised_hands:

As a non-native speaker, I reviewed this PR using the [Docs Translation Review Checklist](https://contribute.docs.astro.build/reviewers/reviewing-translations/) and everything looks good to me (thanks for the extra context in the PR body) 👍

PR #4049, opened 2026-07-07
i18n(fr): update `plugins.mdx` and `themes.mdx`

ArmandPhilippot:
#### Description

Adds changes from #4045 and #4039 to the French translation of `resources/plugins.mdx` and `resources/themes.mdx`.

I'm not 100% sure of "son aspect chromé" for "[with its layout,] chrome[, and]". I mean, with the other colors I don't really see any chrome so it's a bit odd... But, I'm sure this should not be "son chrome" in French as this is a [metal name](https://www.larousse.fr/dictionnaires/francais/chrome/15794).

### DIFF
--- a/docs/src/content/docs/fr/resources/themes.mdx
+++ b/docs/src/content/docs/fr/resources/themes.mdx
@@ -158,7 +158,7 @@ Découvrez ci-dessous un aperçu de tous les thèmes ou essayez-les de manière 
 		{
 			title: 'Starlight mdBook',
 			description:
-				"Un thème qui reproduit l'apparence et l'ergonomie de mdBook de rust-lang, avec sa mise en page, son aspect chromé et ses cinq jeux de couleurs intégrés.",
+				"Un thème qui reproduit l'apparence et l'ergonomie de mdBook de rust-lang, avec sa mise en page, son interface et ses cinq jeux de couleurs intégrés.",
 			href: 'https://starlight-theme-mdbook.netlify.app/',
 			previews: { light: 'mdbook-light.png', dark: 'mdbook-dark.png' },
 		},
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/fr/resources/themes.mdx
```diff
@@ -155,6 +155,13 @@ Découvrez ci-dessous un aperçu de tous les thèmes ou essayez-les de manière
 			href: 'https://axiaobo7788.github.io/starlight-material-design-theme/',
 			previews: { light: 'md3-light.png', dark: 'md3-dark.png' },
 		},
+		{
+			title: 'Starlight mdBook',
+			description:
+				"Un thème qui reproduit l'apparence et l'ergonomie de mdBook de rust-lang, avec sa mise en page, son aspect chromé et ses cinq jeux de couleurs intégrés.",
```
HiDeoo:
```suggestion
				"Un thème qui reproduit l'apparence et l'ergonomie de mdBook de rust-lang, avec sa mise en page, son interface et ses cinq jeux de couleurs intégrés.",
```

I think in this case it's just the UI, not the color/aspect:

- https://www.nngroup.com/articles/browser-and-gui-chrome/
- https://developer.mozilla.org/en-US/docs/Glossary/Chrome

ArmandPhilippot: Oh, right, I totally forgot about this meaning. 🤦🏽 I tried to search for chrome to get some ideas but... my search was a bit naive. (`-google` also removes those pages 😅 ).

HiDeoo: Yeah, I'm pretty sure I never use this meaning too, so confusing 😅

HiDeoo (approved):
Thanks for the quick update 🌟

PR #3996, opened 2026-06-26
i18n(de): update plugins.mdx

trueberryless:
#### Description

This PR updates the German translation of `resources/plugins.mdx` with the changes from #3964.

PR #4057, opened 2026-07-10
docs: move `starlight-md-txt` to plugins

trueberryless:
#### Description

This PR moves the `starlight-md-txt` from the "Community tools" to the "Community plugins" section because it is a Starlight plugin. I think it was added to the wrong section because of a little misinterpretation of ["end of the list"](https://github.com/withastro/starlight/pull/3964#pullrequestreview-4522472586) in #3964. 

> Errare humanum est - Seneca the Younger[^1]

I discovered this coincidentally while I wanted to update the German translation because funnily enough, in #3996 I accidentally added the entry to the "Community plugins" in the DE version by mistake 😆 
Well, if that isn't what you'd call a blessing in disguise... 😅 

[^1]: https://en.wiktionary.org/wiki/errare_humanum_est

### DIFF
--- a/docs/src/content/docs/resources/plugins.mdx
+++ b/docs/src/content/docs/resources/plugins.mdx
@@ -224,6 +224,11 @@ Extend your site with official plugins supported by the Starlight team and commu
 		description="Starlight plugin to create a recipe website."
 	/>
 	<LinkCard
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks for the fix!

PR #4055, opened 2026-07-10
docs: add starlight-page-context-action to community plugins

babblebey:
This pull request adds a new plugin to the list of community plugins. The new plugin, `starlight-page-context-action`, provides page action buttons for documentation pages, offering utility actions like copy/view as markdown, open in AI chat, and scroll to top in an accessible position.

Package - https://www.npmjs.com/package/starlight-page-context-action
Repository - https://github.com/babblebey/starlight-page-context-action
Consumption (Plugin Usage) - https://semantic-release.org/

PR #4062, opened 2026-07-11

jsparkdev:
#### Description

- #4055
- #4057


docs/src/content/docs/ko/resources/plugins.mdx
```diff
@@ -223,11 +223,21 @@ Starlight 팀이 지원하는 공식 플러그인과 Starlight 사용자가 관
...
+	/>
 	<LinkCard
 		href="https://github.com/TechDocsStudio/starlight-biel"
 		title="starlight-biel"
 		description="Biel.ai 기반의 콘텐츠를 바탕으로 답변하는 Ask AI 챗봇을 추가하세요."
 	/>
+	<LinkCard
+		href="https://github.com/babblebey/starlight-page-context-action"
+		title="starlight-page-context-action"
+		description="페이지를 Markdown으로 복사/보기, AI 채팅에서 열기, 위로 스크롤하기와 같은 유틸리티 기능들을 문서에서 항상 사용할 수 있도록 만드세요."
```
trueberryless, line 239: I am not a Korean native speaker, but is `"document"` and `"documentation"` the same in Korean? Because I noticed that when translating this back to English, it means "document", which is also a common gotcha in the German translation (`"Dokument"` vs `"Dokumentation"`).

I'm not sure how you usually translate "docs" in Astro and Starlight docs. If `"문서"` is clear enough, ignore this suggestion from AI:

```suggestion
		description="페이지를 Markdown으로 복사/보기, AI 채팅에서 열기, 위로 스크롤하기와 같은 유틸리티 기능들을 공식 문서에서 항상 사용할 수 있도록 만드세요."
```

trueberryless (approved):
I reviewed this translation as a non-native speaker and it looks good to me overall. I just left a tiny nitpicky comment.

docs/src/content/docs/ko/resources/plugins.mdx
jsparkdev: Although there is a slight nuance difference, they are virtually identical in modern technical documentation. Also, the term "공식 문서" used in Astro and Starlight refers to "official docs."

Since the word "문서" currently used in the code block is an appropriate translation, we won't be applying the suggestion.

Thank you for sharing your thoughts!

HiDeoo (approved):

PR #4068, opened 2026-07-13

ArmandPhilippot:
#### Description

Adds changes from #4057 and #4055 to the French translation of `resources/plugins.mdx`.

This is a bit more verbose than the English version unfortunately but I think "Faire en sorte que [...] soient" sounds better than a literal translation.


HiDeoo (approved):
Looks perfect to me, not quite sure it would be possible to do better/shorter imo 👍

PR #4082, opened 2026-07-19
chore(deps): update `magic-string` to v1

ocavue:
#### Description

Update `magic-string` from v0.30 to v1. `magic-string` v1 is now pure-ESM.

See changelog: https://github.com/Rich-Harris/magic-string/releases/tag/v1.0.0

No changeset file is added because it's not a user-facing change, and this change can be included in the next release with other changes. I could add a changeset file if maintainers prefer.


ocavue:
Blocking by https://github.com/rolldown/rolldown/issues/10354

trueberryless:
Hi @ocavue
Is this still blocked? I see that the linked Pull Request got merged and thought I'll leave a ping - just in case.

delucis:
@ocavue’s change was only just released in Rolldown 1.2.1: https://github.com/rolldown/rolldown/releases/tag/v1.2.1

So we’ll at least need to wait for the 3-day minimum release age window. Not sure if there are other consideration (e.g. do Starlight users also need to be on Rolldown 1.2.1 for things to work?)

ocavue:
vite pins rolldown to a fixed version. So we need to also wait for a new vite version that picks this rolldown version too. I guess it would be vite@8.2.0

delucis:
I merged #4108 which should fix the CI here when the branch gets updated.

> No changeset file is added because it's not a user-facing change, and this change can be included in the next release with other changes. I could add a changeset file if maintainers prefer.

This is not entirely accurate IIUC? Correct me if I’m wrong, but it seems like updating `magic-string` to v1 will break type checking for users who aren’t using a recent enough Vite/Rolldown version? (In theory it’s a Starlight internal issue, but until we merge #3572, Starlight type issues may cause user type issues too.)

What commands would a user need to run to update things in their project for compatibility?

delucis:
Hey @ocavue — what’s still needed here? We’re planning to drop support older Astro versions in the next minor. Might mean we can also do this?

ocavue:
Sorry for the late reply. I don't think we can merge this PR as it is. I can confirm your concern that "Starlight type issues may cause user type issues too".

The latest Astro still accepts vite versions as old as `vite@8.0.13`, and that vite version uses a rolldown version released before my fix. So a user who:

- uses `vite@8.0.13`
- has `exactOptionalPropertyTypes` enabled in `tsconfig.json`
- uses the latest Astro
- uses a Starlight with magic-string v1 (which is what this PR does)

will get an error when running `tsc -b`.

Here is an end-to-end reproduction of the error: https://github.com/issueset/test-starlight-magic-string-ts

Possible solutions:

1. Simply wait for #3572 before merging this. Having two `magic-string` versions
   installed is not the end of the world.
2. Add an `xxx as yyy` hack in this PR. That fixes the issue on the user side, but it is a bit ugly.
3. Bump the minimum `vite` version in the `astro` repo, right before a new minor Astro release. Of course we can do that, and we probably will at some point for other good reasons, but "fixing a `magic-string` compatibility issue in Starlight" does not feel like a strong enough reason on its own.

delucis:
Thanks for the careful reply!

The intention will be that the next minor include #3572 so that’s option 1 in your list. So I guess in that case this will be safe to merge? (We still need to fix the type checks on this PR though.)

ocavue:
> The intention will be that the next minor include #3572 so that’s option 1 in your list. So I guess in that case this will be safe to merge?


Yes, once https://github.com/withastro/starlight/pull/3572 is merged, you can merge this PR too. I just rebased against the latest `main` branch, so the CI should be green now.

delucis:
Awesome. Thanks again for taking care of this 🙌

delucis (approved):
Thanks again! Approving as we’re almost ready to cut the next minor 🚀

ocavue:
Thanks. @delucis, you can press the merge button directly when it's ready, since you know the release workflow better than I do. 😊

delucis:
Done!

For reference, in this case we wanted to merge #3572 first to simplify conflict resolution in other PRs.

Otherwise our main thing is to avoid merging PRs early if they update docs because we publish docs straight from `main` so once stuff lands on `main` it appears in docs. So we aim to merge as close to publish as possible so that docs aren’t “incorrect” for too long (most noticeable in minor releases where we might add a feature e.g.) Your PR doesn’t fall into that category obviously so it was just a case of avoiding any potential conflicts.

Btw, I think you have the option to allow maintainers to edit your branches disabled? I wanted to update the branch to ensure CI still passed with the latest changes from `main` but I wasn’t able to.

ocavue:
> Btw, I think you have the option to allow maintainers to edit your branches disabled?

Weird. I do not remember unchecking that checkbox. To double-check, I just created another PR, and I don't see this option in the GitHub UI (see the screenshot below).


Screenshot 2026-09-02 at 20 49 19
 
Screenshot 2026-09-02 at 20 49 29

- - - 


update: I found the reason. The reason is that I use an org `ocavue-forks` to hold all my forked repo. TIL.

> You cannot find the "Allow edits from maintainers" option because your pull request branch was created from a fork owned by an organization instead of a personal account.
> 
> https://github.com/orgs/community/discussions/5634

delucis:
> update: I found the reason.

Ahh, OK, right that makes sense. Weird that it’s the case though. And yeah also in that new PR I can’t e.g. add a changeset like I can on most branches:

image

But in any case as a maintainer you’d be welcome to push branches directly to this repo rather than using a fork if you wanted. But it’s not a big deal usually. Only a problem if there’s some small time sensitive thing where you weren’t available. Fortunately this time there were not conflicts or whatever.

ocavue:
> But in any case as a maintainer you’d be welcome to push branches directly to this repo 


Sure! Since I have a local clone of Starlight that points to my fork, which was created before I gained push permission from the `withastro` org, I just used that clone to create this PR. I’ll push directly next time!

PR #4087, opened 2026-07-21
Remove `minimumReleaseAgeExclude` from workspace

delucis:
#### Description
We had to introduce quite a few packages to `minimumReleaseAgeExclude` when co-ordinating the release alongside Astro v7 but should now be safe to remove this. This change helps protect us from accidentally installing a newer version of one of these packages unintentionally.


HiDeoo (approved):
Thanks for cleaning up :raised_hands: (and hopefully we have some tooling soon to help dealing with it)

PR #4088, opened 2026-07-21
Remove pagehide event from sidebar persistence script

### LINKED ISSUE #4083: SidebarPersistState listens for 'pageHide' (camelCase), so it never fires
### What

`SidebarPersistState.ts` registers its unload-time save with the wrong casing:

https://github.com/withastro/starlight/blob/main/packages/starlight/components-internals/SidebarPersistState.ts#L72

```ts
addEventListener('pageHide', updateState);
```

DOM event types are case-sensitive, and the event is `pagehide`. This listener
has never fired.

### Repro

Register both casings, navigate away twice, count invocations (counts kept in
`sessionStorage` so they survive the navigation):

```js
sessionStorage.setItem('camel', '0');
sessionStorage.setItem('lower', '0');
addEventListener('pageHide', () => sessionStorage.setItem('camel', String(+sessionStorage.getItem('camel') + 1)));
addEventListener('pagehide', () => sessionStorage.setItem('lower', String(+sessionStorage.getItem('lower') + 1)));
```

After 2 navigations (Chrome 151):

| listener | times fired |
| --- | --- |
| `'pageHide'` | **0** |
| `'pagehide'` | **2** |

### Why it matters

The comment directly above says the intent is to save on navigation *without*
using `beforeunload`, because of bfcache:

```ts
// Store sidebar state before navigating. These will also store it on tab blur etc.,
// but avoid using the `beforeunload` event, which can cause issues with back/forward cache
// on some browsers.
```

That's the right call — but with `pageHide` dead, `visibilitychange` is the only
surviving save path, so there is no redundancy. `pagehide` is the reliable
terminal signal in the page lifecycle precisely because `visibilitychange` →
`hidden` isn't guaranteed to fire before every unload path (Safari's bfcache
path being the usual example). So sidebar scroll/open state can be silently
lost on exactly the navigations `pagehide` exists to cover.

### Fix

One character:

```diff
-addEventListener('pageHide', updateState);
+addEventListener('pagehide', updateState);
```

Happy to open a PR if useful.

### Context

Found while bridging this persistence to Astro's view-transition lifecycle on a
Starlight site (`astro:before-swap` / `astro:after-swap`), since neither
`visibilitychange` nor `pagehide` fires during a `<ClientRouter />` navigation —
the document is never hidden and never unloads. Verified present on `main`
today; we're running 0.36.3, and the line is unchanged as of 0.41.3.

delucis:
#### Description

- Closes #4083
- Closes #4084
- This PR removes our use of the `pagehide` event in our sidebar state persistence script.
- As spotted in #4083 this has never been working as intended due to a typo as `pageHide` in the event name.
- We could fix the casing, but AFAICT we do not need the event at all:
  1. It hasn’t been firing for 1 year without anyone noticing.
  2. The main reason for `pagehide` was due to Safari’s partial `visibilitychange` support. According to [MDN’s data](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#browser_compatibility) Safari fixed this in April 2021 (v14.1 on desktop, v14.5 on mobile), so we no longer need `pagehide` to cover those older browsers which fall outside of our [browser support matrix](https://github.com/withastro/starlight/blob/main/CONTRIBUTING.md#understanding-starlight).
  3. As noted in #4083, even if it _had_ been cased properly, it wouldn’t have fixed their specific issue. It was just a detail they noticed in passing.

### DIFF
--- a/.changeset/whole-beds-add.md
+++ b/.changeset/whole-beds-add.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Simplifies Starlight’s client-side sidebar state persistence script slightly
--- a/packages/starlight/components-internals/SidebarPersistState.ts
+++ b/packages/starlight/components-internals/SidebarPersistState.ts
@@ -69,4 +69,3 @@ target?.addEventListener('click', (event) => {
 addEventListener('visibilitychange', () => {
 	if (document.visibilityState === 'hidden') updateState();
 });
-addEventListener('pageHide', updateState);


HiDeoo (approved):
Makes total sense to me and like you said, Safari 14.5 is way below our minimum supported version.

Great call 🧹

PR #4092, opened 2026-07-21
Support links with any protocol in sidebars

delucis:
#### Description

- Closes #4051
- Currently we special case `http(s)://` links in the sidebar and treat everything else as a relative link, prefixing locales, slashes, etc.
- This PR switches to check if the link contains any protocol component instead of only checking for an HTTP protocol so that protocols like `mailto:` or even app protocols like `vscode://` are supported.
- ~~The implementation may look a bit funny but is essentially doing `/^[^\/]*:/.test()`, i.e. checking if a colon is present in the first segment of a link (`/foo:bar` is a valid relative link, but `foo:bar` is treated as a protocol `foo:`). This implementation seems to be an order of magnitude faster than the current regular expression we use. I ran a [little benchmark](https://jsbm.dev/FxO0fLgzTXLLD) that showed this running about 15x faster than the regular expression, which may matter given sidebars can be a performance-sensitive area.~~
- There is still one use of `isAbsoluteUrl()` in the code base, but it’s for the favicon URL, so I’ve left that as is — other protocols aren’t relevant there ~~and it isn’t performance sensitive in the same way the sidebar is~~.

### DIFF
--- a/packages/starlight/utils/url.ts
+++ b/packages/starlight/utils/url.ts
@@ -1,12 +1,8 @@
 const HTTPProtocolRegEx = /^https?:\/\//;
+const ProtocolRegEx = /^[a-z][a-z\d+.-]*:/i;
 
 /** Check if a string starts with one of `http://` or `https://`. */
 export const isAbsoluteUrl = (link: string) => HTTPProtocolRegEx.test(link);
 
 /** Check if a string contains a protocol (e.g., `http:`, `https:`, `mailto:`). */
-export const hasProtocol = (link: string) => {
-	const colonIndex = link.indexOf(':');
-	if (colonIndex === -1) return false;
-	const slashIndex = link.indexOf('/');
-	return slashIndex === -1 || colonIndex < slashIndex;
-};
+export const hasProtocol = (link: string) => ProtocolRegEx.test(link);
--- a/packages/starlight/utils/url.ts
+++ b/packages/starlight/utils/url.ts
@@ -2,3 +2,6 @@ const HTTPProtocolRegEx = /^https?:\/\//;
 
 /** Check if a string starts with one of `http://` or `https://`. */
 export const isAbsoluteUrl = (link: string) => HTTPProtocolRegEx.test(link);
+
+/** Check if a string contains a protocol (e.g., `http:`, `https:`, `mailto:`). */
+export const hasProtocol = (link: string) => link.includes(':');
--- a/packages/starlight/utils/url.ts
+++ b/packages/starlight/utils/url.ts
@@ -4,4 +4,9 @@ const HTTPProtocolRegEx = /^https?:\/\//;
 export const isAbsoluteUrl = (link: string) => HTTPProtocolRegEx.test(link);
 
 /** Check if a string contains a protocol (e.g., `http:`, `https:`, `mailto:`). */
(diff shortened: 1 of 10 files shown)


packages/starlight/utils/url.ts
```diff
@@ -2,3 +2,11 @@ const HTTPProtocolRegEx = /^https?:\/\//;
...
 /** Check if a string starts with one of `http://` or `https://`. */
 export const isAbsoluteUrl = (link: string) => HTTPProtocolRegEx.test(link);
+
+/** Check if a string contains a protocol (e.g., `http:`, `https:`, `mailto:`). */
+export const hasProtocol = (link: string) => {
+	const colonIndex = link.indexOf(':');
+	if (colonIndex === -1) return false;
+	const slashIndex = link.indexOf('/');
+	return slashIndex === -1 || colonIndex < slashIndex;
+};
```
HiDeoo: Do we really want to check if it contains or starts with a protocol?

I wonder if we won't need to rely on a regex for this as I think `{ label: 'Test', link: 'getting-started?test=https://example.com' }` would be considered as a URL with a protocol with the current implementation? Maybe `/^[a-z][a-z\d+.-]*:/i` based on https://datatracker.ietf.org/doc/html/rfc3986#section-3.1 ?

delucis: Hmm, I'm not sure that's a valid relative link. Just as a quick test, it doesn't render here on GitHub: 

```md
[Test](getting-started?test=https://example.com)
```

[Test](getting-started?test=https://example.com)

(I also tried with `<a>` and got the same result.)

Admittedly, adding a leading slash does make it work: 

[Test with leading slash](/getting-started?test=https://example.com)

and we add leading slashes for people currently, so I suppose at least theoretically if some has a link like your example, then it will work currently and my change would break their usage. 🤔

HiDeoo: I think it's technically valid as a [`path-noscheme`](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) even if sanitized by GH but like you said, I would imagine it's probably very unlikely someone use this or very rare. Altho, I guess making it non-breaking would probably be better since it doesn't really complicate the implementation and remove a few lines?

What do you think?

delucis: Yeah, probably best to be safe. I didn’t mind it being a few lines longer for how much faster the `indexOf` approach is, but to be correct we need more complex logic so might as well use a regular expression.

Honestly I’d almost like one day to remove all the special handling of `link` type entries entirely: use `slug` for prefixing, locales etc.; use `link` for raw input. But that’s probably a more disruptive change (and I’m not sure how to help people migrate for that situation).

HiDeoo: > Honestly I’d almost like one day to remove all the special handling of `link` type entries entirely: use `slug` for prefixing, locales etc.; use `link` for raw input.

Definitely agree, would make things much easier.

delucis: Switched to a regular expression in 34c46d859e11819582214e2a9fb36f23d2708bcd

Thanks for thinking through all the edge cases!

HiDeoo (approved):
Thanks for the update, looks perfect to me :raised_hands:

Sorry for making it slower :sweat_smile:

PR #4095, opened 2026-07-24
Fix union schema `extend` regression

### LINKED ISSUE #4094: `docsSchema({ extend })` doesn't support `ZodUnion`

0.41.4

7.1.3

pnpm

linux

chrome

### Describe the Bug

Passing a `z.union(...)` to `docsSchema({ extend })` causes a type error (`ZodUnion` not assignable to `ZodObject`).

Regression in #3936 

https://stackblitz.com/edit/github-um6khbgu?file=src%2Fcontent.config.ts

### Participation

HiDeoo:
#### Description

- Closes #4094

This PR fixes a regression introduced in #3936 for an edge case I definitely didn't think about: using a union to extend the `docs` schema.

To address this, after unwrapping (optional, nullable, etc.) the user-provided schema, we need to check if it's a union and if it is, we deep-merge each member of the union with the default schema.

### DIFF
--- a/.changeset/rich-falcons-warn.md
+++ b/.changeset/rich-falcons-warn.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Fixes a regression when using a union to [extend](https://starlight.astro.build/reference/frontmatter/#extend) Starlight’s `docsSchema()`.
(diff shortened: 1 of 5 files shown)


delucis (approved):
Oh, it’s even the approach we use for Astro’s docs, so very good to fix 😄 

Thanks for taking care of this @HiDeoo 💜

HiDeoo:
> Oh, it’s even the approach we use for Astro’s docs, so very good to fix 😄

I totally forgot about that. Before merging, I'll try this branch against the docs to make sure I don't spot anything else while at it.

HiDeoo:
Confirmed updating Astro Docs to `v0.41.4` introduces a type error on the `docs` schema and fails to build.

Using the version from this PR fixes both the type error and the build issue.

PR #4107, opened 2026-08-02
i18n(ru): update translations

dragomano:


delucis (approved):
Thank you for the updates @dragomano 🙌

PR #4111, opened 2026-08-04
Use pinned version of `pkg-pr-new` for preview releases

HiDeoo:
#### Description

While reviewing another PR, I noticed [we use `pnpm dlx`](https://github.com/withastro/starlight/blob/1e6d0feca6c9352d3bf4bd404921611a476779f4/.github/workflows/preview-release.yml#L91) to run `pkg-pr-new` for preview releases.

This is [not recommended](https://github.com/stackblitz-labs/pkg.pr.new/blob/main/README.md#publish-packages) because `pnpm dlx` would always resolves to the latest version which could have unexpected behavior due to some upstream breaking changes but also increases supply-chain risks.

> [!CAUTION]
> In CI environments, avoid `npx`, `pnpm dlx`, `yarn dlx`, and `bunx` for this step. Install `pkg-pr-new` as a dependency and execute it from the lockfile (`npm exec`, `pnpm exec`, `yarn`, or `bun run`).

This PR fixes that by installing `pkg-pr-new` as a dev dependency and running it using `pnpm exec`.

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -88,4 +88,4 @@ jobs:
           # Note that `--compact` is used for shorter URLs but requires packages to be published on
           # npm. This means that it's not possible to publish a preview for a new package that has
           # not yet been published on npm.
-          pnpm dlx pkg-pr-new publish --pnpm --compact --packageManager=pnpm --no-template $PACKAGES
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thanks @HiDeoo!

PR #4114, opened 2026-08-04
Refactor Sätteri RTL code support plugin

delucis:
#### Description

- Fixes our Sätteri plugin for supporting code in RTL documents in newer versions of Astro
- Fixes the failing test in #4113 
- Sätteri now has a `ctx.parent()` we can use to get parents so we can remove the more convoluted code for working this out. Strictly speaking these do not work 1:1 — before, we’d skip applying attributes to a `<code>` element even if it was deeply nested in a `<pre>` whereas now we only skip if the `<code>` is a direct child of the `<pre>`. However, in practice, I think this covers all the scenarios we were intending?
- Existing tests here should pass, showing that the changes work for people on older versions of Astro. And I ran the changes against #4113 too to ensure it fixes the failing test there.

### DIFF
--- a/packages/starlight/integrations/satteri.ts
+++ b/packages/starlight/integrations/satteri.ts
@@ -1,6 +1,6 @@
 import { fileURLToPath } from 'node:url';
 import { satteriHeadingIdsPlugin } from '@astrojs/markdown-satteri';
-import type { Element, Properties } from 'hast';
+import type { Properties } from 'hast';
 import type { Paragraph } from 'mdast';
 import { directiveToMarkdown } from 'mdast-util-directive';
 import { toMarkdown } from 'mdast-util-to-markdown';
--- a/packages/starlight/integrations/satteri.ts
+++ b/packages/starlight/integrations/satteri.ts
@@ -136,39 +136,37 @@ function serializeDirective(node: Parameters<typeof toMarkdown>[0]): string {
 	return md.at(-1) === '\n' ? md.slice(0, -1) : md;
 }
 
-function satteriRtlCodeSupportPlugin(allowedPaths: string[]): () => HastPluginDefinition {
-	return () => {
-		return {
-			name: 'starlight-rtl-code-support',
-			element: [
-				{
-					filter: ['pre'],
-					visit(node, ctx) {
-						if (!shouldTransformPath(ctx.fileURL, allowedPaths)) return;
-						if (node.properties && 'dir' in node.properties) return;
-						ctx.setProperty(node, 'dir', 'ltr');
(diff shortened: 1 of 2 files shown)


packages/starlight/integrations/satteri.ts
```diff
@@ -138,21 +138,13 @@ function serializeDirective(node: Parameters<typeof toMarkdown>[0]): string {
 
 function satteriRtlCodeSupportPlugin(allowedPaths: string[]): () => HastPluginDefinition {
 	return () => {
```
HiDeoo: Nit: I guess we could even remove the factory and directly return the plugin now if we want to.

HiDeoo (commented):
> However, in practice, I think this covers all the scenarios we were intending?

I guess so and something like `<pre><span><code>` or similar is probably not common but that makes it a different behavior between the Unified and Sätteri implementation of the same plugin. Would it work to walk `ctx.parent()` until reaching `<pre>` or the root?

packages/starlight/integrations/satteri.ts
delucis: Good idea. Done in 1c5e868e24528a120d592bfa461d55a041ebd019.

delucis:
> Would it work to walk `ctx.parent()` until reaching `\<pre\>` or the root?

I think so? I assume that’s more expensive because you have to serialize the entire parent node tree. I asked Erika [on Discord](https://discord.com/channels/830184174198718474/1070481941863878697/1534237080727588874) for her reaction.

> that makes it a different behavior between the Unified and Sätteri implementation of the same plugin

Yeah, true. We _could_ update the Unified implementation to match Sätteri instead if equivalence is important? That might be cheaper than adding a lot of recursion for “correctness” even though the use cases it covers are basically all unintended.

---

**Edit:** Erika said this kind of check will be cheap so I went ahead and pushed 275e45452b519756dad1a7bfd02155aa73a6255d which walks up the full tree. (For sanity, I also ported all the changes to the #4113 branch to double check it still fixes the test failure there and I confirmed it does.)

HiDeoo (approved):
This looks perfect to me :rocket:

> Erika said this kind of check will be cheap so I went ahead and pushed [275e454](https://github.com/withastro/starlight/commit/275e45452b519756dad1a7bfd02155aa73a6255d) which walks up the full tree. (For sanity, I also ported all the changes to the #4113 branch to double check it still fixes the test failure there and I confirmed it does.)

Amazing, thanks for investigating and updating the PR, definitely feels better imo to have both implementations having the same behavior :raised_hands:

PR #4120, opened 2026-08-13
Replace 3 outdated links to Astro Docs

ArmandPhilippot:
#### Description

- Replace "Markdown plugins" links

We have recently updated the Markdown guide in Astro Docs (https://github.com/withastro/docs/pull/14297) and the section was renamed "Markdown processor plugins". There was two occurrences in Starlight docs.

- Replace an example using Astro DB

Astro DB has been removed in Astro 7 (https://github.com/withastro/docs/pull/13985) and the Astro DB guide now redirects to the integration page with an aside saying "Removed". I think it is better to use an example that is still relevant.

### DIFF
--- a/docs/src/content/docs/guides/authoring-content.mdx
+++ b/docs/src/content/docs/guides/authoring-content.mdx
@@ -561,7 +561,7 @@ Starlight supports all other Markdown authoring syntax, such as lists and tables
 
 ## Advanced Markdown and MDX configuration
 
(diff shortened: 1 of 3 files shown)


delucis (commented):
Oh great thinking! I totally forgot about this when reviewing your changes. Thanks @ArmandPhilippot 🙌

delucis (approved):

PR #4121, opened 2026-08-13
Refactor mobile menu to use Popover API

delucis:
#### Description

This PR refactors Starlight’s mobile menu toggle to use the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) instead of the current JS-powered `<button>` with `aria-expanded` pattern.

- A small amount of JS remains but it’s now only a progressive enhancement for focus trapping. The core functionality of opening/closing the mobile menu and locking body scroll works even if JS breaks for whatever reason.

- The JS is now in the `PageFrame` component as a custom element that attaches to the popover body instead of the menu button. It reacts to global events (`matchMedia()`) and the popover’s `"toggle"` event so it makes more sense to live alongside the element it attaches to. This also means the menu button component is now easier to override as any button with the correct `popovertarget` attribute will work.

- CSS switches from hooking into `[aria-expanded]` to using the [`:popover-open`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:popover-open) pseudo class.

- We no longer add a `data-mobile-menu-expanded` attribute to `<body>` when the menu is open and use `body:has(#starlight__sidebar:popover-open)` in CSS instead, which works without the JS-managed attribute.

- While I was updating the button, I switched from `aria-label` to a visually hidden span of text, which is generally the recommended pattern for labels where possible.

#### Browser compatibility

Use of the Popover API requires slightly bumping our minimum supported browsers:
- Firefox 121 (December 2023) => 125 (April 2024)
- Chromium 111 (March 2023) => 116 (August 2023)
- Safari 16.4 (March 2023) => 17.0 (September 2023) (for both macOS and iOS)
  
According to browsersl.ist, comparing [before](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+111%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+16.4%2C+iOS+%3E%3D+16.4%2C+not+op_mini+all) and [after](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+116%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+125%2C+Safari+%3E%3D+17.0%2C+iOS+%3E%3D+17.0%2C+not+op_mini+all) shows a drop in global coverage of 1 percentage point. The newest minimum supported browser will be Firefox 125, released 29 months ago.

N.B. that umbrella compatibility measures such as [`wf-popover`](https://caniuse.com/wf-popover) show support only arriving later in some browsers. However, IIUC this is due to subfeatures that we are not relying on not being ready yet and the features we require are safe even in these older browser versions:

- [`popover` attribute](https://caniuse.com/mdn-html_global_attributes_popover)
- [`popovertarget` attribute](https://caniuse.com/mdn-html_elements_button_popovertarget)
- [`hidePopover()` method](https://caniuse.com/mdn-api_htmlelement_hidepopover)
- [`"toggle"` event](https://caniuse.com/mdn-api_htmlelement_toggle_event_popover_elements)
  
Safari on iOS had a [long-standing bug](https://bugs.webkit.org/show_bug.cgi?id=267688) in versions 17.0–18.2 which prevented tap-away clicks closing popovers as is expected. However, we do not require that behaviour in Starlight mobile menus as there is nowhere a user expects to click away to and the main control is the button.

#### Accessibility

The button with `popovertarget` and `popover` combo has built-in accessible roles equivalent to our previous `aria-expanded` pattern. I tested with VoiceOver in Firefox, Chrome, and Safari on macOS 26.5.1 and found the behaviour to be a very slight improvement in my opinion compared to the current announcements although both are basically equivalent. Would be great to test in more scenarios!

#### Demo

Here’s a small screen capture of the live Starlight docs and compared to this branch, showing the menu working without JavaScript after these changes:

https://github.com/user-attachments/assets/b73fc7f3-c1ba-4289-b7cf-69d7fd035a90


louisescher (dismissed):
LGTM! Tested on Zen 1.21.14b (Firefox 153.0.3) and Chrome 151.0.7922.109, both on Windows.
Edit: Also tested on Zen and Safari on MacOS, both work.

trueberryless:
Works on Firefox Android as well 👍

.changeset/clean-flies-start.md
```diff
@@ -0,0 +1,21 @@
+---
+'@astrojs/starlight': minor
+---
+
+Simplifies markup for Starlight’s mobile menu toggle
+
+**⚠️ Potentially breaking change:** If you use a theme plugin or custom styles targeting the mobile menu toggle button, you may need to adjust these for the new button. The button is no longer wrapped in a `<starlight-menu-button>` custom element and no longer uses the `aria-expanded` attribute. Instead, you can use `button[popovertarget="starlight__sidebar"]` to target the button and the `:popover-open` pseudo-class to style the menu open state specifically.
```
HiDeoo: > If you use a theme plugin or custom styles

Should we mention component overrides for any of the 2 changes components too?

> The button is no longer wrapped in a `<starlight-menu-button>`

From a basic [GH search](https://github.com/search?q=starlight-menu-button&type=code&p=1) for `starlight-menu-button`, and how much it's used just for styling, I wonder if there is a world where we keep it to reduce breaking changes for people (`aria-expanded` seems to not be used a lot tho in combination with it).

> The JS is no longer a custom element as it only reacts to global events

Not sure I fully understand this part, the `beforetoggle` event is still attached to a specific DOM node. If we were to keep the custom element, I guess this wouldn't break this specific case with a client router navigation?

Not sure I have made a personal opinion yet, but throwing it here in the meantime.

.changeset/long-jokes-post.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/starlight': minor
+---
+
+Refactors Starlight’s mobile menu toggle to work when JavaScript fails or is disabled
+
+⚠️ **BREAKING CHANGE:** This release drops official support for Chromium-based browsers prior to version 116 (released August 2023), Safari-based browsers prior to version 17.0 (released September 2023), and Firefox prior to version 125 (released April 2024). You can find a list of currently supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+116%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+125%2C+Safari+%3E%3D+17.0%2C+iOS+%3E%3D+17.0%2C+not+op_mini+all).
```
HiDeoo:
```suggestion
⚠️ **BREAKING CHANGE:** This release drops official support for Chromium-based browsers prior to version 116 (released August 2023), Safari-based browsers prior to version 17.0 (released September 2023), and Firefox prior to version 125 (released April 2024). You can find a list of currently supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+116%2C+Edge+%3E%3D+116%2C+Firefox+%3E%3D+125%2C+Safari+%3E%3D+17.0%2C+iOS+%3E%3D+17.0%2C+not+op_mini+all).
```

I think the Edge version should be updated here too.

.changeset/long-jokes-post.md
```diff
@@ -0,0 +1,9 @@
...
+
+If you still need to support older browsers, you can avoid upgrading or use a polyfill for the Popover API such as [`@oddbird/popover-polyfill`](https://github.com/oddbird/popover-polyfill). Note however that future Starlight releases may depend on additional browser features covered by our current browser support range without mentioning it explicitly in changelogs.
```
HiDeoo: Didn't look at the polyfill before today, but after checking it, I wonder if we should really recommend it as `:popover-open` [cannot be polyfilled in CSS](https://github.com/oddbird/popover-polyfill#caveats).

We could also support their fallback with some selectors like `:where(:popover-open, .\:popover-open)` I guess but not sure how I feel about that 🤔

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -2,59 +2,30 @@
...
+	// The mobile menu uses the popover API to open and close without JavaScript. We enhance this
+	// behaviour with focus trapping while the menu is open.
+
+	// Reset menu state when the viewport toggles between mobile and desktop sizes.
+	matchMedia('(min-width: 50em)').addEventListener('change', () => trapFocus(false));
+
+	// Update the menu state when the sidebar popover is toggled.
+	document
+		.getElementById('starlight__sidebar')
+		?.addEventListener('beforetoggle', ({ newState }) => trapFocus(newState === 'open'));
```
HiDeoo: Not quite sure what the scenario would be, maybe if another custom modal is opened and trying to prevent the sidebar from opening when this other modal is opened, but technically `beforetoggle` is cancelable so we could end up leaving the page inert if it was actually cancelled.

packages/starlight/components/MobileMenuToggle.astro
```diff
@@ -2,59 +2,30 @@
...
-			if (e.code === 'Escape') {
-				this.setExpanded(false);
-				this.btn.focus();
-			}
-		}
+	// The mobile menu uses the popover API to open and close without JavaScript. We enhance this
+	// behaviour with focus trapping while the menu is open.
+
+	// Reset menu state when the viewport toggles between mobile and desktop sizes.
+	matchMedia('(min-width: 50em)').addEventListener('change', () => trapFocus(false));
```
HiDeoo: I think we still kinda need to close the popover like we closed the menu before, otherwise if you're on mobile viewport, open the menu, resize to desktop viewport (this removes the focus trap), and go back to a mobile viewport, the popover would still be opened but we would remove the trap again.

packages/starlight/components/PageFrame.astro
```diff
@@ -48,7 +48,11 @@ const { hasSidebar } = Astro.locals.starlightRoute;
 		}
 
 		.sidebar-pane {
-			visibility: var(--sl-sidebar-visibility, hidden);
+			/* By default, user agents apply some opinionated styles to popovers, which we don’t want. */
+			border: unset;
+			padding: unset;
+			height: unset;
```
HiDeoo, line 76: Based on the [UA stylesheet](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:popover-open#examples), I think we should also inherit the `color` to avoid `CanvasText` being used.

For comparisons when adding regular text to the popover:

| Before | After |
| --- | --- |
| Image | Image |

CONTRIBUTING.md
```diff
@@ -275,7 +275,7 @@ To add a language, you will need its BCP-47 tag and a label. See [“Adding a ne
 
 - Components that require client-side JavaScript or CSS should use JavaScript/CSS features that are well-supported by browsers.
 
-  You can find a list of supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+111%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+121%2C+Safari+%3E%3D+16.4%2C+iOS+%3E%3D+16.4%2C+not+op_mini+all). To check whether or not a feature is supported, you can visit the [Can I use](https://caniuse.com) website and search for the feature.
+  You can find a list of supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+116%2C+Edge+%3E%3D+111%2C+Firefox+%3E%3D+125%2C+Safari+%3E%3D+17.0%2C+iOS+%3E%3D+17.0%2C+not+op_mini+all). To check whether or not a feature is supported, you can visit the [Can I use](https://caniuse.com) website and search for the feature.
```
HiDeoo:
```suggestion
  You can find a list of supported browsers and their versions using this [browserslist query](https://browsersl.ist/#q=%3E+0.5%25%2C+not+dead%2C+Chrome+%3E%3D+116%2C+Edge+%3E%3D+116%2C+Firefox+%3E%3D+125%2C+Safari+%3E%3D+17.0%2C+iOS+%3E%3D+17.0%2C+not+op_mini+all). To check whether or not a feature is supported, you can visit the [Can I use](https://caniuse.com) website and search for the feature.
```

Same as before for the Edge version.

HiDeoo (changes requested):
Amazing that we can finally switch, thanks for taking care of the update :raised_hands:

Left a few comments, but other than that, everything looks great. I tested on:

- macOS: Chrome - Firefox - Safari
- iOS: Safari
- Windows: Chrome - Edge - Firefox

I also played around with some screen readers and the same browsers:

- macOS VoiceOver
- iOS VoiceOver
- Windows NVDA
- Windows JAWS

This is pretty much always equivalent or better like you mentioned and I didn't spot regressions so far.

---

Totally unrelated to this PR and wouldn't belong here as this is not a new behavior but while testing on Windows (and also noticeable on macOS with scrollbar visibility set to "Always"), we can probably avoid the buttons shift when opening the menu:

https://github.com/user-attachments/assets/a0096f4b-41ed-4448-a2d0-0fa58c9c5f4a

.changeset/clean-flies-start.md
delucis: > Should we mention component overrides for any of the 2 changes components too?

Ah yeah, probably. The `MobileMenuToggle` could be impacted in particular — `PageFrame` is more extreme. (**Edit:** made a change for this in 837b654fa16b7dbf86c76cc5e12c641a193d7b5c)

> I wonder if there is a world where we keep it to reduce breaking changes for people

Yes, I also wondered this. I don’t know — I thought it could be nice to clean up while still pre-v1 as it simplifies the styles a bit, e.g.

```diff
- starlight-menu-button:has(~ :popover-open) button {
+ button:has(~ :popover-open) {
```

But obviously removing it is more disruptive than not.

> the `beforetoggle` event is still attached to a specific DOM node

Yes, that is true, but I meant the button is no longer attached to — the custom element wrapped the button, but the event is attached to the sidebar pane, which is a unique element accessed by `id`. We could wrap the sidebar pane in a new custom element instead if you think it’s a good idea? e.g. the `<div>` here could become `<sl-sidebar-pane>` or something?

https://github.com/withastro/starlight/blob/39a8b15c362817f3b21ca148f56464b9a0433f9e/packages/starlight/components/PageFrame.astro#L13-L17

**Edit:** switched to a custom element as proposed here in 050b7ca8e93d86bbe3bb839a2edf5c5d783f67da now that the popover needs to be programmatically closed as well

.changeset/long-jokes-post.md
delucis: Wow, well spotted 😄

.changeset/long-jokes-post.md
delucis: Ahh, I did not actually dive into the details. In that case, yeah, maybe best to remove that recommendation. I do feel fairly confident that we’re not jumping **too** far ahead in terms of browser support so it’s OK to just inform people and they can decide whether to update or not.

```suggestion
```

packages/starlight/components/PageFrame.astro
delucis: Ah, good catch 👍
```suggestion
			height: unset;
			color: unset;
```

packages/starlight/components/MobileMenuToggle.astro
delucis: Do you think `toggle` would be better? I initially used it but E2E tests failed because technically it means the page is briefly still inert after closing the menu. I suspect this is only an issue for hyper fast users (a.k.a. computers) but I took the easy route out 😅

packages/starlight/components/MobileMenuToggle.astro
delucis: Done in cbf64ca38ae1f00fb6d73e69f64152e9958cc975 — thanks for catching it!

.changeset/clean-flies-start.md
HiDeoo: Hmm, interesting, I didn't think about this idea of a new custom element.

I'm personally still torn about it: we end up breaking all usage of `starlight-menu-button` (which I agree would be a nice cleanup), and now we define a custom element in `MobileMenuToggle` but that custom element is not used in this component (which was the case before) soit  could end up being a bit more confusing for users overriding it.

Definitely a tricky choice 🤔

.changeset/long-jokes-post.md
HiDeoo: > Wow, well spotted 😄

I updated it recently for Astro 7 and I remember there was 2 edits 😄

packages/starlight/components/MobileMenuToggle.astro
HiDeoo: Technically yes, I think this is the proper event to use fo this case afaik.

And for the tests, I guess in this case we would want to add some `await expect(page.locator('.main-frame')).toHaveAttribute('inert', '');` (and the `.not.toHaveAttribute` opposite) after the clicks, as they're [auto-retrying assertions](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) so Playwright will properly retry until the assertion passes or times out (and it's a better approach than using a timeout or something similar).

.changeset/clean-flies-start.md
delucis: > that custom element is not used in this component (which was the case before) soit could end up being a bit more confusing for users overriding it.

We could move the new custom element script to the `PageFrame` if you think that’s clearer? I think it might result in it being included in pages that don’t have sidebars, but that maybe doesn’t matter or could even be better if it means it gets bundled with other page-level scripts. I can test that.

**Edit:** tested this and it makes no difference to output, so went ahead and move the script to `PageFrame` in 5c73fb365f86792a5e7ec5c92a9cc2edbf4e0f20.

> we end up breaking all usage of `starlight-menu-button`

I think we will for most cases anyway, even if we keep it?

If you have an override that replaces the menu button, I guess it will break most likely as the sidebar popover styles no longer work in response to `aria-expanded` set by the old script you maybe copied.

If you have an override that replaces the `PageFrame`, it will break because the sidebar pane won’t have a `popover` attribute.

If you have custom styles that depended on `aria-expanded`, those won’t work anymore.

The only case where it would keep working would be very simple theming like I do e.g. [in my Flexoki theme](https://github.com/delucis/starlight-theme-flexoki/blob/ec48998818dc7cedc36e85ee1f8e7e6fe0102ef8/packages/starlight-theme-flexoki/styles/base.css#L16-L21) that target `starlight-menu-button button` and nothing else. That’s still _something_ of course, but I don’t know if it justifies keeping it enough?

We _could_ add a class name to the button if we think users migrating to target `button[popovertarget="starlight__sidebar"]` feels a bit clunky?

HiDeoo: Yeah, I'm definitely mostly talking on the basic `starlight-menu-button button` case which seems from the GH search to represent a lot of the usage but this would definitely break for more advanced patterns.

`button[popovertarget="starlight__sidebar"]` can definitely feel weird as it's an attribute selector, use what could be a weird double underscore ID so yeah, maybe having a more user-friendly class could be a nice middle ground and ease the migration a bit for the common basic styling cases?

packages/starlight/components/MobileMenuToggle.astro
delucis: Thanks! I switched back to `toggle` and adjusted the tests in a1986b78787561b1572c0d0ee7a5dbb84e56e057.

.changeset/clean-flies-start.md
delucis: Good call. I added a `sl-menu-button` class name in 62964d30722f1662a2077c1b7c050594cb534558 that people can use and updated the migration guidance. I went for the `sl-` prefix as I was a bit concerned that looking at
```diff
- starlight-menu-button
+ .starlight-menu-button
```
it might be very easy to miss the `.` and hopefully the different prefix underlines the difference (we’re not very consistent about using `starlight-` vs `sl-` in any case).

One other advantage of moving the script into `PageFrame` is that it’s now actually very easy to provide an entirely custom menu button component if you want — any `<button popovertarget="starlight__sidebar">` will work.

.changeset/clean-flies-start.md
```diff
@@ -0,0 +1,23 @@
...
+
+In the following example, custom styles for the menu button are updated for the new approach:
+
+```diff
+- starlight-menu-button button {
++ .sl-menu-button {
+  color: var(--sl-color-text);
+}
+
+- [aria-expanded='true'] starlight-menu-button button {
```
HiDeoo: I guess it's either


```suggestion
- starlight-menu-button[aria-expanded='true'] button {
```

or 


```suggestion
- [data-mobile-menu-expanded] starlight-menu-button button {
```

Should we also explicitly mention `data-mobile-menu-expanded` is gone? Looks like it's also [used quite a lot](https://github.com/search?q=data-mobile-menu-expanded&type=code).

Or maybe we show 3 diff examples:

- `starlight-menu-button button` → `.sl-menu-button` (existing)
- `starlight-menu-button[aria-expanded='true'] button` → `.sl-menu-button:has(~ :popover-open)` (existing)
- `[data-mobile-menu-expanded] .myClass` → `body:has(sl-sidebar-pane:popover-open) .myClass` (new?)

packages/starlight/__e2e__/basics.test.ts
```diff
@@ -832,10 +837,12 @@ test.describe('mobile menu focus trap', () => {
 		await page.setViewportSize({ width: 1280, height: 720 });
 
 		// The anchor heading link should be focusable again.
+		await expect(mainFrame).not.toHaveAttribute('inert', '');
 		await anchorHeadingLink.focus();
 		await expect(currentFocus).toHaveText(anchorLinkAccessibleName);
 
 		// Resizing back to a smaller viewport should not re-enable the focus trap.
+		await expect(mainFrame).not.toHaveAttribute('inert', '');
 		await page.setViewportSize({ width: 375, height: 667 });
```
HiDeoo:
```suggestion
		await page.setViewportSize({ width: 375, height: 667 });
		await expect(mainFrame).not.toHaveAttribute('inert', '');
		await expect(page.locator('.sidebar-pane:popover-open')).toHaveCount(0);
```

I think we should assert `inert` after resizing so we don't check twice the desktop version and we check the final state.

Up to you, I also suggested adding a check to ensure the popover is still closed after resizing to mobile which would have caught the previous issue.

.changeset/clean-flies-start.md
HiDeoo: Amazing, thanks for going back and forth with me on this.

.changeset/clean-flies-start.md
delucis: > Should we also explicitly mention `data-mobile-menu-expanded` is gone? Looks like it's also [used quite a lot](https://github.com/search?q=data-mobile-menu-expanded&type=code).

Oof, yes. Boy, this breaking changes/having dependents thing is scary 😅 

I’ll update it to incorporate your fixes and mention that attribute too………

I think it’s definitely better that these things will work without JS but I do always feel bad for these kinds of things knowing there’ll be details people miss when updating, leading to subtly broken sites.

HiDeoo: Yeah, but it's definitely for the better good in the end, and we definitely try to make our changeset as precise and guided as possible.

Altho, I'll admit I'm a bit scared I missed the invalid selector in my first review considering I'm pretty sure I read it 3 times 😅 I blame the heat 🥵

delucis: OK, made those changes. I decided to mention the data attribute in the other changeset because it makes a bit more sense alongside the the “no JS” update compared to this changeset which is about the button. Hopefully that’s clear.

HiDeoo (dismissed):
I think adding the data attribute changes to the second changeset makes perfect sense.

Thanks for all the updates, this now looks perfect to me and ready 🚀 🧹

trueberryless (approved):
The changes look good to me here. Well done, Chris with the speedy implementation and Hideo with the extraordinary review 🥳 

I experimented a bit with the changes in my [Sidebar Swipe plugin](https://starlight-sidebar-swipe.netlify.app/), but quickly realized that it's not possible to use the popover API there because I need to place the sidebar _behind_ the actual content in order for the Discord-like swipe gesture to work. And AFAIU the intend of pop‎<em>over</em> is... well... being _on top of_ the content, which doesn't go hand in hand with my plugin.

I therefore added the `PageFrame.astro` as a new component override and "reverted" the changes from here 😥
If someone knows some ingenious way to make it work cleaner, let me know.

EDIT: Chris suggested a potential solution, which would change the behavior a little bit [on Discord](https://discord.com/channels/830184174198718474/1538869186896330802). 🫶

PR #4122, opened 2026-08-13
Update `size-limit` config

delucis:
#### Description

We noticed while working on #4121 that our current size limit checks won’t reflect changes to the page sidebars because we test against the start template’s landing page.

This PR adds one of the example guide pages in the starter template to our size limit checks so we also track changes in size to the more standard documentation page layouts as suggested by @trueberryless.

While there I’ve also slightly reduced the size limit for the HTML pages to be a bit closer to the current output sizes and trigger a failure earlier if those increase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,13 @@
     {
       "name": "/index.html",
       "path": "examples/basics/dist/index.html",
-      "limit": "8 kB",
+      "limit": "7 kB",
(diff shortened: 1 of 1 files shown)


louisescher (approved):

trueberryless (approved):
Thanks for the quick update!

PR #4126, opened 2026-08-14
i18n(de): update links to Astro docs

trueberryless:
#### Description

This PR updates the German translation of 3 pages in total, which included outdated links to Astro's documentation -- as originally discovered and updated by Armand in #4120 

I decided to keep the links localized (linking to `/de/` instead of `/en/`), although those pages are not yet translated, so the user stays localized in German nonetheless. The HTML fragments are kept valid in English tho.

### DIFF
--- a/docs/src/content/docs/de/guides/authoring-content.mdx
+++ b/docs/src/content/docs/de/guides/authoring-content.mdx
@@ -561,7 +561,7 @@ Starlight unterstützt alle anderen Markdown-Autorensyntaxen, wie Listen und Tab
 
 ## Erweiterte Markdown- und MDX-Konfiguration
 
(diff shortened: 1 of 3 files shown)


delucis (approved):
Thanks @trueberryless! Agreed on the links being `/de/`. Hopefully we can remember to update the fragments if these pages are every updated.

trueberryless:
> Hopefully we can remember to update the fragments if these pages are every updated.

This is indeed also a little concern for me and I'd love to improve this with some cross-domain link checker. We could either think about setting up some Starlight to Astro Docs checks in CI here (because the other way around is pretty rare). One fast and great link checker I heard about is [Lychee](https://lychee.cli.rs/) (which also uses Starlight for the docs - the circle closes 😅).

Or we kindly rely on Armand to remember everything. In Armand we trust 🙏

PR #4133, opened 2026-08-20
Use a page-specific counter for `<Tabs>`  IDs

### LINKED ISSUE #4132: `<Tabs>` ids come from a build-global counter, so a page's ids depend on the rest of the build

0.41.7 (and 0.36.3 for the second half below)

7.2.4 (and 5.18.2 for the second half below)

npm (also reproduced with bun, and by invoking `node ./node_modules/astro/astro.js build` directly, so it isn't a package-manager artifact)

Mac — darwin 25.5.0, Node v26.0.0

n/a — this is about build output, not runtime behaviour

### Describe the Bug

`<Tabs>` ids are handed out by a **build-global** counter, so the ids a page gets are a function of *where that page fell in the build*, not of the page's own content. Two consequences, one of which reproduces on the latest versions:

**A. Adding one page rewrites the tab ids on every page built after it.** (Reproduces on starlight 0.41.7 / astro 7.2.4.) In a 40-page site where every page has one `<Tabs>`, adding a single new page under `aaa/` changes `id="tab-N"` / `href="#tab-panel-N"` on **all 40** pre-existing pages. Removing it restores every id exactly. Nothing about those 40 pages changed.

**B. On the astro 5 line the same commit does not build twice the same.** (starlight 0.36.3 / astro 5.18.2.) Three builds of an unchanged project, each from a cold content-layer store, give three different id assignments — most of the 40 pages get a different id each time (31 and 33 of 40 in the transcript below), with no input change at all.

The difference between A and B is not Starlight: it is that astro ≥ 6 writes the content store sorted by key ([`content/data-store-writer.js`](https://github.com/withastro/astro/blob/d48312502ef33a32aef3f25b6b6035db8b38e189/packages/astro/src/content/data-store-writer.ts)), which makes `getCollection()` order stable, while astro 5 leaves it at the glob loader's async completion order. So astro ≥ 6 masks B; the dependency on build order that causes both is unchanged in Starlight.

Why it matters either way: the HTML of a page is not a function of that page, so build output isn't diffable. Any before/after comparison of output, artifact-diffing deploy, or output-equality check in CI reports pages as changed that did not change. We hit B on a 244-page site while byte-comparing output before and after a refactor that touched no content: 9–14 pages "changed", all of them `<Tabs>` pages, every difference one of these ids.

#### Cause

[`user-components/rehype-tabs.ts#L42-L46`](https://github.com/withastro/starlight/blob/9b8221c0d85903437a26c81d1de5bd9a95a9baa2/packages/starlight/user-components/rehype-tabs.ts#L42-L46) keeps the counter at module scope, so it is process-global for the whole build:

```ts
let count = 0;
const getIDs = () => {
	const id = count++;
	return { panelId: 'tab-panel-' + id, tabId: 'tab-' + id };
};
```

`Tabs.astro` calls `processPanels()` at render time, so the ids a page receives depend on how many `<TabItem>`s were rendered earlier **in the same build**. Page render order is the docs collection order — [`utils/routing/index.ts`](https://github.com/withastro/starlight/blob/9b8221c0d85903437a26c81d1de5bd9a95a9baa2/packages/starlight/utils/routing/index.ts) builds `routes` and `paths` straight from the `getCollection('docs')` array, preserving its order — and `getCollection()` documents no ordering guarantee. On astro 5 that order is the glob loader's completion order under [`pLimit(10)`](https://github.com/withastro/astro/blob/d48312502ef33a32aef3f25b6b6035db8b38e189/packages/astro/src/content/loaders/glob.ts#L307), which is why B needs **more than ~10 content files** to show up at all: under 10 the limiter never queues and completion order matches input order.

B also needs a **cold content-layer store** (`rm -rf node_modules/.astro`) — what a CI checkout always has. The store is persisted, so consecutive warm builds reuse the previous order and agree; that is why this reads as flakiness rather than as a property of the build. In our repo three warm builds agreed on all 40 pages while three cold ones did not.

#### Suggested fix

Tab ids only have to be unique **within a document** — they are same-page `aria-labelledby` and `href="#…"` targets. So the counter doesn't need to be build-global at all: making it per-page removes the dependency on build order entirely and makes a page's ids a function of that page.

`Tabs.astro` already keeps per-page state for the synced-tabs restore script, through a symbol on `Astro.locals` that Astro resets for each page. The same mechanism could carry a per-page tab counter passed into `processPanels(html, offset)` — no new machinery, and it reuses the pattern already in that file.

#### Workaround, for anyone else who lands here

Normalize the ids before comparing two builds — but **renumber per page rather than collapsing them.** Replacing every id with one literal also equates a panel that points at the wrong sibling with one that points at the right one; on a real page, rewriting a single `href="#tab-panel-41"` to `#tab-panel-40` (an orphaned panel and a mis-wired tab) still compared **equal** under a blanket substitution. Mapping each page's numbers onto `0,1,2…` in order of first appearance is one-to-one within the page, so it removes only the build-global starting value:

```sh
for d in b1 b2; do
  find "$d" -name '*.html' -exec perl -0777 -i -pe \
    'my %m; my $i = 0; s{\btab-(panel-)?(\d+)\b}{"tab-" . ($1 // "") . ($m{$2} //= $i++)}ge' {} +
done
diff -rq b1 b2 -x pagefind
```

Keying the map on the number alone keeps `tab-8` and `tab-panel-8` on the same new value. On a 310-page site this took 14 differing files down to 0.

It does not reach the search index, though: Pagefind fragment filenames are content hashes of a fragment whose text embeds the anchor ids, so the same pages churn 32 `.pf_fragment` files plus `pagefind-entry.json` — built from the un-normalized HTML, and nothing done to `dist` afterwards touches it. Only a per-page counter removes that.

Not a StackBlitz: half of this is about **build** output across repeated builds from a cold content-layer store, which a WebContainer preview can't show. The reproduction is a single self-contained script — it writes a stock Starlight project (`docsLoader()`, `docsSchema()`, no plugins, no i18n, `pagefind: false`), builds it three times, then adds one page and builds again:

Copy the script in the block below to `tab-id-repro.sh`, then:

```sh
chmod +x tab-id-repro.sh
./tab-id-repro.sh                                   # starlight 0.41.7 / astro 7.2.4
STARLIGHT=0.36.3 ASTRO=5.18.2 ./tab-id-repro.sh     # the astro 5 line
```

Every page gets exactly one `<Tabs>` with exactly one `<TabItem>`, so a page consumes exactly one id and "which id it got" is precisely "where in the build it rendered".

<details>
<summary><code>tab-id-repro.sh</code></summary>

```bash
#!/usr/bin/env bash
# Starlight <Tabs> ids come from a build-global counter, so they depend on where
# a page falls in the build rather than on the page itself.
#
#   ./tab-id-repro.sh                                  # latest (starlight 0.41.7 / astro 7.2.4)
#   STARLIGHT=0.36.3 ASTRO=5.18.2 ./tab-id-repro.sh    # the astro 5 line
#   PM=bun ./tab-id-repro.sh                           # npm by default
#
# It builds a stock 40-page site three times from a cold content-layer store,
# then adds one page and builds again, printing which page got which id.
set -euo pipefail
PM="${PM:-npm}"
STARLIGHT="${STARLIGHT:-0.41.7}"
ASTRO="${ASTRO:-7.2.4}"
DIR="${DIR:-$(mktemp -d)}"
PAGES="${PAGES:-10}"                  # per directory; >10 files total is what matters
DIRS="${DIRS:-one two three four}"

echo "project: $DIR  (starlight $STARLIGHT, astro $ASTRO, $PM)"
mkdir -p "$DIR/src/content/docs"
cd "$DIR"

cat > package.json <<EOF
{
  "name": "starlight-tab-id-repro",
  "type": "module",
  "private": true,
  "scripts": { "build": "astro build" },
  "dependencies": { "@astrojs/starlight": "$STARLIGHT", "astro": "$ASTRO" }
}
EOF

cat > astro.config.mjs <<'EOF'
import { defineConfig } from "astro/config";
import starlight from "@astrojs/starlight";
export default defineConfig({ integrations: [starlight({ title: "repro", pagefind: false })] });
EOF

cat > src/content.config.ts <<'EOF'
import { defineCollection } from "astro:content";
import { docsLoader } from "@astrojs/starlight/loaders";
import { docsSchema } from "@astrojs/starlight/schema";
export const collections = { docs: defineCollection({ loader: docsLoader(), schema: docsSchema() }) };
EOF

printf -- '---\ntitle: Home\n---\nHome.\n' > src/content/docs/index.md

# One <Tabs> with exactly one <TabItem> per page, so a page consumes exactly one
# id and "which id" is precisely "where in the build this page rendered".
page() {
  mkdir -p "$(dirname "$1")"
  cat > "$1" <<EOF
---
title: $2
---

<Tabs>
  <TabItem label="only">x</TabItem>
</Tabs>
EOF
}
for d in $DIRS; do
  for i in $(seq 1 "$PAGES"); do n=$(printf '%02d' "$i"); page "src/content/docs/$d/p$n.mdx" "$d-$n"; done
done

$PM install >/dev/null 2>&1
echo "$(find src/content/docs -name '*.mdx' | wc -l | tr -d ' ') pages use <Tabs>"

ids() {   # id -> page, one line each
  for f in dist/*/*/index.html; do
    id=$(grep -o 'id="tab-[0-9]*"' "$f" | sed 's/[^0-9]//g' || true)
    [ -n "$id" ] && printf '%03d %s\n' "$id" "${f#dist/}"
  done | sort -n
}
build() { rm -rf dist node_modules/.astro; $PM run build >/dev/null 2>&1; ids > "$1"; }

echo
echo "A. three builds of the same unchanged project, each from a cold store"
for b in 1 2 3; do
  build "b$b.txt"
  echo "   build $b: tab-0..5 -> $(head -6 "b$b.txt" | awk '{print $2}' | sed 's|/index.html||' | tr '\n' ' ')"
done
for b in 2 3; do
  if diff -q b1.txt "b$b.txt" >/dev/null; then echo "   build 1 == build $b"
  else echo "   build 1 != build $b  ($(diff b1.txt "b$b.txt" | grep -c '^<') of $(wc -l < b1.txt | tr -d ' ') pages got a different id)  <-- output is not reproducible"
  fi
done

echo
echo "B. add ONE new page, then build again"
page src/content/docs/aaa/p01.mdx aaa-01
build after.txt
changed=$(join -j2 <(sort -k2 b1.txt) <(sort -k2 after.txt) | awk '$2!=$3' | wc -l | tr -d ' ')
echo "   $changed of $(wc -l < b1.txt | tr -d ' ') pre-existing pages now carry a different tab id"
rm -rf src/content/docs/aaa
build restored.txt
diff -q b1.txt restored.txt >/dev/null && echo "   removing it restores every id exactly"
```

</details>

<details open>
<summary>Output — latest (starlight 0.41.7, astro 7.2.4, npm)</summary>

```
40 pages use <Tabs>

A. three builds of the same unchanged project, each from a cold store
   build 1: tab-0..5 -> four/p01 four/p02 four/p03 four/p04 four/p05 four/p06 
   build 2: tab-0..5 -> four/p01 four/p02 four/p03 four/p04 four/p05 four/p06 
   build 3: tab-0..5 -> four/p01 four/p02 four/p03 four/p04 four/p05 four/p06 
   build 1 == build 2
   build 1 == build 3

B. add ONE new page, then build again
   40 of 40 pre-existing pages now carry a different tab id
   removing it restores every id exactly

```

Three builds agree, so B does not bite here — but one added page moved all 40.

</details>

<details open>
<summary>Output — astro 5 line (starlight 0.36.3, astro 5.18.2, npm)</summary>

```
40 pages use <Tabs>

A. three builds of the same unchanged project, each from a cold store
   build 1: tab-0..5 -> four/p03 four/p05 four/p02 four/p01 four/p04 four/p06 
   build 2: tab-0..5 -> four/p01 four/p02 four/p03 four/p05 four/p06 four/p07 
   build 3: tab-0..5 -> four/p01 four/p03 four/p02 four/p04 four/p05 four/p07 
   build 1 != build 2  (31 of 40 pages got a different id)  <-- output is not reproducible
   build 1 != build 3  (33 of 40 pages got a different id)  <-- output is not reproducible

B. add ONE new page, then build again
   38 of 40 pre-existing pages now carry a different tab id
```

Here the same unchanged project builds three different ways. (The last check prints nothing because the "restored" build is itself a fresh roll of the order.)

</details>

### Participation

delucis:
#### Description

- Closes #4132  
- Uses a per-page counter so that the HTML output by `<Tabs>` does not change on one page as a side effect of adding/removing tabs in other pages.
- Before, tab HTML would include markup like `id="tab-panel-37"` indicating the 37th panel rendered during a build/dev session. These IDs now contain two indexes: one for the instance on the page, another for the panel within that instance, e.g. `id="tab-panel-1-3"` for the fourth panel in the second `<Tabs>` on a page (the counts are zero-indexed).
- Initially when reading #4132 I considered closing it as “working as intended” because tab IDs are an internal API detail and don’t matter too much. However, I think it’s valuable to fix as build systems that diff page content and upload changed files may benefit from this by not needing to reupload multiple files just because of a `<Tabs>` change on another page earlier in the build.
- Made this a patch as I don’t think there are any reasonable ways people could be using the current `id` format but happy to reconsider if someone knows of one.

### DIFF
--- a/packages/starlight/user-components/Tabs.astro
+++ b/packages/starlight/user-components/Tabs.astro
@@ -7,8 +7,6 @@ interface Props {
 }
 
 const { syncKey } = Astro.props;
-const panelHtml = await Astro.slots.render('default');
-const { html, panels } = processPanels(panelHtml);
 
 /**
  * Synced tabs are persisted across page using `localStorage`. The script used to restore the
@@ -37,6 +35,15 @@ if (isSynced) {
 	// @ts-expect-error - See above
 	Astro.locals[didRenderSyncedTabsRestoreScriptSymbol] = true
 }
+
+// Track how many <Tabs> have been rendered on the current page.
+const tabInstanceCountSymbol = Symbol.for('starlight:tab-instance-count');
+const tabInstanceCount
+// @ts-expect-error - Untyped as it’s a local and private API, see above.
+	= Astro.locals[tabInstanceCountSymbol] = ((Astro.locals[tabInstanceCountSymbol]) ?? 0) + 1;
+
+const panelHtml = await Astro.slots.render('default');
+const { html, panels } = processPanels(panelHtml, tabInstanceCount);
 ---
 
 {/* Inlined to avoid a flash of invalid active tab. */}
--- a/packages/starlight/user-components/Tabs.astro
+++ b/packages/starlight/user-components/Tabs.astro
@@ -8,39 +8,45 @@ interface Props {
 
 const { syncKey } = Astro.props;
 
-/**
- * Synced tabs are persisted across page using `localStorage`. The script used to restore the
- * active tab for a given sync key has a few requirements:
- *
- * - The script should only be included when at least one set of synced tabs is present on the page.
- * - The script should be inlined to avoid a flash of invalid active tab.
- * - The script should only be included once per page.
- *
- * To do so, we keep track of whether the script has been rendered using a variable stored using
- * `Astro.locals` which will be reset for each new page. The value is tracked using an untyped
- * symbol on purpose to avoid Starlight users to get autocomplete for it and avoid potential
- * clashes with user-defined variables.
- *
- * The restore script defines a custom element `starlight-tabs-restore` that will be included in
- * each set of synced tabs to restore the active tab based on the persisted value using the
- * `connectedCallback` lifecycle method. To ensure this callback can access all tabs and panels for
- * the current set of tabs, the script should be rendered before the tabs themselves.
- */
-const isSynced = syncKey !== undefined;
+// Some tab functionality has to be co-ordinated across multiple instances for each page. To do
+// this, we keep track of page-level state in symbol-keyed variables in `Astro.locals`, which is a
+// fresh object for each new page. The keys are not typed globally to avoid Starlight users
+// getting autocomplete for them and using symbols reduces the risk of potential clashes with
+// user-defined variables.
+//
+// We store data for:
+// 
+// - The number of tabs rendered on the current page, so that each set can have unique IDs.
+//
+// - Synced tabs, which are persisted across pages using `localStorage`. The script used to restore
+// the active tab for a given sync key has a few requirements:
+//   - The script should only be included when at least one set of synced tabs is present on the page.
+//   - The script should be inlined to avoid a flash of invalid active tab.
+//   - The script should only be included once per page.
+//
+//   The restore script defines a custom element `starlight-tabs-restore` that will be included in
+//   each set of synced tabs to restore the active tab based on the persisted value using the
+//   `connectedCallback` lifecycle method. To ensure this callback can access all tabs and panels for
+//   the current set of tabs, the script should be rendered before the tabs themselves.
+
 const didRenderSyncedTabsRestoreScriptSymbol = Symbol.for('starlight:did-render-synced-tabs-restore-script');
-// @ts-expect-error - See above
-const shouldRenderSyncedTabsRestoreScript = isSynced && Astro.locals[didRenderSyncedTabsRestoreScriptSymbol] !== true;
+const tabInstanceCountSymbol = Symbol.for('starlight:tab-instance-count');
+
+const locals = Astro.locals as App.Locals & {
+	[didRenderSyncedTabsRestoreScriptSymbol]?: boolean;
+	[tabInstanceCountSymbol]?: number;
+};
+
+const isSynced = syncKey !== undefined;
+const shouldRenderSyncedTabsRestoreScript =
+	isSynced && locals[didRenderSyncedTabsRestoreScriptSymbol] !== true;
 
 if (isSynced) {
-	// @ts-expect-error - See above
-	Astro.locals[didRenderSyncedTabsRestoreScriptSymbol] = true
+	locals[didRenderSyncedTabsRestoreScriptSymbol] = true;
 }
 
-// Track how many <Tabs> have been rendered on the current page.
-const tabInstanceCountSymbol = Symbol.for('starlight:tab-instance-count');
-const tabInstanceCount
-// @ts-expect-error - Untyped as it’s a local and private API, see above.
-	= Astro.locals[tabInstanceCountSymbol] = ((Astro.locals[tabInstanceCountSymbol]) ?? 0) + 1;
+locals[tabInstanceCountSymbol] ??= 0;
+const tabInstanceCount = locals[tabInstanceCountSymbol]++;
 
 const panelHtml = await Astro.slots.render('default');
 const { html, panels } = processPanels(panelHtml, tabInstanceCount);
--- a/.changeset/long-adults-wave.md
+++ b/.changeset/long-adults-wave.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/starlight': patch
+---
+
+Internal refactor: numbered `id` attributes in the `<Tabs>` component now increment from `1` on each page instead of using a global counter
--- a/.changeset/long-adults-wave.md
+++ b/.changeset/long-adults-wave.md
@@ -2,4 +2,4 @@
 '@astrojs/starlight': patch
 ---
 
-Internal refactor: numbered `id` attributes in the `<Tabs>` component now increment from `1` on each page instead of using a global counter
+Internal refactor: numbered `id` attributes in the `<Tabs>` component are now page-specific instead of using a global counter
--- a/packages/starlight/__tests__/markdown-processor/rehype-tabs.test.ts
+++ b/packages/starlight/__tests__/markdown-processor/rehype-tabs.test.ts
@@ -38,12 +38,12 @@ test('tab items are processed', () => {
 	const { panels, html } = processPanels(input, 0);
 
 	expect(html).toMatchInlineSnapshot(
-		`"<div id="tab-panel-0" aria-labelledby="tab-0" role="tabpanel" tabindex="0"><p>Random paragraph</p></div>"`
+		`"<div id="tab-panel-0-0" aria-labelledby="tab-0-0" role="tabpanel" tabindex="0"><p>Random paragraph</p></div>"`
 	);
 	expect(panels).toHaveLength(1);
 	expect(panels?.[0]?.label).toBe(label);
-	expect(panels?.[0]?.panelId).toMatchInlineSnapshot('"tab-panel-0"');
-	expect(panels?.[0]?.tabId).toMatchInlineSnapshot('"tab-0"');
+	expect(panels?.[0]?.panelId).toMatchInlineSnapshot(`"tab-panel-0-0"`);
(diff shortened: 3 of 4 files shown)


packages/starlight/user-components/Tabs.astro
```diff
@@ -37,6 +35,15 @@ if (isSynced) {
...
 }
+
+// Track how many <Tabs> have been rendered on the current page.
+const tabInstanceCountSymbol = Symbol.for('starlight:tab-instance-count');
+const tabInstanceCount
+// @ts-expect-error - Untyped as it’s a local and private API, see above.
+	= Astro.locals[tabInstanceCountSymbol] = ((Astro.locals[tabInstanceCountSymbol]) ?? 0) + 1;
+
+const panelHtml = await Astro.slots.render('default');
+const { html, panels } = processPanels(panelHtml, tabInstanceCount);
```
HiDeoo: Nit: I think the new additions may not be the easiest to read with the double assignments. I guess this was done to avoid using 2 `@ts-expect-error` maybe.

I guess another solution could be re-using the approach that we use in `packages/starlight/components/SidebarRestorePoint.astro` (but I cannot suggest a diff for it as part of it are outside of the review range):


```ts
/**
 * ...
 * Existing comment
 * ...
 */
const didRenderSyncedTabsRestoreScriptSymbol = Symbol.for('starlight:did-render-synced-tabs-restore-script');
// Track how many <Tabs> have been rendered on the current page.
const tabInstanceCountSymbol = Symbol.for('starlight:tab-instance-count');

const locals = Astro.locals as App.Locals & {
	[didRenderSyncedTabsRestoreScriptSymbol]?: boolean;
	[tabInstanceCountSymbol]?: number;
};

const isSynced = syncKey !== undefined;
const shouldRenderSyncedTabsRestoreScript =
	isSynced && locals[didRenderSyncedTabsRestoreScriptSymbol] !== true;

if (isSynced) {
	locals[didRenderSyncedTabsRestoreScriptSymbol] = true;
}

const tabInstanceCount = (locals[tabInstanceCountSymbol] ?? 0) + 1;
locals[tabInstanceCountSymbol] = tabInstanceCount;

const panelHtml = await Astro.slots.render('default');
const { html, panels } = processPanels(panelHtml, tabInstanceCount);
```

HiDeoo (dismissed):

I think this is a reasonable change for such use-cases and I think a patch should be fine too.

Left a tiny nit but won't block the merge, up to you if you want to address it or not.

packages/starlight/user-components/Tabs.astro
delucis: Good call. I was thinking about whether it could be typed in a way that wouldn’t leak to users but forgot we already had this approach. I used your suggestion plus some tweaks to improve the comment (I think) to better describe the full behaviour — e144fd78d1f5dc22028df69155920e8bdbd37ad1.

HiDeoo: Amazing, even better, thanks

HiDeoo (approved):

PR #4134, opened 2026-08-20
Use Sätteri 0.10 APIs

HiDeoo:
#### Description

This PR refactors some of the code to use the new Sätteri 0.10 APIs now available in Astro 7.2.4 (but technically 7.2.5 for a fix).

One of the most meaningful change is our components that were previously using rehype and creating processors to parse the default slot content are now using the new Sätteri `htmlToHast()` API.

Regarding the changeset, not quite sure yet the level of details we should use, e.g. should we just say we updated some deps or go as far as mentioning that we can now only check paths to transform once per plugin type thanks to Sätteri conditional plugin bundles.

Note that GitHub displays some renamed files as deletions + additions because they changed significantly (mostly indentations tho). The [rename-only commit](https://github.com/withastro/starlight/pull/4134/changes/4233e31954ce1ff83f4aac03efc9deefff1f3955) may be easier to review separately.

#### Remaining tasks

- [x] Wait for https://github.com/withastro/astro/pull/17766 to be released.
- [x] Will need some updates once https://github.com/withastro/starlight/pull/4133 is merged.
- [x] Adress all `TODO(HiDeoo)` comments.


packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
```diff
@@ -34,16 +34,10 @@ describeEachProcessor('code RTL support', (ctx, name) => {
 		expect(res.code).not.includes('dir="auto"');
 	});
 
-	test('handles `dir="ltr"` on raw `<pre>` HTML per processor', async () => {
+	test('does nto apply `dir="ltr"` on raw `<pre>` HTML', async () => {
```
delucis: Typo nit:
```suggestion
	test('does not apply `dir="ltr"` on raw `<pre>` HTML', async () => {
```

delucis (commented):
This looks great, @HiDeoo! Thanks for taking the time to bring these all up-to-date for the bright and glorious Sätteri future.

The diff is a little harder to read because Git doesn’t recognise things as renames that it probably should. For anyone else curious, I ended up copy-pasting the before/after processor files into https://www.diffchecker.com/ to visualise what actually changed (not much, very clean).

I guess we’ll probably be able to remove the TODO in `pnpm-workspace.yaml` by merge time given this will be in a minor, but everything else looks good to me here 🙌

packages/starlight/__tests__/markdown-processor/code-rtl-support.test.ts
HiDeoo: Ooops, good catch 🕵️

pnpm-workspace.yaml
```diff
@@ -25,3 +25,8 @@ peerDependencyRules:
 
 # Prevent installing packages published in the last 3 days
 minimumReleaseAge: 4320
+# TODO(HiDeoo) Remove
+minimumReleaseAgeExclude:
+  - '@astrojs/markdown-satteri@0.3.8'
+  - '@astrojs/mdx@7.0.8'
+  - 'astro@7.2.5'
```
delucis: I think it should be safe to remove these now 🎉 
```suggestion
```

delucis (dismissed):

delucis (approved):

PR #4138, opened 2026-08-21
Fix localisation of Expressive Code UI

delucis:
#### Description

- When adding support for the Sätteri Markdown processor, we missed that because Sätteri provides a URL in its file object, we were treating it with the same logic we used for the `<Code>` component when trying to detect the locale (which passes `Astro.url`). Because the assumptions are quite different this broke localisation of Expressive Code code blocks when using with the Sätteri processor.
- This PR fixes that by detecting Sätteri’s `file://` URLs and ignoring them.
- I’ve added some new unit tests for `getBlockLocale()` although I’m a bit conflicted because these test don’t actually check the interface between real world inputs and the `getBlockLocale()` method, so I suppose a future change to the `file` object passed by a Markdown processor could still cause a similar regression, but at least this is better than nothing.
- Doing this also brought these files newly into view for Vitest, so dropped our coverage down. I’ve added some additional tests for other parts of the Expressive Code config processing logic to keep coverage high.

(It might be best to hide whitespace changes while reviewing the tests because I wrapped some existing tests in this file in a `describe()` block but otherwise they are unchanged.)

### DIFF
--- a/.changeset/modern-dragons-reply.md
+++ b/.changeset/modern-dragons-reply.md
@@ -2,4 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
-Fixes localisation of code block UI elements
+Fixes localisation of code block UI elements when using the Sätteri Markdown processor
--- a/.changeset/modern-dragons-reply.md
+++ b/.changeset/modern-dragons-reply.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fixes localisation of code block UI elements
--- a/packages/starlight/__tests__/expressive-code/preprocessor.test.ts
+++ b/packages/starlight/__tests__/expressive-code/preprocessor.test.ts
@@ -1,0 +1,218 @@
+import type { ExpressiveCodeTheme, StyleVariant } from 'astro-expressive-code';
+import { describe, expect, test, vi } from 'vitest';
+import { getStarlightEcConfigPreprocessor } from '../../integrations/expressive-code/preprocessor';
+import { StarlightConfigSchema, type StarlightUserConfig } from '../../utils/user-config';
+
+describe('getStarlightEcConfigPreprocessor()', () => {
+	test('warns when using `useStarlightUiThemeColors` with a single theme', async () => {
+		const consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+		await preprocessEcConfig({ themes: ['starlight-light'], useStarlightUiThemeColors: true });
+
+		expect(consoleWarnSpy).toHaveBeenCalledWith(
+			expect.stringContaining(
+				'Using the config option "useStarlightUiThemeColors: true" with a single theme is not recommended.'
+			)
+		);
+
+		consoleWarnSpy.mockRestore();
+	});
+});
+
+describe('Expressive Code config', () => {
+	describe('themeCssSelector()', () => {
+		test('returns a custom theme selector for a single custom theme', async () => {
+			const config = await preprocessEcConfig();
+			if (typeof config.themeCssSelector !== 'function') {
+				throw new Error('Expected `themeCssSelector` to be a function');
+			}
+
+			const selector = config.themeCssSelector({ name: 'test-theme' } as ExpressiveCodeTheme, {
+				styleVariants: [],
+			});
+
+			expect(selector).toBe("[data-theme='test-theme']");
+		});
+
+		test('returns Starlight theme switcher-compatible selectors for a dark/light theme pair', async () => {
+			const config = await preprocessEcConfig();
+			if (typeof config.themeCssSelector !== 'function') {
+				throw new Error('Expected `themeCssSelector` to be a function');
+			}
+
+			const darkStyle = { theme: { name: 'test-theme', type: 'dark' } } as StyleVariant;
+			const lightStyle = { theme: { name: 'test-theme-light', type: 'light' } } as StyleVariant;
+			const context = { styleVariants: [darkStyle, lightStyle] };
+
+			const darkSelector = config.themeCssSelector(darkStyle.theme, context);
+			expect(darkSelector).toBe("[data-theme='dark']");
+			const lightSelector = config.themeCssSelector(lightStyle.theme, context);
+			expect(lightSelector).toBe("[data-theme='light']");
+		});
+
+		test('returns theme name selector if dark/light toggle is disabled', async () => {
+			const config = await preprocessEcConfig({ useStarlightDarkModeSwitch: false });
+			if (typeof config.themeCssSelector !== 'function') {
+				throw new Error('Expected `themeCssSelector` to be a function');
+			}
+
+			const darkStyle = { theme: { name: 'test-theme', type: 'dark' } } as StyleVariant;
+			const lightStyle = { theme: { name: 'test-theme-light', type: 'light' } } as StyleVariant;
+
+			const selector = config.themeCssSelector(darkStyle.theme, {
+				styleVariants: [darkStyle, lightStyle],
+			});
+
+			expect(selector).toBe("[data-theme='test-theme']");
+		});
+
+		test('returns theme name selector for additional themes', async () => {
+			const config = await preprocessEcConfig();
+			if (typeof config.themeCssSelector !== 'function') {
+				throw new Error('Expected `themeCssSelector` to be a function');
+			}
+
+			const darkStyle = { theme: { name: 'test-theme-dark', type: 'dark' } } as StyleVariant;
+			const lightStyle = { theme: { name: 'test-theme-light', type: 'light' } } as StyleVariant;
+			const thirdStyle = { theme: { name: 'test-theme-third', type: 'dark' } } as StyleVariant;
+			const fourthStyle = { theme: { name: 'test-theme-fourth', type: 'light' } } as StyleVariant;
+			const context = { styleVariants: [darkStyle, lightStyle, thirdStyle, fourthStyle] };
+
+			expect(config.themeCssSelector(darkStyle.theme, context)).toBe("[data-theme='dark']");
+			expect(config.themeCssSelector(lightStyle.theme, context)).toBe("[data-theme='light']");
+			expect(config.themeCssSelector(thirdStyle.theme, context)).toBe(
+				"[data-theme='test-theme-third']"
+			);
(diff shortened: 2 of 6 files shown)


trueberryless:
Could we add e2e tests to avoid regressions or is this not possible for some reason?

Oh, I just read [HiDeoo's comment](https://discord.com/channels/830184174198718474/1070481941863878697/1540397192450150441). Would an e2e test be too heavy?

delucis:
> Could we add e2e tests to avoid regressions or is this not possible for some reason?
> 
> Oh, I just read [HiDeoo's comment](https://discord.com/channels/830184174198718474/1070481941863878697/1540397192450150441). Would an e2e test be too heavy?

Yes, it’s possible, but in general we try to avoid end-to-end tests as much as possible because they are slower and only use them for things that are impossible to test in other ways (in-browser interaction for example) and feel fragile enough that the cost of an end-to-end test feels valuable. I’m not totally opposed to it here, but we’re only really interested in the HTML output of Expressive Code in this case, not also running the page in the browser, so it does feel a bit overkill.

.changeset/modern-dragons-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Fixes localisation of code block UI elements
```
HiDeoo, line 5: Should we be a bit more precise? Up to you.


```suggestion
Fixes localisation of code block UI elements when using the Sätteri Markdown processor
```

packages/starlight/__tests__/expressive-code/preprocessor.test.ts
```diff
@@ -0,0 +1,218 @@
+import type { ExpressiveCodeTheme, StyleVariant } from 'astro-expressive-code';
+import { describe, expect, test, vi } from 'vitest';
+import { getStarlightEcConfigPreprocessor } from '../../integrations/expressive-code/preprocessor';
```
HiDeoo, line 3:
```suggestion
import type { StarlightExpressiveCodeOptions } from '../../integrations/expressive-code';
import { getStarlightEcConfigPreprocessor } from '../../integrations/expressive-code/preprocessor';
```

Needed later to fix an issue.

packages/starlight/__tests__/expressive-code/preprocessor.test.ts
```diff
@@ -0,0 +1,218 @@
...
+			const customizedTheme = config.customizeTheme?.({
+				name: 'original-theme',
+				colors: {},
+				styleOverrides: {},
+			} as ExpressiveCodeTheme);
+			expect(customizedTheme?.name).toBe('custom-theme');
+		});
+
+		test('is a no-op if `customizeTheme()` function is not set and `applyStarlightUiThemeColors` is false', async () => {
+			const config = await preprocessEcConfig({ applyStarlightUiThemeColors: false });
```
HiDeoo, line 177:
```suggestion
		test('is a no-op if `customizeTheme()` function is not set and `useStarlightUiThemeColors` is false', async () => {
			const config = await preprocessEcConfig({ useStarlightUiThemeColors: false });
```

The option name is `useStarlightUiThemeColors`.

packages/starlight/__tests__/expressive-code/preprocessor.test.ts
```diff
@@ -0,0 +1,218 @@
...
+			const theme = { name: 'my-theme', colors: {}, styleOverrides: {} } as ExpressiveCodeTheme;
+			const customizedTheme = config.customizeTheme?.(theme);
+			expect(customizedTheme).toBe(theme);
```
HiDeoo, line 180:
```suggestion
			expect(customizedTheme).toStrictEqual({ name: 'my-theme', colors: {}, styleOverrides: {} });
```

I guess here to have caught the option error, we want to compare against a literal as `theme` is mutated.

packages/starlight/__tests__/expressive-code/preprocessor.test.ts
```diff
@@ -0,0 +1,218 @@
...
+		});
+	});
+});
+
+/** Preprocess an Expressive Code config using the Starlight preprocessor. */
+async function preprocessEcConfig(ecConfig: unknown = {}) {
```
HiDeoo, line 186:
```suggestion
async function preprocessEcConfig(ecConfig: StarlightExpressiveCodeOptions = {}) {
```

To more easily catch the invalid option, we can type the config here.

packages/starlight/__tests__/expressive-code/preprocessor.test.ts
```diff
@@ -0,0 +1,218 @@
...
+		root: { label: 'English', lang: 'en' },
+		fr: { label: 'French', lang: 'fr' },
+	});
+	const configPreprocessor = getStarlightEcConfigPreprocessor({
+		docsPath: '/path/to/docs/',
+		starlightConfig,
+		useTranslations,
+	});
+	// @ts-expect-error — We’re skipping some properties of `astroConfig`.
+	return await configPreprocessor({ ecConfig, astroConfig: {} });
```
HiDeoo, line 197:
```suggestion
	return await configPreprocessor({
		ecConfig,
		// @ts-expect-error — We’re skipping some properties of `astroConfig`.
		astroConfig: {},
	});
```

Nit: in case something is ever wrong with `ecConfig`, e.g. it's removed, it's not silenced by `@ts-expect-error`.

packages/starlight/__tests__/expressive-code/theming.test.ts
```diff
@@ -0,0 +1,49 @@
+import { describe, expect, test } from 'vitest';
```
HiDeoo, line 1:
```suggestion
import { describe, expect, test } from 'vitest';
import { ExpressiveCodeTheme } from 'astro-expressive-code';
```

For a change below.

packages/starlight/__tests__/expressive-code/theming.test.ts
```diff
@@ -0,0 +1,49 @@
...
+		const result = preprocessThemes(['nord', 'github-light']);
+		expect(result).toEqual(['nord', 'github-light']);
+	});
+});
+
+describe('applyStarlightUiThemeColors', () => {
+	test('applies the correct colors to a dark theme', () => {
+		const [darkTheme] = preprocessThemes(['starlight-dark']);
+
+		// @ts-expect-error - In theory preprocessThemes can return other shapes, but it’s fine in this case.
```
HiDeoo, line 31: Here we are not testing an invalid input, so we might as well assert it's what we expect.


```suggestion
		expect.assert.instanceOf(darkTheme, ExpressiveCodeTheme);
```

Not using `exepect().toBeInstanceOf()` as this does not narrow the type.

packages/starlight/__tests__/expressive-code/theming.test.ts
```diff
@@ -0,0 +1,49 @@
...
+		expect(result.colors['titleBar.activeBackground']).toBe('var(--sl-color-black)');
+		expect(result.colors['editorGroupHeader.tabsBorder']).toBe(
+			'color-mix(in srgb, var(--sl-color-gray-5), transparent 25%)'
+		);
+	});
+
+	test('applies the correct colors to a light theme', () => {
+		const [lightTheme] = preprocessThemes(['starlight-light']);
+
+		// @ts-expect-error - In theory preprocessThemes can return other shapes, but it’s fine in this case.
```
HiDeoo, line 42:
```suggestion
		expect.assert.instanceOf(lightTheme, ExpressiveCodeTheme);
```

And same here.

HiDeoo (changes requested):
Thanks for the PR :raised_hands:

We already discussed about the fix on Discord, and it still looks good to me.

I left some changes regarding 1 issue in a test and some minor nits which are up to you to apply or not.

> I’ve added some new unit tests for `getBlockLocale()` although I’m a bit conflicted because these test don’t actually check the interface between real world inputs and the `getBlockLocale()` method, so I suppose a future change to the `file` object passed by a Markdown processor could still cause a similar regression, but at least this is better than nothing.

I played around with the idea of creating more of an integration test (rather than a unit or e2e one) that would have caught the issue. I ended up with:

```ts
// packages/starlight/__tests__/expressive-code/preprocessor.test.ts

// Existing imports

import { unified } from '@astrojs/markdown-remark';
import { satteri } from '@astrojs/markdown-satteri';
import { astroExpressiveCode } from 'astro-expressive-code';
import { getCollectionPosixPath } from '../../utils/collection-fs';
import { createPluginTestOptions, docFileURL } from '../test-utils';

// Existing tests

describe('localization', () => {
	const processors = [
		{ name: 'Sätteri', create: () => satteri() },
		{ name: 'Unified', create: () => unified() },
	];

	test.for(processors)('localizes Expressive Code UI with $name', async ({ create }) => {
		const userConfig: StarlightUserConfig = {
			title: 'Expressive Code',
			locales: {
				root: { label: 'English', lang: 'en' },
				fr: { label: 'French', lang: 'fr' },
			},
		};

		const { astroConfig, useTranslations } = await createPluginTestOptions(userConfig);
		const starlightConfig = StarlightConfigSchema.parse(userConfig);

		const processor = create();

		const ecIntegration = astroExpressiveCode({
			customConfigPreprocessors: {
				preprocessAstroIntegrationConfig: getStarlightEcConfigPreprocessor({
					docsPath: getCollectionPosixPath('docs', astroConfig.srcDir),
					starlightConfig,
					useTranslations,
				}),
				preprocessComponentConfig: '',
			},
		});

		await ecIntegration.hooks['astro:config:setup']({
			config: {
				root: astroConfig.root,
				integrations: [ecIntegration],
				markdown: { processor },
			},
			addWatchFile: () => {},
			updateConfig: () => {},
		});

		const renderer = await processor.createRenderer({ syntaxHighlight: false });

		const result = await renderer.render('```js\nconsole.log("bonjour")\n```', {
			fileURL: docFileURL('fr/index.md'),
		});

		expect(result.code).toContain('title="Copier dans le presse-papiers"');
	});
});
```

Any thoughts on this?

delucis:
> I played around with the idea of creating more of an integration test (rather than a unit or e2e one)

Amazing! I added this more or less verbatim in 24b19ddb17486a4f0b067454d3e7f6ea31bce0c7.

Thanks for the detailed review 💖

HiDeoo (approved):
Thanks for the updates, looks ready to :rocket: to me.

PR #4139, opened 2026-08-22
chore: migrate Changesets to v3

trueberryless:
#### Description

This PR migrates from Changesets v2 to v3, following the [migration guide](https://changesets.dev/guide/migration). The main points to highlight:

- The `cli` was updated to `3.0.1` and the `changelog-github` to `1.0.0`
- The `config.json` schema was updated to `4.0.0`
- The `ignore` option in `config.json` was removed because all those packages have `private: true`, which are now [ignored automatically](https://changesets.dev/guide/migration#opt-in-to-versioning-private-packages).
- Other options have been left untouched alto some are already default, as discussed with Chris on Discord.
- [The action](https://github.com/changesets/action) has been updated and pinned to `v2.1.1`, which support the `v3` cli, and all parameters have been renamed.
- I did not split the action into sub-actions to keep the initial migration minimal. I propose to follow-up with another PR once successful publishing is confirmed to switch to the recommended approach from the [Trusted Publishing guide](https://changesets.dev/guide/automating#trusted-publishing).
- Erika also [confirmed (Discord)](https://discord.com/channels/830184174198718474/1070481941863878697/1540373296971784212) that the empty `NPM_TOKEN` workaround regarding trusted publishing (provenance) can now safely be removed, as [she did it in the prettier plugin](https://github.com/withastro/prettier-plugin-astro/pull/470/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34).
- I had to update `pnpm` to `11.1.2` because of some [installation issue with cached time metadata](https://github.com/pnpm/pnpm/issues/11619), which makes the `minReleaseAge` check fail. If you have a better suggestion, I am very open for this as this change is pretty unrelated. Latest pnpm version is currently `11.22`, but I didn't want a minor update in this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,12 +36,9 @@ jobs:
         run: pnpm i
 
       - name: Create Release Pull Request
-        uses: changesets/action@a45c4d594aa4e2c509dc14a9f2b3b67ba3780d0d # v1.9.0
+        uses: changesets/action@8488615a623b1b9c987934bb89eae8af6a946ac1 # v2.1.1
         with:
-          version: pnpm run version
-          publish: pnpm changeset publish
-          commit: '[ci] release'
-          title: '[ci] release'
-        env:
-          GITHUB_TOKEN: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
-          NPM_TOKEN: "" # See https://github.com/changesets/changesets/issues/1152#issuecomment-3190884868
+          version-script: pnpm run version
+          commit-message: '[ci] release'
+          pr-title: '[ci] release'
+          github-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,7 @@ jobs:
         uses: changesets/action@8488615a623b1b9c987934bb89eae8af6a946ac1 # v2.1.1
         with:
           version-script: pnpm run version
+          publish-script: pnpm changeset publish
           commit-message: '[ci] release'
           pr-title: '[ci] release'
           github-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "typescript": "^6.0.3",
     "typescript-eslint": "^8.59.2"
   },
-  "packageManager": "pnpm@11.1.2",
+  "packageManager": "pnpm@11.22.0",
   "size-limit": [
     {
       "name": "/index.html",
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -36,12 +36,10 @@ jobs:
...
-          commit: '[ci] release'
-          title: '[ci] release'
-        env:
-          GITHUB_TOKEN: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
-          NPM_TOKEN: "" # See https://github.com/changesets/changesets/issues/1152#issuecomment-3190884868
+          version-script: pnpm run version
+          publish-script: pnpm changeset publish
+          commit-message: '[ci] release'
+          pr-title: '[ci] release'
+          github-token: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
```
delucis, line 45: Should this have changed? I left it as-is when migrating my personal repos, but maybe there’s a good reason you updated it?
```suggestion
        env:
          GITHUB_TOKEN: ${{ secrets.FREDKBOT_GITHUB_TOKEN }}
```

package.json
```diff
@@ -29,7 +29,7 @@
     "typescript": "^6.0.3",
     "typescript-eslint": "^8.59.2"
   },
-  "packageManager": "pnpm@11.1.1",
+  "packageManager": "pnpm@11.1.2",
```
delucis: I think it may in fact be best to update to a more recent version — the early versions of v11 had various issues, but I’ve experienced somewhat more reliable behaviour in the later minor releases. What do you think?
```suggestion
  "packageManager": "pnpm@11.22.0",
```

delucis (commented):
Thanks for taking care of this @trueberryless! Left a couple of questions, but this mostly looks good to me 🙌

.github/workflows/release.yml
trueberryless: The GitHub token should be passed via `github-token` parameter since v2 as documented [here](https://github.com/changesets/action#api). I [discussed with Bluwy](https://github.com/changesets/changesets/pull/2259#issuecomment-5381979006) whether to document this in the migration guide, but since it is not necessary anymore if you used the default token before, we decided to leave it out completely.

package.json
trueberryless: Sure we can do that. I just didn't want to bump a minor initially in this PR since the change is unrelated to the Changeset migration itself.

.github/workflows/release.yml
delucis: Thanks for double checking!

delucis (approved):
Thanks for taking care of this for us @trueberryless 💖

PR #4143, opened 2026-08-24
[i18nIgnore] docs: reduce 3rd party content weight in search results

HiDeoo:
#### Description

Based on this [Discord message](https://discord.com/channels/830184174198718474/1070481941863878697/1541171518132977664) where a user searched for "changelog" in the documentation and miss it as a plugin ranks higher than the official documentation, this PR experiments with reducing the weight of 3rd party content in search results to favor official documentation.

Right now, the weight is set to 0.75, and considering it's a quadratic scale, this should roughly be a 45% reduced weight for 3rd party content.

This fixes the "changelog" case, but I'm keeping it as a draft for now as I want to play around with more examples and see if it's a good balance or not.

Probably easier to review with whitespace changes disabled, as it's mostly just wrapping some sections in a new `<ThirdPartyContent>` component.


delucis (commented):
Neat idea to use a component for this. I’d wondered if we should revive the idea a frontmatter option for weighting (see #2371), but it does make sense that this can sometimes apply just to a page section, not only at the page level.

This LGTM. Maybe once we confirm the approach, it would make sense to apply to all languages and use an `i18nIgnore` prefix?

HiDeoo:
> Neat idea to use a component for this. I’d wondered if we should revive the idea a frontmatter option for weighting (see #2371), but it does make sense that this can sometimes apply just to a page section, not only at the page level.

I did indeed took a look at https://github.com/withastro/starlight/pull/2371 but ended up switching to a component version as I didn't want to reduce the weight for official plugins and official videos. But I think we could probably revive and put this PR over the finish line at some point at seems a fair use case.

> This LGTM. Maybe once we confirm the approach, it would make sense to apply to all languages and use an `i18nIgnore` prefix?

Great idea, will do that once I'm done checking a few more search terms and before marking as ready for review 👍

HiDeoo:
Updated the PR:

- Did a few more tests using some terms that are both used in 3rd party content but also in the official docs, and I didn't find any downside to the current approach. I guess there is a world where the `0.75` weight becomes not enough, e.g. with "changelog" if many more plugins were to use it, but we can always adjust the weight if that happens.
- For showcase and themes pages, I moved the new component in `<ShowcaseSites>` and `<ThemeGrid>` as they always represent 3rd party content and this avoid editing files where they are used.
- Updated all translation files and added the `[i18nIgnore]` prefix to the PR title so that we merge it with it.

delucis (dismissed):
Looks great! There’s one conflict now in the Korean page but other than that, perfect.

> I guess there is a world where the `0.75` weight becomes not enough, e.g. with "changelog" if many more plugins were to use it, but we can always adjust the weight if that happens.

Yeah. We could also pick a lower weight to start, but if `0.75` is working well at the moment it’s probably fine to stick with that and adjust as and when necessary.

HiDeoo:
Fixed conflicts in German, French, and Korean so we should be good to go.

Thanks for the review.

delucis (approved):
LGTMBMC!

<sup>(Let's Get This Merged Before More Conflicts!)</sup>

PR #4141, opened 2026-08-24
docs: add starlight-custom-navigation to plugins showcase

frostybee:
Adds [starlight-custom-navigation](https://github.com/frostybee/starlight-custom-navigation) to  the community plugins list.

This plugin adds side navigation strips, floating buttons, keyboard shortcuts, and swipe gestures  for moving between pages.

PR #4130, opened 2026-08-17
Add starlight-pydocs and starlight-quiz

ewels:
#### Description

Adds new LinkCard entries for two new plugins to the starlight [community plugins](https://starlight.astro.build/resources/plugins/) section:

- [starlight-pydocs](https://github.com/ewels/starlight-pydocs), a Starlight version of [mkdocstrings](https://mkdocstrings.github.io/) that uses Griffe for Python static analysis and then builds the Starlight pages from that JSON.
- [starlight-quiz](https://github.com/ewels/starlight-quiz), a Starlight version of [mkdocs-quiz](https://ewels.github.io/mkdocs-quiz/)

Convention seemed to be to link the GitHub source repositories, so that's what I did. If you prefer, I can link the docs websites instead:

- https://ewels.github.io/starlight-pydocs/
- https://ewels.github.io/starlight-quiz/

Thank you for Astro and Starlight! 🌟

PR #4147, opened 2026-08-25

ArmandPhilippot:
#### Description

Adds changes from #4130 and #4141 to the French translation of `resources/plugins.mdx`.

Note that I'm not aware of any common translation for "side navigation strips".
I chose "bandeaux de navigation latéraux" because using "barre" would be confusing (sidebar, not exactly the same thing)... but we could use "bande". I don't have a strong opinion on this.


delucis (approved):
LGTM! For what it’s worth, I don’t think “navigation strips” is a common phrase/concept in English either, so I would definitely feel free to choose the translation flexibly.

Up to you if you want to wait for @HiDeoo’s more expert feedback on that topic or not. Thanks @ArmandPhilippot 🙌

PR #4148, opened 2026-08-25
perf: optimize route and sidebar lookups

ematipico:
#### Description

Replaced repeated route and sidebar scans with maps, sets, and small caches. This reduces build work on documentation sites with many pages and avoids attaching unused route data to static paths.

This PR doesn't affect anything visual


delucis (dismissed):
These changes all make sense to me! Although it’s a bit hard to reason about exactly what the performance improvement will look like. Maybe as suggested by @HiDeoo on Discord we should wait to merge this after setting up some benchmarks so we can see clearly how we’ve improved.

delucis (approved):
Approving again because I added a changeset.

ematipico:
Ran the benchmarks on my machine:
- `route_data` 35.8x faster
- `sidebar` 1,409x faster

PR #4150, opened 2026-08-25
perf: clean up language select lifecycle

ematipico:
#### Description

Language selector listeners now attach when the component connects and are removed when it disconnects. This prevents `change` and `pageshow` handlers from accumulating across client-side navigations.

This change is more a proposal. It doesn't affect Starlight today, however if this component is placed inside a `ClientRouter`, the event handlers aren't cleaned after a navigation, so they leak. If you are interested, I have more PRs similar to this.

### DIFF
--- a/packages/starlight/components/LanguageSelect.astro
+++ b/packages/starlight/components/LanguageSelect.astro
@@ -31,26 +31,43 @@ function localizedPathname(locale: string | undefined): string {
 
 <script>
 	class StarlightLanguageSelect extends HTMLElement {
-		constructor() {
-			super();
+		private controller: AbortController | undefined;
+
+		connectedCallback() {
+			this.controller?.abort();
+			this.controller = new AbortController();
+			const { signal } = this.controller;
 			const select = this.querySelector('select');
 			if (select) {
-				select.addEventListener('change', (e) => {
-					if (e.currentTarget instanceof HTMLSelectElement) {
-						window.location.pathname = e.currentTarget.value;
-					}
-				});
-				window.addEventListener('pageshow', (event) => {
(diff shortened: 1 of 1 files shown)


delucis (commented):
I’ll wait for others’ to chime in, but generally the current behaviour is intended: we don’t support the `ClientRouter` officially intentionally, because it’s not really needed for docs and introduces complexity to the code and increased client JS size which the vast majority of users won’t benefit from. (That’s why there’s no simple option for adding it.)

So my instinct is to say that we would either need to comprehensively go through all our client-side scripts to ensure compatibility with the router and decide it’s an officially supported thing, potentially with a config option to toggle it on without overrides, or otherwise keep the current smaller, simpler scripts and continue to recommend people avoid the client router.

HiDeoo:
> but generally the current behaviour is intended: we don’t support the `ClientRouter` officially intentionally, because it’s not really needed for docs and introduces complexity to the code and increased client JS size which the vast majority of users won’t benefit from.

I personally agree with this approach, and we have a plugin that allow users interested in using the client router and view transitions to do so quite easily. It's also maintained by Martin, the view transition master, so I definitely trust it to provide the best experience and support possible.

We have also put some efforts into [some documentation](https://events-3bg.pages.dev/jotter/starlight/) to easily guide users through the various options available to them, from using the `<ClientRouter>` to using the native View Transition API so unless the `<ClientRouter>` becomes a more widely requested feature in documentation, I would personally prefer to keep the current approach for now and the opt-in plugin for those who want to use it.

ematipico:
Perfect, good to know. Thank you

PR #4152, opened 2026-08-25
[i18nIgnore] docs: update theme links

trueberryless:
#### Description

This PR updates all my theme website links across all languages to the generic `netlify.app` domain since I myself migrated from `trueberryless.org` to `felixs.dev`.


delucis (approved):
Thanks for the update! I added the `i18nIgnore` keyword so this doesn’t trigger Lunaria changes.

PR #4157, opened 2026-08-26
chore: add benchmarks

ematipico:
#### Description

This PR benchmarks to starlight, which can be run locally and via CodSpeed.

They are four benchmarks:
- many starlight components in one page
- locales
- deep nested sidebar
- `generateRouteData`

The latter is a "functional" benchmark, meaning it needs a different setup that the others need, and that's why there are two separate configs.


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22.23.1
+          cache: pnpm
+      - run: pnpm i --filter @astrojs/starlight --filter @bench/components --filter @bench/large-sidebar --filter @bench/locales
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@562878fc153dee6c0f3719c99b1799334cfc3ccc # v5
```
delucis: I couldn’t find this SHA in their repo? This would be the most recent release from before our 3-day embargo window:

```suggestion
        uses: CodSpeedHQ/action@4296e51e7041e24dadb86d1d6e8b9320d223dbe8 # v5.0.3
```

.gitignore
```diff
@@ -22,3 +22,5 @@ test-results/
 
 # Vercel output
 .vercel
+
+.idea
```
delucis, line 26: Was this change intentional?

packages/starlight/__tests__/test-config.ts
```diff
@@ -13,6 +13,7 @@ export async function defineVitestConfig(
 		build?: Pick<AstroConfig['build'], 'format'>;
 		trailingSlash?: AstroConfig['trailingSlash'];
 		command?: 'dev' | 'build' | 'preview';
+		snapshotSerializers?: boolean;
```
delucis: This is a bit of a nit, but it might be clearer how this works if we name this `disableSnapshotSerializers` and set it to `true` in the benchmarks currently passing `false`.

packages/starlight/vitest.functional.bench.config.ts
```diff
@@ -0,0 +1,26 @@
+import codspeedPlugin from '@codspeed/vitest-plugin';
+import { defineVitestConfig } from './__tests__/test-config';
+
+const testConfig = await defineVitestConfig(
+	{
+		title: 'Starlight benchmarks',
+		sidebar: [{ label: 'Reference', items: [{ autogenerate: { directory: 'reference' } }] }],
+	},
+	{ snapshotSerializers: false }
```
delucis: Might also be good to have a comment here explaining why these serializers cause issues and need to be disabled.

packages/starlight/vitest.e2e.bench.config.ts
```diff
@@ -0,0 +1,13 @@
+import codspeedPlugin from '@codspeed/vitest-plugin';
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	plugins: [codspeedPlugin()],
+	benchmark: {
```
delucis: There’s a type error here currently: https://github.com/withastro/starlight/actions/runs/32977984781/job/98207323537?pr=4157#step:7:12

> Object literal may only specify known properties, and 'benchmark' does not exist in type 'ViteUserConfigExport'.

packages/starlight/vitest.e2e.bench.config.ts
```diff
@@ -0,0 +1,13 @@
...
+		include: ['__bench_e2e__/*.bench.ts'],
+	},
+	test: {
+		fileParallelism: false,
+		hookTimeout: 600_000,
```
delucis: This is 10 minutes. I guess we don’t expect things to take that long so maybe a lower timeout could make sense?

packages/starlight/vitest.functional.bench.config.ts
```diff
@@ -0,0 +1,26 @@
...
+	return {
+		...config,
+		plugins: [...(config.plugins ?? []), codspeedPlugin()],
+		benchmark: {
+			include: ['__bench_fn__/*.bench.ts'],
+		},
+		test: {
+			...config.test,
+			fileParallelism: false,
+			hookTimeout: 600_000,
```
delucis: Same comment here, maybe should be a lower timeout?

HiDeoo:
I did not review all files yet, but I have to run for now so sharing what I found so far.

- We need to add new directories in the `packages/starlight/.npmignore` file.
- Looks like this introduces a peer dependency issue as `@codspeed/vitest-plugin` [does not yet officially support Vite 8](https://github.com/CodSpeedHQ/codspeed-node/blob/main/packages/vitest-plugin/package.json#L43) ([draft PR opened](https://redirect.github.com/CodSpeedHQ/codspeed-node/pull/74) but not yet merged)
- I wonder if we should experiment with the `walltime` mode for builds ([reference](https://codspeed.io/docs/features/macro-runners#when-to-use-macro-runners)).
  > Macro runners are ideal when you need to measure
  > - System calls and I/O operations that are excluded from the CPU Simulation instrument
  > - End-to-end performance including network, disk, and system interactions
  > - Walltime benchmarks where actual execution time matters
  > - Integration tests that require a complete system environment
- To follow up on [Chris's message](https://discord.com/channels/830184174198718474/1070481941863878697/1542201934465863751), I _think_ there may be an issue with the way the PR setup things where in CodSpeed, our `setup` and `teardown` may not even called. There is a [old closed and unmerged PR to fix this](<https://github.com/CodSpeedHQ/codspeed-node/pull/46>) and the current code [here](<https://github.com/CodSpeedHQ/codspeed-node/blob/main/packages/vitest-plugin/src/analysis.ts#L42-L59>) and [here](<https://github.com/CodSpeedHQ/codspeed-node/blob/main/packages/vitest-plugin/src/analysis.ts#L72-L84>) still seems to only call `beforeAll`, `beforeEach`,  `afterEach`, and `afterAll`(but not `setup` and `teardown` used in the PR).

Depending on the time it ends up taking in the end, we should also see if we need to consider triggering benchmarks on labels like pre-releases to avoid running it on too many PRs unnecessarily, e.g. PRs adding or updating translations, adding an icon, CSS updates, etc.

delucis (commented):
Thanks for the idea to set this up and leading the effort @ematipico 🙌 

I left some detailed comments.

Also one high-level thought: at least in our initial run, the three benchmarks for `astro build` gave basically identical results (24.9–25.0 seconds each). I wonder if that suggests we could just combine the three into one and get just as much benefit? It suggests there is a negligible difference between the components/sidebars/locales customizations so I guess we’re mostly measuring other parts of the build 3 times? What do you think?

.gitignore
ematipico: Yes, because I use JetBrains IDEs

delucis (approved):
Thanks for cleaning that up @ematipico! If @HiDeoo agrees, I’d like to get this merged so we can test #4148 with it.

HiDeoo (approved):
Of course, let's go ahead so we can measure improvements in https://github.com/withastro/starlight/pull/4148 and later we can re-use this PR as a template for re-experimenting with build benchmarks.

Thanks for setting all this up :raised_hands:

PR #4159, opened 2026-08-27
feat: add npmx icon

trueberryless:
#### Description

Although I can already feel custom icons tingling in my fingertips, I would like to add one of my favorite icons of 2026: `npmx`.


delucis (approved):
Wow, that’s a very confusing icon 😅 

I guess some people know that `./` can mean npmx? Happy to add this and we can update it if they ever switch to something a bit more recognisable.

(Will merge when we do our next release so that the icon doesn’t end up in docs prematurely.)

PR #4160, opened 2026-08-27
Add Expressive Code missing translations based on Astro Docs

ArmandPhilippot:
#### Description

Follow-up on https://github.com/withastro/docs/pull/14460#issuecomment-5402445200

Expressive Code only has English and German translations built-in. Astro Docs has more translations: we should take advantage of this to upstream the missing translations to Starlight.

I only included the missing ones: [`ar`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ar.yml#L58-L60), [`de`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/de.yml#L60-L62), [`hi`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/hi.yml#L60-L62), [`it`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/it.yml#L60-L62), [`ja`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ja.yml#L60-L62), [`ko`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ko.yml#L60-L62), [`pt`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/pt-BR.yml#L60-L62), [`zh-CN`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/zh-CN.yml#L60-L62), and [`zh-TW`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/zh-TW.yml#L60-L62).

The following are not included:
- `es`: [Astro Docs](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/es.yml#L61-L63) / [Starlight](https://github.com/withastro/starlight/blob/23207221169b31e0cc210451a39a19c80fa1f69e/packages/starlight/translations/es.json#L27-L29) (seems identical)
- `pl`: [Astro Docs](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/pl.yml#L58-L60) / [Starlight](https://github.com/withastro/starlight/blob/23207221169b31e0cc210451a39a19c80fa1f69e/packages/starlight/translations/pl.json#L28-L30) (seems identical)
- `ru`: [Astro Docs](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ru.yml#L60-L62) / [Starlight](https://github.com/withastro/starlight/blob/23207221169b31e0cc210451a39a19c80fa1f69e/packages/starlight/translations/ru.json#L28-L30) (`expressiveCode.copyButtonTooltip` seems different)
- edit: I forgot about French... well, the Starlight translation looks better to me so... https://github.com/withastro/docs/pull/14477

And, I haven't checked if the translation are correct... I trust Astro Docs translators. 😅

### DIFF
--- a/packages/starlight/translations/de.json
+++ b/packages/starlight/translations/de.json
@@ -24,9 +24,6 @@
   "aside.tip": "Tipp",
   "aside.caution": "Achtung",
   "aside.danger": "Gefahr",
-  "expressiveCode.terminalWindowFallbackTitle": "Terminal-Fenster",
-  "expressiveCode.copyButtonTooltip": "In Zwischenablage kopieren",
-  "expressiveCode.copyButtonCopied": "Kopiert!",
   "fileTree.directory": "Ordner",
   "builtWithStarlight.label": "Erstellt mit Starlight",
   "heading.anchorLabel": "Abschnitt betitelt „{{title}}“"
--- a/packages/starlight/translations/de.json
+++ b/packages/starlight/translations/de.json
@@ -24,6 +24,9 @@
   "aside.tip": "Tipp",
   "aside.caution": "Achtung",
   "aside.danger": "Gefahr",
+  "expressiveCode.terminalWindowFallbackTitle": "Terminal-Fenster",
+  "expressiveCode.copyButtonTooltip": "In Zwischenablage kopieren",
+  "expressiveCode.copyButtonCopied": "Kopiert!",
   "fileTree.directory": "Ordner",
   "builtWithStarlight.label": "Erstellt mit Starlight",
   "heading.anchorLabel": "Abschnitt betitelt „{{title}}“"
--- a/.changeset/clean-breads-decide.md
+++ b/.changeset/clean-breads-decide.md
@@ -2,4 +2,4 @@
 "@astrojs/starlight": patch
 ---
 
-Adds the missing translations for Expressive Code in the following locales: `ar`, `de`, `hi`, `it`, `ja`, `ko`, `pt`, `zh-CN`, and `zh-TW`.
+Adds the missing translations for Expressive Code in the following locales: `ar`, `hi`, `it`, `ja`, `ko`, `pt`, `zh-CN`, and `zh-TW`.
--- a/.changeset/clean-breads-decide.md
+++ b/.changeset/clean-breads-decide.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/starlight": patch
+---
+
+Adds the missing translations for Expressive Code in the following locales: `ar`, `de`, `hi`, `it`, `ja`, `ko`, `pt`, `zh-CN`, and `zh-TW`.
--- a/packages/starlight/translations/ar.json
+++ b/packages/starlight/translations/ar.json
@@ -24,6 +24,9 @@
   "aside.tip": "نصيحة",
   "aside.caution": "تنبيه",
   "aside.danger": "تحذير",
+  "expressiveCode.terminalWindowFallbackTitle": "نافذة طرفيّة",
+  "expressiveCode.copyButtonTooltip": "نسخ إلى الحافظة",
+  "expressiveCode.copyButtonCopied": "تم النسخ!",
   "fileTree.directory": "Directory",
   "builtWithStarlight.label": "طوِّر بواسطة Starlight",
   "heading.anchorLabel": "Section titled “{{title}}”"
--- a/packages/starlight/translations/hi.json
+++ b/packages/starlight/translations/hi.json
@@ -24,6 +24,9 @@
   "aside.tip": "संकेत",
   "aside.caution": "सावधानी",
   "aside.danger": "खतरा",
+  "expressiveCode.terminalWindowFallbackTitle": "टर्मिनल विंडो",
+  "expressiveCode.copyButtonTooltip": "क्लिपबोर्ड पर कॉपी करें",
+  "expressiveCode.copyButtonCopied": "कॉपी हो गया!",
   "fileTree.directory": "Directory",
   "builtWithStarlight.label": "Starlight द्वारा निर्मित",
   "heading.anchorLabel": "Section titled “{{title}}”"
--- a/packages/starlight/translations/it.json
+++ b/packages/starlight/translations/it.json
@@ -24,6 +24,9 @@
   "aside.tip": "Consiglio",
   "aside.caution": "Attenzione",
   "aside.danger": "Pericolo",
+  "expressiveCode.terminalWindowFallbackTitle": "Finestra del terminale",
+  "expressiveCode.copyButtonTooltip": "Copia",
+  "expressiveCode.copyButtonCopied": "Copiato!",
   "fileTree.directory": "Directory",
   "builtWithStarlight.label": "Built with Starlight",
   "heading.anchorLabel": "Sezione intitolata “{{title}}”"
--- a/packages/starlight/translations/ja.json
+++ b/packages/starlight/translations/ja.json
@@ -24,6 +24,9 @@
   "aside.tip": "ヒント",
   "aside.caution": "注意",
   "aside.danger": "危険",
+  "expressiveCode.terminalWindowFallbackTitle": "ターミナルウィンドウ",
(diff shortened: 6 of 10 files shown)


packages/starlight/translations/de.json
```diff
@@ -24,6 +24,9 @@
   "aside.tip": "Tipp",
   "aside.caution": "Achtung",
   "aside.danger": "Gefahr",
+  "expressiveCode.terminalWindowFallbackTitle": "Terminal-Fenster",
+  "expressiveCode.copyButtonTooltip": "In Zwischenablage kopieren",
+  "expressiveCode.copyButtonCopied": "Kopiert!",
```
delucis: Should we include these? Expressive Code already includes exactly these itself:

https://github.com/expressive-code/expressive-code/blob/e0f1321544cf4a60bd07633499c5aaae300f8308/packages/%40expressive-code/plugin-frames/src/index.ts#L75-L79

I’d guess we can rely on Expressive Code for the German ones.

```suggestion
```

delucis (commented):
Awesome! Thanks for following up on this and taking the time to figure out which ones it made sense to copy.

I only spotted one potential change, although, like you, I can’t judge the translation quality itself. But I also trust the Astro Docs process and these should definitely be better than no translation.

packages/starlight/translations/de.json
trueberryless: My two cents on this: I think it adds a little bit of confusion if they are missing only for some languages just because they are already provided from somewhere else. Moreover, defining them here would mean independence and control of the translation strings.

But at the same time we also don't re-translate Pagefind's i18n strings, so I agree that we should stick to this pattern.

trueberryless:
Thanks for the contribution and detailed PR description!

The PR overall looks really solid. However, one thing I noticed were differences between translations of the tooltip: [Expressive Code itself uses `"Copy to clipboard"`](https://github.com/expressive-code/expressive-code/blob/e0f1321544cf4a60bd07633499c5aaae300f8308/packages/%40expressive-code/plugin-frames/src/index.ts#L71). According to Deepl, these languages also translate back to `"Copy to clipboard"`: [`ar`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ar.yml#L59), [`de`](https://github.com/expressive-code/expressive-code/blob/e0f1321544cf4a60bd07633499c5aaae300f8308/packages/%40expressive-code/plugin-frames/src/index.ts#L77), [`es`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/es.yml#L62), [`hi`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/hi.yml#L61), [`ja`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ja.yml#L61), [`ko`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ko.yml#L61), [`pl`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/pl.yml#L59), [`pt-BR`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/pt-BR.yml#L61), [`zh-CN`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/zh-CN.yml#L61), and [`zh-TW`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/zh-TW.yml#L61). 

However, these two languages seem to only translate the string `"Copy"`: [`it`](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/it.yml#L61), [`ru`](https://github.com/withastro/starlight/blob/23207221169b31e0cc210451a39a19c80fa1f69e/packages/starlight/translations/ru.json#L29).

Russian could easily be fixed by taking over the translation string from [Astro docs](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/ru.yml#L61):

image

For Italian, we might want to get help from a native-speaker from Discord 🤔

packages/starlight/translations/de.json
ArmandPhilippot: My bad, apparently... I hesitated, but I thought I saw the strings in the English file, so I figured we should do the same for German. But it seems I got mixed up between all those files (maybe I checked the Astro Docs version 😅 ). So, yeah, it makes sense to me to not include the German translations! I'll update the changeset.

trueberryless: One more thing I just noticed is a small differences in the translation string between [Astro docs](https://github.com/withastro/docs/blob/b9773d37fae129f1b807b2ce33a0932092a6aff0/src/content/i18n/de.yml#L61) ( `"In Zwischenablage kopieren"`) and [Expressive Code](https://github.com/expressive-code/expressive-code/blob/e0f1321544cf4a60bd07633499c5aaae300f8308/packages/%40expressive-code/plugin-frames/src/index.ts#L77) (`"In die Zwischenablage kopieren"`) -- notice the missing `"die"` in the Astro docs. This begs the question: Which one should we take?

Well... what a coincidence that I am German native-speaker and believe that `"In Zwischenablage kopieren"` would fit better for a tooltip because it is short and sufficient + the English version also does not include a `"the"`. This would mean that we should override the translation strings in Starlight or adapt it upstream -- depending on the outcome of the original intention of this thread.

ArmandPhilippot:
@trueberryless Yeah, I noticed the difference. However, I don't know what the common usage is in those languages: just because a machine translation tool suggests something doesn't mean it's the expression actually used. Perhaps the concept of a "clipboard" doesn't add any value for them?

But, if we think this is problematic vs falling back to English; I don't mind removing them.

packages/starlight/translations/de.json
ArmandPhilippot: Ah, I already applied Chris suggestion. I'm not a German speaker so I can't really give my opinion on this... But, if you think the German translation is wrong I guess you can suggest a change on Expressive Code directly? Or make a following PR to correct that in Starlight-only?

Again, I don't speak German so maybe Chris opinion is more appropriated here. 😄

trueberryless: I'll create a follow-up PR of this follow-up PR, so we do not block this PR. 😊

HiDeoo: If we remove them and German needs some fixes, I guess this should be a PR in Expressive Code so all users can benefit from the improvements so this wouldn't block this PR anw I guess.

delucis: Yeah, ultimately the ideal is to move stuff upstream so all Expressive Code users benefit. We discussed this in the past with Hippo and because the defaults are set in a slightly weird way with the programmatic `addLocale()` API he preferred to wait and let Starlight collect definitions for now.

Eventually I would guess the best would be to add some data files to Expressive Code to make adding and accepting more languages there an easier thing to do.

delucis (approved):
Thanks @ArmandPhilippot 🙌

Re: @trueberryless’s last comment — I definitely trust the human translators over machine translation, so I’m happy to accept these and can always tweak based on native speaker feedback if we get some in the future.

PR #4173, opened 2026-09-02
Update dependencies

delucis:
#### Description

This PR updates the Astro MDX integration to v8, dropping a bunch of dependencies.

It also requires bumping the minimum supported Astro version from 7.2.5 to 7.2.10 and the dependencies for Sätteri/Remark integration too. We were already bumping these for the current release, so I’ve updated that changeset to reflect this.


Adammatthiesen (approved):
LGTM

PR #4175, opened 2026-09-02
[i18nIgnore] Test v1 beta of `prettier-plugin-astro`

delucis:
#### Description

This is a draft PR updating to `prettier-plugin-astro` v1 (beta) and seeing the results.

I’ve had to make some manual tweaks here:

1. Improve some of the results, cleaning up whitespace etc. after things shifted around
2. Fix some bugs currently in the beta. Most notably CSS at-rules are being collapsed so all `<style>` blocks with `@media`, `@layer` or `@supports` were receiving a buggy formatting result. Definitely means we can’t use the beta just yet.


HiDeoo:
Just adding a note that it would be nice to investigate if we can remove some entries from our [`.prettierignore` file](https://github.com/withastro/starlight/blob/main/.prettierignore#L19-L24) and see if some things around some components have improved or if we can fix them maybe while at it.

packages/starlight/src/components/SidebarPersister.astro
```diff
@@ -66,8 +65,7 @@ declare global {
 			scroller.scrollTop = window._starlightScrollRestore;
 			delete window._starlightScrollRestore;
 		})();
-	</script>
-</sl-sidebar-state-persist>
+	</script></sl-sidebar-state-persist>
```
delucis, line 68: This seems to be the last outstanding issue — there’s a whitespace bug too, but I think it should be fixed by https://github.com/withastro/prettier-plugin-astro/pull/488 when that’s released.

PR #4180, opened 2026-09-03
Update to Vitest 5

HiDeoo:
#### Description

This PR updates the monorepo to use [Vitest 5](https://vitest.dev/blog/vitest-5.html).

On top of updating the dependencies and following the [migration guide](https://vitest.dev/guide/migration/), it also applys the following changes:

- Enables [`fsModuleCache`](https://vitest.dev/config/fsmodulecache.html) which is no longer experimental. On my machine, with a warm cache, this saves about 4 seconds on running all tests.
- Disables a diagnostic about isolation (the code includes a comment explaining why).
- Refactors how functional benchmarks are run. Vitest 5 now warns us about issues with the way we were running benchmarks that can make results unreliable. I applied changes [following their documentation](https://vitest.dev/guide/benchmarking#module-runner-overhead) and try to comment as much as possible the way it works.
  image

I'll undraft the PR once the new packages satisfies our `minimumReleaseAge` and all remaining tasks are addressed.

#### Remaining tasks

- [ ] Address all `TODO(HiDeoo)` comments
- [ ] Wait for `@codspeed/vitest-plugin` to support Vitest 5 ([PR opened](https://github.com/CodSpeedHQ/codspeed-node/pull/86))
- [ ] Vitest 5 also includes somes fixes that should allow us to fix our `coverage.include` and clean up our `coverage.exclude` settings, but it would probably be best to do this in another PR. I'll investate and report back.

### DIFF
--- a/packages/starlight/__bench_fn__/benchmark-functions.ts
+++ b/packages/starlight/__bench_fn__/benchmark-functions.ts
@@ -1,0 +1,15 @@
+/**
+ * Vitest runs benchmarks in Node.js using Vite's module runner which transforms all module exports
+ * into getters. With many repeated cross-module calls, this makes results unreliable because
+ * export getters add overhead.
+ *
+ * To avoid the issue, we re-export every function we want to benchmark from this module and bundle
+ * it as recommended in the Vitest documentation. During benchmarks, we reference the bundled
+ * functions using an alias.
+ *
+ * @see https://vitest.dev/guide/benchmarking#module-runner-overhead
+ */
+
+export { getSidebar } from '../src/utils/navigation';
+export { getRouteBySlugParam } from '../src/utils/routing';
+export { generateRouteData } from '../src/utils/routing/data';
--- a/packages/starlight/__tests__/basics/pagefind.test.ts
+++ b/packages/starlight/__tests__/basics/pagefind.test.ts
@@ -1,14 +1,10 @@
 import { fileURLToPath } from 'node:url';
-import { afterEach, expect, test, vi } from 'vitest';
+import { expect, test, vi } from 'vitest';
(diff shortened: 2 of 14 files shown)


packages/starlight/__tests__/basics/pagefind.test.ts
```diff
@@ -1,15 +1,11 @@
 import { fileURLToPath } from 'node:url';
-import { afterEach, expect, test, vi } from 'vitest';
+import { expect, test, vi } from 'vitest';
 import * as pagefind from 'pagefind';
 import { starlightPagefind } from '../../src/integrations/pagefind';
 import { TestAstroIntegrationLogger } from '../test-plugin-utils';
 
 vi.mock('pagefind');
 
-afterEach(() => {
```
HiDeoo, line 9: https://vitest.dev/guide/migration/#clearmocks-is-enabled-by-default

packages/starlight/__bench_fn__/benchmark-functions.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * To avoid the issue, we re-export every function we want to benchmark from this module and bundle
+ * it as recommended in the Vitest documentation. During benchmarks, we reference the bundled
+ * functions using an alias.
+ *
+ * @see https://vitest.dev/guide/benchmarking#module-runner-overhead
+ */
+
+export { getSidebar } from '../src/utils/navigation';
+export { getRouteBySlugParam } from '../src/utils/routing';
```
trueberryless: I assume you included `getRouteBySlugParam` here to keep the imports in `generate-route-data.bench.ts` organised?

PR #4187, opened 2026-09-08
docs: deploy docs with built Starlight

HiDeoo:
#### Description

As a follow-up to some discussion, this PR experiments with deploying the docs using a built version of Starlight.

To do that, we temporarily [overrides the Netlify build command using a configuration file](https://docs.netlify.com/build/configure-builds/file-based-configuration/) for deploy previews.

The PR also fixes an issue with the `build:dist` command where a same version of pnpm can be either a JavaScript entry point or a standalone binary (e.g. when using pnpm v12 as a package manager manager). I hit this issue locally as I now use pnpm v12 to automatically install pnpm v11.22.0 used in the monorepo but we would ultimately need the fix once switching the monorepo to pnpm v12.

If we want to move on with the approach, we should remove the override in this PR and instead update the Netlify build command.

#### Observations

- The override worked properly:
  image
- The entire "Building" Netlify step took 1mn9s instead of 1mn for the previous deploy preview.
- Building Starlight and doing the pnpm shenanigans to use this version took 14s.

Difficult to properly compare with one deploy preview but the additional time is expected and seems totally reasonable.

#### Remaining tasks

- [x] Address `TODO(HiDeoo)` comments
- [x] Update Netlify build command to `cd docs && pnpm build:dist`

### DIFF
--- a/docs/build-dist.mjs
+++ b/docs/build-dist.mjs
@@ -13,6 +13,9 @@ const docsPackageJson = new URL('package.json', docsDir);
 const tarballPath = join(tmpdir(), `starlight-${process.pid}.tgz`);
 const pnpmPath = process.env.npm_execpath ?? '';
 
+// Depending on the version, pnpm can be a JavaScript entry point or a standalone binary.
+const isPnpmJavaScript = /\.(?:c|m)?js$/i.test(pnpmPath);
+
 if (!pnpmPath || !process.env.npm_config_user_agent?.startsWith('pnpm/')) {
 	throw new Error('The script must be run using pnpm');
 }
@@ -53,5 +56,9 @@ try {
  * @param {...string} args - Arguments to pass to the pnpm command.
  */
 function runPnpm(cwd, ...args) {
-	execFileSync(process.execPath, [pnpmPath, ...args], { cwd, stdio: 'inherit' });
+	execFileSync(
+		isPnpmJavaScript ? process.execPath : pnpmPath,
+		isPnpmJavaScript ? [pnpmPath, ...args] : args,
+		{ cwd, stdio: 'inherit' }
+	);
 }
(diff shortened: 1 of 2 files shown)


delucis (commented):
Sorry, I missed that this had been marked ready for review somehow!

Seems fine to me and over all, probably OK to accept the small amount of extra build time to have the extra layer of verifying things are working as expected.

I guess the flow here would be this?

1. Remove the `netlify.toml`.
2. Merge this PR.
3. Update the config in the Netlify dashboard.

HiDeoo:
No worries and rush, thanks for the review :raised_hands:

> I guess the flow here would be this?
> 
> 1. Remove the `netlify.toml`.
> 2. Merge this PR.
> 3. Update the config in the Netlify dashboard.

Hmm, I would have initially thought about swapping 2 and 3 so that after updating the build command in the Netlify dashboard, merging this PR would immediately use the new configuration (rather than having to wait for another PR to see if everything work)?

delucis:
> Hmm, I would have initially thought about swapping 2 and 3 so that after updating the build command in the Netlify dashboard, merging this PR would immediately use the new configuration

Ah yeah, sure, that makes sense if we do them close together so that Netlify isn’t trying to run `build:dist` before we have the chance to merge this fix. OK, so we can check off 1 and then I can do the second two tasks together.

delucis (approved):
OK, I can update Netlify and merge this. I think it’s safe to ignore the Lunaria issue and fix that right after.

delucis:
I updated Netlify to `cd docs && pnpm build:dist` and will merge now!

delucis:
And just confirming that the latest deployment from `main` completed successfully and used the `build:dist` command as expected: https://app.netlify.com/projects/astro-starlight/deploys/6aabf7382072040008109545

PR #4190, opened 2026-09-09
Support any `scopedStyleStrategy`

delucis:
#### Background

- Since #615 and Starlight v0.9.0, we have forced `scopedStyleStrategy: "where"`. At the time, the higher specificity of the `"attribute"` default scoping strategy caused a lot of issues so we avoided it by sticking with the Astro v2 default of `"where"`.
- Since we released #2322 the issues with specificity are much reduced: user styles will always override our `@layer` styles making support for different scoping strategies more viable.

#### This PR

- This PR currently removes the line where we set `scopedStyleStrategy` and leave it to user discretion. This means by default users will switch to Astro’s default of `"attribute"`.
- Various small changes to styles were needed to correctly support the different scoping strategies with one case being the most common:
  - icons that had a class name passed down — these do not receive the parent scope when using the `"attribute"` strategy, so needed to be updated to use a parent selector and `:global()`.
- Because of some specificity/ordering issues **within** our `starlight.components` layer, I had to extract the `<Icon>` styles and move them to a global `.sl-icon` class name instead of just depending on an Astro-scoped `svg` element selector. I also added this class name to the icons generated by Markdown asides for consistency.
- I tested all English docs pages and two Farsi docs pages with [starlight-visual-diff](https://github.com/HiDeoo/starlight-visual-diff) to ensure consistency between all strategies and the existing appearance, but some more testing and thinking about the impact of this change is still very welcome.

#### Open questions

- It could be less disruptive to continue setting `scopedStyleStrategy` for users in the cases where they do not set it themselves. It would make Starlight’s default different from Astro though, so for now I went with removing it and documenting how to restore it in the changeset.
- Supporting multiple strategies may potentially mean there are visual differences between strategies, which we don’t notice in testing for changes made in the future. Might be worth us considering how best to handle that to ensure we maintain consistency.


delucis:
Just went through this again and marked it ready for review. I think it should be a reasonably safe change thanks to `@layer` insulating users from our styles. And for anyone it does cause issues for, the path back to the previous behaviour is very simple.

> Supporting multiple strategies may potentially mean there are visual differences between strategies, which we don’t notice in testing for changes made in the future. Might be worth us considering how best to handle that to ensure we maintain consistency.

On this point I still don’t have an answer and it may be OK? AFAICT the _increase_ in specificity of switching to the default strategy was riskier than the _decrease_ of using `"where"`. At least it seems that currently switching back to `"where"` was still totally fine without any specific effort or care to do so.

Any approach to asserting this in tests would presumably mean running something like [starlight-visual-diff](https://github.com/HiDeoo/starlight-visual-diff) in CI and that feels like a pretty big hammer for something that very often won’t apply (a small % of PRs touch styling). So maybe it’s fine to leave that as a manual thing to bear in mind.

PR #4191, opened 2026-09-10
Fix broken Astro CLI link in Portuguese guide

haddybhaiya:
The Portuguese getting-started guide links to an Astro CLI page that no longer exists. This updates it to the working Portuguese Astro documentation link.


delucis:
Thanks for the PR @haddybhaiya. The reason this link is outdated is because the entire page is outdated. We’ll wait until a translator who speaks Portuguese has time to update it in full (or potentially we may decide to cull some pages if we feel they are too outdated), but we don’t aim to patch small details in them like this.

haddybhaiya:
Fair enough, Thanks for clarification @delucis

PR #4192, opened 2026-09-10
i18n: add freshness to Lunaria dashboard

HiDeoo:
#### Description

This PR adds a new freshness section to the Lunaria dashboard that contains data that can be queried to evaluate our documentation translations.

- This initially started during various Talking & Doc'ing sessions by running git commands and analyzing the data.
- It evolved into some manual scripts running such commands and computing various metrics.
- This PR now integrates these metrics into the Lunaria dashboard.

The changes in this PR extracts required data from Lunaria when building the i18n dashboard, serializes it, and add a form to display controls to compute various freshness metrics client-side based on this data and various thresholds.

- A translation threshold (months) to show locales with outdated translations last updated more than this many months ago.
- A recent activity window (months) to show how many translation files were updated within this many months.

With time, maybe more metrics and controls will be added but for now, this addition provides the metrics we have used in the past to evaluate translation freshness.

We don't deploy previews of the Lunaria dashboard, so to see the changes locally, you can run the following commands:

```
cd docs
pnpm lunaria:build && pnpm exec lunaria preview
```

Also sharing a screenshot of the new freshness section for reference.

lunaria-freshness

### DIFF
--- a/docs/lunaria/components.ts
+++ b/docs/lunaria/components.ts
@@ -1,4 +1,6 @@
-import { html } from '@lunariajs/core';
+import { readFileSync } from 'node:fs';
+import { html, type LocalizationStatus } from '@lunariajs/core';
+import type { LunariaConfig } from '@lunariajs/core/config';
 
 export const TitleParagraph = () => html`
 	<p>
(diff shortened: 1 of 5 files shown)


docs/lunaria/styles.css
```diff
@@ -45,3 +45,244 @@ sup {
...
+	clip: rect(0, 0, 0, 0);
+	white-space: nowrap;
+	border-width: 0;
+}
+
+translation-freshness {
+	--freshness-color-bar-text: hsl(218, 35%, 14%);
+	--freshness-color-up-to-date: var(--ln-color-done);
+	--freshness-color-outdated-within: var(--ln-color-outdated);
+	--freshness-color-outdated-beyond: hsl(353, 89%, 63%);
```
delucis, line 65: It might be my eyes but I found it quite hard to spot the difference initially between the orange and red shades here:

part of a bar chart with an orange section and a red section. the colours are very similar.

Maybe the `outdated-within` variant should be a lighter yellow shade or something like that? For example:
```suggestion
	--freshness-color-outdated-within: hsl(30, 100%, 75%);
	--freshness-color-outdated-beyond: var(--ln-color-outdated);
```

the freshness UI showing a brighter orange-yellow for the middle segment of the bar graphs and the current dark orange for the end section

docs/lunaria/components.ts
```diff
@@ -16,3 +18,171 @@ export const TitleParagraph = () => html`
...
+							<span class="freshness-outdated-within" data-outdated-within-bar></span>
+							<span class="freshness-outdated-beyond" data-outdated-beyond-bar></span>
+						</div>
+					</div>
+					<div class="freshness-activity">
+						<div class="freshness-bar" aria-hidden="true">
+							<span class="freshness-recent" data-activity-bar></span>
+						</div>
+						<p data-activity-count aria-hidden="true"></p>
+					</div>
```
delucis, line 183: It took me a while to realise that the two columns are essentially different perspectives on the same set of documents, i.e. both bars have the same number of total documents even though they are not the same length.

I wonder if a visualisation like this might make that a bit clearer?

Image

This would also make it more obvious what proportion of the up to date docs were recently updated.

delucis: (in which case, I guess the header could be updated to something like this)

Image

docs/lunaria/styles.css
```diff
@@ -45,3 +45,244 @@ sup {
...
+	.freshness-chart-rows {
+		list-style: none;
+		padding: 0;
+	}
+
+	.freshness-legend {
+		display: grid;
+		font-weight: bold;
+		gap: 0.75rem 1rem;
+		grid-template-columns: repeat(auto-fit, minmax(min(100%, 10rem), 1fr));
```
delucis, line 142: Bit of a nit but the grid layout here feels a bit odd as it leaves a lot of whitespace after the first item and there’s no reason to align stuff with the things below:

Image

Might be better as a flex?
```suggestion
	.freshness-legend {
		display: flex;
		flex-wrap: wrap;
		font-weight: bold;
		gap: 0.75rem 1rem;
```

Image

delucis (commented):
Looks great @HiDeoo! Lots of extra insight here which is really useful.

One general comment: the UI does not seem to be fully responsive. If I resize the viewport, at smaller sizes it overflows, e.g.

Image

Would be nice to fix that.

docs/lunaria/components.ts
delucis: Another option would be to just integrate all four segments into a single bar, so for this one you’d do:

```
[ 3 |    9    | 3 |         20         ]
```

- Up to date ≤ 3 months: 3
- Up to date > 3 months: 9
- Outdated ≤ 12 months: 3
- Outdated > 12 months: 20

delucis: Or we go the whole hog and use two histograms which is essentially what this is trying to show:

```
UP TO DATE             OUTDATED
0–3 months  ███        0–3 months
3–6 months  ██         3–6 months  ██
6–9 months  █████      6–9 months
9–12 months            9–12 months █
>12 months  ██         >12 months  ████████████████████
```

trueberryless: One unrelated suggestion I had regarding the up-to-date column is to only consider outdated pages. This means, pages that never change don't influence the percentage value, which would make sense to me as they are considered up-to-date already. I find the current visualisation a bit misleading because let's assume a language has everything up-to-date, but no English document changed since 3 months, it would show as 0% instead of 100%.

trueberryless (commented):
Thanks for the PR, HiDeoo! The additional section look really good in general. I just have some initial questions:

- Did you talk with Yan about the idea to add such a feature section to Lunaria itself (potentially opt-in) or do you have a reason why this should be Starlight specific and not available to all Lunaria users?
- Why did you choose the granularity months in the overview and days in the detailed view? I mean it makes sense that an overview gives less granularity than the detailed section, but it adds a little unintuitivity in my opinion, as you have to convert the ranges in your head.

HiDeoo:
> [!WARNING]
> The Lunaria action is currently failing due to https://github.com/withastro/starlight/pull/4203.

Updated the branch with the Lunaria update and suggestions from @delucis. I also updated the screenshot in the PR preview.

Regarding the graph layout suggestions, I tried all of them and decided to go with the [first one](https://github.com/withastro/starlight/pull/4192#discussion_r3988284459). It's a huge improvement over the previous layout but was also the only one that felt intuitive to me. Open to more discussions if needed.

> One general comment: the UI does not seem to be fully responsive. If I resize the viewport, at smaller sizes it overflows, e.g.

Ah, forgot to mention that in the PR. I actually noticed that while initially working on the PR and even have a todo about it but I noticed this actually already exists in the [live version](https://i18n.starlight.astro.build/).

Only did a small investigation so far but it looks like the issue is related to tables and if we hide them, the layout and new section is properly responsive.

lunaria-responsive

Considering the tables in the new section use the same markup for tables as Lunaria itself, I think a better fix would be to fix the issue in Lunaria itself for all users (and update our markup accordingly once the fix is available) as a fix in this PR would probably be only for the new tables but the existing Lunaria one would still have the issue (I guess we may need some kind of wrapper around the tables).

We can also wait for an upstream fix and update before merging this PR. I'll try to get to the todo tomorrow morning.

> One unrelated suggestion I had regarding the up-to-date column is to only consider outdated pages. This means, pages that never change don't influence the percentage value, which would make sense to me as they are considered up-to-date already. I find the current visualisation a bit misleading because let's assume a language has everything up-to-date, but no English document changed since 3 months, it would show as 0% instead of 100%.

I don't think that exact example applies here as a locale with everything up to date would be filtered out and not visible at all in this section.

The activity bar shows recent edits, rather than whether a locale is maintained or up to date. For example, let's imagine a locale with 100 translated pages: 99 of them are up to date and 1 is an old outdated page. If no translations have been edited recently, the activity bar would show 0% (and the status bar would show 99% up to date) and that is accurate as this locale did not have recent changes, which is the goal of this bar.

> Did you talk with Yan about the idea to add such a feature section to Lunaria itself (potentially opt-in) or do you have a reason why this should be Starlight specific and not available to all Lunaria users?

Yes. This PR adds exactly what we have used in the past for Starlight, and nothing more (and will be updated if we need more stats). Lunaria will in the future have a plugin for this kind of statistics once Yan will have a chance to implement it, think about all users and not just Starlight, which stats would make sense for everyone, maybe have it configurable, consider all Lunaria users and use-cases, and such things like that but this is not the goal of the PR.

> Why did you choose the granularity months in the overview and days in the detailed view? I mean it makes sense that an overview gives less granularity than the detailed section, but it adds a little unintuitivity in my opinion, as you have to convert the ranges in your head.

When tweaking thresholds, I found it easier to reason about them in terms of months rather than days. Once I focus on a locale that I want to investigate based on the graphs, I click on the locale label to jump to it's detailed view where I want to see the more granular day-driven data. That's pretty much it so far. As this section is primarily meant to be used by maintainers only when the team decides to evaluate our translations, we would have to see if maintainers find this approach intuitive or if adjustments are needed.

delucis (approved):
Thanks for the detailed explanations @HiDeoo!

I think this all makes sense and yes, the table is indeed the issue impacting responsiveness, so we can fix it upstream and merge this PR without worrying about it.

I checked this out locally and it worked great as expected.

The only other thing that popped into my head while looking was that maybe on the detail tables, the colour coding could also be helpful to highlight outdatedness more visually, e.g.

table showing Hindi document freshness, each row in the table starts with a small block of colour to indicate outdatedness

But currently most pages seem to be in the “beyond” category, so probably would only makes sense with a different threshold or when we have a different mix of outdated stats to look at.

In any case this could be a future enhancement so doesn’t have to happen in this PR.

HiDeoo:
I opened a [Lunaria PR](https://github.com/lunariajs/lunaria/pull/186) for the responsive issue. Once a fix is agreed upon and released, I'll make a PR bumping Lunaria in the monorepo and updating our table markup accordingly.

PR #4197, opened 2026-09-12
chore: remove Astro 7 upgrade banner [i18nIgnore]

trueberryless:
#### Description

This PR removes the "Upgrade to Astro 7" banner on the Starlight landing page of all locales because the Astro v7 release was already [almost 3 months ago (June 22)](https://github.com/withastro/astro/releases/tag/astro%407.0.0) and this would give the landing page a fresh look before Astro v8 is released.


gameroman (approved):

delucis (approved):
Yeah, I think this is reasonable. Overall, usage of versions pre v7 is still fairly high (roughly 50:50 ≥7:<7) so people _will_ still be upgrading, but it’s not brand new anymore, so hopefully people are OK with figuring it out.

PR #4198, opened 2026-09-13
remove `popover` attribute from sidebar-pane on wide viewports

### LINKED ISSUE #4193: Sidebar uses a closed popover on wide viewports

0.42.0

7.3.2

pnpm

Mac

Chrome, Firefox, Safari

### Describe the Bug

Starting with 0.42.0 (#4121), the sidebar navigation now uses the `popover` attribute:

https://github.com/withastro/starlight/blob/88ad3c2630487ba227a7b4ccbffc01a2bdf623a5/packages/starlight/src/components/PageFrame.astro#L35

This works well for mobile/narrow viewports, where the sidebar's visibility is controlled by a `<button>` that has the `popovertarget` attribute.

---

The problem is that on wider viewports, this popover is never opened.

The `PageFrame` component instead overrides the default `display` to make the element visible: https://github.com/withastro/starlight/blob/88ad3c2630487ba227a7b4ccbffc01a2bdf623a5/packages/starlight/src/components/PageFrame.astro#L113-L115

But the underlying popover is still closed. (Technically, the ["popover visibility state"](https://html.spec.whatwg.org/multipage/popover.html#popover-visibility-state) is `hidden`.)

```JS
document.querySelector("sl-sidebar-pane").matches(":popover-open") // false
```

The `display: block` declaration is very easily overridden by a [CSS reset](https://mayank.co/notes/popover-default-styles/) if it uses an `!important` declaration in a lower layer. With this setup, upgrading to Starlight 0.42.0 silently broke my sidebars.

Workarounds: I can patch `@astrojs/starlight`, or override `PageFrame`, or use client-side JS to call `showPopover()`.

Ideally, the `popover` attribute is only set for the mobile UI where it's needed.

https://stackblitz.com/edit/github-wkrtu2bm?file=src%2Fstyles.css

### Participation

mayank99:
Closes #4193 

### Changes

- Updates the `<sl-sidebar-pane>` element definition to remove/re-add the [`popover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/popover) property based on whether the media query matches.
- This also replaces the existing `hidePopover` call, which is automatically handled when the `popover` property is changed.

### Testing

Tested the dev server by resizing the viewport across the `50em` breakpoint.
- On wide/desktop viewports, there is no `popover` attribute (fixing the linked issue).
- On narrow viewports, the `popover` attribute returns and the mobile menu continues to work.


delucis (approved):
Thanks for taking the time to make this change @mayank99 💜

PR #4204, opened 2026-09-17
Update Lunaria action version in workflow

yanthomasdev:
#### Description

Updates the Lunaria action version in the workflow to fix the failing runs issue


delucis (approved):
Lunaria Gets To Merge 🚀🌛

HiDeoo (approved):
Thanks :raised_hands: If this hash is good enough for `docs`, it'll be good enough for us 😄

PR #4208, opened 2026-09-19
docs: add Pinlyx Docs to the showcase

azorkai:
## Description

Adds [Pinlyx Docs](https://docs.pinlyx.com/) to the showcase page. It is the public documentation for the Pinlyx CRM API (REST reference, guides, MCP server and webhooks), built with Starlight `0.36` and `starlight-openapi`.

- Screenshot added at `docs/src/assets/showcase/docs.pinlyx.com.png` (800 × 450, PNG, named after the domain)
- Entry appended at the end of the list in `docs/src/components/showcase-sites.astro`

No changeset needed, docs site only.


HiDeoo (approved):
Thanks for the great new addition to the showcase :raised_hands:

